PR #5, opened 2020-06-26
build(storybook): deploy Storybook to Netlify

### LINKED ISSUE #4: Add github pages
Need to add github pages to the build
- needs to contain the collective story book for all packages
- needs to be updated when the build is done with the changes from the packages

SimonFinney:
Closes #4 

## Changelog

**New**

- Add Storybook build commands
- Add Netlify deploy configuration


netlify.toml
```diff
@@ -0,0 +1,5 @@
+# Netlify build and deploy settings - https://docs.netlify.com/configure-builds/file-based-configuration/#sample-file
+[build]
+base = "packages/core"
+command = "yarn build"
```
dbrugger, line 4: maybe better to do something like `yarn build-storybook` here to prevent unnecessary build steps to sneak in for netlify as they get added to the general build target

PR #8, opened 2020-07-01
build(main): add entry points, permit Storybook formatting

SimonFinney:
## Changelog

**New**

Add entry points for `packages/security`

**Changed**

Apply Prettier formatting to Storybook source files

**Removed**

Storybook source files from `.prettierignore`

### DIFF
--- a/packages/security/package.json
+++ b/packages/security/package.json
@@ -3,7 +3,6 @@
   "private": true,
   "version": "0.0.0",
   "license": "Apache-2.0",
-  "main": "src/index.js",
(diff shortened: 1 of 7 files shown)


packages/security/package.json
```diff
@@ -3,6 +3,7 @@
   "private": true,
   "version": "0.0.0",
   "license": "Apache-2.0",
+  "main": "src/index.js",
```
dbrugger, line 6: should the `main` not refer to the dist entrypoint, and have a `module` equivalent? https://github.com/carbon-design-system/carbon/blob/master/packages/react/package.json#L6-L7

PR #10, opened 2020-07-01
docs(storybook): add sectionTitle configs to move components easily

dbrugger:
#### Changelog

**New**

- add `sectionTitle` in `packages/<section>/config.js`

**Changed**

- replaced hardcoded section titles with the string from the config
- moved dummy stories into dummy component folders

### DIFF
--- a/packages/automation/src/Dummy/Dummy.stories.js
+++ b/packages/automation/src/Dummy/Dummy.stories.js
@@ -7,8 +7,10 @@
 
 import React from 'react';
 
-export const Default = () => <>Story</>;
+import { sectionTitle } from '../../config';
+
+export const Default = () => <>Dummy Story</>;
 
 export default {
-  title: 'Automation/Story',
+  title: `${sectionTitle}/Dummy`,
 };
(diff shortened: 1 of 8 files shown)


packages/automation/src/Dummy/Dummy.stories.js
```diff
@@ -7,8 +7,10 @@
...
 import React from 'react';
 
-export const Default = () => <>Story</>;
+import { sectionTitle } from '../../config';
+
+export const Default = () => <>Dummy Story</>;
 
 export default {
-  title: 'Automation/Story',
+  title: `${sectionTitle}/Dummy`,
```
SimonFinney, line 15: Is there still a way for us to unify the formatting for these stories centrally in `core`? For example, while the section title might live in each individual package, we can still ensure that the `${sectionTitle}/Story` string is consistent across each.

dbrugger: i cannot think of an easy / straightforward way really without creating overhead of this relatively simple formatter

PR #25, opened 2020-07-20
docs(scroll-gradient): add `ScrollGradient` and story

SimonFinney:
## Changelog

**New**

- Add `ScrollGradient` component and styles
- Add basic `ScrollGradient` story with knobs


SimonFinney:
@cameroncalder Potentially! We'd need to create an issue in https://github.com/carbon-design-system/ibm-security to tackle as it's not currently in place today

lee-chase (changes requested):
Conflict should be resolved in favour of always using the current Carbon version. All packages should use the current version as otherwise, conflicts such as bx--grid--narrow not being available because the security package required 10.12.0 of @carbon/grid.

P.S. Also worry that we are not bringing code across to this package. I do not think we should publish with wrappers.

PR #46, opened 2020-08-04
Add browser support to README

dbrugger:
#### Changelog

**New**

- add list of supported browsers to README


asfordmatt:
I'd recommend saying the latest version of those browsers. 
Also do we need to include ESR versions of Firefox?

dbrugger:
Thank you @asfordmatt for your input on this! I've added mention of _latest_ and removed ESR for now so we can start with a slim baseline but happy to get further input!

PR #78, opened 2020-08-14
feat: add experimental package

lee-chase:
Add an experimental package to provide a place for work to progress without impacting on other packages.


packages/core/package.json
```diff
@@ -20,6 +20,7 @@
     "@carbon/type": "^10.12.0",
     "@storybook/addon-actions": "^6.0.6",
     "@storybook/addon-docs": "^6.0.6",
+    "@storybook/addon-controls": "^6.0.6  ",
```
SimonFinney, line 23:
```suggestion
    "@storybook/addon-controls": "^6.0.6",
```

packages/experimental-package/src/global/styles/_carbon-imports.scss
```diff
@@ -0,0 +1,14 @@
...
+// LICENSE file in the root directory of this source tree.
+//
+
+// @import './carbon-settings';
+
+// Global styles
+// @import 'carbon-components/scss/globals/grid/grid';
+
+// Carbon components
+@import 'carbon-components/scss/components/button/button';
```
SimonFinney, line 14: I wonder if this import should live within the Sass files of `ExampleComponent` if it's being explicitly used by that?

lee-chase: That should be OK as Carbon builds in a use once mechanism.

A note of caution perhaps with regards to including SCSS imports in the react component. Some webpack builds treat components as separate compilation units which results in many copies of SCSS prior to any minification.

PR #81, opened 2020-09-02
feat: add event handler to experimental example

lee-chase:
Adds event handler for demonstration purposes to the experimental example component

### DIFF
--- a/packages/experimental-package/src/components/ExampleComponent/ExampleComponent.stories.js
+++ b/packages/experimental-package/src/components/ExampleComponent/ExampleComponent.stories.js
@@ -6,6 +6,7 @@
 //
 
 import React, { useState } from 'react';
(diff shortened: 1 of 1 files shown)


SimonFinney:
Thanks @lee-chase - would it be possible to emit these events via the 'Actions' panel in Storybook? https://www.npmjs.com/package/@storybook/addon-actions

lee-chase:
> Thanks @lee-chase - would it be possible to emit these events via the 'Actions' panel in Storybook? Here's an example of it currently - https://ibm-cloud-paks.netlify.app/?path=/story/security-decorator--default

Yes, you're right @SimonFinney I was rather looking for a pattern to use on the modified tabs and went a bit far here.

SimonFinney (approved):
LGTM! 💯

PR #82, opened 2020-09-04
feat: add storysource addon

lee-chase:
Sometimes stories/usage of a component requires state outside of the component. The ModifiedTabs for example expect the component user to maintain the list of tabs and content.

Unfortunately neither‘Show info’ nor ‘Docs’ show the actual story.

While less than perfect, the storysource addon does show the full source.


SimonFinney (approved):
LGTM! Thank you 🎉

PR #84, opened 2020-09-10
chore: add eslint with carbon config

lee-chase:
Part of #6 

Adds eslint and carbon config

#### Testing / Reviewing

Ran and no issues highlighted. Switched to airbnb config to see issues before switching back.


davidmenendez:
I pulled this down, ran the `lint` command, and saw only `done` output, but when i look at a component like the `ExampleComponent` in VS code I see a lot of lint error highlighting showing up. so i guess my question is what is the intended result from running `yarn lint`?

lee-chase:
> I pulled this down, ran the `lint` command, and saw only `done` output, but when i look at a component like the `ExampleComponent` in VS code I see a lot of lint error highlighting showing up. so i guess my question is what is the intended result from running `yarn lint`?

Needed quotes around 'packages/**/*.js'

lee-chase:
Are you able to take another look at this @davidmenendez

davidmenendez (approved):
looks good now 👍

PR #89, opened 2020-09-16
Update modified tabs

lee-chase:
Updates modified tabs adding a second story with a save dialog prompt. 

Also added the unsavedContnet indicator, could move to another PR.


davidmenendez (approved):
LGTM

PR #119, opened 2020-10-02
feat: added remove component

### LINKED ISSUE #61: Remove
Catalog # 18

[Link to Design ](https://www.carbondesignsystem.com/community/patterns/remove-pattern/)

[View remove implementation](https://ibm-cloud-cognitive.netlify.app/?path=/story/experimental-removalmodal--with-state-manager)

This may have working code, check with Vikki Paterson.

**Maintainer**
Vikki Paterson

## Tasks

- [x] Initial review of design/existing code

### Working in experimental package

- [x] Create a shell component which
  - [x] Create ComponentName/ComponentName.js component file that renders props
        in raw form (no styling or sub components, plain text OK)
  - [x] Adds prop-types with comments for DocGen to
        ComponentName/ComponentName.js
  - [x] Add Doc file e.g. ComponentName.mdx with a brief description of the
        component.
- [x] First engineering review/playback

### Continuing in experimental package

- [x] Storybook - stories to cover design scenarios, story with long, min and
      max content, etc
- [x] Doc tab in storybook derived from design docs with embedded stories where
      appropriate
- [x] Code to include prop-types with comments for automated DocGen
      documentation
- [x] Tests (written in React Test Library) to cover behavior from the
      documented in the components mdx file.

### Review and promote

- [ ] Design review
- [ ] Add design review issues addressed, issues raised or discounted.
- [ ] Engineering review - Submit PR to promote component to production package
- [ ] All engineering issues addressed, issues raised or discounted.

davidmenendez:
Closes #61 

Adds the removal modal

#### Changelog

```
A       packages/experimental-package/src/components/Remove/RemovalModal.js
A       packages/experimental-package/src/components/Remove/RemovalModal.stories.js
A       packages/experimental-package/src/components/Remove/_index.scss
A       packages/experimental-package/src/components/Remove/_removal-modal.scss
A       packages/experimental-package/src/components/Remove/_storybook-styles.scss
A       packages/experimental-package/src/components/Remove/index.js
```

#### Testing / Reviewing

when using the `with confirmation` variation verify that the primary button (submit) is only active when the text input matches the resource name


davidmenendez:
work in progress.

i've added a WIP label to designate works in progress.

davidmenendez:
This is ready for a code review. I didn't add any tests because if we are switching to RTL I didn't want to waste time with enzyme. i'll work on updating testing and adding tests here afterwards.

lee-chase:
There are 0 code changes @davidmenendez think something might have gone wrong

lee-chase (changes requested):
No file changes shown

davidmenendez:
@lee-chase thanks for pointing that out. fixed.

packages/experimental-package/src/components/RemoveModal/RemovalModal.js
```diff
@@ -0,0 +1,69 @@
...
+  textConfirmation,
+}) => {
+  const [userInput, setUserInput] = useState('');
+  const onChangeHandler = (e) => {
+    setUserInput(e.target.value);
+  };
+  const primaryButtonDisabled = textConfirmation && userInput !== resource;
+  return (
+    <Modal
+      danger
```
moores2, line 25: The `iconDescription` prop of the modal needs to be set (from a client prop) to avoid the hardcoded, English 'Close' title for the close icon.

packages/experimental-package/src/components/RemoveModal/RemovalModal.js
```diff
@@ -0,0 +1,69 @@
...
+      className={`removal-modal ${className}`}
```
moores2, line 26: All classnames in the JS and in the SCSS should be prefixed with the global prefix. See ExampleComponent.

packages/experimental-package/src/components/RemoveModal/RemovalModal.js
```diff
@@ -0,0 +1,69 @@
...
+      modalLabel={subheader}
+      primaryButtonText={primaryButtonText}
+      primaryButtonDisabled={primaryButtonDisabled}
+      secondaryButtonText={secondaryButtonText}
+      onRequestSubmit={onRequestSubmit}
+      onRequestClose={onRequestClose}>
+      <p className="removal-modal-body">{body}</p>
+      {textConfirmation && (
+        <TextInput
+          id="remove-confirmation"
```
moores2, line 38: If an id is used in a component it should be unique so as not to clash with other ids that may be in the page. There should be a common mechanism for generating these ids for our components to use.

packages/experimental-package/src/components/RemoveModal/RemovalModal.js
```diff
@@ -0,0 +1,69 @@
...
+          invalidText="A valid value is required"
```
moores2, line 39: There should be no hardcoded English text in a component, except possibly for default values.

packages/experimental-package/src/components/RemoveModal/_removal-modal.scss
```diff
@@ -0,0 +1,18 @@
+@import '../../global/styles/carbon-settings';
+@import 'carbon-components/scss/components/modal/modal';
+
+.removal-modal .bx--modal-container {
+  background: $ui-03;
+}
+
+.removal-modal .bx--btn--secondary {
+  background: $ui-05;
```
moores2, line 9: The secondary button text is not visible when using the g90 and g100 themes.

moores2 (dismissed):

davidmenendez:
@moores2 thanks for the feedback. changes made. the things you pointed out are things i will definitely think about when adding future code and components 👍

lee-chase (changes requested):
I would like to find a way of simplifying wrapped components which may otherwise leave us with a maintenance issue. I have tried to do this with the PageHeaderUtility components ActionBarItems and PageHeaderItems, which may or may not be a good way of doing this.

Can we have a get together to discuss
@moores @davidmenendez

packages/experimental-package/src/components/RemoveModal/RemovalModal.js
```diff
@@ -0,0 +1,85 @@
+import React, { useState } from 'react';
+import { Modal, TextInput } from 'carbon-components-react';
+import PropTypes from 'prop-types';
+import { expPrefix } from '../../global/js/settings';
+
+export const RemovalModal = ({
+  body,
```
lee-chase, line 7: Can we not repeat Carbon props somehow?

packages/experimental-package/src/components/RemoveModal/RemovalModal.js
moores2: Carbon components React has a shared common function like this: 
```
let lastId = 0;

export default function(prefix = 'id') {
  lastId++;
  return `${prefix}${lastId}`;
}
```
it tends to be used like this:
```
import uid from '../../somewhere-common/uniqueId';
...
this.uid = this.props.id || uid();
...
<input
  ...
  id={this.uid}
/>
<label htmlFor={this.uid} ... >
  ...
</label>
```

All their components use it when they need an id, instead of forcing the user to supply one. In this case we just need one because it's required by `<TextInput>`. It's of no value to the client of the component, and forcing them to supply it via a prop causes them needless extra work. 

We don't have a common function for generating ids for internal use by our components yet, but we probably should add one to the repo.

davidmenendez: good idea. i'll implement that functionality. would you prefer we implement our own id generating function or use a library like [uuid](https://www.npmjs.com/package/uuid)?

lee-chase: In Carbon Vue we went with a an optional id. In effect it gets out of the way if the user specifies one. Can we do something like that?

https://github.com/carbon-design-system/carbon-components-vue/blob/master/packages/core/src/mixins/uid-mixin.js

packages/experimental-package/src/components/RemoveModal/RemovalModal.js
lee-chase: Raised this issue to discuss how we create shallow wrappers around Carbon components

https://github.com/carbon-design-system/ibm-cloud-paks/issues/138

packages/experimental-package/src/components/RemoveModal/RemovalModal.js
lee-chase: React uses the uniqueId in the new treeview component as follows which effectively allows the user to pass a value in, or get a generated value.

```
  const { current: id } = useRef(rest.id || uniqueId());
```

packages/experimental-package/src/components/RemoveModal/RemovalModal.js
```diff
@@ -0,0 +1,88 @@
...
+      modalLabel={subheader}
+      primaryButtonText={primaryButtonText}
+      primaryButtonDisabled={primaryButtonDisabled}
+      secondaryButtonText={secondaryButtonText}
+      onRequestSubmit={onRequestSubmit}
+      onRequestClose={onRequestClose}>
+      <p className={`${expPrefix}-removal-modal-body`}>{body}</p>
+      {textConfirmation && (
+        <TextInput
+          id={idRef.current}
```
lee-chase, line 47: This alongside other suggestion. If the user passes an Id we must apply to the outer element as per standard HTML.

```suggestion
          id={`${idRef.current}-confirm-input`}
```

packages/experimental-package/src/components/RemoveModal/RemovalModal.js
```diff
@@ -0,0 +1,88 @@
...
+import PropTypes from 'prop-types';
+import { expPrefix } from '../../global/js/settings';
+import uuidv4 from '../../global/js/utils/uuidv4';
+
+export const RemovalModal = ({
+  body,
+  className,
+  header,
+  iconDescription,
+  id,
```
lee-chase, line 12:
```suggestion
  id={idRef.current}
```

packages/experimental-package/src/components/RemoveModal/RemovalModal.js
```diff
@@ -0,0 +1,88 @@
...
+  inputLabelText,
+  inputPlaceholderText,
+  onRequestClose,
+  onRequestSubmit,
+  open,
+  primaryButtonText,
+  resource,
+  secondaryButtonText,
+  subheader,
+  textConfirmation,
```
lee-chase, line 23:
```suggestion
  textConfirmation,
  ...other,
```

packages/experimental-package/src/components/RemoveModal/RemovalModal.js
```diff
@@ -0,0 +1,88 @@
...
+  const onChangeHandler = (e) => {
+    setUserInput(e.target.value);
+  };
+  const primaryButtonDisabled = textConfirmation && userInput !== resource;
+  return (
+    <Modal
+      danger
+      className={`${expPrefix}-removal-modal ${className}`}
+      iconDescription={iconDescription}
+      open={open}
```
lee-chase, line 36:
```suggestion
      open={open}
      id={id}
```

packages/experimental-package/src/components/RemoveModal/RemovalModal.js
```diff
@@ -0,0 +1,88 @@
...
+      modalHeading={header}
+      modalLabel={subheader}
+      primaryButtonText={primaryButtonText}
+      primaryButtonDisabled={primaryButtonDisabled}
+      secondaryButtonText={secondaryButtonText}
+      onRequestSubmit={onRequestSubmit}
+      onRequestClose={onRequestClose}>
```
lee-chase, line 43:
```suggestion
      onRequestClose={onRequestClose}
      {...other}>
```

lee-chase (approved):
LGTM David. 

I think you can remove the id from propTypes and defaultProps and include it as a test case and maybe a story. You can add args to the story that are not listed in propTypes or defaultProps.

I assume next up is the creation of stories and tests. For stories we have chosen to represent each of the examples given in the design page as a story. I think Simon has done a pretty good job so far in PageHeader.mdx although there are some issues to iron out. For tests Simon is looking at switching to React test library, so may appreciate the example. 

P.S. Document blocks in RemoveModal.propTypes will show up in the RemoveModal.mdx render on the Docs tab.

PR #127, opened 2020-10-06
feat(cdai): add build setup

asfordmatt:
Set up build step for `cdai` components using Babel


lee-chase (approved):
LGTM although as Simon introduced the build for the repository it may be worth him reviewing this too.

I'm guessing this is a straight copy of the IDE build and as a result, I'm happy with it.

The main build produces a bundle, as opposed to individually compiled components (unlike Carbon) which we may wish to look in to.

packages/cdai/package.json
```diff
@@ -12,6 +12,31 @@
     "cdai"
   ],
   "scripts": {
-    "storybook": "start-storybook -p 6006"
+    "storybook": "start-storybook -p 6006",
+    "build": "node scripts/build.js"
+  },
+  "dependencies": {
+    "@babel/plugin-proposal-class-properties": "^7.10.4",
```
SimonFinney, line 19: Should these be `devDependencies` versus dependencies, or will they be included at runtime?

SimonFinney (approved):
LGTM! 🥇 I had one question on package dependencies, but I'm also wondering if the eventual intention is to include the `packages/cdai` as part of the overall bundle in:

- https://github.com/carbon-design-system/ibm-cloud-paks/blob/master/packages/cloud-paks/src/index.js
- https://github.com/carbon-design-system/ibm-cloud-paks/blob/master/packages/cloud-paks/src/index.scss

packages/cdai/package.json
asfordmatt: We're having this discussion/debate in the CDAI guild :)
Currently they are in the same place as in the 'ide-components' repo.
The argument is that such plugins insert code which is executed at runtime, therefore that should count as a dependency instead of devDependency. Some security scanning tools ignore devDependencies, so this approach is safer, as a vulnerability in such a plugin could potentially affect the running application.

asfordmatt:
@SimonFinney would be great to discuss the above - whether we publish CDAI components under a separate namespace, or altogether with the others.

Issue #59, opened 2020-08-05
Page Header


carrenelloyd:
## Issue description

Catalog # 16

Design maintainer: Cameron Calder
Dev maintainer: Lee Chase

[Link to Design ](https://pages.github.ibm.com/cdai-design/pal/components/page-header/usage/)
Replace existing design with new design below

[Link to Design Issue](https://github.ibm.com/CDAI-design/pal/issues/102)

[Working code](https://pages.github.ibm.com/cloud-integration-design/ide-components/?path=/story/production-ready-ide-navigation-components-ide-page-header-component--with-breadcrumb-heading-subtitle-and-tabs-properties)

[View page header implementation](https://ibm-cloud-cognitive.netlify.app/?path=/story/experimental-pageheader--all-attributes-set)

## Implementation tasks

- [x] Evaluate existing components. Is it a good starting point for the new header.
- [x] Boolean property to control background appearance
- [x] Breadcrumb and action bar (optional, jsx) - Array of BreadcrumbItem
- [x] Title (optional, prop string 56) and Actions (optional, jsx)
- [x] Subtitle/description area (string, optional)
- [x] Available space (optional, jsx)
- [x] Tabs / Content switcher (optional slot, jsx) (story to include content of tabs)
- [x] Stories (1 story per every design example)
- [x] Tests
  - [x] Convert tests to React Test Library
- [x] Documentation
  - [x] Fix scroll behaviour in Doc page.
- [x] Outstanding issues:
  - [x] No easing when the primary action buttons jump to action bar during scrolling.
  - [x] Primary action buttons possibly jump too early - perhaps they should jump just as they hit the action bar and not before.
  - [x] In this [story](http://localhost:3000/?path=/story/experimental-pageheader--with-background-breadcrumbitems-title-pageactions-tags)  when you scroll the tags off the page it leaves too much whitespace below the primary page action buttons.
  - [x] The title wraps and it should not.
  - [x] The title should be limited to 56 chars and truncated and an ellipse added if it's longer that that, with the full text in the browser hover tooltip. Could be hard because of localisation.

## Review items
- [x] Ready for initial review.
- [x] Fix background color on load (maybe use ResizeObserver package)
- [x] Review responsive behaviour.

## Design review
- [x] Ensure spacing between type and objects matches spacing tokens used in design specs under the Structure section below
- [x] Tabs should be offset 16 to the left so the type in the tabs aligns with the Page title
[...]

PR #130, opened 2020-10-07
feat: further heading scroll tweaks

lee-chase:
Relates to  #59

Some minor tweaks to the heading scroll, think I need to move the height stuff into state variables but will do so in a separate PR.

#### Changelog

M       packages/experimental-package/src/components/PageHeader/PageHeader.js
M       packages/experimental-package/src/components/PageHeader/_page-header.scss


davidmenendez (approved):
The code looks good to me and seems to be working. My only suggestion is that due to the ever growing complexity of this component I personally would suggest adding some comments in the code to help give better visibility and guidance into what's going on. i guess it's not strictly necessary, but i think others contributors (myself included) would benefit from having a better understanding at what's going on at a glance.

i will approve, but please consider my feedback. thanks!

PR #132, opened 2020-10-07
feat: header to simpler wrapped buttons

lee-chase:
Part of  #59

Utility components PageAction and ActionBarItem simplified to wrap button. These are restricted UI types specified by the design.

#### Changelog

removed files.

M       packages/experimental-package/src/components/ExampleComponent/ExampleComponent.js
M       packages/experimental-package/src/components/PageHeader/ActionBarItem.js
A       packages/experimental-package/src/components/PageHeader/ActionBarItem.stories.js
A       packages/experimental-package/src/components/PageHeader/PageActionItem.js
A       packages/experimental-package/src/components/PageHeader/PageActionItem.stories.js
M       packages/experimental-package/src/components/PageHeader/PageHeader.stories.js

#### Testing / Reviewing

Verified still working in the storybook. Ran linter and tests.

### DIFF
--- a/packages/experimental-package/src/components/PageHeader/ActionBarItem.js
+++ b/packages/experimental-package/src/components/PageHeader/ActionBarItem.js
@@ -34,13 +34,21 @@ export const ActionBarItem = (props) => {
   );
 };
 
-ActionBarItem.propTypes = {
-  /**
-   * Specifies class(es) to be applied to the top-level PageHeader node.
-   * Optional.
-   */
-  className: PropTypes.string,
-};
-ActionBarItem.defaultProps = {
-  className: '',
-};
+const reservedProps = [
+  'hasIconOnly',
+  'kind',
+  'size',
+  'tooltipPosition',
+  'tooltipAlignment',
+  'type',
+];
+const propTypes = { ...Button.propTypes };
+const defaultProps = { ...Button.defaultProps };
+
+reservedProps.forEach((prop) => {
+  delete propTypes[prop];
+  delete defaultProps[prop];
+});
+
+ActionBarItem.propTypes = { ...propTypes };
+ActionBarItem.defaultProps = { ...defaultProps };
--- a/packages/experimental-package/src/components/PageHeader/ActionBarItem.js
+++ b/packages/experimental-package/src/components/PageHeader/ActionBarItem.js
@@ -6,6 +6,7 @@
 //
 
 import React from 'react';
+import PropTypes from 'prop-types';
 
 import cx from 'classnames';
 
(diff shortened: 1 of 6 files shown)


packages/experimental-package/src/components/PageHeader/ActionBarItem.js
```diff
@@ -6,72 +6,48 @@
...
+  'kind',
+  'size',
+  'tooltipPosition',
+  'tooltipAlignment',
+  'type',
+];
+const propTypes = { ...Button.propTypes };
+const defaultProps = { ...Button.defaultProps };
+
+reservedProps.forEach((prop) => {
```
davidmenendez, line 47: i'm not sure i'm understanding what the intent is here. is this to make sure that certain props passed from the user to `ActionBarItem` don't make it to the underlying `Button` component?

packages/experimental-package/src/components/PageHeader/PageActionItem.js
```diff
@@ -0,0 +1,44 @@
...
+      {children}
+    </Button>
+  );
+};
+
+const reservedProps = ['size', 'type'];
+const propTypes = { ...Button.propTypes };
+const defaultProps = { ...Button.defaultProps };
+
+reservedProps.forEach((prop) => {
```
davidmenendez, line 38: same question as my previous.

packages/experimental-package/src/components/PageHeader/ActionBarItem.js
lee-chase: Yes.

davidmenendez: hmmm interesting. i don't believe i've encountered this pattern before. so you're just copying the same prop types and default props from `Button` to `ActionBarItem` and then removing props that we don't want the user to pass. i'm a little torn on this. i think this is an interesting way to cut down on redundant code, but I think this hurts the optics. if i'm trying to find out what the proptypes are for this component I don't think I'd want to have to go to a completely separate repo to find them. i mean i guess you can find this information from the storybook, but code wise I think it hurts readability.

i think i'd like to see some other peoples opinion on this pattern.

lee-chase: Me either it does have some benefits in that it reduces the components exposure to changes in Carbon. It should even carry on working even if PropTypes and storybook become out of sync with Carbon.

It is frustrating that it's necessary to declare any of the PropTypes, they're only required for storybook.

moores2 (approved):

packages/experimental-package/src/components/PageHeader/ActionBarItem.js
lee-chase: @davidmenendez can we move forward with this and discuss wrapping in #138 ?

PR #136, opened 2020-10-08
chore: restrict linting to exclude built files

lee-chase:
Files built and/or copied by builds should not be linted.

#### Changelog

The CDAI package copies SCSS files rather than compile to css, this results in double linting of these files.

Changed the blob for stylelint and eslint to target only `package/*/src/**` files.

M       package.json

#### Testing / Reviewing

Added and then removed eslint and errors to check both stylelint and eslint function as expected.


SimonFinney (approved):
LGTM! 🎉

PR #141, opened 2020-10-12
Convert enzyme test to RTL for PageHeader

moores2:
Progresses #59

Converts enzyme tests to React Testing Library.

#### Changelog

M       packages/experimental-package/src/components/PageHeader/PageHeader-test.js

**Changed**

- Automated tests


asfordmatt:
@moores2 
it's recommended to use `screen` instead of destructure from `render`, and to use `@testing-library/user-event` instead of `fireEvent`:

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library

moores2:
> @moores2
> it's recommended to use `screen` instead of destructure from `render`, and to use `@testing-library/user-event` instead of `fireEvent`:
> 
> https://kentcdodds.com/blog/common-mistakes-with-react-testing-library

I can see that avoiding destructuring of query functions is goodness, but I'm struggling to see how to avoid destructuring `container`. Is there another way to get access to a top level DOM node so that you can use `querySelectorAll()` etc over the whole of the rendered DOM tree?

asfordmatt:
> I can see that avoiding destructuring of query functions is goodness, but I'm struggling to see how to avoid destructuring `container`. Is there another way to get access to a top level DOM node so that you can use `querySelectorAll()` etc over the whole of the rendered DOM tree?

Well "Using container to query for elements" is on that avoid list :)
The idea with RTL is to test the component the way the user interacts with it, such as by using the "*ByRole" queries. I would suggest writing a user-style test, and then if you really feel the need to, put all the class name checking in a separate test.

moores2:
> > I can see that avoiding destructuring of query functions is goodness, but I'm struggling to see how to avoid destructuring `container`. Is there another way to get access to a top level DOM node so that you can use `querySelectorAll()` etc over the whole of the rendered DOM tree?
> 
> Well "Using container to query for elements" is on that avoid list :)
> The idea with RTL is to test the component the way the user interacts with it, such as by using the "*ByRole" queries. I would suggest writing a user-style test, and then if you really feel the need to, put all the class name checking in a separate test.

  I understand the user interaction focused philosophy of RTL, but my challenge is that the PageHeader is essentially a layout component with negligible user interaction. It shows and hides client content (such as navigation controls and action bars etc.) and re-lays them out as the page scrolls. As such, checking that the right content/containers are present in the DOM and have the correct classes applied is a useful thing to do, and necessary to if we are going to be checking for code coverage of the tests in the future.

  If this component were something like a date picker then I agree that user interaction based testing would be the thing to focus on, but I don't see how to avoid tests that inspect the DOM for a layout component.

moores2:
@asfordmatt I've removed the destructuring of the return from render() and also the use of fireEvent().

lee-chase (approved):
LGTM, perhaps documentation of test best practices would be a useful issue to raise.

PR #145, opened 2020-10-13
Removal modal mdx

davidmenendez:
Adds an MDX file for `RemovalModal` and updates some of the proptypes and documentation

#### Changelog

removed files.

M       packages/experimental-package/src/components/RemovalModal/RemovalModal.js
A       packages/experimental-package/src/components/RemovalModal/RemovalModal.mdx
M       packages/experimental-package/src/components/RemovalModal/RemovalModal.stories.js

**New**

- MDX documentation for `RemovalModal`

### DIFF
--- a/packages/experimental-package/src/components/RemovalModal/RemovalModal.js
+++ b/packages/experimental-package/src/components/RemovalModal/RemovalModal.js
@@ -57,21 +57,65 @@ export const RemovalModal = ({
 };
 
 RemovalModal.propTypes = {
+  /**
+   * The conent to be displayed in the body of the modal
+   */
   body: PropTypes.string.isRequired,
+  /**
+   * Optional classname
+   */
   className: PropTypes.string,
+  /**
+   * The text displayed at the top of the modal
+   */
   header: PropTypes.string.isRequired,
+  /**
+   * Provide a description for "close" icon that can be read by screen readers
+   */
   iconDescription: PropTypes.string,
-  id: PropTypes.string,
+  /**
+   * Message showed when user input fails validation
+   */
   inputInvalidText: PropTypes.string,
+  /**
+   * Label for text box
+   */
   inputLabelText: PropTypes.string,
+  /**
+   * Placeholder for text box
+   */
   inputPlaceholderText: PropTypes.string,
+  /**
+   * Callback function that runs when user closes the modal
+   */
   onRequestClose: PropTypes.func,
+  /**
+   * Callback function that runs when user submits the modal
+   */
   onRequestSubmit: PropTypes.func,
+  /**
+   * Specify whether the Modal is currently open
+   */
   open: PropTypes.bool.isRequired,
+  /**
+   * Specify the text for the primary button
+   */
   primaryButtonText: PropTypes.string,
+  /**
+   * The name of the resource being acted upon
+   */
   resource: PropTypes.string.isRequired,
+  /**
+   * Specify the text for the secondary button
+   */
   secondaryButtonText: PropTypes.string,
+  /**
+   * Specify the text for the subheader
+   */
   subheader: PropTypes.string,
+  /**
+   * Specify whether or not to show the text confirmation input
+   */
   textConfirmation: PropTypes.bool,
 };
 
@@ -82,6 +126,8 @@ RemovalModal.defaultProps = {
   inputInvalidText: '',
   inputLabelText: '',
   inputPlaceholderText: '',
+  onRequestClose: () => {},
+  onRequestSubmit: () => {},
   primaryButtonText: '',
   secondaryButtonText: '',
   subheader: '',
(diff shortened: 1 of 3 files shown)


davidmenendez:
While making these changes I was thinking about the discussion about spreading wrapped carbon component proptypes and defaultprops to the wrapper component. i think in some cases this can be easily achieved and perhaps should be done for sake of convenience, but then for a component like `RemovalModal`, which uses multiple carbon components, I think it becomes more tricky to just spread everything because there may be an overlap in prop names used between components. for instance `Modal` and `TextInput` both can take a classname, so would that class be used for both components? Do we compare every carbon component for duplicate prop names and assign them a custom prop name in the wrapper component?

Another problem becomes the spreading. If you have two carbon components you would still have to manually track all the props and assign them their own object.

```
const RemovalModal = props => {
  const modalProps = {
    className: props.modalClassName,
    body,
    ...
  };
  const inputProps = {
    className: props.inputClassName,
    placeholder,
    ...
  };
  ...
  <Modal {...modalProps}>
    <TextInput {...inputProps} />
    ...
  </Modal>
}
```

Obviously i'm up for more discussion, but I think this is a scenario that we haven't considered. In this particular case it makes sense to me to just manually pass all required props to the carbon components.

dbrugger:
i think the key here is to avoid to overly flatten when composing

lee-chase:
I would not suggest doing this for multiple components just the outermost wrapped component. In the case of remove, the modal dialog and the parts in normally exposes, but not the additional content such as a text box.

davidmenendez:
i guess i just don't see a need to explicitly declare how we want to handle this sort of thing since there's no perfect catch all solution. i think we should leave it to the discretion of the author to decide which is the best method for handling how propTypes and defaultProps are shared from the parent to the children. i think in the long run we are bound to run into issues where carbon updates something and leaves us responsible for updating our code also, so I don't know if it makes sense to use that as a reason not to explicitly write out the props. as long as we are careful about locking in our carbon versions it's something that shouldn't catch us off guard and that we can prepare for.

lee-chase:
The main reason for me is consistency. It makes maintenance easier and builds confidence with users when they dig into the source.

davidmenendez:
i'm willing to try to find a solution that satisfies everyone, but i'm running into an issue with storybook and using spread proptypes. if i implement the following code

```
// RemovalModal.js
...
const modalProps = { ...Modal.propTypes };

RemovalModal.propTypes = {
  ...modalProps,
};

RemovalModal.defaultProps = {
  ...modalProps,
};
```

it doens't seem to work with storybook docs.

Screen Shot 2020-10-13 at 3 31 22 PM

here's that same code but with an additional manually added prop to see what happens

```
const modalProps = { ...Modal.propTypes };

RemovalModal.propTypes = {
  ...modalProps,
  test: PropTypes.bool,
};

RemovalModal.defaultProps = {
  ...modalProps,
  test: true,
};
```

Screen Shot 2020-10-13 at 3 33 54 PM

According to the storybook docs it uses [react-docgen](https://storybook.js.org/docs/react/api/argtypes#automatic-argtype-inference) to generate this table, so i'm looking around for a solution, but if it looks like i'm doing something wrong i'd love to know what i'm doing wrong.

lee-chase:
> i'm willing to try to find a solution that satisfies everyone, but i'm running into an issue with storybook and using spread proptypes. if i implement the following code
> 
> ```
> // RemovalModal.js
> ...
> const modalProps = { ...Modal.propTypes };
> 
> RemovalModal.propTypes = {
>   ...modalProps,
> };
> 
> RemovalModal.defaultProps = {
>   ...modalProps,
> };
> ```
> 
> it doens't seem to work with storybook docs.
> 
> Screen Shot 2020-10-13 at 3 31 22 PM
> 
> here's that same code but with an additional manually added prop to see what happens
> 
> ```
> const modalProps = { ...Modal.propTypes };
> 
> RemovalModal.propTypes = {
>   ...modalProps,
>   test: PropTypes.bool,
> };
> 
> RemovalModal.defaultProps = {
>   ...modalProps,
>   test: true,
> };
> ```
> 
> Screen Shot 2020-10-13 at 3 33 54 PM
> 
> According to the storybook docs it uses [react-docgen](https://storybook.js.org/docs/react/api/argtypes#automatic-argtype-inference) to generate this table, so i'm looking around for a solution, but if it looks like i'm doing something wrong i'd love to know what i'm doing wrong.

That is correct, every property you want to support in storybook currently needs prop-types. As per discussion issue and page header.

This is a storybook issue I believe, not checked to see if one should be raised

davidmenendez:
after searching around for a solution to this i currently am unable to find one. this may just be some weird bug or limitation of react-docgen or storybook at the moment. at the very least i think this update should be good to be merged and we can continue to discuss what we think is the best way to handle propTypes and defaultProps inheritance, although i have also been searching for this type of pattern of inheriting propsTypes and defaultProps from one component to another and i haven't been able to find anything on this topic, which i find strange. if someone can point me towards anything so i can learn more i'd really appreciate it.

lee-chase (approved):
LGTM guessing adding more info to docs is next up

PR #151, opened 2020-10-15
Create New Component or Pattern

lee-chase:
Proposed checkbox list for new components or patterns.

The list could be copied to existing component / pattern creation issues.


davidmenendez (approved):

SimonFinney (approved):
LGTM! 🔥  The only thing I can think of is whether we need to consolidate with / supersede https://github.com/carbon-design-system/ibm-cloud-paks/blob/master/.github/ISSUE_TEMPLATE/feature-request-or-enhancement.md

.github/ISSUE_TEMPLATE/new-component-or-pattern-template.md
```diff
@@ -0,0 +1,31 @@
...
+### Links to other materials
+
+e.g. sketch files, icons, images, prototypes
+
+## Design Maintainer(s)
+
+- Owner/main maintainer(s)
+- Second/backup maintainer(s)
+
+## Tasks
```
matthew-chirgwin, line 20: I agree with all the tasks detailed below. One thing I would like to suggest is a change to the ordering as follows (along with rationale):

All done in the experimental package:

1) Contribute a Readme describing the purpose of the component, and where it is expected to be used. The README should also contain the prop types, with descriptions/help text. This should link to supporting design docs/issues where appropriate to provide context. This should be reviewed (lightly) by engineering and design to make sure it will implement all requirements the component has, and that it will be consumeable.
2) Contribute 'outline' Storybook entries and tests (in RTL) to cover the described behaviour from the Readme. These can be empty initially, with the content provioded as the component is implemented. (The Storybook entry could reference/use the readme from 1) as part of its doc).
3) Implement the stories/test/component (ideally little and often for ease of engineering review). Iterate until complete.
4) When code complete, final design and engineering review. If approved, component promoted to production package.

matthew-chirgwin: I would also suggest some of the steps here should go onto a PR review checklist as well

lee-chase: @matthew-chirgwin  

With regards to 1.  Is the README rendered redundant by the 'ComponentName.mdx' file? If we use prop-types to with DocGen the API gets documented for us.

See https://ibm-cloud-paks.netlify.app/?path=/docs/experimental-pageheader--all-attributes-set which is produced by the `PageHeader.md`x and DocGen. 

So instead of writing a README an MDX file plus a  shell component with prop-types could suffice.

matthew-chirgwin: Works for me - main thing is the purpose/context/usage is detailed somewhere 👍 (and thus can be discussed!)

lee-chase:
> LGTM! 🔥 The only thing I can think of is whether we need to consolidate with / supersede https://github.com/carbon-design-system/ibm-cloud-paks/blob/master/.github/ISSUE_TEMPLATE/feature-request-or-enhancement.md

I'm tempted to say the list would be less onerous, or perhaps tweaked for feature-request-or-enhancement. We could perhaps come up with wording that suits both.

Issue #54, opened 2020-08-05
Empty states


carrenelloyd:
Catalog # 7

[Link to Design ](https://pages.github.ibm.com/cdai-design/pal/patterns/empty-state/usage)

**Maintainer**
Vikki Paterson

**Dev Maintainer**
@matthewgallo 

https://github.com/carbon-design-system/ibm-cloud-paks/projects/2#card-45884682

[View empty states implementation](https://ibm-cloud-cognitive.netlify.app/?path=/story/experimental-emptystate--default)
## Tasks
- [x] Initial review of design/existing code

### Working in experimental package
- [x] Create a shell component which
  - [x] Create ComponentName/ComponentName.js component file that renders props in raw form (no styling or sub components, plain text OK)
  - [x] Adds prop-types with comments for DocGen to ComponentName/ComponentName.js
  - [x] Add Doc file e.g. ComponentName.mdx with a brief description of the component.
- [x] First engineering review/playback
### Continuing in experimental package
- [x] Storybook - stories to cover design scenarios, story with long, min and
      max content, etc
- [x] Doc tab in storybook derived from design docs with embedded stories where
      appropriate
- [x] Code to include prop-types with comments for automated DocGen
      documentation
- [x] Tests  (written in React Test Library) to cover behavior from the documented in the components mdx file.
### Review and promote
- [ ] Design review
- [ ] Add design review issues addressed, issues raised or discounted.
- [ ] Engineering review - Submit PR to promote component to production package
- [ ] All engineering issues addressed, issues raised or discounted.

lee-chase:
@vikkipaterson can you comment on the status of the design it was marked as 'status: refining' by @carrenelloyd

lee-chase:
Dev thoughts
This appears to be a redesign of https://pages.github.ibm.com/cloud-integration-design/ide-components/?path=/story/production-ready-ide-empty-state-component--default

@vikkipaterson I am a bit confused by the design. When describing sizes it shows a flow of image, title, description with optional CTA button below. Then when it moves to describe empty state types it shows sub-text, title, image.

Lift and shift with some styling tweaks. 

Clarify with @vikkipaterson whether a single call to action is permissible. The current implementation allows for an array of links https://pages.github.ibm.com/cloud-integration-design/ide-components/?path=/story/production-ready-ide-empty-state-component--default

vikkipaterson:
[...]

PR #159, opened 2020-10-22
PR for #54: Empty states

matthewgallo:
Contributes to #54 

This adds a new `<EmptyState />` component to the experimental package. I used the [empty state page in the CD&AI PAL](https://pages.github.ibm.com/cdai-design/pal/patterns/empty-state/usage/) site as design reference and have reviewed some of this work with Chloe Poulter and @vikkipaterson.

#### Changelog

**New**

`/packages/experimental/src/components/EmptyState/EmptyStateIllustrations`
`/packages/experimental/src/components/EmptyState/_empty-state.scss`
`/packages/experimental/src/components/EmptyState/_index.scss`
`/packages/experimental/src/components/EmptyState/EmptyState-test.js`
`/packages/experimental/src/components/EmptyState/EmptyState.js`
`/packages/experimental/src/components/EmptyState/EmptyState.mdx`
`/packages/experimental/src/components/EmptyState/EmptyState.stories.js`
`/packages/experimental/src/components/EmptyState/index.js`

**Changed**

`/packages/experimental/src/components/_index.scss` --> added new component styles

#### Testing / Reviewing

You can run `yarn test` to see the new test I wrote for this component. As well, as reviewing the new storybook component `EmptyState`

image
image

Just let me know if I missed something with the contribution process as this is my first PR, thanks!


packages/experimental/src/components/EmptyState/EmptyState.js
```diff
@@ -0,0 +1,97 @@
...
+      )}
+    </div>
+  );
+};
+
+EmptyState.propTypes = {
+  /**
+   * Empty state action button
+   */
+  action: PropTypes.shape({
```
davidmenendez, line 57: I'm curious about your choice to pass an object instead of passing the attributes separately. I guess it's more of a personal preference, but why do that instead of something like `actionText` `actionType` etc? I personally think it's more clear to the consumer and less prone to bugs to pass props individually instead of as an object. I don't know if anyone else wants to comment on this, but I think it would be interesting to discuss.

matthewgallo: Yeah, I agree that it would probably be clearer. At first I didn't know if there were going to be multiple actions but I think the design will only ever call for one so I can change this

matthewgallo: Fixed [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/159/commits/8a7e632d37f2f5232d0350324b01f07443480a44)

davidmenendez (approved):
LGTM. This definitely needs additional tests added, but I think that can wait for a future PR. I don't believe we've talked formally about adding test coverage, but I think it's useful and a practice we should look to implement in the future.

CptPilot:
This appears to be missing the link option which is available in the component this is intended to supersede and is also referenced in the pattern on the Carbon website:

https://www.carbondesignsystem.com/patterns/empty-states-pattern/

Was that intentional?

lee-chase (changes requested):
Have a few questions which I figured I'd stick in a request changes.

1. Could we find a way to switch the illustration theme based on the current theme? Ideally, the icon would be themeable. Perhaps this is something we could bat back to the design maintainer.
2. I may be wrong but I believe events should be 'onXXX' e.g 'connection'
3. The default illustration should be listed as 'No Data'
4. CTA Buttons - suggests more than one button. If following the pattern of the page header we should create a CtaButton component and accept an array of them.

lee-chase:
> This appears to be missing the link option which is available in the component this is intended to supersede and is also referenced in the pattern on the Carbon website:
> 
> https://www.carbondesignsystem.com/patterns/empty-states-pattern/
> 
> Was that intentional?

The design does not call for one, perhaps a question for @vikkipaterson along with the theme question above.

matthewgallo:
@lee-chase I spoke with Chloe Poulter and she said that there is only ever supposed to be one action button for this component

lee-chase:
> @lee-chase I spoke with Chloe Poulter and she said that there is only ever supposed to be one action button for this component

Can we ask Chloe to comment here and ideally update the design page to indicate a single button.

chloepoulterdesign:
Shoot. I didn't check the carbon pattern - some part of me thought they had adopted the Cloud & Cognitive one. 
Technically, our job is to componentise the cloud & cognitive empty state - and I've never seen anyone implement the extra link - BUT for maximum adoption we should add that link. 

Is it possible to get it added as an option? 
So sorry I didn't spot that when you got in touch.

lee-chase:
@chloepoulterdesign could you also comment on whether the illustration is themed independently of the Carbon theme and the design naming a section 'CTA Buttons' noting the plural. @matthewgallo indicated that this should be 'CTA Button'

packages/experimental/src/components/EmptyState/EmptyState.js
```diff
@@ -0,0 +1,107 @@
...
+  illustration,
+  illustrationSize,
+  illustrationTheme,
+  onActionEvent,
+}) => {
+  return (
+    <div className={`${expPrefix}-empty-state`}>
+      {illustration && typeof illustration === 'string' && (
+        <img
+          src={require(`./EmptyStateIllustrations/${illustrationTheme}/${illustration}.svg`)}
```
lee-chase, line 29: Please remove use of require.

lee-chase (changes requested):

chloepoulterdesign:
I am 99% sure it is one optional button and one optional link.

There are 6 out of the box illustrations provided [here](https://pages.github.ibm.com/cdai-design/pal/patterns/empty-state/usage) which a designer can choose from, *or* design one of their own, *or according to Carbon*, have *no* image.

lee-chase:
@chloepoulterdesign are the icons intended to be themed by; Carbon theme, or come in light/dark options or are there effectively 12 icons?

packages/experimental/src/components/EmptyState/EmptyState.js
matthewgallo: Hey Lee, so I did it this way to prevent having to import all 12 svgs in the component and only ever using 1 of them. If I remove this line, we will have to add 12 import statements for each svg version to the top of this component. And we'll also end up with a bunch of eslint errors such as...
```
'noDataSvg' is declared but its value is never read.ts(6133)
'noDataSvg' is defined but never used. Allowed unused vars must match /^_/u.eslintno-unused-vars
```
Is it ok to keep this or are you thinking about another way?

matthewgallo:
@chloepoulterdesign, @lee-chase and @CptPilot, I have added the optional link as discussed above ^

packages/experimental/src/components/EmptyState/EmptyState.js
lee-chase: I hadn't realised dynamic imports were not a thing yet https://github.com/tc39/proposal-dynamic-import

While we could put a placeholder image in a folder and strip off its name and construct the path that way it would seem to be a bit excessive.

Leave as is and if there is a better pattern found it can be added later.

chloepoulterdesign:
It looks like they are themed @lee-chase so I would expect it to be carbon themes. Will ping this to gregor to confirm

lee-chase:
** Slack conversation Oct 23rd 2020 **

Lee Chase  1:16 PM
Figured I’d try and close the loop on this one.
Looking at the PR the outstanding questions are:
Number of buttons. It would be nice to be 100% sure Chloe. I’d take a PR to the pal site removing the S to indicate that. It sounds trivial but it has the potential to require structurally different HTML
Are there 12 standard illustrations or 6 with dark and light as an illustration theme switch.
If 2 = 6then does every custom illustration have to come in dark and light varieties?
My personal preference is for 2  to be 12 and we have 12 options for the user to choose from or they provide their own. Which means the illustrationThem switch can be removed.
With regards to 1 we can move forward with a single button but I do not believe we can close out a final dev review process without a definitive answer. (edited) 






Chloe Poulter:speech_balloon:  1:20 PM
Yeah absolutely happy for the documentation to be changed.
I've reached out to David Levinson because Gregor is out. I suspect it's carbon themed.
That's a whole new world designers are going to have to cope with in future - yes, they will be expected to provide 2 illustrations wherever the same one will not work in both.
The key is that we need to support the theme switcher pattern either now or in future - as UIs will all be able to switch between light and dark by user preference soon
1:22
confirmation from David
image.png 
image.png



Matt Gallo  1:22 PM
Right, and to just add to what Chloe said...each illustration has a dark and light version. The developer using this component just has to know their app's own current theme value and then pass the illustration theme prop like so
illustrationTheme={appTheme === 'carbon--g100' ? dark : light}

Lee Chase  1:26 PM
The component itself cannot, with certainty, detect which theme. So I think we are stuck with a manual switch. It’s not that it is impossible, just cumbersome and flaky.
:+1:
1


Matt Gallo  1:26 PM
Right, I gave it a default prop of light

Lee Chase  1:26 PM
Will you submit a PR to update the documentation Chloe?
1:27
So I think you should have a code example in the docs noting ‘g100’ and ‘g90’ as dark themes.
1:27
Sorry @mattgallo last line was for you.

Chloe Poulter:speech_balloon:  1:28 PM
cool

Matt Gallo  1:28 PM
Do you mean, just saying that if you are using carbon-g100 or carbon-g90 just to be sure to use the dark themed illustrations?

Lee Chase  1:29 PM
Not sure I would say be sure, but just explain how you might switch illustration theme.

Matt Gallo  1:30 PM
oh ok, kinda like that example I posted above?

Lee Chase  1:31 PM
Yes. One more thing, you need to accept a custom image from the user. Or it would seem a pair light and dark.
1:33
Any objections to me copying the above thread over to the github issue?

Matt Gallo  1:33 PM
Not at all
New
1:34
Well, if we allow the user to set their own custom illustration I think they're responsible for handling the logic of passing a light/dark version based on the current theme of their app. Similar to the illustrationTheme prop above

Lee Chase  1:35 PM
Again an example / doc in storybook should be added.
:+1:
1


Matt Gallo  1:35 PM
customIllustration={appTheme === 'carbon-g100' ? myDarkThemeIllustration : myLightThemeIllustration}

Lee Chase  1:36 PM
I would re-use the illustration property, but that’s just my personal preference.
1:36
@Chloe OK to copy to github?

Chloe Poulter:speech_balloon:  1:36 PM
oh! yes, go ahead!

lee-chase (approved):

PR #162, opened 2020-10-27
feat: add saving component

asfordmatt:
- Add Saving component from CDAI
- move some dependencies to devDependencies as per recent discussion


vladbalanescu (approved):
Awesome, looking good to me 👍

Issue #30, opened 2020-07-28
Import


SimonFinney:
[**Link to design**](https://www.carbondesignsystem.com/community/patterns/import-pattern/)

Maintainers: [Vikki Paterson](https://github.com/vikkipaterson) , [Chloe Poulter](https://github.com/chloepoulterdesign)

Dev Maintainer: @davidmenendez 

Catalog number: 12

## Tasks

- [x] Initial review of design/existing code

## Release reviewer(s)

- @matthewgallo 
- @sydrosa 

### Working in experimental package

- [x] Create a shell component which
  - [x] Create ComponentName/ComponentName.js component file that renders props
        in raw form (no styling or sub components, plain text OK)
  - [x] Adds prop-types with comments for DocGen to
        ComponentName/ComponentName.js
  - [x] Add Doc file e.g. ComponentName.mdx with a brief description of the
        component.
- [x] First engineering review/playback

### Continuing in experimental package

- [x] Storybook - stories to cover design scenarios, story with long, min and
      max content, etc
- [x] Doc tab in storybook derived from design docs with embedded stories where
      appropriate
- [x] Code to include prop-types with comments for automated DocGen
      documentation
- [x] Tests (written in React Test Library) to cover behavior from the
      documented in the components mdx file.

### Review and promote

- [x] Design review
- [x] Add design review issues addressed, issues raised or discounted.
- [x] Engineering review - Submit PR to promote component to production package
- [x] All engineering issues addressed, issues raised or discounted.

carrenelloyd:
[**Link to design**](https://www.carbondesignsystem.com/community/patterns/import-pattern/)

Maintainers: [Vikki Paterson](https://github.com/vikkipaterson) , [Chloe Poulter](https://github.com/chloepoulterdesign)

Catalog number: 12

SimonFinney:
@carrenelloyd Would you know if there's redlines or existing code available for this?

SimonFinney:
Existing code - https://pages.github.ibm.com/cloud-integration-design/ide-components/?path=/story/production-ready-ide-importing-importing--importing-multiple-files

lee-chase:
Initial Dev review.

@vikkipaterson @chloepoulterdesign this looks very much like this with various container options and action buttons.

https://pages.github.ibm.com/cloud-integration-design/ide-components/?path=/story/production-ready-ide-importing-uploading--extension-validation

[...]

PR #163, opened 2020-10-28
feat: Import component

davidmenendez:
Contributes to #30 

implements the import modal

currently a work in progress.


davidmenendez:
you can test the import file functionality with this image api https://picsum.photos/

vladbalanescu (approved):
LGTM

lee-chase (commented):
Does this only allow the uploading of one file at a time @davidmenendez?

davidmenendez:
@lee-chase it does because it wasn't a clear requirement for me. I can include that though if that is a requirement.

davidmenendez:
I will rework this to include multiple files, but just for transparency I wanted to give my thoughts on why I think that there is conflicting design surrounding this. 

yes, there is some wording in the designs that indicate that multiple files could be added, but there's a lot of ux that indicates that that's not the case. if you look at the design (https://www.carbondesignsystem.com/community/patterns/import-pattern/) and image with the text underneath that reads "Example of a loaded file in the import modal" the drag and drop field as well as the text input and add file button are all in a disabled state once a file has been uploaded. it's the same thing with the "Example of a file imported by URL" image. it's also strangely enough the same thing in "Example of files loaded in the import dialog". even though there are multiple files being uploaded the inputs are still disabled. The only time you don't see this interaction is under "Import from a URL (full page or side-panel)". With that all being said i guess i'll just reach out to the designers to discuss what exactly is the intended ux for this component because based on what i'm seeing i don't think it's clear what's supposed to be intended behavior.

davidmenendez:
component has been updated to support multiple files. please review new changes.

davidmenendez:
Added an additional commit to fix the failing tests. it involves commenting out some code in the `setupFilesAfterEnv.js` config. i talked to Josh Black and there is a known issue in the latest version of carbon components that's causing tests with certain carbon components to fail. his suggestion was just to stub the offending `console.log` function. i assume we'll be able to revert that change once Josh has solved the issue on carbons end.

packages/experimental/src/components/ImportModal/ImportModal.js
```diff
@@ -0,0 +1,316 @@
...
+      onRequestClose={onRequestClose}
+      className={`${expPrefix}--import-modal`}>
+      <div className={`${expPrefix}--import-modal-inner`}>
+        <p className={`${expPrefix}--import-modal-body`}>{modalBody}</p>
+        <p className={`${expPrefix}--import-modal-label`}>{fileDropHeader}</p>
+        <FileUploaderDropContainer
+          accept={validFileTypes}
+          labelText={fileDropLabel}
+          onAddFiles={onAddFile}
+          multiple
```
lee-chase, line 151: Should be a boolean property of the modal for added flexibility. Otherwise looks good to me.

lee-chase (approved):

Issue #12, opened 2020-07-01
Publish @carbon/ibm-cloud-paks npm module


dbrugger:

SimonFinney:
Acceptance criteria to be defined in Slack

lee-chase:
Publishing/modifying npm packages is restricted to owners. There is only one level of access.

https://docs.npmjs.com/cli/owner

A list of owners for carbon-components-react, for instance, is as follows. It is of note that they have a carbon-bot email address which I suspect is for automated publishing. Confirmation will be sort from @joshblack 

$npm owner ls carbon-components-react
abbeyhrt <Abagail.Hart@ibm.com>
adavila <adavila91@gmail.com>
carbon-bot <carbondesignsystem@gmail.com>
dakahn <da.anguiano@ibm.com>
emyarod <afw5059@gmail.com>
jnm2377 <josefinanoemi4@gmail.com>
joshblack <joshblack@us.ibm.com>
tjegan <tj.egan@icloud.com>
vpicone <vp@vincepic.one>

lee-chase:
Without an automated build (as is the case for @carbon/vue) 'lerna publish' can be used to manage the publication of multiple packages from a repo.

https://www.npmjs.com/package/@lerna/publish

joshblack:
Let me know if we can help out with this, if publishing from CI we can provide an NPM_TOKEN for the carbon-bot.

Similarly, we can create a group of users under the `@carbon` scope and give them access to the relevant packages.

Lerna is a great tool for publishing, especially with the conventional commit flag. Let me know if you want to chat about it at all!

SimonFinney:
For the first release, I'd be happy with enough configuration that enables us to automate the publishing of a package and tag based on conventional commits.

joshblack:
@SimonFinney very nice, `lerna version` has a `--conventional-commits` option for this: https://github.com/lerna/lerna/tree/master/commands/version#--conventional-commits

Similarly, you can use `--conventional-prerelease` and `--conventional-graduate` for a prerelease workflow.

It should help generate relevant tags and release changelogs (and CHANGELOG.md files) too!

lee-chase:
There's also the option to use `lerna publish`, which includes a call to `lerna version` from the command line to publish to npm.

SimonFinney:
[...]

PR #164, opened 2020-10-29
build(release): add release and publish step

SimonFinney:
Contributes to #12 

#### Description

@carbon-design-system/ibm-cloud-cognitive-development - Draft PR to discuss the release and publishing process. For the short-term, I'm proposing we publish and release on each merge to `master` and figure out any additional channels, configuration, and steps required as they arise in the future.

#### Changelog

**New**

Add draft release GitHub Action workflow similar to https://github.com/carbon-design-system/carbon/blob/master/.github/workflows/release.yml


.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - 12/publish-step
```
SimonFinney, line 6: This will be swapped out for `master` commented out below before merged

.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Install
+        run: yarn --offline
+
+      - name: Build
+        run: yarn build
+
+      - name: Continuous integration check
+        run: yarn ci-check
+
+      # Dry run - `yarn lerna version --conventional-commits --no-changelog --no-git-tag-version --no-push --yes`
```
SimonFinney, line 33: This will simulate what happens when the publish command below is run, without tagging and pushing the committed changes

.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: echo 'publish' # `yarn lerna publish --conventional-commits --no-changelog --yes`
```
SimonFinney, line 37: This will be swapped out for the commented-out command before merging - it does everything the dry run above does and additionally publishes the packages to npm

.github/workflows/publish.yml
```diff
@@ -0,0 +1,38 @@
+name: Publish
+
+on:
+  push:
+    branches:
+      - 12/publish-step
```
SimonFinney, line 6: This will be swapped out for master commented out below before merging

.github/workflows/publish.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Install
+        run: yarn --offline
+
+      - name: Build
+        run: yarn build
+
+      - name: Continuous integration check
+        run: yarn ci-check
+
+      # Dry run - `yarn lerna version --conventional-commits --no-changelog --no-git-tag-version --no-push --yes`
```
SimonFinney, line 33: We can use this locally to simulate what happens when the publish command below is run, without tagging and pushing the committed changes

.github/workflows/publish.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: echo 'publish' # `yarn lerna publish --conventional-commits --no-changelog --yes`
```
SimonFinney, line 37: This will be swapped out for the commented-out command before merging - it does everything the dry run above does and additionally publishes the packages to npm

.github/workflows/publish.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Build
+        run: yarn build
+
+      - name: Continuous integration check
+        run: yarn ci-check
+
+      # Dry run - `yarn lerna version --conventional-commits --no-changelog --no-git-tag-version --no-push --yes`
+      - name: Publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
SimonFinney, line 36: We need to get the correct npm access token and add it to https://github.com/carbon-design-system/ibm-cloud-cognitive/settings/secrets before merging

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Continuous integration check
+        run: yarn ci-check
+
+      - name: Release
+
+        # uses: actions/create-release@latest # https://github.com/actions/create-release
+
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: echo 'release'
```
SimonFinney, line 36: These will be swapped out for the commented-out code in this step before merging - it will create the release in GitHub based on the tag pushed from https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/164/files#r514282150

lee-chase (commented):
There are steps which may well relate to https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/48 that we also need to address.

1. Creation of a changelog. While there are automated tools that can help https://www.npmjs.com/package/auto-changelog simply using the commit history can be problematic without a formal PR/commit comment/name process. As a result a human created changelog can be a bit more penetrable to the average user.
2. Will package.json files be updated to reflect the version just published
3. Lerna can automatically throw a tag, will we be using this or a manual process?

SimonFinney:
- For 1, if we want changelogs per package to be generated and updated with Lerna, we can remove the `--no-changelog` flag. Lerna doesn't have an option for rolling up all package changes and generating release notes, so it would either require manually inputting for now (although I can't see this being an issue just yet given we have no consumers 😆 ) or in the future writing a script (similar to https://github.com/carbon-design-system/carbon/blob/master/packages/cli/src/changelog.js) as part of `release.yml` to do this for us - I'm also all for getting into a practice of paying closer attention to merge / squash commit messages to ensure these can be better utilized as a part of generating these notes
- For 2 and 3, as I understand it, https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/164/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7R37 will bump, tag, and push - I'm proposing for this to be automated on merges to `master` for now, but if we want more control later down the line, we could opt for a manual process

lee-chase (approved):
LGTM

PR #186, opened 2020-11-04
feat: ide-remove

asfordmatt:
Moved IdeRemove component from CDAI


packages/cdai/src/components/IdeRemove/IdeRemove.feature
```diff
@@ -0,0 +1,25 @@
+Feature: IDE Remove component
```
vladbalanescu, line 1: Do we not want a licence header here 🤔 ?

vladbalanescu (approved):
LGTM 👍

PR #191, opened 2020-11-06
chore: update example block class

lee-chase:
A minor tweak to the example component to use variables for the block class.

#### Changelog

M       packages/experimental/src/components/ExampleComponent/ExampleComponent.js
M       packages/experimental/src/components/ExampleComponent/_example-component.scss


vladbalanescu (approved):
LGTM 👍

Issue #183, opened 2020-11-02
Tearsheet


moores2:
# Tearsheet component

Dev maintainer: Dave Clark (@dcwarwick)

## Brief description

The Tearsheet component facilitates rich modal interactions such as resource creations that require multiple steps. It is akin to a dialog that is anchored to the bottom of the viewport. They come in 2 sizes, narrow and wide, each with specific layouts. They can be stacked, facilitating nested interactions. 

![image](https://user-images.githubusercontent.com/6977424/97885467-4f815600-1d1f-11eb-85e3-699446392d78.png)

## Details

[Link to INTERIM Design](https://ibm.ent.box.com/notes/713596972705?s=0fvfln65s2oh0mqrducnqknz29jead23)

### Links to other materials

The Tearsheet is a pre-req for issue #52. 

## Design Maintainer(s)

- Owner/main maintainer(s): Mitchell Bernstein 
- Second/backup maintainer(s)

## Tasks

- [x] Initial review of design/existing code

### Working in experimental package

- [ ] Create a shell component which
  - [x] Create ComponentName/ComponentName.js component file that renders props
        in raw form (no styling or sub components, plain text OK)
  - [x] Adds prop-types with comments for DocGen to
        ComponentName/ComponentName.js
  - [x] Add Doc file e.g. ComponentName.mdx with a brief description of the
        component.
- [ ] First engineering review/playback

### Continuing in experimental package

- [ ] Storybook - stories to cover design scenarios, story with long, min and
      max content, story to demonstrate right-to-left support, etc.
- [x] Doc tab in storybook derived from design docs with embedded stories where
      appropriate
- [x] Code to include prop-types with comments for automated DocGen
      documentation
- [ ] Tests (written in React Test Library) to cover behavior from the
      documented in the components mdx file.

### Review and promote

- [ ] Design review
- [ ] Add design review issues addressed, issues raised or discounted.
- [ ] Engineering review - Submit PR to promote component to production package
- [ ] All engineering issues addressed, issues raised or discounted.

mitchellbernstein:
Pattern guidance for tearsheet published: https://pages.github.ibm.com/cdai-design/pal/components/tearsheet/usage

gentryb:
@lee-chase  CPD (on-prem) is scoping priorities for 4.0 release. Several experience are looking to utilize the tearsheet component. Do you have a ball park time-line for when it will be completed?

moores2:
[...]

PR #193, opened 2020-11-09
Add wide version of Tearsheet

moores2:
Contributes to #183 

Initial version of the wide Tearsheet

#### Changelog

A       packages/experimental/src/components/Tearsheet/Tearsheet.js
A       packages/experimental/src/components/Tearsheet/Tearsheet.stories.js
A       packages/experimental/src/components/Tearsheet/TearsheetShell.js
A       packages/experimental/src/components/Tearsheet/TearsheetShell.stories.js
A       packages/experimental/src/components/Tearsheet/_index.scss
A       packages/experimental/src/components/Tearsheet/_storybook-styles.scss
A       packages/experimental/src/components/Tearsheet/_tearsheet.scss
M       packages/experimental/src/components/_index.scss

**New**

- Wide Tearsheet


packages/experimental/src/components/Tearsheet/Tearsheet.js
```diff
@@ -0,0 +1,170 @@
...
+  className: PropTypes.string,
+  /**
+   * The description for the close icon. Optional.
+   */
+  closeIconDescription: PropTypes.string,
+  /**
+   * Specifies the description of the Tearsheet.
+   * Optional.
+   */
+  description: PropTypes.node,
```
lee-chase, line 106: Why is this not PropTypes.string?

lee-chase (approved):
1. Should the ghost button take up the full space available?
2. The tabs appear to be offset to the left by 16pxish
image
3. It appears to be 8, 10 and 10 pixels too short
image
4. Probably asked before but can we get storybook actions from the tabs

packages/experimental/src/components/Tearsheet/Tearsheet.js
moores2: So, that rich content can be provided if needed. It is standard practice for the ModalHeader etc to allow nodes for label, title etc, so is consistent with that.

moores2:
> 1. Should the ghost button take up the full space available?
It only appears to. It's actually 25% of the width as per the other buttons and the spec. However, being a ghost the size is only apparent on hover.
> 2. The tabs appear to be offset to the left by 16pxish
>    image
Will fix.
> 3. It appears to be 8, 10 and 10 pixels too short
>    image
The heights of the individual elements in the style spec do not add up to the total shown. Since it is interim this will be deferred until it becomes final and we can review it properly.
> 4. Probably asked before but can we get storybook actions from the tabs
Will investigate.

Issue #56, opened 2020-08-05
Generate an API key


carrenelloyd:
Catalog # 10

[Link to Design ](https://www.carbondesignsystem.com/community/patterns/generate-an-api-key/)

**Maintainers**
Vikki Paterson, Chloe Poulter

[Working code](https://pages.github.ibm.com/cloud-integration-design/ide-components/?path=/story/production-ready-ide-api-key-generation-component--with-core-configuration)

lee-chase:
@vikkipaterson can you comment as to the status of the design it was marked as 'status: refining' by @carrenelloyd

davidmenendez:
@vikkipaterson @chloepoulterdesign this component is now ready for design review https://ibm-cloud-cognitive.netlify.app/?path=/story/experimental-apikeymodal--standard

chloepoulterdesign:
Thanks David!

In `API Key modal > standard`:
1. I think the verb on the button should be `Generate` - ideally `Generate API key` rather than the current `Add API key`
2. I believe that while text in a modal should have a 20% margin on the right hand side, fields should be full-width.
3. The helper text which reads `Providing the application name will help you idenfity your api key later.` should have less padding above it (it should be closer to the field) and API should be capitalised. 
3. The "generating" state isn't quite right - when a user has filled in the name and clicks "Generate API Key", both the "Generate API Key" button and the "Application name" field should have a disabled-state - so that the user can't click the button or change the contents of the field while the key is being generated
4. Have you taken any direction on the keyboard tabbing order? I think as a keyboard user, when the modal opens, I would expect the "application name" field to have focus - so i can start typing immediately. Then, when I'm done typing, I would want to hit "enter" to activate the "generate an API key" button. Alternatively, if I hit "tab", I would expect focus to be applied to the "Generate an API key" button so that I could activate it next.
5. While the key is generating, I think the loading text with the spinner should just say "Generating..."
6. Once the key is generated, I think the primary button text needs to be "Copy & close" per some complaints about its' current text in [Slack](https://ibmapim.slack.com/archives/C0M053VPT/p1599658565344400?thread_ts=1599589238.319800&cid=C0M053VPT)

In `API Key modal > minimal`:
1. The button to open the modal is missing
2. There is no loading state in this option - what if I click the button to launch the modal, but it takes a moment to generate and then opens? I think we need to include the "Generate API key" button, which becomes a "Generating" message, and then opens the modal.
[...]

PR #200, opened 2020-11-10
feat: apikey modal

davidmenendez:
Contributes to #56 

adds the api key modal

#### Changelog

A       packages/experimental/src/components/APIKeyModal/APIKeyDownloader.js
A       packages/experimental/src/components/APIKeyModal/APIKeyModal.js
A       packages/experimental/src/components/APIKeyModal/APIKeyModal.mdx
A       packages/experimental/src/components/APIKeyModal/APIKeyModal.stories.js
A       packages/experimental/src/components/APIKeyModal/_api-key-modal.scss
A       packages/experimental/src/components/APIKeyModal/_index.scss
A       packages/experimental/src/components/APIKeyModal/_storybook-styles.scss
A       packages/experimental/src/components/APIKeyModal/index.js


davidmenendez:
Ready for review. I know the original issues also shows a multi-step design, but I think that particular design warrants it's own consideration and possibly it's own component. I'll be taking a look at that particular design in the next iteration after this is merged in.

packages/experimental/src/components/APIKeyModal/APIKeyModal.stories.js
```diff
@@ -0,0 +1,121 @@
...
+/* eslint-disable react/prop-types */
+const ModalStateManager = ({
+  renderLauncher: LauncherContent,
+  children: ModalContent,
+}) => {
+  const [open, setOpen] = useState(false);
+  const [apiKey, setApiKey] = useState('');
+  const [loading, setLoading] = useState(false);
+  const setKeyHandler = async (name) => {
+    console.log(`some api call for ${name}`);
```
lee-chase, line 58: Remove console log

packages/experimental/src/components/APIKeyModal/APIKeyModal.js
```diff
@@ -0,0 +1,236 @@
...
+   */
+  successBody: PropTypes.node,
+  /**
+   * modal header for a successful api key creation
+   */
+  successHeader: PropTypes.string,
+  /**
+   * specifies if the api key input should use the password type toggle
+   */
+  visibilityToggler: PropTypes.bool,
```
lee-chase, line 207: A poor variable name implies a method to be called.
```suggestion
 aipKeyVisibility: PropTypes.bool,
```

lee-chase (changes requested):
- What is the expected behaviour on clicking copy? There is no feedback on clicking copy, please check back with the designer on this issue.
- The multi-step behaviour should probably follow a similar pattern to that of TearSheets please discuss with @moores2 
- In the 'standard' story none of the controls are active
- When viewing code in the Docs tab the functionality of the `<ModalStatemanager>` hides valuable information from the user. Both this and the previous issue would be solved by adding the functionality of the ModalStateManager into the template.
- 
- Can we stop the scroll bar jumping up and down when it has the loading animation.

davidmenendez:
> When viewing code in the Docs tab the functionality of the <ModalStatemanager> hides valuable information from the user. Both this and the previous issue would be solved by adding the functionality of the ModalStateManager into the template.

to give some context- the reason for this configuration is because the way that the modal is rendered it needs to be able to have access to some kind of state to toggle it being open or not. the reason for this is because if it didn't when you view the modal in the docs section it just appears and there's no way to close it. the implementation i've done is based on how carbon has setup stories around modals. i can try to find a better solution to this, but in the event you're solution doesn't work and I can't find another solution to get the carbon modal to play nice with storybook the alternative solution would be to not put the modal in the docs section, which is the entire reason for having to build a state manager wrapper.

packages/experimental/src/components/APIKeyModal/APIKeyModal.stories.js
```diff
@@ -0,0 +1,121 @@
...
+    const timeout = () => new Promise((resolve) => setTimeout(resolve, 2000));
+    await timeout();
+    setApiKey('111-111-111-111');
+    setLoading(false);
+  };
+  return (
+    <>
+      {!ModalContent || typeof document === 'undefined'
+        ? null
+        : ReactDOM.createPortal(
```
lee-chase, line 69: Using a portal should not be necessary to make the modal appear over other content.

davidmenendez: the portal is necessary for the modal component to show up properly in the docs section. if you don't it causes issues with the way the modal is rendered due to styling take a look here.

Screen Shot 2020-11-17 at 2 25 08 PM


this is also how the modal should be rendered according to carbon and generally how react recommends rendering a modal component. this code should be more realistic in terms of how it's most likely to actually be consumed. i was able to rework the way the modal story code works though so now it should be able to use controls.

davidmenendez:
@lee-chase 

> What is the expected behaviour on clicking copy? There is no feedback on clicking copy, please check back with the designer on this issue.

I've contacted the designers on slack and i'm awaiting their response.

> When viewing code in the Docs tab the functionality of the <ModalStatemanager> hides valuable information from the user. Both this and the previous issue would be solved by adding the functionality of the ModalStateManager into the template.

I'm not sure what information is being hidden could you please elaborate. I did do some reworking on the story and the modal state wrapper so that the controls should work. As I explained above getting the modal to work with the docs section of storybook is tricky because it forces us to introduce state to the story in order to be able to show and hide the modal. also in order to properly show the "flow" of creating a key in the standard story we're forced to use state to establish a loading a state. I could remove this code, but then there's no way to have the user walk through create => wait => result
```
const setKeyHandler = async (resourceName) => {
  setLoading(true);
  const timeout = () => new Promise((resolve) => setTimeout(resolve, 2000));
  await timeout();
  setApiKey('111-111-111-111');
  setLoading(false);
};
...
<APIKeyModal
  {...args}
  apiKey={apiKey}
  loading={loading}
  onRequestClose={onCloseHandler}
  onRequestSubmit={setKeyHandler}
  open={open}
/>
```

packages/experimental/src/components/APIKeyModal/APIKeyModal.stories.js
lee-chase: @davidmenendez the narrow tearsheet, produced by @moores2, without the use of a portal. 

Can you swap notes and figure out the best route forward.

lee-chase:
> @lee-chase
> 
> > What is the expected behaviour on clicking copy? There is no feedback on clicking copy, please check back with the designer on this issue.
> 
> I've contacted the designers on slack and i'm awaiting their response.
> 
> > When viewing code in the Docs tab the functionality of the  hides valuable information from the user. Both this and the previous issue would be solved by adding the functionality of the ModalStateManager into the template.
> 
> I'm not sure what information is being hidden could you please elaborate. I did do some reworking on the story and the modal state wrapper so that the controls should work. As I explained above getting the modal to work with the docs section of storybook is tricky because it forces us to introduce state to the story in order to be able to show and hide the modal. also in order to properly show the "flow" of creating a key in the standard story we're forced to use state to establish a loading a state. I could remove this code, but then there's no way to have the user walk through create => wait => result
> 
> ```
> const setKeyHandler = async (resourceName) => {
>   setLoading(true);
>   const timeout = () => new Promise((resolve) => setTimeout(resolve, 2000));
>   await timeout();
>   setApiKey('111-111-111-111');
>   setLoading(false);
> };
> ...
> <APIKeyModal
>   {...args}
>   apiKey={apiKey}
>   loading={loading}
>   onRequestClose={onCloseHandler}
>   onRequestSubmit={setKeyHandler}
>   open={open}
> />
> ```

When viewing code on standard I see
```js
<TemplateWithState
  apiKey="123-456-789"
  apiKeyInputId="apiKeyInput"
  apiKeyLabel="API key"
  apiKeyVisibility
  copyButtonText="Copy"
  createButtonText="Generate API key"
  createHeader="Generate an API key"
  downloadBodyText="This is your unique API key and is non-recoverable. If you lose this API key, you will have to reset it."
  downloadLinkText="Download as JSON"
  downloadable
  downloadableFileName="apikey"
  loadingMessage="your key is being created. please wait..."
  modalBody="Optional description text. To connect securely to {{product}}, your application or tool needs an API key with permission to access the cluster and resources."
  nameHelperText="Providing the application name will help you idenfity your api key later."
  nameInputId="nameInput"
  nameLabel="Name your application"
  namePlaceholder="Application name"
  nameRequired
  onRequestClose={function noRefCheck(){}}
  onRequestSubmit={function noRefCheck(){}}
  open
  secondaryButtonText="Close"
  successBody={<p>This is your unique API key and is non-recoverable. If you lose this API key, you will have to reset it.</p>}
  successHeader="API key successully created"
/>
```

On the narrow tearsheet I see
```js
<>
  <style>
    {`.exp-tearsheet { opacity: 0 };`}
  </style>
  <Button
    disabled={false}
    kind="primary"
    onClick={() => {}}
    size="default"
    tabIndex={0}
    tooltipAlignment="center"
    tooltipPosition="top"
    type="button"
  >
    Open Tearsheet
  </Button>
  <TearsheetNarrow
    buttons={<div className="tearsheet-stories__buttons"><Button className="tearsheet-stories__button-50" disabled={false} kind="secondary" onClick={function noRefCheck() {}} size="default" tabIndex={0} tooltipAlignment="center" tooltipPosition="top" type="button">Cancel</Button><Button className="tearsheet-stories__button-50" disabled={false} kind="primary" onClick={function noRefCheck() {}} size="default" tabIndex={0} tooltipAlignment="center" tooltipPosition="top" type="button">Create</Button></div>}
    className="client-class-1 client-class-2"
    closeIconDescription="Close the tearsheet"
    description={<span>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor <strong>incididunt ut labore</strong> et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</span>}
    hasCloseIcon
    height="normal"
    label="The label of the tearsheet"
    onClose={function noRefCheck() {}}
    preventCloseOnClickOutside
    title="Title of the tearsheet"
  >
    <div className="tearsheet-stories__dummy-content-block">
      Main content
    </div>
  </TearsheetNarrow>
</>
```

davidmenendez:
@lee-chase `NarrowTearSheet` and `TearSheet` both experience the styling issue I mentioned when you do not to use a portal

Screen Shot 2020-11-18 at 9 53 54 AM
Screen Shot 2020-11-18 at 9 54 12 AM

i'll try to find a CSS solution to this problem.

lee-chase:
> @lee-chase `NarrowTearSheet` and `TearSheet` both experience the styling issue I mentioned when you do not to use a portal
> 
> Screen Shot 2020-11-18 at 9 53 54 AM
> 
> Screen Shot 2020-11-18 at 9 54 12 AM
> 
> i'll try to find a CSS solution to this problem.

You are right @davidmenendez looking at the Docs tab the content comes after and is all position relative so it will be hard to get in front of it. 

@moores2 still needs to fix his docs tab.

It is possible to work around with a couple of z-index (one on sbdocs-preview and one on story-experimental-tearsheetnarrow--all-attributes-set or a modal decorator could be created)

I'll approve the PR but @davidmenendez you still need to find a way for the 'code' button to show working code as @moores has.

lee-chase (approved):

davidmenendez:
@lee-chase thanks for reviewing and working with me on this.

After a lot of trial and error I found a solution updating the preview decorator in our storybook settings. please take a look and let me know if this satisfies your concerns.

PR #207, opened 2020-11-11
build(babel): revert Rollup to Babel, add common configuration

### LINKED ISSUE #131: Core Babel configuration

SimonFinney:
Resolves #131 

#### Changelog

**New**

- Add common JavaScript Babel build script to `core`
- Add a task for copying assets

**Changed**

- Revert Rollup configuration to Babel until we have a better understanding of how to leverage it
- Update component internal image directories and references to `assets` naming convention, similar to `cdai`
- Hoist `cpx` package to root `package.json`


lee-chase (commented):
LGTM. 

I prefer the individual component publish.

I do however think we need to publish the src folder as part of the package.

We also have an issue with packages/feature-flags which is not a package but is used in packages/experimental. We probably need to check the Carbon solution for feature-flags.

SimonFinney:
@lee-chase Interesting - what's the benefit of publishing the `src` folder as part of the package? I would've assumed if someone wanted to view the source they could access it via the repo

SimonFinney:
Converting to draft as we need to handle illustrations for the `EmptyState`

lee-chase:
@SimonFinney when you publish the source folder (as we have published the SCSS) it allows the consumer to choose their own babel/polyfill configuration.

SimonFinney:
I couldn't find a similar practice across any of Carbon's packages, and I also haven't any first-hand experience with that requirement - is this something that's common across your org's offerings, and if so, can we create a follow-up issue to address to unblock this?

lee-chase:
> I couldn't find a similar practice across any of Carbon's packages, and I also haven't any first-hand experience with that requirement - is this something that's common across your org's offerings, and if so, can we create a follow-up issue to address and unblock this?

Your right, must just be me ;-) I could have sworn it was common practice.

lee-chase (approved):
LGTM

PR #227, opened 2020-11-23
fix: updates deps for cdai

asfordmatt:
need to have matching dependencies for consumers of CDAI components


vladbalanescu (approved):
LGTM

PR #232, opened 2020-11-24
fix: use tilde for node_module imports 2

asfordmatt:
missed 2 from https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/231


vladbalanescu (approved):
👍

PR #238, opened 2020-11-25
build: run tests independently from root task

SimonFinney:
#### Changelog

**New**

Add dummy `security` test to support configuration - this will be updated in the follow-up to #233 

**Changed**

- Move `accessibility-checker` configuration into `core` and use in root and `security`
- Use same Babel configuration for building and testing
- Run any package test scripts from the root - JavaScript tests were being run, but not individual package SCSS tests
- Optimise `security` build and test scripts
- Refine continuous integration check

**Removed**

- 'Lint' and 'Test' step from continuous integration check to reduce turnaround time as these already seemed covered by `ci-check`
- `security` Rollup configuration references


package.json
```diff
@@ -24,7 +24,7 @@
     "lint": "yarn lint:es && yarn lint:style",
     "lint:es": "eslint 'packages/*/src/**/*.js'",
     "lint:style": "stylelint 'packages/*/src/**/*.scss' --report-needless-disables --report-invalid-scope-disables",
-    "test": "jest",
+    "test": "lerna run test --stream && jest",
```
SimonFinney, line 27: @carbon-design-system/ibm-cloud-cognitive-development I believe this will cause duplication for any package that also runs `jest` further downstream, so I'd like to discuss what a good approach for this might be

lee-chase (approved):
LGTM

PR #244, opened 2020-11-30
chore(release): correct repository information in package.jsons

dbrugger:
Contributes to #

Some repository links are computed wrongly and e.g. include the package subdirectory when pointing to issues or PRs.

#### Changelog

**Changed**

- `packages/cli/src/commands/sync/packageJson.js`: use extended repository information that includes the directory instead of a concatenated string (see https://docs.npmjs.com/cli/v6/configuring-npm/package-json)
- ran `yarn sync`

### DIFF
--- a/packages/automation/package.json
+++ b/packages/automation/package.json
@@ -3,7 +3,11 @@
   "private": true,
   "version": "0.0.2",
   "license": "Apache-2.0",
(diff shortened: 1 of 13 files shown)


SimonFinney (approved):
LGTM! 🔥

Issue #134, opened 2020-10-07
Notifications


Levinson1:
Pattern design documentation: https://pages.github.ibm.com/cdai-design/pal/patterns/notifications

Dev maintainer: Matt Gallo (@matthewgallo)

matthewgallo:
## Tasks

- [ ] Initial review of design/existing code

### Working in experimental package

- [ ] Create a shell component which
  - [ ] Create ComponentName/ComponentName.js component file that renders props
        in raw form (no styling or sub components, plain text OK)
  - [ ] Adds prop-types with comments for DocGen to
        ComponentName/ComponentName.js
  - [ ] Add Doc file e.g. ComponentName.mdx with a brief description of the
        component.
- [ ] First engineering review/playback

### Continuing in experimental package

- [ ] Storybook - stories to cover design scenarios, story with long, min and
      max content, etc
- [ ] Doc tab in storybook derived from design docs with embedded stories where
      appropriate
- [ ] Code to include prop-types with comments for automated DocGen
      documentation
- [ ] Tests (written in React Test Library) to cover behavior from the
      documented in the components mdx file.

### Review and promote

- [ ] Design review
- [ ] Add design review issues addressed, issues raised or discounted.
- [ ] Engineering review - Submit PR to promote component to production package
- [ ] All engineering issues addressed, issues raised or discounted.

matthewgallo:
https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/254 for initial work

mauifrancis78:
@matthewgallo Ok finally got my Git up and working. Here are some points of feedback I see so far along with screen shots:

Spacing/padding (which I provided a screen for)
What happened to the bottom bar “view all” “settings icon”?
Individual notification dimiss controls
Read/unread states
Empty state spacing

![Notification active (click state)](https://user-images.githubusercontent.com/75501685/101216299-e2761e80-3644-11eb-96d1-813d6fc8f41c.png)

![Notification default](https://user-images.githubusercontent.com/75501685/101216301-e30eb500-3644-11eb-8550-b650c529383a.png)

![Notification hover](https://user-images.githubusercontent.com/75501685/101216302-e30eb500-3644-11eb-88cc-2e2933610b36.png)

![Notification_empty state](https://user-images.githubusercontent.com/75501685/101216826-d2127380-3645-11eb-80c5-1c518770c9a0.png)

matthewgallo:
[...]

PR #254, opened 2020-12-02
PR for #134: Notifications

matthewgallo:
Contributes to #134 

This adds the new Notifications component, it is based off of the designs seen [here](https://pages.github.ibm.com/cdai-design/pal/patterns/notifications/#notifications-view-all-page). This code is only for the Notifications panel, in the design there is also an entire page for Viewing all notifications. This component also makes use of the new `<EmptyState />` component, when there are not any notifications.

#### Changelog

`packages/experimental/src/components/Notifications/index.scss`
`packages/experimental/src/components/Notifications/_notifications.scss`
`packages/experimental/src/components/Notifications/index.js`
`packages/experimental/src/components/Notifications/Notifications.stories.js`
`packages/experimental/src/components/Notifications/Notifications_data.js`
`packages/experimental/src/components/Notifications/Notifications-test.js`
`packages/experimental/src/components/Notifications/Notifications.js`
`packages/experimental/src/components/Notifications/utils.js`
`packages/experimental/src/components/Notifications/Notificaitons.mdx`

**New**

- Added new Notifications component based on designs seen [here](https://pages.github.ibm.com/cdai-design/pal/patterns/notifications)

#### Testing / Reviewing

Run `yarn test` and see that all tests pass for this component.

image


packages/experimental/src/components/Notifications/Notifications.js
```diff
@@ -0,0 +1,293 @@
...
+              Dismiss all
+            </Button>
+          </div>
+          <ToggleSmall
+            className={`${prefix}-notifications-do-not-disturb-toggle`}
+            id={`${prefix}-notifications-do-not-disturn-toggle-component`}
+            labelA="Do not disturb"
+            labelB="Do not disturb"
+            onToggle={(event) => onDisableNotificationChange(event)}
+            aria-label="Do not disturb toggle"
```
lee-chase, line 207: Is an aria-label needed when other labels are visible?

packages/experimental/src/components/Notifications/Notifications.js
```diff
@@ -0,0 +1,293 @@
...
+              }}>
+              Dismiss all
+            </Button>
+          </div>
+          <ToggleSmall
+            className={`${prefix}-notifications-do-not-disturb-toggle`}
+            id={`${prefix}-notifications-do-not-disturn-toggle-component`}
+            labelA="Do not disturb"
+            labelB="Do not disturb"
+            onToggle={(event) => onDisableNotificationChange(event)}
```
lee-chase, line 206: Event should be 'onDoNotDisturbChange' to match other naming.

packages/experimental/src/components/Notifications/Notifications.js
```diff
@@ -0,0 +1,293 @@
...
+
+  return (
+    shouldRender && (
+      <div
+        className={[`${prefix}-notifications-panel-container`].join(' ')}
+        style={{ animation: `${open ? 'fadeIn 250ms' : 'fadeOut 250ms'}` }}
+        onAnimationEnd={onAnimationEnd}>
+        <div className={`${prefix}-notifications-header-container`}>
+          <div className={`${prefix}-notifications-header-flex`}>
+            <h6 className={`${prefix}-notifications-header`}>Notifications</h6>
```
lee-chase, line 190: A h6?

lee-chase (changes requested):
- * Please do not use Knobs in your story, switch to use Controls instead which is largely automatic if you include Proptypes.
- Animate reveal/close when clicking read more.
- Should this close on click outside?
- Can we show a notification being received, perhaps on a timer?
- * All text (labels, titles, aria-labels, content) should be from properties.
- How will this work in conjunction with the HeaderGlobalAction notification icon? Will we be providing a separate component to display a status there?
- How does do not disturb work?
- Can we use a decorator to show this in context with a UI shell header and notification icon?

* These need addressing, the other items can be questions asked of your designer (Slack and in Issue).

packages/experimental/src/components/Notifications/Notifications.js
matthewgallo: I used this just to get the right type size, but I can change it to an `h1` and use the carbon mixin to get the correct type size that way.

packages/experimental/src/components/Notifications/Notifications.js
matthewgallo: Not 100% sure I just thought it was safest to add it just in case.

matthewgallo:
Just to address some points of your points @lee-chase:
- I think it is a good idea for this to close when clicking outside
- The Do not disturb functionality should probably be left up to the consumer of this component, that's why I created the `onDoNotDisturbChange` prop that returns the current value of the toggle so they can make whatever api request they need to "quiet" the notifications.
- Yes, I can add a new story that automatically generates new notifications coming in

lee-chase (approved):
LGTM although you are causing a scroll bar on open.

Issue #148, opened 2020-10-14
About screen modal


Levinson1:
New pattern in need of code
https://pages.github.ibm.com/cdai-design/pal/patterns/about-modal/usage

[View about modal implementation](https://ibm-cloud-cognitive.netlify.app/?path=/story/experimental-aboutmodal--default)

Design Maintainer: Pascal Potvin (@pascal-potvin)
Dev Maintainer: Matt Gallo (@matthewgallo)

## Tasks

- [x] Initial review of design/existing code

### Working in experimental package

- [x] Create a shell component which
  - [x] Create ComponentName/ComponentName.js component file that renders props
        in raw form (no styling or sub components, plain text OK)
  - [x] Adds prop-types with comments for DocGen to
        ComponentName/ComponentName.js
  - [x] Add Doc file e.g. ComponentName.mdx with a brief description of the
        component.
- [x] First engineering review/playback

### Continuing in experimental package

- [x] Storybook - stories to cover design scenarios, story with long, min and
      max content, etc
- [x] Doc tab in storybook derived from design docs with embedded stories where
      appropriate
- [x] Code to include prop-types with comments for automated DocGen
      documentation
- [x] Tests (written in React Test Library) to cover behavior from the
      documented in the components mdx file.

matthewgallo:
Hey @Levinson1, is there a design contact for this component?

Levinson1:
Hey @matthewgallo . Designer for this one is @pascal-potvin .

matthewgallo:
https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/187

pascal-potvin:
This looks really good! A couple of small things here.

**Scroll fade**: The fade should be the exact same colour as the background, which goes from 100% opacity to 0% opacity. It seems to be currently set as black.

**Tabs:** They seem off in the Storybook example. We should be using the _default tabs_ and not the _container tabs_ 

![image](https://user-images.githubusercontent.com/38508793/99675370-23671400-2a45-11eb-90b6-ce8595ace3ca.png)

matthewgallo:
Thanks for the review @pascal-potvin, I think the deployed storybook is using a different `carbon-components-react` version, where the Tab component had a class change. This is what it looks like with the correct version

![image](https://user-images.githubusercontent.com/10215203/99702959-81572400-2a64-11eb-93db-f8853fad7050.png)

[...]

PR #256, opened 2020-12-03
fix: use correct carbon token for about modal gradient

matthewgallo:
Contributes to #148

Use correct carbon color token on about modal scroll gradient


lee-chase (approved):
LGTM

PR #266, opened 2020-12-09
Provide an initial state for <IdeImporting> component

polinaouk:
e8d00780-3970-11eb-8745-6c7d3d1bfeef

We have a special case of using `IdeImporting` component. When we imported the file, user can switch between tabs and then return back to the imported file. However, this sections gets rerendered with IdeImporting component.

In order to make it make work, this changes allow to provide an initial state of files.

**New**
`IdeImporting` component has initial state for files

#### Testing / Reviewing

- open storybook
- you will just see already imported file in a new story


vladbalanescu (approved):
LGTM 👍  

Defering to @asfordmatt for the final merge 😉

asfordmatt (approved):
Looks good - adds the new prop without breaking existing consumers.
In an ideal world the props would all have explaining comments, but most don't currently in this component.

PR #271, opened 2020-12-10
Use the displayName instead of a type for IdeCreate

vladbalanescu:
Background:
---

Use the `displayName` instead of `type` for `IdeCreate` CDAI children component checking


s100 (changes requested):
I'm fairly sure this won't work, `IdeCreateStep` is a React component (*i.e.* a function) whereas `c.displayName` will be a string, so this comparison will never be satisfied. However... I think this change won't be necessary at all if we have https://github.ibm.com/cp4i/cp4i-operand-create/pull/357 merged.

vladbalanescu:
Closing based on above comment

PR #278, opened 2020-12-14
fix: uuidv4 for ssr

lee-chase:
Make unique id generator ssr safe

#### Changelog

m packages/experimental/src/global/js/utils/uuidv4.js


davidmenendez:
thanks for bringing this up. i hadn't considered if `crypto` was a module in node itself. my only concern is if using `math.random` is suitable enough or if we should consider going with a well established package like `uuid` (https://www.npmjs.com/package/uuid) that we can trust will give us safe and secure values.

thoughts?

lee-chase:
> thanks for bringing this up. i hadn't considered if `crypto` was a module in node itself. my only concern is if using `math.random` is suitable enough or if we should consider going with a well established package like `uuid` (https://www.npmjs.com/package/uuid) that we can trust will give us safe and secure values.
> 
> thoughts?

Depends what we are using it for. Collisions expected between 20 and 50 million apparently.
https://gist.github.com/jed/982883

As we are only using it for HTML IDs I think we will be OK. Not recommended for UUIDs for propper crypto/database work.

davidmenendez:
that's what I thought, but i'd be remiss if i didn't bring it up with everyone else.

davidmenendez (approved):

PR #280, opened 2020-12-15
Added grid as a separate import to fix broken consumers

vladbalanescu:
Changes:
---

Grid is now a seprate package which needs an explicit import in order to fix broken consumer' builds:
image


package.json
```diff
@@ -33,6 +33,7 @@
   },
   "devDependencies": {
     "@carbon/bundler": "^10.9.0",
+    "@carbon/grid": "^10.17.0",
```
SimonFinney, line 36: @vladbalanescu Should this be a production dependency in `cdai/package.json`?

vladbalanescu: @SimonFinney Done, mind re-reviewing?

SimonFinney (approved):
LGTM! 🔥

PR #282, opened 2020-12-15
feat: add side panel component

matthewgallo:
Contributes to https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/64

This PR adds the side panel component (two types: slide over and slide in). See [here](https://pages.github.ibm.com/cdai-design/pal/components/side-panel/usage) for the Side Panel designs.

#### Changelog

**New**

```
packages/experimental/src/components/SidePanel/SidePanel-test.js
packages/experimental/src/components/SidePanel/SidePanel.js
packages/experimental/src/components/SidePanel/SidePanel.mdx
packages/experimental/src/components/SidePanel/SidePanel.stories.js
packages/experimental/src/components/SidePanel/_index.scss
packages/experimental/src/components/SidePanel/_side-panel.scss
packages/experimental/src/components/SidePanel/_storybook-styles.scss
packages/experimental/src/components/SidePanel/constants.js
packages/experimental/src/components/SidePanel/index.js
packages/experimental/src/global/js/utils/keyboardNavigation.js
packages/experimental/src/global/js/utils/wrapFocus.js
```

**Changed**

```
packages/experimental/src/components/_index.scss
```

#### Testing / Reviewing

Run `yarn test packages/experimental/src/components/SidePanel/SidePanel-test.js` to see all the tests for this component.


davidmenendez:
there's a lot to go through here, so it'll take me some time. one of the things i noticed right away though is a lot of this

```
<div className={['class-a', 'class-b', condition ? 'class-c' : 'class-d'].join(' ')]} />
```

we're already utilizing the `classnames` package to handle className management. could you please switch to using `classnames`? if you search the codebase for `import cx from 'classnames';` you'll find some examples of how to use it (if you aren't already familiar). this is also my personal opinion, but i think it's better to strip logic from the HTML whenever possible. take the snippet above for example. i'd instead refactor it to be something like

```
import cx from 'classnames';
const componentClasses = cx(
  'class-a',
  'class-b',
  {
    'class-c': condition,
    'class-d': condition,
  }
);
...
<div className={componentClasses} />
```

personally i think separating the className logic from the HTML makes it easier to read.

packages/experimental/src/components/SidePanel/SidePanel.js
```diff
@@ -0,0 +1,585 @@
...
+  }, [open, animateTitle, animationComplete]);
+
+  // error handling for primary action buttons
+  useEffect(() => {
+    if (primaryActions && primaryActions.length) {
+      if (
+        (size === 'small' || size === 'extraSmall' || size === 'medium') &&
+        primaryActions.length > 3
+      ) {
+        throw new Error(
```
davidmenendez, line 113: hmmm i'm not sure about throwing errors here. the first issue being translations. we're trying to avoid returning any strings from our components directly to the consumer because we're currently not doing any string translating on our end. secondly, even though i know we're trying to guide the user by providing them with established patterns i don't think we need to be yelling at them from the console about their "misusage". i think it would become very involved on our end to try to police every single use case when they should be referencing the design usage guidelines that we link in the docs.

we might want to open this up to discussion in slack with other people.

davidmenendez (approved):
I had one comment that I think we should discuss with other people. aside from that everything looks good to me. i'll approve and leave it up to you to decide what you want to do with my comment.

packages/experimental/src/components/SidePanel/SidePanel.js
matthewgallo: Generally, I agree that we should probably leave it up to the consumer to just use the right configuration. _However_, I just thought it might be better to throw these errors in this case because there is very specific design guidelines as to the button combinations that you can have in this component.

davidmenendez: yeah that makes sense and i definitely considered the fact that these components are more pattern based and should adhere to the design guidelines. i guess i'm just unsure if that's something we should have to worry about or the consumer. and as for the translation stuff i think @lee-chase may have an opinion about it

lee-chase: Hmm, food for thought that is for sure. 

I don't like leaving this up to our consumer. I think I would generate a console.error if more than three are supplied and truncate the array if the guidance is specific.

There is this suggestion here that creating your own propType function makes, which may make even more sense. I think from my recent typeScript study its a tuple so there may be better solutions out there https://stackoverflow.com/questions/42659728/react-proptypes-array-with-specific-length

lee-chase: I think I'm leaning toward writing your own PropTypes validator

matthewgallo: Hey @lee-chase, I'm just getting around to this. I added a custom prop validator and removed this logic from the `useEffect`.

lee-chase (approved):
LGTM

PR #298, opened 2021-01-19
fix: keyboard support for apikey modal

davidmenendez:
Contributes to #56

adds keyboard support. name input should now focus automatically when opened and pressing enter/return on the keyboard should now cause the `onSubmit` function to trigger.

#### Changelog

M       packages/experimental/src/components/APIKeyModal/APIKeyModal.js
M       packages/experimental/src/components/APIKeyModal/APIKeyModal.stories.js

#### Testing / Reviewing

standard example- open the modal. the name input should be focused. enter a name and hit return on the keyboard. the modal should properly submit.

multi-step example- open the modal. the name input should be focused. enter a name fill out the rest of the form. hit return when the resource name on the second step is complete. the modal should properly submit.


rodet:
Hi @davidmenendez , the current code looks good from a quick look.

I tested the preview and noticed that the Enter key doesn't work when the focus is on the radio button. Shouldn't the Enter key be activated while on any control inside the form? At least this is what I would expect.

image

davidmenendez:
@rodet interesting. i personally don't think I would expect that sort of behavior.

@chloepoulterdesign can you give me your input on Stephen's comment.

> I tested the preview and noticed that the Enter key doesn't work when the focus is on the radio button. Shouldn't the Enter key be activated while on any control inside the form? At least this is what I would expect.

chloepoulterdesign:
I agree with @rodet , I think that was one of the items I requested for desired behaviour?

lee-chase (approved):
LGTM

davidmenendez:
There seems to be some technical limitation here with Carbon. I am unable to trigger an `onSubmit` function from the surrounding `<Form />` component or the `<RadioButtonGroup />` component when i'm focused on a radio button. i'm going to merge this in for now and follow up with Carbon about this. if this is something that can't be fixed to meet the desired functionality at least you can still tab to the primary button to go to the next step, so keyboard functionality still exists.

PR #318, opened 2021-02-03
chore: re-org storybook tree

lee-chase:
Proposed prefix to storybook folders to distinguish between newer and older designs in the storybook.

#### Changelog

removed files.

M       packages/automation/config.js
M       packages/cdai/config.js
M       packages/cloud-cognitive/CHANGELOG.md
M       packages/cloud-cognitive/package.json
M       packages/common-services/config.js
M       packages/core/.storybook/preview.js
M       packages/core/CHANGELOG.md
M       packages/core/package.json
M       packages/experimental/CHANGELOG.md
M       packages/experimental/config.js
M       packages/experimental/package.json
M       packages/experimental/src/components/APIKeyModal/APIKeyModal.stories.js
M       packages/experimental/src/components/AboutModal/AboutModal.stories.js
M       packages/experimental/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.stories.js
M       packages/experimental/src/components/ContextHeader/ContextHeader.stories.js
M       packages/experimental/src/components/EmptyState/EmptyState.stories.js
M       packages/experimental/src/components/ExampleComponent/ExampleComponent.stories.js
M       packages/experimental/src/components/ExportModal/ExportModal.stories.js
M       packages/experimental/src/components/ImportModal/ImportModal.stories.js
M       packages/experimental/src/components/ModifiedTabs/ModifiedTabs.stories.js
M       packages/experimental/src/components/Notifications/Notification.stories.js
M       packages/experimental/src/components/PageHeader/ActionBarItem.stories.js
M       packages/experimental/src/components/PageHeader/PageActionItem.stories.js
M       packages/experimental/src/components/PageHeader/PageHeader.stories.js
M       packages/experimental/src/components/RemoveDeleteModal/RemoveDeleteModal.stories.js
M       packages/experimental/src/components/SidePanel/SidePanel.stories.js
M       packages/experimental/src/components/TagSet/TagSet.stories.js
M       packages/experimental/src/components/Tearsheet/Tearsheet.stories.js
M       packages/experimental/src/components/Tearsheet/TearsheetNarrow.stories.js
M       packages/experimental/src/components/Tearsheet/TearsheetShell.stories.js
M       packages/experimental/src/components/Tearsheet/TearsheetStacking.stories.js
M       packages/experimental/src/components/Tearsheet/TearsheetStackingNarrow.stories.js
M       packages/experimental/src/global/js/settings.js
M       packages/security/CHANGELOG.md
M       packages/security/config.js
M       packages/security/package.json
M       packages/security/src/ComboButton/ComboButton.js
M       packages/security/src/ComboButton/_combo-button.scss


richardpilot:
Can we link to a discussion or add the relative discussion to this PR as to why we're adding dates to the storybook? I'm not sure what it gets us other than making it look out of date. It's 2021 not 2020!

lee-chase:
> Can we link to a discussion or add the relative discussion to this PR as to why we're adding dates to the storybook? I'm not sure what it gets us other than making it look out of date. It's 2021 not 2020!

The discussion at the last tech sync revolved around the desire of design to highlight current cloud PAL designs over the existing implementations. In addition, I was instructed that the IDE/CDAI components were not to be considered a V1 of the latest designs but a separate package.

Not wanting to stray into the realms of using words like deprecated I suggested 2020 and 2021 as buckets in storybook to place the stories. I'd be happy to find another form of numbers or words e.g. pre-existing / current. 

This is storybook only, there is no change to the published packges.

SimonFinney:
> > Can we link to a discussion or add the relative discussion to this PR as to why we're adding dates to the storybook? I'm not sure what it gets us other than making it look out of date. It's 2021 not 2020!
> 
> The discussion at the last tech sync revolved around the desire of design to highlight current cloud PAL designs over the existing implementations. In addition, I was instructed that the IDE/CDAI components were not to be considered a V1 of the latest designs but a separate package.
> 
> Not wanting to stray into the realms of using words like deprecated I suggested 2020 and 2021 as buckets in storybook to place the stories. I'd be happy to find another form of numbers or words e.g. pre-existing / current.
> 
> This is storybook only, there is no change to the published packges.


I personally prefer messaging over dates here - if the goal is compatibility for pre-existing components, I'm happy with 'Deprecated' because it's at your own risk and should be .

SimonFinney:
> > Can we link to a discussion or add the relative discussion to this PR as to why we're adding dates to the storybook? I'm not sure what it gets us other than making it look out of date. It's 2021 not 2020!
> 
> The discussion at the last tech sync revolved around the desire of design to highlight current cloud PAL designs over the existing implementations. In addition, I was instructed that the IDE/CDAI components were not to be considered a V1 of the latest designs but a separate package.
> 
> Not wanting to stray into the realms of using words like deprecated I suggested 2020 and 2021 as buckets in storybook to place the stories. I'd be happy to find another form of numbers or words e.g. pre-existing / current.
> 
> This is storybook only, there is no change to the published packges.

/cc @Levinson1 for additional design context.

I personally prefer messaging over dates here - if the goal is compatibility for pre-existing components, I'm happy with 'Deprecated' because it's at your own risk and should deter non-legacy consumers.

[Navigation ordering](https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy#sorting-stories) is important here too, For example:

```
├── Cloud & Cognitive
│   ├── Production
│   │   ├── ComponentX
│   ├── Experimental
│   │   ├── ComponentY
├── Deprecated
│   ├── CD&AI
│   │   ├── DeprecatedComponentX
│   ├── Security
│   │   ├── DeprecatedComponentY
```

davidmenendez:
This seems like a tricky situation to navigate. I _think_ I understand their aversion to using numerical versioning labels (v1, v2, etc), but I'm not a fan of using years because I don't think it does the best job of saying "you should use this instead of that". If we're trying to push users to use one version over another I think the messaging has to be clear and in my opinion I don't think using a year system accomplishes that.

I don't think I'm completely informed on the thought process behind these decisions, but I think if we aren't clear on what version the majority of people should be using that it will lead to confusion on adoption of our components. If for whatever reason some teams have to use legacy versions then I don't think it should matter if they have to navigate to the `legacy` or `deprecated` tab of components.

Just my two cents.

lee-chase:
Well, you can never have too many options. How about the terms Canary (instead of experimental) and Support instead of deprecated.

├── Cloud & Cognitive
│   ├── Production
│   │   ├── ComponentX
│   ├── Canary
│   │   ├── ComponentY
├── Support
│   ├── CD&AI
│   │   ├── DeprecatedComponentX
│   ├── Security
│   │   ├── DeprecatedComponentY

lee-chase:
Agreed Canary and Legacy.
Raise issue to remove the experimental package and use Canary.

packages/core/.storybook/preview.js
```diff
@@ -42,6 +42,26 @@ const decorators = [
 const parameters = {
   controls: { expanded: true, hideNoControlsWarning: true },
   layout: 'centered',
+  options: {
```
SimonFinney, line 45: Would it make more sense to leverage the [`order` property](https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy#sorting-stories) on `storySort` here? I feel it would be both easier on the eyes and to maintain

packages/experimental/config.js
```diff
@@ -6,3 +6,5 @@
  */
 
 export default 'Experimental';
+
+export const sectionTitle = 'Legacy/Automation';
```
SimonFinney, line 10: If I recall correctly, per https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/328 we shouldn't need this anymore.

```suggestion
```

packages/experimental/src/global/js/settings.js
```diff
@@ -3,5 +3,13 @@ import { settings } from 'carbon-components';
 
 const pkgPrefix = 'exp';
 const carbonPrefix = settings.prefix;
+const storybookPrefixCanary = 'Cloud & Cognitive/Canary';
+const storybookPrefixReleased = 'Cloud & Cognitive/Released';
```
SimonFinney, line 7: As these are a Storybook construct, these ought to live in `packages/cloud-cognitive/config.js` or `packages/experimental/config.js`, similar to the convention of other packages, otherwise, they'll be included as part of the distributable.

packages/core/.storybook/preview.js
```diff
@@ -42,6 +42,26 @@ const decorators = [
...
+          return -1;
+        if (
+          b[1].kind.startsWith('Cloud & Cognitive/Released/') &&
+          a[1].kind.startsWith('Cloud & Cognitive/Canary/')
+        )
+          return 1;
+        console.log(a[1].kind, b[1].kind);
+        return a[1].id.localeCompare(b[1].id, undefined, { numeric: true });
+      }
+    },
```
SimonFinney, line 63: Would it make more sense to leverage the [`order` property](https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy#sorting-stories) on `storySort` here? I feel it would be both easier on the eyes and to maintain

packages/experimental/src/components/_pending-a-released-component.stories.js
```diff
@@ -0,0 +1,39 @@
...
+  illustration: CustomIllustration,
+  linkText: 'View components in progress instead.',
+  linkUrl: '/?path=/story/cloud-cognitive-canary-emptystate--default',
+  subtext: (
+    <p>
+      If you have not read the Hitch Hikers Guide to the Universe, DON&apos;T
+      PANIC.
+    </p>
+  ),
+};
```
SimonFinney, line 39: I might be out of the loop on this, but my gut says that unless there has been an explicit desire from design to include something like this as a placeholder, I would hold off from doing so.

SimonFinney (changes requested):
Thanks for your patience @lee-chase ! I have changes, comments, and thoughts for your consideration 🔥

packages/core/.storybook/preview.js
lee-chase: Tried this, but it did not work. Only seems to work on first level meaning canary would be above released.

```
  options: {
    storySort: {
      order: ['Cloud & Cognitive/Released', 'Cloud & Cognitive/Canary', 'Legacy']
    },
  },
```

packages/experimental/src/global/js/settings.js
lee-chase: Moved to config

SimonFinney (approved):
LGTM! 🎉

Issue #272, opened 2020-12-10
Release review


SimonFinney:
This issue is to facilitate a discussion around defining a formal process for conducting release reviews on individual components and patterns.

Comments, discussion, ideas, and thoughts in the thread.

/cc @carbon-design-system/ibm-cloud-cognitive , @carbon-design-system/ibm-cloud-cognitive-design, @carbon-design-system/ibm-cloud-cognitive-development

@matthew-chirgwin @vladbalanescu @asfordmatt @moores2 @davidmenendez @matthewgallo @lee-chase @SimonFinney 

As a starting point the technical sync meeting discussed:
- 100% code coverage
- E2E testing
- API - Are Dev happy with props, methods and events. A specific query about API bloat due to the need to allow all strings to be changeable. The topic of providing translation for fixed strings e.g. OK.
- Use of or tested in a product or product team?

lee-chase:
Personally, I think we should aim for 100% test coverage and check the storybook and it's doc tab.

With regards to the API we should answer the question. Are we likely to need to break this before v1.

SimonFinney:
Thanks @lee-chase - I'm all for defining a threshold for unit tests to ensure we have baseline coverage, as long we're not sacrificing the quality of tests in favor of hitting the numbers. Perhaps we should expand the criteria to include whether it's meeting appropriate use cases and that all inputs and outputs work as expected for consumers?

One thought on the API - I wonder if we should also verify against consistency with similar props in Carbon, as well as our own codebase.

lee-chase:
In summary:
- consistency (the APIs look similar, internally and to Carbon) - NOTE: Carbon sometimes uses node instead of string prop to add flexibility
- testing: coverage above X% yet to be defined
- stories covers designed scenarios
- design has signed off the current version
- Doc tab covers designed scenarios.
- Doc tab includes metrics like test coverage, usage when data available, status any missing or incomplete functionality
- Codesandbox examples - may need a pro account.

lee-chase:
Being discussed here to progress https://ibm.ent.box.com/notes/772354757139

lee-chase:
Andrea Garcia -

> List of standards to be met in order to consider a component or pattern implementation complete and to make it available to the public
> 
> https://github.ibm.com/IBMPrivateCloud/BedrockServices/blob/master/AdopterGuides/CommonUXStandardsAdoptionGuide.md

lee-chase:
[...]

PR #320, opened 2021-02-05
docs(packages): update `README.md` package descriptions

SimonFinney:
Contributes to #272 (https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/272#issuecomment-773944198)

The main thing I'd like feedback on (@asfordmatt / @vladbalanescu ) is that I've explicitly excluded `cdai` and `security` from this table to avoid potential confusion with non-legacy consumers that these packages can be adopted for non-compatibility purposes.

Alternatively, these packages could be included as part of the table but should still indicate the above intent.

#### Changelog

**Changed**

Update `README.md` to describe supported packages similarly to https://github.com/carbon-design-system/carbon#getting-started


lee-chase:
Perhaps we could tie this in with organising content into 2020/2021 folders or pre-existing/current.

lee-chase (approved):

PR #325, opened 2021-02-09
fix: more header review updates

lee-chase:
Contributes to #59

Addresses further page header review items.
- Corrects breadcrumb overflow behaviour to always show the current item
- Corrects unwrapIfFragment to take account of multiple levels
- Corrects measurement of breadcrumb based on priority of last, first then reverse order.

#### Changelog

M       packages/experimental/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
M       packages/experimental/src/components/BreadcrumbWithOverflow/_breadcrumb-with-overflow.scss
M       packages/experimental/src/components/PageHeader/PageHeader.js
M       packages/experimental/src/components/PageHeader/_page-header.scss
M       packages/experimental/src/global/js/utils/unwrap-if-fragment.js


packages/experimental/src/global/js/utils/unwrap-if-fragment.js
```diff
@@ -1,14 +1,31 @@
...
-    children && children.type === React.Fragment
-      ? children.props.children
-      : children;
-  for (let index in realChildren) {
-    newChildArray.push(realChildren[index]);
+
+  const isFragment = (item) => item && item.type === React.Fragment;
+
+  const addChildren = (_children) => {
+    for (let index in _children) {
```
dcwarwick, line 13: Do we believe in "let ... of" ? Or is there some reason to avoid? It would be nice here.

```suggestion
    for (let child of _children) {
      if (isFragment(child) {
        addChildren(child.props.children);
      } else {
        newChildArray.push(child);
      }
    }
```

Just a thought.

packages/experimental/src/global/js/utils/unwrap-if-fragment.js
```diff
@@ -1,14 +1,31 @@
...
-  const realChildren =
-    children && children.type === React.Fragment
-      ? children.props.children
-      : children;
-  for (let index in realChildren) {
-    newChildArray.push(realChildren[index]);
+
+  const isFragment = (item) => item && item.type === React.Fragment;
+
+  const addChildren = (_children) => {
```
dcwarwick, line 12: I see what you're doing, with "_children" (although hiding the outer variable isn't actually a problem as we don't need it -- but it could get confusing!). But "_children" doesn't follow the variable name conventions we usually apply. Maybe use something like "items"?

packages/experimental/src/global/js/utils/unwrap-if-fragment.js
```diff
@@ -1,14 +1,31 @@
...
+      if (isFragment(_children[index])) {
+        addChildren(_children[index].props.children);
+      } else {
+        newChildArray.push(_children[index]);
+      }
+    }
+  };
+
+  // nothing, one fragment, one or more children
+  if (isFragment(children)) {
```
dcwarwick, line 23: This could be abbreviated to `addChildren([children]);`. Just a thought.

dcwarwick (changes requested):
One error, and a few bits of the code are really quite complex and could be usefully clarified. I've added some review comments to the code.

packages/experimental/src/global/js/utils/unwrap-if-fragment.js
lee-chase: > This could be abbreviated to `addChildren([children]);`. Just a thought.

It could but would perhaps not be clear why on this line that addChildren() will unpack the array.

packages/experimental/src/global/js/utils/unwrap-if-fragment.js
lee-chase: Makes sense.

packages/experimental/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -71,124 +71,166 @@ export const BreadcrumbWithOverflow = ({
...
-            <BreadcrumbOverflowMenu
-              overflowItems={newOverflowBreadcrumbItems}
-              key={`$displayed-breadcrumb-${internalId}-overflow`}
-            />
-          );
-        }
-      }
+    // skip items in overflow + 1 for first item
+    for (
+      let i = newOverflowBreadcrumbItems.length + 1;
```
dcwarwick, line 131: I don't think this works for a one-item breadcrumb: it will not insert the final (and only) breadcrumb item.
```suggestion
      let i = newOverflowBreadcrumbItems.length + overflowStart;
```

packages/experimental/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -71,124 +71,166 @@ export const BreadcrumbWithOverflow = ({
...
-          spaceAvailable += breadcrumbWidthsIncludingMargin[i];
+          if (spaceAvailable >= breadcrumbWidthsIncludingMargin[index]) {
+            spaceAvailable -= breadcrumbWidthsIncludingMargin[index];
+            willFit += 1;
+          } else {
+            break;
+          }
+        }
+        // address overflow breadcrumb if needed
+        if (willFit > 1 && willFit < sizingBreadcrumbItems.length - 1) {
```
dcwarwick, line 206: Very minor, but you don't need to test willFit > 1 here, and I think it's slightly clearer without: if willFit < sizingBreadcrumbItems.length - 1 then we are not displaying all the items so we need room for the overflow, and the next loop says while willFit > 1 and there isn't enough room for the overflow reduce willFit.
```suggestion
        if (willFit < sizingBreadcrumbItems.length - 1) {
```

packages/experimental/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -71,124 +71,166 @@ export const BreadcrumbWithOverflow = ({
...
+          // -1 for overflow item
+
+          while (willFit > 1 && spaceAvailable < overflowWidth) {
+            willFit -= 1;
+
+            // Highly unlikely any useful breadcrumb-item is smaller
```
dcwarwick, line 212:
```suggestion
            // Highly unlikely any useful breadcrumb-item is smaller than the overflow menu, but we loop anyway just in case
```

packages/experimental/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -71,124 +71,166 @@ export const BreadcrumbWithOverflow = ({
...
+
+            // item removed is based on last item added which is the current value of willFit
+            const itemToRemove = displayItemIndex(
+              sizingBreadcrumbItems.length - 1,
```
dcwarwick, line 216: Shouldn't be minus one. Also, better to match lines 191 and 194 I think.
```suggestion
              breadcrumbWidthsIncludingMargin.length,
```

packages/experimental/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -71,124 +71,166 @@ export const BreadcrumbWithOverflow = ({
...
+            willFit += 1;
+          } else {
+            break;
+          }
+        }
+        // address overflow breadcrumb if needed
+        if (willFit > 1 && willFit < sizingBreadcrumbItems.length - 1) {
+          // -1 for overflow item
+
+          while (willFit > 1 && spaceAvailable < overflowWidth) {
```
dcwarwick, line 209: Actually, you don't need to test willFit > 1 here either. Might be slightly clearer to allow willFit to go back down to 0 if there really is insufficient space. Because in line 224 (at time of making this comment) we're going to set displayCount to 1 anyway if willFit ends up <1.
```suggestion
          while (spaceAvailable < overflowWidth) {
```

packages/experimental/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -71,124 +71,166 @@ export const BreadcrumbWithOverflow = ({
...
         }
       }
-    }
 
-    setDisplayCount(
-      maxVisibleBreadcrumbItems
-        ? Math.min(willFit, maxVisibleBreadcrumbItems)
-        : willFit
-    );
+      if (willFit < 1) {
```
dcwarwick, line 224:
```suggestion
      if (willFit <= 1) {
```

packages/experimental/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -248,7 +290,7 @@ BreadcrumbWithOverflow.propTypes = {
    */
   className: PropTypes.string,
   /**
-   * maximum visible breadcrumb-items
+   * maximum visible breadcrumb-items values less than 1 treated as 1
```
dcwarwick, line 293:
```suggestion
   * maximum visible breadcrumb-items (values less than 1 are treated as 1)
```

dcwarwick (changes requested):
Much clearer logic! I did find a couple of errors I think, and so also added a few minor tidying-up suggestions that I probably wouldn't have made otherwise :-)

dcwarwick (approved):

packages/experimental/src/global/js/utils/unwrap-if-fragment.js
matthewgallo: Hey @lee-chase and @dcwarwick, it looks like this loop here is causing the PageHeader component to break. I suppose there needs to be another check before the loop to make sure the item passed to `addChildren()` is iterable.

Screen Shot 2021-02-18 at 12 53 44 PM

@AlexanderMelox ran into this issue when using the PageHeader component in the most recent build version.

Issue #301, opened 2021-01-20
Cards


Levinson1:
# Name of component or pattern - Cards

## Brief description - Cards group related pieces of information and invite the user to take action on the content. The user should be able to attain a quick overview of relevant and actionable content. Cards have a limited set of actions launched from them; this creates a singular focus for the user. 

Add a brief description.

## Details

https://pages.github.ibm.com/cdai-design/pal/components/card/overview

### Links to other materials

e.g. sketch files, icons, images, prototypes

## Design Maintainer(s)

- Owner/main maintainer(s) - Johanna Koval
- Second/backup maintainer(s) - Kacie Eberhart, Pascal Potvin

## Tasks

- [x] Initial review of design/existing code

### Working in experimental package

- [x] Create a shell component which
  - [x] Create ComponentName/ComponentName.js component file that renders props
        in raw form (no styling or sub components, plain text OK)
  - [x] Adds prop-types with comments for DocGen to
        ComponentName/ComponentName.js
  - [x] Add Doc file e.g. ComponentName.mdx with a brief description of the
        component.
- [x] First engineering review/playback

### Continuing in experimental package

- [x] Storybook - stories to cover design scenarios, story with long, min and
      max content, story to demonstrate right-to-left support, etc.
- [x] Doc tab in storybook derived from design docs with embedded stories where
      appropriate
- [x] Code to include prop-types with comments for automated DocGen
      documentation
- [x] Tests (written in React Test Library) to cover behavior from the
      documented in the components mdx file.

### New production dependencies

- [x] Verify the dependency has previously been
      [pedigree reviewed](https://pedigree-service.wdc1a.cirrus.ibm.com)
- [x] Verify the dependency
      [doesn't contain any vulnerabilities](https://snyk.io/vuln)
- [x] Verify the dependency [bundle size](https://bundlephobia.com) is
      acceptable
- [x] Verify the dependency is [actively maintained](https://www.npmtrends.com)

### Review and promote

-
- [x] Design review
- [x] Add design review issues addressed, issues raised or discounted.
- [x] Engineering review - Submit PR to promote component to production package
- [ ] All engineering issues addressed, issues raised or discounted.

PR #326, opened 2021-02-09
fix: card updates

davidmenendez:
Contributes to #301

* updates to allow different media positioning (https://pages.github.ibm.com/cdai-design/pal/components/card/expressive/style#3.-media-(optional))
* adds `pictogram` option
* changes how `media` in consumed. previously assumed a url and provided an `` tag for that url, but now the consumer will provide the media in component form.

#### Changelog

M       packages/experimental/src/components/Card/Card.js
M       packages/experimental/src/components/Card/Card.stories.js
M       packages/experimental/src/components/Card/_card.scss
M       packages/experimental/src/components/Card/_storybook-styles.scss

### DIFF
--- a/packages/experimental/src/components/Card/Card.js
+++ b/packages/experimental/src/components/Card/Card.js
@@ -5,7 +5,6 @@ import PropTypes from 'prop-types';
 import { pkgPrefix } from '../../global/js/settings';
 
 export const Card = ({
-  actionOnclick,
   actionButtonText,
   actionIcon: ActionIcon,
   pictogram: Pictogram,
@@ -15,6 +14,7 @@ export const Card = ({
   label,
   media,
   mediaPosition,
+  onActionClick,
   onClick,
   title,
 }) => {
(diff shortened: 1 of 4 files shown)


packages/experimental/src/components/Card/Card.js
```diff
@@ -5,53 +5,55 @@ import PropTypes from 'prop-types';
 import { pkgPrefix } from '../../global/js/settings';
 
 export const Card = ({
-  actionButtonFn,
+  actionOnclick,
```
lee-chase, line 8: By convention should this not be 'onActionClick'

lee-chase (changes requested):
See commnet.

packages/experimental/src/components/Card/Card.js
davidmenendez: yes i agree. will change.

davidmenendez:
@lee-chase fixed

lee-chase (approved):

PR #334, opened 2021-02-11
fix(Notifications): use correct value to prevent horizontal scroll

matthewgallo:
Contributes to #134 

This fixed the horizontal scroll bar that was showing in the Notifications panel.

#### Changelog
**Changed**

`packages/experimental/src/components/Notifications/_notifications.scss`


andrea-gm (approved):
scrollbar is gone, thanks! looks good

PR #338, opened 2021-02-12
feat: add action bar tests

lee-chase:
Contributes to #375 

- Adds tests for ActionBar
- Removes ActionBar (not ActionBarItem) from export.
- Removes prop no longer used. inOverflow was used internally when the overflow used tooltips.
- Adds overflowAriaLabel
- Commentary similar to ExampleComponent and AboutModal examples in prep for release.

#### Changelog

M       packages/cloud-cognitive/src/components/AboutModal/_storybook-styles.scss
M       packages/cloud-cognitive/src/components/ActionBar/ActionBar.js
A       packages/cloud-cognitive/src/components/ActionBar/ActionBar.test.js
M       packages/cloud-cognitive/src/components/ActionBar/ActionBarItem.js
M       packages/cloud-cognitive/src/components/ActionBar/_action-bar.scss
M       packages/cloud-cognitive/src/components/ActionBar/_index.scss
M       packages/cloud-cognitive/src/components/ActionBar/_storybook-styles.scss
M       packages/cloud-cognitive/src/components/index.js
M       packages/cloud-cognitive/src/global/js/package-settings.js


packages/cloud-cognitive/src/components/ActionBar/ActionBar.test.js
```diff
@@ -0,0 +1,91 @@
...
+
+    expect(
+      screen.getByText(/Action 01/) && screen.getByText(/Action 10/)
+    ).toBeTruthy();
+  });
+
+  it('Renders an action bar with overflow items', () => {
+    const overflowAriaLabel = 'Overflow aria label';
+    // not enough room so should see an overflow.
+    // const { container } =
```
matthewgallo, line 67: This can be removed since you're using `screen.getByLabelText` instead of `container.getByLabelText`

packages/cloud-cognitive/src/components/ActionBar/ActionBarItem.js
```diff
@@ -27,8 +36,8 @@ export let ActionBarItem = ({ inOverflow, ...otherProps }) => {
         hasIconOnly: true,
         kind: 'ghost',
         size: 'field',
-        tooltipPosition: inOverflow ? 'left' : 'bottom',
-        tooltipAlignment: inOverflow ? 'center' : 'end',
+        tooltipPosition: 'bottom',
+        tooltipAlignment: 'end',
         type: 'button',
       }}></Button>
```
matthewgallo, line 42: This can be changed to a self closing tag which will look/feel a bit more logical being that it is an icon only button like this...
```js
<Button
    someProp={value}
/>
```

packages/cloud-cognitive/src/components/ActionBar/ActionBarItem.js
```diff
@@ -5,18 +5,27 @@
...
 const componentName = 'ActionBarItem';
 const blockClass = `${pkg.prefix}-action-bar-item`;
 
-export let ActionBarItem = ({ inOverflow, ...otherProps }) => {
+// NOTE: the component SCSS is not imported here: it is rolled up separately.
+
+/**
+ * The ActionBarItem is used in the page header to populate the action bar
+ */
+export let ActionBarItem = ({ ...otherProps }) => {
```
matthewgallo, line 28: If this is going to be made a publicly available component (given that it's still part of `package-settings.js`), we should wrap it in `React.forwardRef()` so that the consumer can pass it a `ref` if they want or need.

If this is just an internal component just disregard (would this component need to be removed from the list of components in `package-settings.js` if that is the case?)

lee-chase: Good catch

packages/cloud-cognitive/src/components/ActionBar/ActionBar.test.js
lee-chase: Removed thannks.

packages/cloud-cognitive/src/components/ActionBar/ActionBarItem.js
lee-chase: Agreed

matthewgallo (approved):

Issue #349, opened 2021-02-17
Script to scaffold component shell breaks build


matthewgallo:
## What package(s) are you using?
carbon/ibm-cloud-cognitive-experimental

## Detailed description

It appears that the new script to scaffold components imports the old prefix but uses the new one. This is causing things to break across different components.

matthewgallo:
#350 fixed this

PR #350, opened 2021-02-17
fix: script to generate component shell using correct prefix now

matthewgallo:
Contributes to #349

@AlexanderMelox noticed that the build was breaking it and appears to be caused by 

#### Changelog

**Changed**

`packages/experimental/scripts/generate/templates/DISPLAY_NAME.js`


SimonFinney (approved):
LGTM 🔥

PR #359, opened 2021-02-23
fix: added decorator to card story

davidmenendez:
Contributes to #301 

Added a decorator function to the card story for a more realistic grid experience

#### Changelog

M       packages/experimental/src/components/Card/Card.stories.js


AlexanderMelox (approved):
Looks good, I have a question on the cards prop for the cards. Why would the user have to pass how many cards they are rendering? Wouldn't they just map over and create them. Makes this dependent of the length of what they are iterating over. 

Other than that LGTM 👍🏻

lee-chase (approved):

davidmenendez:
> Why would the user have to pass how many cards they are rendering? Or maybe if you're using this for the purpose of storybook, I don't think the cards prop should live on the component but rather as a control option in storybook.

This is just for the storybook story and i decided to add that as a way to show multiple cards just as an option because, specifically, in regards to cards there's mention in the design docs about using cards in a grid context and how you might use multiple cards next to each other. that's why i added a story specifically to display how a card looks with multiple cards.

but yes i forgot about the controls. i'll check that out and evaluate if that works for me before merging 👍

davidmenendez:
@AlexanderMelox good call on the control. that worked perfectly 👍

PR #373, opened 2021-02-26
issue-333/status-icons

sydrosa:


andrea-gm (commented):
Looks great, my only concern is the 1,000+ files changed. 

Also, making the theme required would be good for lazy devs :)

sydrosa:
> Looks great, my only concern is the 1,000+ files changed.
> 
> Also, making the theme required would be good for lazy devs :)

@andrea-gm -- I worked with Lee this morning to remove the 1000's of yarn updates. Those are gone. Cleaned up commit history as well.

With regards to lazy devs :P I went ahead and set all the `defaultProps` to `null`, so that developers are forced to put input a theme, size and type prop.

packages/experimental/src/components/StatusIcon/StatusIcon.js
```diff
@@ -0,0 +1,193 @@
...
+    'x-large': <Renew32 />,
+  },
+  {
+    type: 'pending',
+    small: <Time16 />,
+    medium: <Time20 />,
+    large: <Time24 />,
+    'x-large': <Time32 />,
+  },
+];
```
dcwarwick, line 131: This looks quite expensive: we're instantiating forty-four render functions, when only one will eventually be used. OK, JavaScript functions are not very heavyweight, but it still seems slightly unnecessary. What we could do instead is stash the imported components directly in the object, then assign the one that is needed into a variable with a capital letter (see the corresponding suggestion below) and use it in the JSX. Also, if we use the icon types as keys rather than as values in each object we can do a direct lookup rather than having to do an array `find()`.

```suggestion
const icons = {
  'fatal': {
    small: Misuse16,
    medium: Misuse20,
    large: Misuse24,
    'x-large': Misuse32,
  },
  'critical': {
    small: ErrorFilled16,
    medium: ErrorFilled20,
    large: ErrorFilled24,
    'x-large': ErrorFilled32,
  },
  'major-warning': {
    small: WarningAltInvertedFilled16,
    medium: WarningAltInvertedFilled20,
    large: WarningAltInvertedFilled24,
    'x-large': WarningAltInvertedFilled32,
  },
  'minor-warning': {
    small: WarningAltFilled16,
    medium: WarningAltFilled20,
    large: WarningAltFilled24,
    'x-large': WarningAltFilled32,
  },
  'undefined': {
    small: UndefinedFilled16,
    medium: UndefinedFilled20,
    large: UndefinedFilled24,
    'x-large': UndefinedFilled32,
  },
  'unknown': {
    small: UnknownFilled16,
    medium: UnknownFilled20,
    large: UnknownFilled24,
    'x-large': UnknownFilled32,
  },
  'normal': {
    small: CheckmarkFilled16,
    medium: CheckmarkFilled20,
    large: CheckmarkFilled24,
    'x-large': CheckmarkFilled32,
  },
  'info': {
    small: InformationSquareFilled16,
    medium: InformationSquareFilled20,
    large: InformationSquareFilled24,
    'x-large': InformationSquareFilled32,
  },
  'in-progress': {
    small: Renew16,
    medium: Renew20,
    large: Renew24,
    'x-large': Renew32,
  },
  'running': {
    small: Renew16,
    medium: Renew20,
    large: Renew24,
    'x-large': Renew32,
  },
  'pending': {
    small: Time16,
    medium: Time20,
    large: Time24,
    'x-large': Time32,
  },
};
```

packages/experimental/src/components/StatusIcon/StatusIcon.js
```diff
@@ -0,0 +1,193 @@
...
+
+import { pkg } from '../../settings';
+
+import cx from 'classnames';
+
+export const StatusIcon = ({ type, theme, size, className, ...rest }) => {
+  const [icon, setIcon] = useState([]);
+  const [iconSize, setIconSize] = useState('');
+  const [iconTheme, setIconTheme] = useState(null);
+
```
dcwarwick, line 141: This suggestion corresponds to the suggestion above for stashing the imported components directly in an object rather than making render functions for them. This also uses the fact we can use the sizes as keys rather than needing to do an array `find`.

```suggestion
  const IconComponent = icon[iconSize];
  
```

packages/experimental/src/components/StatusIcon/StatusIcon.js
```diff
@@ -0,0 +1,193 @@
...
+    size && setIconSize(size);
+  }, [size]);
+
+  useEffect(() => {
+    theme && setIconTheme(theme);
+  }, [theme]);
+
+  return (
+    <div className={classNames} {...rest}>
+      {icon && iconTheme && iconSize ? icon[`${iconSize}`] : null}
```
dcwarwick, line 162: This is the final part of the suggestion above to stash the components in an object rather than render functions. Here, finally, if and when we have an icon component identified, we create a render function for it and insert it into our rendered output.

```suggestion
      {IconComponent && <IconComponent />}
```

dcwarwick (changes requested):

dcwarwick (approved):

dcwarwick:
@sydrosa Once a PR is reviewed/approved you can go ahead and merge it.
@lee-chase Just a note that Syd's new component will need Canarifying once the merge goes through.

Issue #380, opened 2021-02-26
Cards: Expressive type - Design feedback


andrea-gm:
Please size the issue once design feedback is received.

JohannaSuarezKoval:
The expressive card looks great! 
I have a few comments on things that need to be adjusted following the component guidelines. Please feel free to reach out if any of these points need more clarity:

1. Since card size increments only by 4 columns, could we get controls in the storybook by size (small-4col, medium-8col, large-12col, max-16col)  instead of allowing number input? This way, we avoid things like 1, 2, 3, 5, etc columns cards. Side Panel has a similar control for size.
<img width="1434" alt="Screen Shot 2021-03-01 at 4 51 54 PM" src="https://user-images.githubusercontent.com/5809654/109563977-77ef4600-7aae-11eb-874d-64939920427f.png">

2. Ratios determine image size in the expressive card, so we shouldn't include fixed pixel (600x400) images as an example. I wonder if we can include some controls for image ratios. This will vary depending on the placement. For the left placement, the image has to be part of the card size column count.  Please see these examples: https://ibm.invisionapp.com/d/?redirHash=#/console/15373359/319889226/preview?scrollOffset=0

3. There should be a 2px spacing between the label and title.
<img width="649" alt="Screen Shot 2021-03-01 at 5 02 12 PM" src="https://user-images.githubusercontent.com/5809654/109565041-e7196a00-7aaf-11eb-8d5d-349a3ff7fcdb.png">

4. Button size for expressive cards has been determined to be field size (40px). Right now it is default size (48px).
<img width="823" alt="Screen Shot 2021-03-01 at 5 02 44 PM" src="https://user-images.githubusercontent.com/5809654/109565097-f993a380-7aaf-11eb-92af-3da2a5e7ac78.png">

5. Ghost buttons are the default button style for the secondary action. Can we leave that as the selected default in "secondaryButtonKind"?

6. Clickable cards need the hover state please see reference here: https://pages.github.ibm.com/cdai-design/pal/components/card/expressive/usage#action-zones

7. Clickable cards shouldn't include internal links, so I don't think the last variation is necessary.

8. In addition to having a `label + title` option, could we also have include variations of `tittle+caption` and `just label` in the title block?
<img width="544" alt="Screen Shot 2021-03-01 at 5 05 47 PM" src="https://user-images.githubusercontent.com/5809654/109565405-67d86600-7ab0-11eb-92b3-9020fdb4968c.png">


[...]

PR #396, opened 2021-03-03
fix: card design revisions v1

davidmenendez:
Contributes to #380 

handles all the feedback listed here https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/380#issuecomment-788403005

#### What did you change?

Updates to the `Card` component

Run the storybook and verify the changes align with the design feedback and doesn't break anything

### DIFF
--- a/packages/experimental/src/components/Card/_storybook-styles.scss
+++ b/packages/experimental/src/components/Card/_storybook-styles.scss
@@ -1,4 +1,8 @@
+/* stylelint-disable */
+
 @import './index';
+@import '../../global/styles/project-settings';
+@import '../../global/styles/carbon-settings';
 
 // make the root full width to get a better idea of how the cards
 // look in a real grid situation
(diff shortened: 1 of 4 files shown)


packages/experimental/src/components/Card/_storybook-styles.scss
```diff
@@ -1,8 +1,45 @@
...
+.media-ratio-43 .media-box {
+  padding-bottom: 75%;
+}
+
+.media-ratio-32 .media-box {
+  padding-bottom: 66.66%;
+}
+
+.media-ratio-169 .media-box {
+  padding-bottom: 56.25%;
```
matthewgallo, line 34: I wonder if we could use [`aspect-ratio`](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio) here?

lee-chase: Or even https://react.carbondesignsystem.com/?path=/story/aspectratio--playground

davidmenendez: oh wow i can't believe they already had a component for this built. imagine my surprise. i'll go ahead and see if I can use either of these to improve this. thanks!

davidmenendez:
@lee-chase @matthewgallo thanks for the feedback. i was able to use the carbon `AspectRatio` component. It really cleaned everything up. Usage of it in the `media left` story is still tricky, because there needs to be some sort of defined width, so I still had to leave a set width.

matthewgallo (approved):
Looks good! One small thing I noticed was that there are not any focus states for the `with action icon` and `clickable card with onclick` stories. But I'm guessing they're coming in a future PR

Issue #411, opened 2021-03-05
Web terminal: Add entrance/exit animations 


matthewgallo:
## What will this achieve?
This will add entrance and exit animations when this components mounts/unmounts as specified in the design.

## How will success be measured?
When the component is rendered, it animates in/out.

## Additional information

- Designs
https://pages.github.ibm.com/cdai-design/pal/patterns/web-terminal
- Existing code
https://ibm-cloud-cognitive.netlify.app/?path=/story/cloud-cognitive-canary-webterminal--default

matthewgallo:
Complete

PR #412, opened 2021-03-05
feat(WebTerminal): add entrance/exit animation to panel

matthewgallo:
Contributes to #411 

This PR adds mount/unmount animations to the web terminal panel.

#### What did you change?
I added mount/unmount animations to the web terminal component.

Verified in storybook that the interaction is applied and behaving as expected.


dcwarwick:
One question: could we not add `animation: webTerminalEntrance 250ms` to the `${pkgPrefix}-web-terminal--open` class in scss, and `animation: webTerminalExit 250ms` to the `${pkgPrefix}-web-terminal--closed` class? Is there some reason these need to be explicitly added as styles?

matthewgallo:
The mount/unmount animation does not work if we implement it that way (it just appears/disappears). I found this mount/unmount approach [here](https://czaplinski.io/blog/super-easy-animation-with-react-hooks/) and seems to be the best way without help from other libraries such as framer motion.

dcwarwick (approved):

Issue #385, opened 2021-03-01
AboutModal: adjustments to API to ensure consistency


dcwarwick:
- [x] For consistency with Carbon and to ensure we're setting good precedents, let's make the following adjustments to the props for AboutModal:

- body: Change type to node, to enable inclusion of rich content if necessary (eg symbols) and to be consistent with similar Carbon props eg ModalHeader title. Change prop name to 'content' to be consistent with the design spec and terminology used in ComposedModal etc.
- copyrightText, legalText: Change type to node, as for body.
- links: Change type to array of Link (Carbon link component).
- logo: Change type to node. 
- onRequestClose: Change name to onClose and pass to ComposedModal.onClose. This will also require the intervening DIV to be removed so that ModalHeader and ModalFooter are immediate children of the ComposedModal, as this is the pattern that ComposedModal requires to work properly.
- productName: Change type to node, as for body. Change name to title to be consistent with ModalHeader. Add commentary to clarify that the title for AboutModal should (usually) be the product or service name.
- technologiesUsed and versionNumber are covered by #383 
- theme is covered by #384

- [x] Also ensure all props have useful DocGen comments.
- [x] Also ensure the appropriate props are mandatory or have defaults.

PR #417, opened 2021-03-08
Issue 385

moores2:
Contributes to #385

Addresses issues documented in issue #385

Does not address the problem with the resize detector, thus scrollable content has no fade at the end yet.


dcwarwick (changes requested):
One modification suggested: we can drop the default values -- all of them I think -- as they are not semantically required and the component will be quite happy with any/all of the non-required props coming through as undefined. And especially in the case of links, onClose, technologiesUsed, where this will save creating an empty array/function unnecessarily.

dcwarwick (approved):

PR #426, opened 2021-03-10
feat: bump carbon versions to latest

lee-chase:
#### What did you change?

Bumped carbon from x.27 to x.30 released recently

Ran up storybook


dcwarwick (commented):
The yarn offline mirror does not match yarn.lock file. yarn.lock calls for flatpickr-4.6.9, and that's not in the offline mirror, so the CI tasks fail.

PR #428, opened 2021-03-10
Card design feedback round 3

davidmenendez:
Contributes to #380 

address design feedback for https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/380#issuecomment-794440531


dcwarwick (approved):
I'm not quite sure what those size options are meant to show. But I see that this is exactly the set of options the designer has asked for, so happy to approve :-)

davidmenendez:
@dcwarwick those correspond to carbon grid column sizes https://www.carbondesignsystem.com/guidelines/2x-grid/implementation#responsive-options in the story context it's a way to display the card according to a certain column size

dcwarwick:
@davidmenendez Yes, but they seem a curious selection of options to me. sm-4, md-8 and max-16 will all do exactly the same thing, although theoretically responding to different breakpoints. lg-12 suddenly switches to filling 3/4 of the width rather than the full width. Going forward we might want to consider what the options for the stories need to be. But right now these are the four sizes the designer asked for so whatevs it's fine by me :-)

davidmenendez:
oh yes haha i agree. but that's what design has request 🤷

PR #430, opened 2021-03-11
AboutModal: change technologiesUsed and versionNumber props to additionalInfo prop

### LINKED ISSUE #383: AboutModal: support display of "additional information" using tabs
The component has a "technologiesUsed" prop which does indeed add one tab, but "technologies used" is only an example of the kind of additional information that a user of the About Modal might need to add. The additional information also needs to include the version number when multiple tabs are being used.

- [x] The component should enable **one or more tabs**, with supplied names and content, to be added. Version number then becomes just one example of how to use this, and should be illustrated with code sample.
- [ ] As part of that, there should be no translatable strings left in the component code.
- [ ] Also ensure that the spacings are all controlled using Carbon spacing tokens and not literal values.

moores2:
Resolves  #383 

  Generalises the tabs in the AboutModal to support arbitrary additional information, instead of just version number and technologies used.  It also re-organises the SCSS and reduces it in size.


packages/experimental/src/components/AboutModal/AboutModal.js
```diff
@@ -69,52 +68,47 @@ export const AboutModal = ({
...
-          <>
-            <p className={`${pkg.prefix}-about-modal-version-label`}>
-              Version number
-            </p>
-            <p className={`${pkg.prefix}-about-modal-version-number`}>
-              {versionNumber}
-            </p>
-          </>
-        )}
+        {additionalInfo &&
```
dcwarwick, line 71:
```suggestion
        {additionalInfo && additionalInfo.length > 0
```

Just to protect against some numpty doing additionalInfo=[]. Empty arrays aren't falsy.

packages/experimental/src/components/AboutModal/_about-modal.scss
```diff
@@ -13,131 +13,111 @@
...
 
-.#{$pkg-prefix}-about-modal .bx--modal-header,
-.#{$pkg-prefix}-about-modal .bx--modal-content {
-  grid-row: auto;
-  padding: 0 $spacing-05;
+.#{$pkg-prefix}-about-modal .bx--modal-footer {
+  position: relative;
+  flex-direction: column;
+  justify-content: center;
+  height: rem(72px);
```
dcwarwick, line 86: This was 64px before. Why the change to 72px? To my eye, 64px looks a closer match to the design drawing.

dcwarwick (changes requested):

packages/experimental/src/components/AboutModal/_about-modal.scss
moores2: The design spec spacing section calls for 2x padding of 16px plus spacing of 4px plus 2 lines of text at 18px = 72px. At the moment the top and bottom padding is too small, but the flex box centring makes this unobvious.

packages/experimental/src/components/AboutModal/AboutModal.js
moores2: Done.

dcwarwick (approved):

Issue #432, opened 2021-03-11
Sidepanel component does not receive className prop


AlexanderMelox:
## What will this achieve?

I don't want to wrap the `<SidePanel />` component around a div for a className. Also, it might seem weird because the Side panel is position fixed and simply giving a div a className around it doesn't position it fixed too.

## How will success be measured?

If I can pass a custom `className` prop to the SidePanel component

matthewgallo:
#433 now allows the SidePanel to accept a `className` prop that will add any custom class names.

PR #433, opened 2021-03-11
fix(SidePanel): add classname prop to side panel

matthewgallo:
Contributes to #432 

Allow classname to be passed to side panel component.

### DIFF
--- a/packages/experimental/src/components/SidePanel/SidePanel.js
+++ b/packages/experimental/src/components/SidePanel/SidePanel.js
@@ -49,6 +49,7 @@ export const SidePanel = !pkg.isComponentEnabled(componentName)
       currentStep,
       onNavigationBack,
       selectorPrimaryFocus,
+      className,
     }) => {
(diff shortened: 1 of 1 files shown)


packages/experimental/src/components/SidePanel/SidePanel.js
```diff
@@ -320,7 +321,9 @@ export const SidePanel = !pkg.isComponentEnabled(componentName)
           <>
             <div
               id={`${blockClass}-outer`}
-              className={mainPanelClassNames}
+              className={cx(mainPanelClassNames, {
+                [className]: className,
+              })}
```
AlexanderMelox, line 326: I think it will be neater to do this

```
className={cx(mainPanelClassNames, className)}
```

Also how it saw it in other components

AlexanderMelox (dismissed):
Would change the way you are adding custom classNames you can just add it to the cx function at the end, it will all merge together in the end

AlexanderMelox (approved):

Issue #368, opened 2021-02-26
Improve responsive behaviour of page-header


lee-chase:
The page header has various responsive requirements.

https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/59#issuecomment-701408228

- Care should be taken when dropping page Actions onto a new line so that it does not enter the Breadcrumb/ActionBarItem row too early.
- Tags wrap above the tabs when vsible
- Action bar items currently collapse to a comboButton. The design indicates an overflow, this might clash with ActionBarItems when in Breadcrumb/ActionBarItem row. Check with @cameroncalder 
- Breadcrumb should become back arrow on very narrow screens. Should overflow of breadcrumb be shown at this point? Check with @cameroncalder 
- Should ActionBar and PageActions combine to a single overflow when space is tight? Check with @cameroncalder 

**From epic**

Responsive behavior examples

1312
![1312 breakpoint example](https://user-images.githubusercontent.com/50155706/94694685-b5cb2100-02fa-11eb-9168-abc4433039d6.png)

1056
![1056 breakpoint example](https://user-images.githubusercontent.com/50155706/94694727-bebbf280-02fa-11eb-92cc-bb120e8a2ec6.png)

672
![672 breakpoint example](https://user-images.githubusercontent.com/50155706/94694756-c67b9700-02fa-11eb-9a93-d86b52c9ca09.png)

320
![320 breakpoint example](https://user-images.githubusercontent.com/50155706/94694771-cb404b00-02fa-11eb-9130-ce95ac3a41e5.png)

Follows Carbon breadcrumb overflow guidance (one path at the front and two at the end, space providing)
https://www.carbondesignsystem.com/components/breadcrumb/usage/#overflow-content

_*Note, CP4S use-case: For breadcrumbs requiring overflow, two paths are placed at the front of the breadcrumb and one at the end because of the nature of the application IA in relation to the platform. This provides a quick path back to the root of the current application in addition to the platform homepage)_

cameroncalder:
> * Action bar items currently collapse to a comboButton. The design indicates an overflow, this might clash with ActionBarItems when in Breadcrumb/ActionBarItem row. Check with @cameroncalder

Yup, agree comboButton would be the better example to show

> * Breadcrumb should become back arrow on very narrow screens. Should overflow of breadcrumb be shown at this point? Check with @cameroncalder

For mobile, think we'll be ok with just the back arrow

[...]

PR #435, opened 2021-03-12
Page header fixes

lee-chase:
Contributes to #368 making responsivity fixes to page header

#### What did you change?

- Prevent negative displayCount of ActionBarItems
- Moved Carbon grid SCSS use inside of SidePanel and PageHeader as it is not non-generating and prevented or SCSS overriding som grid SCSS in page header.
- Removed flex override for pageActionButtons which was causing scroll position to jump and preventing the smooth transition of title into breadcrumb.

Purely within storybook.


packages/experimental/src/components/SidePanel/_side-panel.scss
```diff
@@ -7,6 +7,7 @@
 
 @import '../../global/styles/carbon-settings';
 @import '../../global/styles/project-settings';
+@import 'carbon-components/scss/globals/grid/grid';
```
matthewgallo, line 10: @lee-chase how come we need to add the grid styles to the side panel again?

matthewgallo (approved):
👍  I did just have a question about the benefit of removing the grid styles from `_carbon-settings.scss`.

PR #387, opened 2021-03-01
Feature flags proposal

lee-chase:
Contributes to #364 

Proposed feature flags changes, while this impacts on many files they most of them are to cope with the changes as opposed to the actual change. 

In order to review the proposal, it is best to view the following files.
- `packages/core/.storybook/preview.js ensures storybook` presents all components and not the canary.
- `packages/experimental/config.js` provides a neat way of getting the storybook prefix and slug (a little out of scope)
- `packages/experimental/scripts/generate/templates/DISPLAY_NAME-canary-test.js` CLI component canary test. *NOTE: This might be overkill as tests on src/index.js do a similar thing.*
- `packages/experimental/scripts/generate/templates/DISPLAY_NAME-test.js` CLI component test with component enabled, note the delayed import
- `packages/experimental/scripts/generate/templates/DISPLAY_NAME.js` CLI component rendering canary or component
- `packages/experimental/scripts/generate/templates/DISPLAY_NAME.stories.js` CLI story (a little out of scope)
- `packages/experimental/src/__tests__/index-all-canary-test.js` Index test all for canary (with TODO to include other components).
- `packages/experimental/src/__tests__/index-release-test.js` index test to check items that should be canary be default produce a canary
- `packages/experimental/src/components/ExampleComponent/ExampleComponent-canary-test.js` same as CLI canary test
- `packages/experimental/src/components/ExampleComponent/ExampleComponent-test.js` same as CLI test
- `packages/experimental/src/components/ExampleComponent/ExampleComponent.js` update to ExampleComponent
- `packages/experimental/src/components/ExampleComponent/ExampleComponent.mdx` a little out of scope is this MDX file using getStorybookSlug
- `packages/experimental/src/components/ExampleComponent/ExampleComponent.stories.js` a little out of scope update to use storybookPrefix
- `packages/experimental/src/components/_Canary/` The whole folder something to display when a component is not released
- `packages/experimental/src/global/js/package-settings.cmn.js` package-settings in commonJS form as a follow-up PR will generate SCSS from it and Carbon provides utils that need commonJS.
- `packages/experimental/src/global/js/package-settings.js` ESModule wrapper around commonJS file
- `packages/experimental/src/global/js/settings.js` the place where the package exports all it's build time settings including carbon.
- `packages/experimental/src/global/js/settings.spec.js` tests settings.js
- `packages/experimental/src/index.js` expose pkg to our consumers

PR #436, opened 2021-03-12
feat: warn on feature flag use

lee-chase:
Contributes to #387 

Adds a warning on feature flag use and simplifies enableAll

#### What did you change?

Package settings

Storybook and browser dev tools.


packages/experimental/src/global/js/package-settings.js
```diff
@@ -47,17 +47,56 @@ const defaults = {
...
-  for (const key in flags) {
-    flags[key] = enabled;
-  }
-};
+const component = { ...defaults.component };
+const feature = { ...defaults.feature };
+const warningMessage = (type, property) =>
+  `IBM Cloud Cognitive (WARNING): ${type} "${property}" enabled via feature flags. This component has not yet completed it's review process.`;
+const warningMessageAll = (type) =>
+  `IBM Cloud Cognitive (WARNING): All canary ${type.toLowerCase()} have been enabled through use of setAll${type}`;
```
dcwarwick, line 55: Ingenious, but I think it would be better to have separate messages for components and features. The message about components not having completed their review process is, I think, crucial, but doesn't apply to features so would seem a bit odd there.

```suggestion
const warningMessageComponent = property =>
  `IBM Cloud Cognitive (WARNING): Component "${property}" enabled via feature flags. This component has not yet completed its review process.`;
const warningMessageFeature = property =>
  `IBM Cloud Cognitive (WARNING): Feature "${property}" enabled via feature flags.`;
const warningMessageAllComponents =
  'IBM Cloud Cognitive (WARNING): All components enabled through use of setAllComponents. This includes components that have not yet completed their review process.';
const warningMessageAllFeatures = (type) =>
  'IBM Cloud Cognitive (WARNING): All features enabled through use of setAllFeatures';
```

packages/experimental/src/global/js/package-settings.js
```diff
@@ -47,17 +47,56 @@ const defaults = {
...
 const settings = {
+  _allComponents: false,
+  _allFeatures: false,
   // flags
   prefix: defaults.prefix,
-  component: { ...defaults.component },
-  feature: { ...defaults.feature },
+  component: new Proxy(component, {
+    set(target, property, value) {
+      console.warn(warningMessage('Component', property, value));
```
dcwarwick, line 64: I think the message only makes sense when a component is enabled, not if one gets disabled.

```suggestion
      if (value) {
        console.warn(warningMessageComponent(property));
      }
```

packages/experimental/src/global/js/package-settings.js
```diff
@@ -47,17 +47,56 @@ const defaults = {
...
+      if (typeof propValue !== 'undefined') {
+        return propValue;
+      } else {
+        return false; // unknown component
+      }
+    },
+  }),
+  feature: new Proxy(feature, {
+    set(target, property, value) {
+      console.warn(warningMessage('Feature', property, value));
```
dcwarwick, line 83:
```suggestion
      if (value) {
        console.warn(warningMessageFeature(property));
      }
```

packages/experimental/src/global/js/package-settings.js
```diff
@@ -69,11 +108,35 @@ const settings = {
     return flags.feature[feature];
   },
   setAllComponents: (enabled) => {
-    setAll(settings.component, enabled);
+    if (enabled) {
+      console.warn(warningMessageAll('Components'));
```
dcwarwick, line 112:
```suggestion
      console.warn(warningMessageAllComponents);
```

packages/experimental/src/global/js/package-settings.js
```diff
@@ -69,11 +108,35 @@ const settings = {
...
+    }
+    settings._allComponents = enabled;
   },
   setAllFeatures: (enabled) => {
-    setAll(settings.feature, enabled);
+    if (enabled) {
+      console.warn(warningMessageAll('Features'));
```
dcwarwick, line 118:
```suggestion
      console.warn(warningMessageAllFeatures);
```

dcwarwick (changes requested):

dcwarwick (approved):

PR #447, opened 2021-03-16
Apply common file names across components

### LINKED ISSUE #440: ES and LIB bundles include tests
This may be due to inconsistent use of .spec.js, .test.js and other variants for test file names.

dcwarwick:
Resolves #440

Ensure all files use the common file names/formats set out in code guidelines. As a consequence of the previous fix, this means the tests and stories in particular will no longer appear in our build outputs.

#### What did you change?

Renamed various test and story files.

Ran a build and checked outputs. Ran storybook and checked functional.


lee-chase (commented):
@dcwarwick CDAI uses .spec.js, using that format would allow us to refine `packages/jest-config/index.js` to match only one file format.

dcwarwick:
> @dcwarwick CDAI uses .spec.js

Carbon-react uses 'test', though. Albeit they use `-test.js` (which doesn't work for them) rather than `.test.js`. Both 'test' and 'spec' seem to be current practices, and most configs are set up to include either, which is also true of ours:
image

packages/experimental/src/components/ExampleComponent/_storybook-styles.scss
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright IBM Corp. 2021, 2021
+// Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2: This looks a bit odd, but I think it's just because I copied/renamed the file from elsewhere. All the other files in this component are marked 2020, 2020.

packages/experimental/src/components/ContextHeader/_index.scss
```diff
@@ -1,7 +1,8 @@
 //
-// Copyright IBM Corp. 2021, 2021
+// Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2: This looks a bit odd, but I think it's just because I copied/renamed the file from elsewhere. All the other files in this component are marked 2020, 2020.

matthewgallo:
I'm submitting my vote for `.test.js` 🎉

lee-chase:
Fair enough @dcwarwick although it might give @matthewgallo the impression this is a democracy ;-)

lee-chase (approved):

PR #448, opened 2021-03-16
fix: useless fragment issue in unwrap

lee-chase:
When confronted with a useless fragment https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-useless-fragment.md

unwrap-if-fragment was failing.

#### What did you change?

- Added test for useless fragment.
- Updated unwrap-if-fragment to work with useless fragments.
 
Ran test and verified storybook use case

```
const breadcrumbItems = (
  <>
    <BreadcrumbItem href="#">Breadcrumb 1</BreadcrumbItem>
  </>
);
```


dcwarwick (approved):
Ah, this is the unwrap-if-fragment fix that also got scooped up into #450...

Issue #393, opened 2021-03-02
AboutModal: standards and consistency


dcwarwick:
We should follow the Carbon BEM pattern for class names etc, which is: `prefix--component-name__element-name--modifier`. Also, the pattern of using a constant `blockClass` in JS and `$block-class` in SCSS to capture `prefix--component-name` helps avoid slips. Also, all public components should use forwardRef to ensure they can be refed when needed, and should support className. Also, all unused properties should be passed through to a suitable DOM node by default.

- [ ] Fix AboutModal to apply the above standards.
- [x] Add a CODE_GUIDELINES.MD to .github with the above standards and other relevant patterns in for future reference.
- [x] Ensure example components and CLI match code guidelines.

PR #455, opened 2021-03-16
Update example component and CLI js

dcwarwick:
Contributes to #393 

Brings the Javascript files for the ExampleComponent and the CLI templates in line with our code guidelines.

#### What did you change?

ExampleComponent js, template js, and CODE_GUIDELINES.md

Ran a build and ran storybook.


packages/experimental/src/components/ExampleComponent/ExampleComponent.js
```diff
@@ -53,83 +66,114 @@ export const ExampleComponent = !pkg.isComponentEnabled(componentName)
...
+            /* Pass through any other property values as HTML attributes. */
+            ...rest
+          }>
           <Button
+            className={`${blockClass}__secondary-button`}
             kind={secondaryKind}
             onClick={handleSecondaryClick}
             size={size}
-            disabled={props.disabled}>
+            disabled={rest.disabled}>
```
davidmenendez, line 91: shouldn't `disabled` just be defined at the top with the other props?

davidmenendez (changes requested):
looks good. just one comment.

packages/experimental/src/components/ExampleComponent/ExampleComponent.js
dcwarwick: Yes, I think so. It looked a bit odd to me too, this. Not sure why it was done. Fixed.

matthewgallo:
Can we add the generated component to the list of default components in `package-settings.js` when the CLI script runs? I think that would make the developer experience of using that script a little better.

dcwarwick:
> Can we add the generated component to the list of default components in `package-settings.js` when the CLI script runs? I think that would make the developer experience of using that script a little better.

That's a nice idea. Needs investigating. It's come up once before, so I've opened #460 to address it.

lee-chase (approved):

davidmenendez (approved):

Issue #449, opened 2021-03-16
HTTPErrors: Add 404 and other HTTPError components


matthewgallo:
## What will this achieve?
This will add the remaining HTTPError state components.

## How will success be measured?
When there are 3 total HTTPError components (403/404/other) in the storybook.

## Additional information

- [Designs](https://pages.github.ibm.com/cdai-design/pal/patterns/http-errors/usage)
- [Existing code](https://ibm-cloud-cognitive.netlify.app/?path=/story/cloud-cognitive-canary-httperrors-httperror403--with-all-props-set)

PR #456, opened 2021-03-16
PR for #449: Add remaining http error components

matthewgallo:
Contributes to #449 

This PR adds the two remaining HTTPError components.

#### What did you change?
I added individual mdx files for each of these new components within the HTTPErrors storybook directory.

Within storybook


davidmenendez (commented):
just out of curiosity, why go with having multiple error components instead of a single one with conditional returns?

matthewgallo:
I would have preferred that approach but this stems from the convo on the empty state PR, as we don’t want to have a single component that will end up importing all of the illustrations in one place. This approach keeps the illustrations separate

davidmenendez:
that's what i figured. man i hate that. oh well. looks good to me 👍

davidmenendez (approved):

Issue #462, opened 2021-03-17
Clean up test output


dcwarwick:
There are a lot of console messages that are emitted, mostly by CDAI components, during the test run `yarn test`. This makes it very difficult to work out what tests are actually failing! If those console messages are not needed they should be removed, and if they are needed then the tests can absorb the console output rather than have it dumped at the console. Ideally this would be done by maintainers of those components, but we should probably do some/all of it for our own sanity purposes.

PR #466, opened 2021-03-17
chore: ignore console log during test

lee-chase:
Contributes to #462

It has become increasingly harder to review test output, mainly because of all the console output. This PR removes it from CDAI.

Made the CDAI tests quiet. While we could just turn off console output for all the tests by mocking console.log/warn/error I have only done this for the IdeDataTable as it is harder to see how all required props get supplied.

I was unable to see how to prevent `UnhandledPromiseRejectionWarning`

#### What did you change?

- IdeButton - children type changed to node on the assumption that the test pass of two elements is correct.
- IdeCreate.test - Used jest spy on console and checked it was called.
- IdeDataTable.spec - mocked console.log/warn/error in beforeAll as it was hard to see how all required props could be added easily.
- IdeFilter - The use of `type="filter"` was out of date. No prop changes just handling of previous behaviour correctly. IdeFilter should add a prop 'filter'.
- IdeFilter.spec - corrected test to look for prop().filter
- IdeImporting - stopped processing async code when unmounted, perhaps test should have waited.
- IdeImporting.test - Removed dumping of log info to console. Console output should be tested if important.
- IdeSlideOverPanel.spec - added requiredProps (onClose) to uses of IdeSlideOverPanel

Ran tests checked no console.logs.


dcwarwick (approved):

davidmenendez (approved):

vladbalanescu (approved):
LGTM 👍

Georgegriff:
We need to take another pass at this pr  these changes https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/466/files#diff-6a018a5d6be2bf646860de69f748340880e2df9624a853a2b2f322297924b8ce caused this regression (double close buttons) https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/713

Issue #457, opened 2021-03-16
Storybook props tables not rendering


matthewgallo:
## What package(s) are you using?

- [x] Cloud & Cognitive
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

The props table within the docs tab in storybook are not rendering for all components I've noticed recently.

davidmenendez:
This issue is being caused by the canary conditional at the top of the component. this can be fixed by moving the conditional to inside the component function. for example-

```
export const APIKeyModal = () => {
  ...
  if (!pkg.isComponentEnabled(componentName)) return <Canary component={componentName} />
  return ...
};
```

cc @lee-chase @dcwarwick

dcwarwick:
The downside of moving the conditional to inside the component function is that various bits of React, like useState and useEffect, cannot be used inside conditional branches: they must occur in the same sequence in all circumstances. So if the conditional is moved inside the component function, it needs to move all the way to where we are just about to return the render function, and can't enclose any preparatory work. This is quite intrusive.

If we are going to have to execute most of the component function, then perhaps we should just accept that and go for maximum convenience. It occurs to me that we don't really need to worry about performance or tree-shaking, since using a component that will render as Canary is not a production scenario -- the key thing is that components that are NOT Canary should spend an absolute minimum of time and distraction checking that they are released. I experimented with a number of ways of arranging this, and the pattern I thought looked neatest and had the shortest path for production use is to export the component with a `let` rather than a `const` and then simply reassign a Canary if we're not released. Thus:
```js
export let Component = (prop1, prop2, ...) => { ..... return <stuff.....>; }

if (!pkg.isComponentEnabled(componentName)) {
  Component = () => <Canary .../>
}
```

In fact, to minimise what needs to appear in each file, I've made a little function in settings.js to do this check. That means each component doesn't need to import Canary and include the logic. It just calls one function in settings.js, and if the component is enabled this imposes a minimum of extra weight (one function return on top of checking the status). If the component is not enabled we've created then discarded a component needlessly, but as noted above this is already a failure situation and not part of a production scenario. Tree-shaking should ensure that all components not used are not packaged up, whether they would have rendered as Canary or not.

PR #473, opened 2021-03-18
A proposal to fix the Storybook failures caused by indirect export of components

dcwarwick:
Contributes to #457 

See notes in #457. After finding I was pushing the logic further and further along, it seemed best to move it out of the component function altogether. ExampleComponent and AboutModal illustrate this. I've tested them, and they seem satisfactory. I can't see any downsides to this. If this looks good, I'll apply to the other components likewise.

#### What did you change?

ExampleComponent.js and AboutModal.js. Also, a new function to settings.js to check whether a component is enabled and return a Canary otherwise, and a small update to Canary to avoid an import loop (this is also one reason the function is in settings.js rather than global/js/package-settings.js).

Ran build, ran tests, and ran Storybook, all with and without disabling the turning on of the component flag. Checked that component renders as expected in both cases, and that Storybook props and controls are working.


lee-chase (commented):
I think it is tidier than my thought, which was to move the canary out to the index.js for each component.

A couple of minor things:
1. I think `AboutModal = pkg.checkComponentEnabled(AboutModal);` should go after assigning component name.
2. settings.js is also modified in #472

lee-chase (approved):

PR #482, opened 2021-03-19
chore: improve generate cli

### LINKED ISSUE #460: CLI should add new component name to list of components
Can we add the generated component to the list of default components in package-settings.js when the CLI script runs? I think that would make the developer experience of using that script a little better.

lee-chase:
Closes #460

Updates the generate cli to add the new component to components/index.js and global/js/package-settings.js

#### What did you change?

packages/experimental/scripts/generate/index.js - to amend the script
packages/experimental/src/global/js/package-settings.js - to add a comment targetted by the script

Tried out the CLI


dcwarwick (changes requested):
Should we not also add `@import './ComponentName/index';` to `_index.scss`.

dcwarwick (approved):

PR #483, opened 2021-03-19
Apply code guidelines to AboutModal

### LINKED ISSUE #393: AboutModal: standards and consistency

dcwarwick:
Closes #393

#### What did you change?

AboutModal component files

Ran build and ran storybook. Added tests for remaining guidelines items and verified all tests run ok.

### DIFF
--- a/packages/experimental/src/components/AboutModal/AboutModal.test.js
+++ b/packages/experimental/src/components/AboutModal/AboutModal.test.js
@@ -5,22 +5,56 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { fireEvent, render, screen } from '@testing-library/react';
 import React from 'react';
-import { Link } from 'carbon-components-react';
+import { fireEvent, render, screen } from '@testing-library/react';
 
+import { pkg, carbon } from '../../settings';
 import '../../enable-all'; // must come before component is imported (directly or indirectly)
 
+import uuidv4 from '../../global/js/utils/uuidv4';
+
+import { Link } from 'carbon-components-react';
 import { AboutModal } from '.';
+
 import ExampleLogo from './_story-assets/example-logo.svg';
 import ansibleLogo from './_story-assets/ansible-logo.png';
 import grafanaLogo from './_story-assets/grafana-logo.png';
 import jsLogo from './_story-assets/js-logo.png';
 
-const onCloseReturnsTrue = jest.fn(() => true);
-const onCloseReturnsFalse = jest.fn(() => false);
+const blockClass = `${pkg.prefix}--about-modal`;
+const componentName = AboutModal.displayName;
 
-const content = 'This is example content';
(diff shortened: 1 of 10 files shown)


packages/experimental/src/components/AboutModal/AboutModal.test.js
```diff
@@ -5,207 +5,196 @@
...
+    expect(container.querySelector(`.${blockClass}`)).not.toBeNull();
+  });
+
+  /*it('has no accessibility violations', async () => {
+    const { container } = render(
+      <AboutModal open {...{ content, logo, title }} />
+    );
+    await expect(container).toBeAccessible(componentName);
+    await expect(container).toHaveNoAxeViolations();
+  });*/
```
matthewgallo, line 98: Should we remove this test since it's commented out?

packages/experimental/src/components/AboutModal/AboutModal.test.js
```diff
@@ -5,207 +5,196 @@
...
+    IBM <span>{titleText}</span>
   </>
 );
+const versionNumber = `1.3.${uuidv4()}`;
 
-describe('AboutModal', () => {
-  test('renders title and content', () => {
-    render(<AboutModal open logo={logo} content={content} title={title} />);
+describe(componentName, () => {
+  it('renders a component AboutModal', () => {
```
matthewgallo, line 85: I noticed all of the test declarations changed from `test` to `it`. However I've seen other components use `test` in the library. Should we from this point forward stick to using `it`?

dcwarwick: Yes, that's the plan. They are synonyms, but Carbon use `it` so we may as well follow suit.

packages/experimental/src/components/AboutModal/AboutModal.test.js
dcwarwick: Ah, good catch. No, we should fix the accessibility failures! I meant to come back to that. Will fix and push.

matthewgallo (approved):

packages/experimental/src/components/AboutModal/AboutModal.test.js
dcwarwick: Fixed.

PR #493, opened 2021-03-19
Fixed AboutModal accessibility test failure and re-enable test

dcwarwick:
Contributes to #393

AboutModal failed an accessibility check for not labelling the dialog, which is actually behaviour inherited from ComposedModal. It turns out ComposedModal will label the dialog automatically from a `label` but not from a `title`. I don't understand the logic of that, but switching to use `label` is a simple way to resolve that accessibility issue.

#### What did you change?

AboutModal.js and .scss and .test.js

Ran all tests.


matthewgallo (approved):
Sorry @dcwarwick I realized after I approved your last PR it was going to merge automatically without this change because of the auto merge label

PR #501, opened 2021-03-22
Epic: Create Modal

sydrosa:
Contributes to #344 

This PR is for the Create Modal, in line with the C&CS create flow patterns.

## What will this achieve?
Prep `create modal` component for release review.

When component follows review guidelines.


AlexanderMelox:
I've noticed that the [designs](https://www.carbondesignsystem.com/community/patterns/create-flows/#modal) for the create modal don't have a close icon? To be consistent with every other component that creates something we should include this for the user. For example the sidepanel always has an x for close, are we going to have to change the prop for every single component to allow an x to close or not?

packages/cloud-cognitive/src/components/CreateModal/CreateModal.stories.js
```diff
@@ -0,0 +1,271 @@
...
+    styles,
+    docs: {
+      page: mdx,
+    },
+  },
+};
+
+const Template = ({ children, ...args }) => {
+  return (
+    <CreateModal {...args} open>
```
AlexanderMelox, line 44: I found a problem with this, if you leave the prop `open` here, in the storybook controls I can't change it. It will always stay open.

AlexanderMelox: I would just let {...args} do its thing, and in the default args for this story just set open:true.

packages/cloud-cognitive/src/components/CreateModal/CreateModal.stories.js
```diff
@@ -0,0 +1,271 @@
...
+  story: PropTypes.object,
+  storyInitiallyOpen: PropTypes.bool,
+  ...CreateModal.propTypes,
+};
+
+TemplateWithFormValidation.propTypes = {
+  story: PropTypes.object,
+  storyInitiallyOpen: PropTypes.bool,
+  ...CreateModal.propTypes,
+};
```
AlexanderMelox, line 150: You don't have to give the storybook stories prop types because they are not public-facing and just adds more code.

packages/cloud-cognitive/src/components/CreateModal/CreateModal.stories.js
```diff
@@ -0,0 +1,271 @@
...
+      />
+      <TextArea
+        id="3"
+        placeholder="Placeholder text"
+        labelText="Text area label"
+        helperText="Optional helper text"
+      />
+    </>
+  ),
+};
```
AlexanderMelox, line 264: I feel like the ways the stories are created here are very verbose and not how the docs mention creating them. doing it this way also made all the controls be a json output when everything should be controlled through the storybooks template API and not our own. 

https://storybook.js.org/docs/react/writing-stories/args

packages/cloud-cognitive/src/components/CreateModal/_storybook-styles.scss
```diff
@@ -0,0 +1,50 @@
...
+
+.bx--list-box__menu-icon {
+  top: $spacing-03;
+}
+
+.example-wrapper {
+  display: grid;
+  grid-gap: $spacing-05;
+  grid-template-columns: 1fr 1fr;
+}
```
AlexanderMelox, line 50: I noticed these styles are affecting some elements inside the modal. This file will only showup on storybook and not the library code. Should these be permanent styles to those components being used inside the modal?

packages/cloud-cognitive/src/components/CreateModal/CreateModal.stories.js
sydrosa: @AlexanderMelox that is correct -- it's supposed to stay `open` just as it is in the Carbon storybook for modal, found here: https://react.carbondesignsystem.com/?path=/story/components-composedmodal--default

packages/cloud-cognitive/src/components/CreateModal/_storybook-styles.scss
sydrosa: Good catch! Technically, the C&CS guidelines for forms inside of a modal are that they take up 100% width of the component. However, `DatePicker` is the one component that is _not_ 100% width, and breaks this rule. Since it is a C&CS guideline, but not a Carbon one, I had to add this additional code. Because this component is a "dumb" component, and I did not want to explicitly add a width to any date pickers in it, as this could be completely up to design.

packages/cloud-cognitive/src/components/CreateModal/CreateModal.stories.js
AlexanderMelox: Okay! I just found it weird when I tried clicking cancel in the first two stories for create modal and nothing happened. Maybe we should have somewhere in the story name like Opened Create modal?

packages/cloud-cognitive/src/components/CreateModal/_storybook-styles.scss
AlexanderMelox: Gotcha 👍🏻

packages/cloud-cognitive/src/components/CreateModal/CreateModal.test.js
```diff
@@ -0,0 +1,132 @@
...
+        key="form-field-1"
+        id="1"
+        labelText="Text input label"
+        helperText="Helper text goes here"
+        placeholder="Placeholder"
+      />
+    </CreateModal>
+  );
+
+describe(componentName, () => {
```
matthewgallo, line 52: This we were just talking about testing guidelines, I noticed you don't have a test that `additional HTML attributes are passed through to an appropriate node`. See [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/main/.github/CODE_GUIDELINES.md) for details and an example.

packages/cloud-cognitive/src/components/CreateModal/CreateModal.test.js
```diff
@@ -0,0 +1,132 @@
+/**
+ * Copyright IBM Corp. 2020, 2021
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { fireEvent, render, screen } from '@testing-library/react';
```
matthewgallo, line 9: Rather than `fireEvent.click`, our guidelines suggest using `userEvent` from `@testing-library/user-event`. This has been updated in `ExampleComponent.test.js` and is the recommended way to go now.

```js
import userEvent from '@testing-library/user-event';
```

packages/cloud-cognitive/src/components/CreateModal/_create-modal.scss
```diff
@@ -0,0 +1,72 @@
+//
```
matthewgallo, line 1: This file needs to use the import once approach. All of your sass should be wrapped in a mixin and you'll then include the mixin at the end of the file. Check out the `_example-component.scss` for reference.

packages/experimental/src/global/js/package-settings.js
```diff
@@ -0,0 +1,82 @@
+//
```
matthewgallo, line 1: This file shouldn't exist anymore. It is located in `packages/cloud-cognitive/global/js/package-settings.js` now

matthewgallo:
One more thing @sydrosa, don't forget to import your `_create-modal.scss` in `packages/cloud-cognitive/src/components/index.scss` and `CreateModal.js` in `packages/cloud-cognitive/src/components/index.js`. There was section about this that was just added to the [code guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/main/.github/CODE_GUIDELINES.md#organisation-and-files) this morning.

andrea-gm:
> I've noticed that the [designs](https://www.carbondesignsystem.com/community/patterns/create-flows/#modal) for the create modal don't have a close icon? To be consistent with every other component that creates something we should include this for the user. For example the sidepanel always has an x for close, are we going to have to change the prop for every single component to allow an x to close or not?

Close and Cancel are different. Close says you can come back to it. Cancel says you’re going to lose it. None of the Create components will have an X icon. This hasn't rolled up to Carbon yet, but it should soon. Anytime the action is Cancel, there should be no X icon.

packages/cloud-cognitive/src/components/CreateModal/CreateModal.stories.js
sydrosa: @AlexanderMelox we went over this in scrum and all agree that this isn't the best way to show the stories -- I went ahead and changed it :D

packages/cloud-cognitive/src/components/CreateModal/CreateModal.js
```diff
@@ -0,0 +1,167 @@
...
+  /**
+   * Specify an optional className to be applied to the modal root node
+   */
+  className: PropTypes.string,
+  /**
+   * The description of the CreateModal serves to provide more information about the modal.
+   */
+  description: PropTypes.node.isRequired,
+  /**
+   * Specifies a handler for disabling or enabling the primary button. This is important for form validation
```
matthewgallo, line 122: It might be more clear to say `Specifies a boolean for...`

packages/cloud-cognitive/src/components/CreateModal/CreateModal.js
```diff
@@ -0,0 +1,167 @@
...
+   * Returning `true` prevents the primary button from being clicked until required fields are completed.
+   */
+  disableSubmit: PropTypes.bool,
+  /**
+   * Specifies whether the submission of the modal can be clicked or not.
```
matthewgallo, line 127: It looks like this describes the `disableSubmit` prop, not `onClose`

matthewgallo (approved):
Just had two comments around doc gen prop descriptions, but everything else looks good. 👍

matthewgallo (approved):

Issue #434, opened 2021-03-11
[BUG] Side panel handleBlur throws error


AlexanderMelox:
## What package(s) are you using?

- [x] Cloud & Cognitive
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

https://user-images.githubusercontent.com/12755042/110857884-60bb1000-8287-11eb-930a-6a2646281ec5.mov

When clicking on an open on click for the Side panel it throws an error 

Looks like these are the problems that I got from the call stack
```
  at wrapFocus (https://h0eig.csb.app/node_modules/
carbon/ibm-cloud-cognitive-experimental/es/global/js/utils/wrapFocus.js:28:36

    at handleBlur (https://h0eig.csb.app/node_modules/
carbon/ibm-cloud-cognitive-experimental/es/components/SidePanel/SidePanel.js:205:29
```

 View the call stack 👇🏻 
<details>

```
Cannot read property 'call' of undefined
    at wrapFocus (https://h0eig.csb.app/node_modules/
carbon/ibm-cloud-cognitive-experimental/es/global/js/utils/wrapFocus.js:28:36
    at handleBlur (https://h0eig.csb.app/node_modules/
carbon/ibm-cloud-cognitive-experimental/es/components/SidePanel/SidePanel.js:205:29
HTMLUnknownElement.callCallback
https://h0eig.csb.app/node_modules/react-dom/cjs/react-dom.development.js:2746:18
Object.invokeGuardedCallbackDev
https://h0eig.csb.app/node_modules/react-dom/cjs/react-dom.development.js:2770:20
invokeGuardedCallback
https://h0eig.csb.app/node_modules/react-dom/cjs/react-dom.development.js:2804:35
invokeGuardedCallbackAndCatchFirstError
https://h0eig.csb.app/node_modules/react-dom/cjs/react-dom.development.js:2807:29
executeDispatch
https://h0eig.csb.app/node_modules/react-dom/cjs/react-dom.development.js:5339:7
processDispatchQueueItemsInOrder
https://h0eig.csb.app/node_modules/react-dom/cjs/react-dom.development.js:5359:11
processDispatchQueue
https://h0eig.csb.app/node_modules/react-dom/cjs/react-dom.development.js:5368:9
dispatchEventsForPlugins
https://h0eig.csb.app/node_modules/react-dom/cjs/react-dom.development.js:5376:7
eval
[...]

PR #511, opened 2021-03-23
fix(SidePanel): should fix focus trap which is breaking the side panel

matthewgallo:
Contributes to #434 

This will _hopefully_ fix the side panel focus trap bug that @AlexanderMelox found. Thanks @joshblack for the help 🙏 

#### What did you change?
Line [63-66 from wrapFocus](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/packages/cloud-cognitive/src/global/js/utils/wrapFocus.js#L63-L66) is trying to convert an array like node list to an array using `Array.slice.call`, but it should have been `Array.prototype.slice.call`, which can also be achieved by using the static method `Array.from`.

- Everything still functions as expected in the storybook for the side panel, however this bug has only been able to be reproduced when using the packaged version of the component.
- New tests for `wrapFocus.js` pass, `jest /utils/`

### DIFF
--- a/packages/cloud-cognitive/src/global/js/utils/__tests__/wrap-focus.test.js
+++ b/packages/cloud-cognitive/src/global/js/utils/__tests__/wrap-focus.test.js
@@ -1,0 +1,150 @@
+import wrapFocus from '../wrapFocus';
+
+/**
+ * Copyright IBM Corp. 2020, 2021
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+describe('wrapFocus', () => {
+  let node;
+  let spyInnerModal;
+  let spyButton0;
+  let spyButton2;
+
+  beforeEach(() => {
+    node = document.createElement('div');
+    node.tabIndex = '-1';
+    node.innerHTML = `
+      <button id="outer-preceding"></button>
+      <span
+        id="start-sentinel"
+        tabIndex="0"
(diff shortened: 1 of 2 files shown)


dcwarwick (changes requested):
Good catch, and thanks @joshblack for helping track this down.

I notice that we've reworked this logic from the carbon original, presumably to avoid a dependency on lodash for one function alone. That's fine, but I also notice that carbon have a set of tests for wrapFocus.js, and when I run those on the code before this change, the tests fail because of the Array.slice mistake. This change fixes them(*). I feel there's a moral in this story somewhere :-) So I suggest we copy carbon's tests over too -- the file wrapFocus-test.js can just be dropped in to our own /packages/cloud-cognitive/src/global/js/utils/\_\_tests\_\_ folder and will start doing its job right away.

(*) when I say it fixes the tests, it fixes all but one. carbon have added the following lines into wrapFocus.js, with a test case to go along with them. I suggest we also add these lines into our wrapFocus, even if we don't need this function, just so we continue to align with the wrapFocus.js in carbon for easier future comparisons. If these lines are added to our version, all the tests then pass.

image

matthewgallo:
Thanks @dcwarwick for the feedback. I added back the last argument for the floating menus and added the tests for this utility from carbon and made sure that everything passes (`jest /utils/`). I definitely agree we should have had these tests for this helper 😅

dcwarwick (approved):
Looks good - many thanks

Issue #510, opened 2021-03-23
Flip in-progress and running icons


sydrosa:
More information here: 

https://github.ibm.com/CDAI-UX-Engineering/uxe-agency-work/issues/98

PR #512, opened 2021-03-23
feat: animation for in-progress icon

sydrosa:
Contributes to #510 

This PR addresses changes suggested by @Levinson1 and @carrenelloyd where we differentiate two Status Icons, `in-progress` and `running` from the "refresh/reload" icons by having it flip clockwise (previously it was animated and pointing counter clockwise).

Linking the two other issues across `/pal` and `uxe-agency`:
[pal](https://github.ibm.com/CDAI-design/pal/issues/443)
[uxe-agency](https://github.ibm.com/CDAI-UX-Engineering/uxe-agency-work/issues/98)

#### What did you change?
`_status_icon.scss`

Added additional transform properties to the two mentioned icons.

Tested locally in dev environment and ran test suite

### DIFF
--- a/packages/cloud-cognitive/src/components/StatusIcon/_status-icon.scss
+++ b/packages/cloud-cognitive/src/components/StatusIcon/_status-icon.scss
@@ -76,10 +76,10 @@ $block-class: #{$pkg-prefix}-status-icon;
 
 @keyframes rotating {
   0% {
-    transform: rotate(360deg);
+    transform: scaleY(-1) rotate(360deg);
   }
   100% {
-    transform: rotate(0deg);
+    transform: scaleY(-1) rotate(0deg);
   }
 }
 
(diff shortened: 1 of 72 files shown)


packages/cloud-cognitive/src/components/StatusIcon/_status-icon.scss
```diff
@@ -112,6 +112,10 @@ $block-class: #{$pkg-prefix}-status-icon;
             animation: rotating 8s infinite linear;
```
matthewgallo, line 112: For `infinite` animations it might be worth adding a `prefers-reduced-motion` check. See [here](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)

```css
@media (prefers-reduced-motion) {
  .some-icon-with-animation {
    animation: rotating 8s infinite linear;
  }
}
```

matthewgallo (approved):

packages/cloud-cognitive/src/components/StatusIcon/_status-icon.scss
sydrosa: I added a media query that removes the animation if a user's settings are set to reduced motion:

```
...
 @each $themekey in $themes {
      .#{$block-class}--#{$theme}.#{$block-class}--#{$themekey}-#{$icon} {
        svg {
          @if $icon == in-progress {
            @media (prefers-reduced-motion) {
              animation: none;
            }
            // stylelint-disable-next-line carbon/motion-token-use, declaration-property-unit-whitelist
            animation: rotating 8s infinite linear;
            // stylelint-disable-next-line carbon/theme-token-use
            fill: clr($icon, $theme);
          } @else if $icon == running {
            transform: scaleY(-1);
...
```

Issue #69, opened 2020-08-05
User profile images


carrenelloyd:
Catalog # 27

[Link to Design ](https://pages.github.ibm.com/cdai-design/pal/patterns/user-profile-images/)

**Maintainers**
Marion Brülls

## Tasks

- [x] Initial review of design/existing code

### Working in experimental package

- [x] Create a shell component which
  - [x] Create ComponentName/ComponentName.js component file that renders props
        in raw form (no styling or sub components, plain text OK)
  - [x] Adds prop-types with comments for DocGen to
        ComponentName/ComponentName.js
  - [x] Add Doc file e.g. ComponentName.mdx with a brief description of the
        component.
- [ ] First engineering review/playback

### Continuing in experimental package

- [ ] Storybook - stories to cover design scenarios, story with long, min and
      max content, etc
- [ ] Doc tab in storybook derived from design docs with embedded stories where
      appropriate
- [ ] Code to include prop-types with comments for automated DocGen
      documentation
- [ ] Tests (written in React Test Library) to cover behavior from the
      documented in the components mdx file.

### Review and promote

- [ ] Design review
- [ ] Add design review issues addressed, issues raised or discounted.
- [ ] Engineering review - Submit PR to promote component to production package
- [ ] All engineering issues addressed, issues raised or discounted.

lee-chase:
Blocked as the status of the design it was set to 'status: refining' by @carrenelloyd

Invited Marion Brülls to join the conversation here.

Levinson1:
Design completed and live

lee-chase:
Dev thoughts.

Would 'Avatar' be a better name for the component? Calling it UserProfileImage may discourage group use and strictly speaking it may not be an image.

lee-chase:
2. IBM Profiles ask the user to provide
- First name
- Inital (optional)
- Last name
-Display name

Would it not be better to use the first two capitalized letters should be used from the display name.

qbi11y:
@lee-chase @carrenelloyd who do I need to speak with to come up to speed on this issue?

andrea-gm:
[...]

PR #519, opened 2021-03-24
69 user profile image

qbi11y:
Contributes to #69
PR adds the initial structure for the User Profile Image component.


packages/cloud-cognitive/src/components/UserProfileImage/UserProfileImage.js
```diff
@@ -0,0 +1,168 @@
...
+  Group24,
+  Group32,
+} from '@carbon/icons-react';
+
+//import { expPrefix } from '../../global/js/settings';
+import cx from 'classnames';
+
+const blockClass = `${pkg.prefix}-user-profile-avatar`;
+
+export const UserProfileImage = !pkg.isComponentEnabled(componentName)
```
dcwarwick, line 30: Sorry, the Canary mechanism has changed, after we found this was breaking Storybook. Have a look at ExampleComponent to see what's needed. The good news is it's a lot simpler :-) Basically:
 * no need to import Canary
 * use `export let` rather than `export const`
 * no other gorp needed -- straight into the component definition
 * then add the following line further down the file:
```js
// Return a placeholder if not released and not enabled by feature flag
UserProfileImage = pkg.checkComponentEnabled(UserProfileImage, componentName);
```

dcwarwick (changes requested):

packages/cloud-cognitive/src/components/UserProfileImage/UserProfileImage.js
dcwarwick: Actually, let's merge this and I'll update the file -- I went through doing all the other components so I'm practised :-)

dcwarwick (approved):

Issue #525, opened 2021-03-25
Card refactor experiment


davidmenendez:
Based on discussions we've had regarding the Card component we decided to try a pattern of having a base component and named themed components.

The benefit of doing this would be 1. PropTypes can be assigned more accurately based on the theme or mode (expressive vs productive) and 2. A component can be imported by name and doesn't require a prop to establish a mode. this may be a more clear way to import a specific component

```
import { Card };
...
return <Card productive {...props}
```
vs
```
import { ProductiveCard };
...
return <ProductiveCard {...props} />
```

the purpose of this issue is to open up discussion and dialog regarding pros and cons to having this sort of pattern and the possible benefits to implementing this pattern in this library.

PR #526, opened 2021-03-25
feat: card refactor

davidmenendez:
Contributes to #525 

This PR addresses refactoring the Card component to use this base & themed component pattern. Now the card component has been broken up into three separate ones. `Card`, `ExpressiveCard`, and `ProductiveCard` where the themed components are just wrappers for card.

```
import { Card }
...
export const ExpressiveCard = ({ props specific to the expressive card }) => <Card {...props} />
```

So the main idea here is to split up the components and their dependencies based on their specific use case. I removed the card from storybook because I think if we went this route we would encourage users to use the themed component and not the base, which is also why I removed the prop comments from card. There is some duplication of props and prop comments between the two themed components, but since we're relying on storybook for documentation I think it's just something we have to live with even though code duplication isn't ideal, but now when you look at the storybook for each component they now display the props that they actually depend on. previously when everything was in a single place we had received some feedback that looked like

> why is X a control in expressive when X is only applicable to productive

Now that's no longer a problem.

Using this pattern now all the shared code is found in the `Card` folder and anything specifically related to the themed components is found in their respective folder, which is mainly css. There is also some general cleanup going on here.


lee-chase:
@dcwarwick looks like you did not complete the review on this PR.

Sorry @davidmenendez I think we said we'd submitted a review.

packages/cloud-cognitive/src/components/Card/_card.scss
```diff
@@ -1,9 +1,14 @@
+//
+// Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2:
```suggestion
// Copyright IBM Corp. 2020, 2021
```

packages/cloud-cognitive/src/components/Card/_storybook-styles.scss
```diff
@@ -1,33 +1,19 @@
+//
+// Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2:
```suggestion
// Copyright IBM Corp. 2020, 2021
```

packages/cloud-cognitive/src/components/ExpressiveCard/ExpressiveCard.js
```diff
@@ -0,0 +1,146 @@
+import React from 'react';
```
dcwarwick, line 1:
```suggestion
/**
 * Copyright IBM Corp. 2020, 2021
 *
 * This source code is licensed under the Apache-2.0 license found in the
 * LICENSE file in the root directory of this source tree.
 */

import React from 'react';
```

packages/cloud-cognitive/src/components/ExpressiveCard/ExpressiveCard.test.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2:
```suggestion
 * Copyright IBM Corp. 2020, 2021
```

packages/cloud-cognitive/src/components/ExpressiveCard/_expressive-card.scss
```diff
@@ -0,0 +1,13 @@
+//
+// Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2:
```suggestion
// Copyright IBM Corp. 2020, 2021
```

packages/cloud-cognitive/src/components/ExpressiveCard/_index.scss
```diff
@@ -0,0 +1,10 @@
+//
+// Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2:
```suggestion
// Copyright IBM Corp. 2020, 2021
```

packages/cloud-cognitive/src/components/ExpressiveCard/_storybook-styles.scss
```diff
@@ -0,0 +1,23 @@
+//
+// Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2:
```suggestion
// Copyright IBM Corp. 2020, 2021
```

packages/cloud-cognitive/src/components/ExpressiveCard/index.js
```diff
@@ -0,0 +1,8 @@
+//
+// Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2:
```suggestion
// Copyright IBM Corp. 2020, 2021
```

packages/cloud-cognitive/src/components/ProductiveCard/ProductiveCard.js
```diff
@@ -0,0 +1,117 @@
+//
+// Copyright IBM Corp. 2020, 2020
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
```
dcwarwick, line 6:
```suggestion
/**
 * Copyright IBM Corp. 2020, 2021
 *
 * This source code is licensed under the Apache-2.0 license found in the
 * LICENSE file in the root directory of this source tree.
 */
```

packages/cloud-cognitive/src/components/ProductiveCard/ProductiveCard.stories.js
```diff
@@ -0,0 +1,102 @@
+//
+// Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2:
```suggestion
// Copyright IBM Corp. 2020, 2021
```

packages/cloud-cognitive/src/components/ProductiveCard/ProductiveCard.test.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2:
```suggestion
 * Copyright IBM Corp. 2020, 2021
```

packages/cloud-cognitive/src/components/ProductiveCard/_index.scss
```diff
@@ -0,0 +1,10 @@
+//
+// Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2:
```suggestion
// Copyright IBM Corp. 2020, 2021
```

packages/cloud-cognitive/src/components/ProductiveCard/_productive-card.scss
```diff
@@ -0,0 +1,34 @@
+@import '../../global/styles/carbon-settings';
```
dcwarwick, line 1:
```suggestion
//
// Copyright IBM Corp. 2020, 2021
//
// This source code is licensed under the Apache-2.0 license found in the
// LICENSE file in the root directory of this source tree.
//

@import '../../global/styles/carbon-settings';
```

packages/cloud-cognitive/src/components/ProductiveCard/_storybook-styles.scss
```diff
@@ -0,0 +1,20 @@
+//
+// Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2:
```suggestion
// Copyright IBM Corp. 2020, 2021
```

packages/cloud-cognitive/src/components/ProductiveCard/index.js
```diff
@@ -0,0 +1,8 @@
+//
+// Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2:
```suggestion
// Copyright IBM Corp. 2020, 2021
```

dcwarwick (changes requested):

davidmenendez:
@dcwarwick 

> We don't need to explicitly destructure and rebuild props, because we can just collect the entire set of props and pass them through. 

I was thinking about this one a lot. my thinking is that this is another step to ensure the consumer is unable to pass an undesirable prop (accidentally or otherwise). since they're just wrappers for the card component it would still be possible to send productive only props to expressive and vice versa if we simply pass it as `...props`. it does mean more lines on our end, but i think going about it this way helps ensure less bugs and unwanted behavior for users.

dcwarwick:
@davidmenendez It's a good thought. The thing is, one of our guidelines calls for us to pass all 'extra' props through as HTML attributes onto the main HTML element, which the Carbon components also do. That means we need to include a spread `...rest` at the end of the props destructuring, and then add `...rest` when passing them through to the wrapped component. In fact, that was the review comment I originally started to make -- then we realised that if we're doing that then we might as well drop all the explicit destructuring, since any props passed in that we don't explicitly destructure will get passed across anyway.

davidmenendez:
that's a good point. i hadn't considered that. what do you think about creating a utility that strips undesired props? something like

```
/**
 * props should be an object of properties
 * blacklist should be an array of undesired property names
 */
const stripProps = (props, blacklist) => // object not containing blacklisted props
```

i think doing something like this will help cut down on bugs and consumer bug reports because ultimately with this kind of pattern there is a certain level of prop sharing between the base component and the wrapper components

davidmenendez:
@dcwarwick while I was able to add this new utility function to solve the prop problem I don't think we can remove the proptypes since ultimately these changes are here to benefit storybook and our documentation. so we will still have some redundancy there, but better safe than sorry, right? 😂

dcwarwick (approved):

packages/cloud-cognitive/src/global/js/utils/props-helper.js
```diff
@@ -0,0 +1,18 @@
...
+// Copyright IBM Corp. 2020, 2021
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+// helpers functions for component props
+
+// remove undesired props from component props object
+export const stripUnwantedProps = (props, blacklist) => {
```
lee-chase, line 11: @dcwarwick @davidmenendez  FYI 526 introduces a second pattern for stripping props, we should probably pick one and go with it.
This is from actionBarItem
const reservedProps = [
  'hasIconOnly',
  'kind',
  'size',
  'tooltipPosition',
  'tooltipAlignment',
  'type',
];
// Base props on Carbon Button
const propTypes = { ...Button.propTypes };
const defaultProps = { ...Button.defaultProps };
// Remove reserved props
reservedProps.forEach((prop) => {
  delete propTypes[prop];
  delete defaultProps[prop];
});

dcwarwick (changes requested):

packages/cloud-cognitive/src/components/ExpressiveCard/ExpressiveCard.test.js
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { ExpressiveCard } from '.';
+import '../../enable-all'; // must come before component is imported (directly or indirectly)
```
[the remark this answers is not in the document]
davidmenendez: done

dcwarwick (dismissed):

davidmenendez:
@lee-chase @dcwarwick i've made a small additional update to move each card to a shared `Cards` folder in storybook to improve organization. 

Screen Shot 2021-03-31 at 4 55 18 PM

These storybook changes have been approved by design and they are very happy with the improved UX.

dcwarwick (approved):

PR #529, opened 2021-03-26
fix: ideStoryError

lee-chase:
Remove IDESlideOverPanel browser console error from story

#### What did you change?

Added id to textinput and textarea components in story

Ran up storybook


matthewgallo (commented):
The textinput can be a self closing tag. And do you think we need to add some kind of prefix given that `text-id` is a somewhat generic id, just so we avoid conflicting ids in the future?

dcwarwick (approved):

lee-chase:
> The textinput can be a self closing tag. And do you think we need to add some kind of prefix given that `text-id` is a somewhat generic id, just so we avoid conflicting ids in the future?

Deliberately making a minimal change as this is CDAI code.

PR #551, opened 2021-03-31
chore(SidePanel): update per review guidance to prep for review

matthewgallo:
Contributes to #490 

This preps the side panel for review. Restructured class names usage based on our guidelines, sass updates (for `import-once`), forward ref capability added, tests to ensure props spreading and passing a ref work.

#### What did you change?
`SidePanel.js`
`SidePanel.test.js`
`_side-panel.scss`
`utils.js`
Everything still works in storybook and all tests for this component pass `jest /SidePanel/`

### DIFF
--- a/packages/cloud-cognitive/src/components/SidePanel/utils.js
+++ b/packages/cloud-cognitive/src/components/SidePanel/utils.js
@@ -1,5 +1,6 @@
 export const changeArrayPosition = (arr, originalPosition, newPosition) => {
-  let cutOut = arr.splice(originalPosition, 1)[0];
-  arr.splice(newPosition, 0, cutOut);
-  return arr;
+  const newArr = [...arr];
+  const cutOut = newArr.splice(originalPosition, 1)[0];
+  newArr.splice(newPosition, 0, cutOut);
+  return newArr;
 };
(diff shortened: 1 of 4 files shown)


packages/cloud-cognitive/src/components/SidePanel/utils.js
```diff
@@ -0,0 +1,5 @@
+export const changeArrayPosition = (arr, originalPosition, newPosition) => {
+  let cutOut = arr.splice(originalPosition, 1)[0];
```
davidmenendez, line 2: we should try not to mutate data whenever possible and unfortunately `splice` is not immutable. i'd recommend just adding this one additional line to create a new array
```
const changeArrayPosition = (arr, originalPosition, newPosition) => {
  const newArr = [...arr];
  const cutOut = arr.splice(originalPosition, 1)[0];
  arr.splice(newPosition, 0, cutOut);
  return arr;
};
```
and now the original data is still intact.

davidmenendez (changes requested):

packages/cloud-cognitive/src/components/SidePanel/utils.js
matthewgallo: Fixed [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/551/commits/e99dd8850916e6c6d11bee807590610ea2a23e8e)

davidmenendez (approved):

PR #552, opened 2021-03-31
Update code guidelines to simpler handling of className

dcwarwick:
Contributes to #393

`cx({ [className]: className })` turns out to be a tadge unnecessary, so `cx(blockClass, className, ...)` is simpler and cleaner.

#### What did you change?

CLI, ExampleComponent, AboutModal, and Canary

Ran build and tests.


lee-chase (changes requested):
There really ought to be an option to approve with feedback. Feel free to merge once accepting or dismissing the suggestion.

PR #559, opened 2021-04-05
Create ActionSet to do action button layouts

dcwarwick:
Contributes to #487

Creates a new ActionSet component based on the primary actions logic from SidePanel. This lays out action buttons, and also provides validation logic for the structured prop. I also looked through the layout logic and streamlined it a little. NB I have left most of the style rules in SidePanel, as I am not sure how many of them will be general and how many are specific to the SidePanel design: I may move some of those style rules into the ActionSet when I investigate how to use the ActionSet for the TearSheet action buttons.

#### What did you change?

SidePanel files, including tests and stories. Also created ActionSet files, including tests, and added action-set scss to main component roll-up.

Ran build, tests, and ran storybook and checked multiple configurations. I may not have checked every relevant configuration, though, so do verify this!

### DIFF
--- a/packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
+++ b/packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
@@ -185,21 +185,16 @@ export let SidePanel = React.forwardRef(
       }
     }, [slideIn, pageContentSelector, placement, shouldRender, size]);
 
-    const setSizeClassName = (panelSize, actions) => {
-      let sizeClassName;
-      if (!actions) {
-        sizeClassName = `${blockClass}__container-`;
-      } else {
-        sizeClassName = `${blockClass}__actions-`;
-      }
+    const setSizeClassName = (panelSize) => {
+      let sizeClassName = `${blockClass}__container-`;
       switch (panelSize) {
-        case 'extraSmall':
+        case 'xs':
           return (sizeClassName = `${sizeClassName}-extra-small`);
-        case 'small':
+        case 'sm':
           return (sizeClassName = `${sizeClassName}-small`);
-        case 'medium':
+        case 'md':
           return (sizeClassName = `${sizeClassName}-medium`);
-        case 'large':
+        case 'lg':
           return (sizeClassName = `${sizeClassName}-large`);
         case 'max':
           return (sizeClassName = `${sizeClassName}-max`);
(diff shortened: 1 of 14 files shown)


packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
```diff
@@ -577,11 +447,16 @@ SidePanel.propTypes = {
    * Sets the primary action buttons for the side panel
    */
   primaryActions: PropTypes.oneOfType([
-    validateActions(),
+    ActionSet.validateActions(
```
matthewgallo, line 450: Should we add all of the possible sizes of the side panel here? Looks like `large` and `max` are missing

dcwarwick: I didn't want to have the ActionSet explicitly know about extraSmall/small/medium/large/max, as these apply to the SidePanel but might not apply in other places where the ActionSet is useful. So instead I abstracted the decisions that were being made into two settings:
* When validating the actions, you were applying tighter constraints for extraSmall/small/medium sizes (limit of three buttons, no mixing of ghost buttons), so I gave the validator a parameter which selects whether to use the 'small' constraints or the normal constraints. We use the 'small' constraints when props.size is extraSmall/small/medium. I've added a comment to try and make that rather mysterious line a bit clearer!
* When you're going to stack the buttons, you reverse the order of them, so I gave ActionSet a 'reverse' prop.
These mean that ActionSet doesn't know specifically about when to switch the order or when to apply the tighter constraints -- that's SidePanel's decision.

dcwarwick: On further examination of the Tearsheet and the styling, I think it will be better to replace the `reverse` prop and the small parameter with actually having `size` on the ActionSet, so the SidePanel just passes its own size through. That also enables ActionSet to take the scss from current SidePanel. I will update this PR when I've done that.

dcwarwick:
OK, this is now ready for review again :-)

There are also stories for ActionSet now.

@matthewgallo when discussing this with @lee-chase yesterday he pointed out that Carbon components variously use some or all of 'xs', 'sm', 'md', 'lg', 'xl' and 'max' for sizes, so that's what I put on ActionSet, and for simplicity updated SidePanel to use those too -- we're not sure we like these short names, but it aligns us with Carbon so meh.

matthewgallo (approved):

matthewgallo:
I like the idea of sticking to Carbon's size naming, good idea!

Issue #382, opened 2021-02-26
Cards: Productive type - Design feedback


andrea-gm:

davidmenendez:
Productive card storybook can be found here http://localhost:3000/?path=/story/cloud-cognitive-canary-card--productive

currently there are only two stories- one for the default mode and one with overflow actions. everything else should be controllable from the controls.

andrea-gm:
@davidmenendez you mentioned you're waiting on the designer, correct? Would it make sense to move this issue into "On hold" or "In progress" ?

JohannaSuarezKoval:
Thanks again for the hard work with this component!


The productive card is a bit more complex than the expressive, so we need to have some variations. 

- The productive card needs to be a separate component from the expressive in the storybook. They are two different types and used in different contexts. Therefore the controls are different for each. For example, the expressive card's action placement doesn't change top or bottom like a Productive card. 
Can we add a "Card" folder and inside have "Expressive Card" and "Productive Card" components separated. 

- There are three different action bar placements. 
By default, all primary actions should be at the top, with the option of having a supplemental action bar at the bottom (just a ghost to the left)
The bottom bar for both the sumplemental and primary actions is optional for just special cases. 
So, can we have two variations like `Standard action bar` with a control to turn on and off the supplemental bar (the bottom one with the left ghost button).  And `Complex bottom bar` variation with all actions at the bottom. 

- The icons should be icon buttons in the productive card.

- The productive card also has similar variations in the title block: 
Title only
With label
With description
Label only

- There are also three main variations in the content that change the margin of the content block.
**Regular content:** just text
<img width="591" alt="Screen Shot 2021-03-18 at 10 57 48 AM" src="https://user-images.githubusercontent.com/5809654/111647450-e20b2900-87d8-11eb-9c5d-ec5dbc532f50.png">

**Chart content:** content starts with a chart in which there is an 8px top margin
<img width="590" alt="Screen Shot 2021-03-18 at 10 57 53 AM" src="https://user-images.githubusercontent.com/5809654/111647479-e6cfdd00-87d8-11eb-9174-1ef5cfacdf80.png">

**Stacked content:** These are clickable content blocks with dividers. Margin changes too
<img width="591" alt="Screen Shot 2021-03-18 at 10 57 58 AM" src="https://user-images.githubusercontent.com/5809654/111647530-f2230880-87d8-11eb-897c-0e2769ce1ac8.png">


[...]

PR #560, opened 2021-04-05
Card design feedback

davidmenendez:
Contributes to #382

### DIFF
--- a/packages/cloud-cognitive/src/components/ExpressiveCard/ExpressiveCard.stories.js
+++ b/packages/cloud-cognitive/src/components/ExpressiveCard/ExpressiveCard.stories.js
@@ -122,6 +122,7 @@ WithActionIcon.args = {
       id: '1',
       icon: ArrowRight24,
       onClick: () => {},
+      iconDescription: 'Next',
     },
   ],
   primaryButtonText: '',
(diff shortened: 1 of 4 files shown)


packages/cloud-cognitive/src/components/ExpressiveCard/ExpressiveCard.stories.js
```diff
@@ -120,7 +120,8 @@ WithActionIcon.args = {
   actionIcons: [
     {
       id: '1',
-      icon: <ArrowRight24 onClick={() => {}} />,
+      icon: ArrowRight24,
```
matthewgallo, line 123: We should add an `iconDescription` prop here as well since that is something that is used in the button [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/560/files#diff-8ef3dc50c7bfa3e42e6d8aec034961f6e564544c5e691af803b7408512ec81a6R75).

matthewgallo: Nevermind, that's productive only I'm just seeing ^

matthewgallo (approved):

packages/cloud-cognitive/src/components/ExpressiveCard/ExpressiveCard.stories.js
davidmenendez: you're only half wrong here haha. i did need to add an `aria-label` prop to the icon, which i've updated 👍

matthewgallo (approved):

packages/cloud-cognitive/src/components/ExpressiveCard/ExpressiveCard.stories.js
matthewgallo: haha ok, glad that worked out

PR #561, opened 2021-04-05
fix: card design feedback

davidmenendez:
Contributes to https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/382#issuecomment-813592978 & https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/382#issuecomment-813605208

please note the usage of `stylelint-disable` i was unable to use a carbon token with a negative value. i tried all the different ways i could think of to get it to evaluate to a negative value, but to no avail. if you know how to get a negative number using a carbon token please let me know.

### DIFF
--- a/packages/cloud-cognitive/src/components/Card/Card.js
+++ b/packages/cloud-cognitive/src/components/Card/Card.js
@@ -86,6 +86,13 @@ export let Card = ({
     return icons;
   };
 
+  const getClickableProps = () => ({
+    onClick,
+    onKeyDown: onClick,
+    role: 'button',
+    tabIndex: '0',
+  });
+
   const getCardProps = () => {
     const clickable =
       (!!onClick && !productive) ||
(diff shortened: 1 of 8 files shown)


dcwarwick:
@davidmenendez In AboutModal, Simon used `top: calc(-1 * #{$spacing-09});`, which the linter seems to be ok with. Would that work for you here?

davidmenendez:
@dcwarwick that did work! i feel like i tried something like this and it didn't work for some reason. i must have missed a space or something. i know some of the css functions are very picky about that. thanks!

dcwarwick:
@davidmenendez My only other comment would be to flag the prop name 'onClickZone'. Props named 'onXxx' are generally the actual handlers -- I wonder if 'onClickZone' should be called 'clickZone', or similar? In any case, I'll approve this pr now and you can give the name some thought.

dcwarwick (approved):

davidmenendez:
@dcwarwick updated

dcwarwick (dismissed):

dcwarwick (approved):

Issue #64, opened 2020-08-05
Side panel


carrenelloyd:
Catalog # 21

[Link to Design ](https://pages.github.ibm.com/cdai-design/pal/components/slide-in-panel/)

See also
https://pages.github.ibm.com/ibmcloud/pal/components/side-panel/usage/

## Tasks

- [x] Initial review of design/existing code

### Working in experimental package

- [x] Create a shell component which
  - [x] Create ComponentName/ComponentName.js component file that renders props
        in raw form (no styling or sub components, plain text OK)
  - [x] Adds prop-types with comments for DocGen to
        ComponentName/ComponentName.js
  - [x] Add Doc file e.g. ComponentName.mdx with a brief description of the
        component.
- [x] First engineering review/playback

### Continuing in experimental package

- [x] Storybook - stories to cover design scenarios, story with long, min and
      max content, etc
- [x] Doc tab in storybook derived from design docs with embedded stories where
      appropriate
- [x] Code to include prop-types with comments for automated DocGen
      documentation
- [x] Tests (written in React Test Library) to cover behavior from the
      documented in the components mdx file.

### Review and promote

- [x] Design review
- [x] Add design review issues addressed, issues raised or discounted.
- [x] Engineering review - Submit PR to promote component to production package
- [x] All engineering issues addressed, issues raised or discounted.

lee-chase:
Blocked as the status of the design it was set to 'status: refining' by @carrenelloyd

vikkidpaterson:
@carrenelloyd what was the reason behind the refining status tag - do you have a link to anything around the refinements underway?

lee-chase:
@gatfora @carrenelloyd wondering what the state of the design is here?

Levinson1:
Design is complete and published
https://pages.github.ibm.com/cdai-design/pal/components/side-panel/usage

andrea-gm:
@matthewgallo will meet with the design owner Kacie to get design approval

andrea-gm:
@matthewgallo can you provide an update from your meeting with Kacie?

matthewgallo:
Yes definitely, here are some of my notes:
```
1. Margin right on title increase to $spacing-07
2. Add `title` html attribute to title in side panel
3. Add underline to bottom of fixed section (1px solid $decorative)
4. Allow size of panel to be controlled by Storybook controls
5. Add optional labelText prop with label-01 type token
   - bottom padding $spacing-02
  ```

[...]

PR #568, opened 2021-04-07
PR for #64: add side panel subtitle micro interaction

matthewgallo:
Contributes to #64 

In the design for the side panel component, there is a micro interaction in which the subtitle animates up out of the view when scrolling occurs (see attached gif). This PR adds that interaction.
side-panel-scroll-header

#### What did you change?
`SidePanel.js`
`SidePanel.stories.js`
`_side-panel.scss`

Verified in storybook


lee-chase:
In the design, the subtitle fades and moves with scroll. 

In the implementation, the subtitle jumps off the screen at the same time as the title morphs from its larger size to smaller. This detracts for the title transformation making the whole thing feel a bit rushed.

Should the subtitle just fade and disappear under the title with scroll?

lee-chase (changes requested):
See comment, probably one for the designer but the interaction is different to that of the title and we are aiming for consistency.

davidmenendez (dismissed):

matthewgallo:
Closing in favor of #585

matthewgallo:
The title animation issue will be addressed in a future PR

PR #571, opened 2021-04-07
test(styles): add styles public API test

SimonFinney:
**Affected issue -** contributes to #577

As part of an investigation into integrating Carbon for Cloud & Cognitive CSS into Security applications already containing both an installation of Carbon and Carbon for IBM Security, a number of style conflicts and duplications were brought to the surface.

I will be providing some additional requirements and potential recommendations over the next few weeks, but as a first step I'd like to propose a workflow for better identifying and mitigating unwanted side effects from the CSS distributable of this library - moving forward, it will enable the team to understand the impact of any style additions and removals (the public API) and have better control over them in pull requests, [for example](https://github.com/SimonFinney/ibm-cloud-cognitive/compare/css-snapshot...SimonFinney:remove-carbon-utility).

It's worth noting that this PR only contains a test to cover the public API for the package's styles - in the future, I recommend having a similar safeguard for the [JavaScript output](https://github.com/carbon-design-system/carbon/blob/main/packages/react/__tests__/PublicAPI-test.js).

#### Proposed changes

- Add styles public API test
- Add initial styles snapshot
- Update contributing guidelines to accommodate workflow change

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -56,10 +56,14 @@ See example component enabled via feature flags on
 For more information on how this affects components see
 [CANARY_STRUCTURE.md](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/CANARY_STRUCTURE.md).
 
-## 4. Test your JavaScript code
+## 4. Test your code
 
 If you're contributing to our JavaScript code, test your changes by running
 `yarn test`.
+
+If you're contributing to our styles, the continuous integration check will fail
+because of a public API snapshot change. Update the snapshot by running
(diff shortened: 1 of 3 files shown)


dcwarwick:
@SimonFinney I'm not quite sure what this is trying to do, and it looks like it will simply cause virtually every build to fail so the team will end up running --update-snapshot on every run. All our PRs also really should relate to an open issue. Can you open an issue to explain the pain points, and what it is that is needed, so we can have a bit of a discussion around ways to address it? This PR can then be tagged as contributing to that issue.

dcwarwick:
These changes have now been incorporated into #595, so closing this one.

Issue #580, opened 2021-04-08
Side panel: issue with panel height 


matthewgallo:
## What package(s) are you using?

- [x] Cloud & Cognitive

## Detailed description

> Describe in detail the issue you're having.
When a side panel has content that does not span the full height of the panel, the actions container element has unintended style issues.

> Is this issue related to a specific component?
SidePanel

> What did you expect to happen? What happened instead? What would you like to
> see changed?
The actions container element should not have height issues, no matter how much content is in the panel.
> What browser are you working in?
Chrome
> What version of the @carbon/ibm-cloud-cognitive packge are you using?
Latest

## Steps to reproduce the issue

1. Use side panel with small amount of content
2. See styling issues

<img width="1714" alt="Screen Shot 2021-04-08 at 4 17 53 PM" src="https://user-images.githubusercontent.com/10215203/114093719-740eca80-9889-11eb-99f3-83407bc4ebac.png">

matthewgallo:
https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/585

matthewgallo:
We need to better account for how much bottom padding the main content gets to properly account for the action buttons. In addition to not rendering the action set container if there are no buttons, currently that element is still rendered and it's empty but has a top border which should not be the case.

matthewgallo:
https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/593

matthewgallo:
Complete

PR #585, opened 2021-04-09
PR for #580: side panel height issue

matthewgallo:
**#568 Should be merged first**

Contributes to #580 

Fixes the positioning of the actions container.

#### What did you change?
`SidePanel.js`
`_side-panel.scss`
`SidePanel.stories.js`
Storybook, addition of new side panel story with minimal content


packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
```diff
@@ -91,25 +91,55 @@ export let SidePanel = React.forwardRef(
       }
     }, [selectorPrimaryFocus, open, animationComplete]);
 
-    // Title animaton
+    // Title and subtitle animaton
     useEffect(() => {
       if (open && animateTitle && animationComplete) {
         const sidePanelOuter = document.querySelector(`#${blockClass}-outer`);
+        const sidePanelSutitleElement = document.querySelector(
```
dcwarwick, line 98: Sorry to be the pedant in the room :-) but can we fix this to `sidePanelSubtitleElement`?! I haven't done it as a suggestion here because it occurs in a number of places.

packages/cloud-cognitive/src/components/SidePanel/_side-panel.scss
```diff
@@ -93,14 +93,16 @@ $max-panel-size: 75%; // set max-width on max panels to 75%
   }
 
   .#{$block-class}__container {
+    --#{$block-class}--subtitle-opacity: 1;
+    --#{$block-class}--title-font-size: 1.25rem;
```
dcwarwick, line 97: Should we be getting this value from a carbon token?

dcwarwick (changes requested):

packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
matthewgallo: Doh, sorry about that. Yes

packages/cloud-cognitive/src/components/SidePanel/_side-panel.scss
matthewgallo: Ideally yes, but the font size for the title is now controlled by this css variable which gives us a smoother title animation. And I can't set the default value of this css variable to a sass variable.

ie this does not work unfortunately

```scss
--#{$block-class}--title-font-size: $productive-heading-03;
```

dcwarwick (approved):

Issue #547, opened 2021-03-31
Review fixes for StatusIcon


dcwarwick:
## What will this achieve?

To complete the review of StatusIcon:

- [x] Add DocGen comments for the props.
- [x] Rename the `type` prop to `kind` to follow carbon current convention.
- ~~[ ] Add defaults (size='medium', theme='light') for size and theme, and make them non-required (leave type as required), for consistency with carbon's approach.~~ In previous discussion it was thought preferable to make these required to protect against lazy developers :-)
- [x] Add a `iconDescription` prop, string type, required, and pass it to the icon as a `<title>` child, exactly as per carbon-react NotificationIcon component (Notification.js line 227).
- [x] Use `forwardRef` to forward a ref to the icon.
- [x] Put ...rest before other props on the div (i.e. `<div {...rest} className={classNames}>`.
- ~~[ ] The `IconComponent` may need to be wrapped in a `useEffect` to ensure it reacts correctly to changes in size, theme, type.~~
- [x] blockClass (in js and scss) should be `${pkg.prefix}--status-icon` (two hyphens before block name).
- [x] Add a class to the IconComponent, something like `${blockClass}__icon`.
- [x] Use the class on the IconComponent svg to replace use of the element type `svg` in our SCSS. Also, eliminate nesting in the SCSS per annoying carbon rule.
- [x] Include blockClass itself as a class on the main div.
- [x] Add tests for all the props (size, kind, theme, iconDescription) to check they produce the intended output.
- [x] Add a codesandbox for StatusIcon (https://codesandbox.io/dashboard/all/?workspace=7d09c446-5d37-4748-92d5-3d510c6572d4)
- [x] Use 'sm', 'md', 'lg' and 'xl' as values for `size` for consistency with Carbon.
- [x] Is the containing `<div>` really necessary? Let's see if we could merge `...rest` and the block classes directly onto the IconComponent and eliminate the wrapping `<div>`.
- [x] Add tests for `...rest` and `ref` being passed through, as per ExampleComponent.
- [x] Make sure StatusIcon is in index.js, _index.scss and package-settings.js as per updated code guidelines.

sydrosa:
Hi @dcwarwick I'm cc'ing @andrea-gm since we had this conversation re: #3 : `add defaults for size and theme` since we don't want a light/medium sized icon to render if someone adds a type. We want to protect against lazy developers who wouldn't change the size or theme if they see something rendered.

[...]

PR #586, opened 2021-04-09
Status icons resolve issues

sydrosa:
Contributes to #547 

This PR addresses the concerns of #547

#### What did you change?

- Removed prop validation for `iconDescription` and updated DocGen/removed spelling errors, etc
- Reworked testing: added testing for all required proptypes, including `iconDescription`, `kind`, `theme` and `size`.
- Updated component to include `forwardRef`
- Added className to the `IconComponent`
- Eliminated all SCSS nesting
- Added DocGen comments
- Small fixes, updates, etc.

Ran IBM accessibility checker, wrote tests, verified visually in Storybook.


packages/cloud-cognitive/src/components/StatusIcon/StatusIcon.js
```diff
@@ -120,46 +120,54 @@ const icons = {
...
-    [`${blockClass}--${iconTheme}-${type}`]: type,
-    [className]: className,
-  });
+const isValidIconDescription = () => (props) => {
+  if (props.iconDescription.toLowerCase() != props.kind) {
+    throw new Error(
+      'The "icon description" prop must match the status icon "kind" prop'
+    );
+  } else return;
+};
```
AlexanderMelox, line 132: If both are these are to be the same, is there a reason why i would have to pass down both? Or can you generate the iconDescription from the kind by capitalizing the first word?

AlexanderMelox: You can also do this in the prop types which I think will make sense because you're validating the props.

Screen Shot 2021-04-09 at 9 34 57 AM

dcwarwick (changes requested):

packages/cloud-cognitive/src/components/StatusIcon/StatusIcon.js
dcwarwick: In fact, I don't think we can link these at all: `iconDescription` needs to be translatable, so we can't generate it from the kind. We could perhaps generate a default icon description from the kind.

AlexanderMelox: Yeah because here it's indicating that the kind and iconDescription have to be the same, what if we didnt? I really don't know the design decision but was just wondering.

packages/cloud-cognitive/src/components/StatusIcon/StatusIcon.js
```diff
@@ -120,46 +120,54 @@ const icons = {
...
-  const [iconTheme, setIconTheme] = useState(null);
-  const IconComponent = icon[iconSize];
-
-  const classNames = cx({
-    [`${blockClass}--${iconTheme}`]: iconTheme,
-    [`${blockClass}--${iconTheme}-${type}`]: type,
-    [className]: className,
-  });
+const isValidIconDescription = () => (props) => {
+  if (props.iconDescription.toLowerCase() != props.kind) {
```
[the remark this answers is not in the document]
AlexanderMelox: Good catch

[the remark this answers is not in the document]
sydrosa: Solid! Totally didn't think about that 😅

dcwarwick (approved):

Issue #590, opened 2021-04-09
Saving prep for release review


davidmenendez:
add testing and docs for the `Saving` component.

PR #591, opened 2021-04-09
fix: adds tests and updates docs for saving

davidmenendez:
Contributes to #590 

primarily adds testing and updates docs. secondarily adds icon description prop for buttons.


davidmenendez:
i also refactored some of the code to merge the 3 different functions that get a value based off the `type` prop into a single one.

dcwarwick (approved):

PR #593, opened 2021-04-12
fix(SidePanel): correct how bottom padding of content is calculated

matthewgallo:
Contributes to #580 

This PR removes the hard coded bottom padding of the main content that accounts for the height of the actions container. The height of the actions container can differ depending on what combinations of buttons are used, thus the padding bottom for this is now set via a css variable that checks the height of the actions container beforehand.

I needed to pass the `condensed` prop the ActionSet so that the buttons maintained the proper height after I removed the hard coded height on the actions container. Since, the height can be variable based on the button combination, the buttons themselves should explicitly set their height, rather than the container element.

#### What did you change?
`SidePanel.js`
`_side-panel.scss`
`ActionSet.js`
`_action-set.scss`
Storybook

### DIFF
--- a/packages/cloud-cognitive/src/components/ActionSet/ActionSet.js
+++ b/packages/cloud-cognitive/src/components/ActionSet/ActionSet.js
@@ -31,7 +31,6 @@ export const ActionSet = React.forwardRef(
       // The component props, in alphabetical order (for consistency).
       actions,
       className,
-      condensed,
       size,
       // Collect any other property values passed in.
       ...rest
@@ -78,10 +77,7 @@ export const ActionSet = React.forwardRef(
             kind={action.kind}
             className={cx([
               `${blockClass}__action-button`,
-              {
-                [`${blockClass}__ghost-button`]: action.kind === 'ghost',
-                [`${blockClass}__action-button--condensed`]: condensed,
-              },
+              { [`${blockClass}__ghost-button`]: action.kind === 'ghost' },
             ])}>
             {action.label}
             {action.loading && <InlineLoading />}
@@ -163,11 +159,6 @@ ActionSet.propTypes = {
    * Sets an optional className to be added to the side panel outermost element.
    */
   className: PropTypes.string,
-
-  /**
-   * Determines if the action set buttons will receive the condensed version
-   */
-  condensed: PropTypes.bool,
 
   /**
    * Sets the size of the action set. Different button arrangements are used
--- a/packages/cloud-cognitive/src/components/ActionSet/ActionSet.js
+++ b/packages/cloud-cognitive/src/components/ActionSet/ActionSet.js
@@ -31,6 +31,7 @@ export const ActionSet = React.forwardRef(
       // The component props, in alphabetical order (for consistency).
       actions,
       className,
+      condensed,
       size,
       // Collect any other property values passed in.
       ...rest
@@ -77,7 +78,10 @@ export const ActionSet = React.forwardRef(
             kind={action.kind}
             className={cx([
               `${blockClass}__action-button`,
-              { [`${blockClass}__ghost-button`]: action.kind === 'ghost' },
+              {
+                [`${blockClass}__ghost-button`]: action.kind === 'ghost',
+                [`${blockClass}__action-button--condensed`]: condensed,
+              },
             ])}>
             {action.label}
             {action.loading && <InlineLoading />}
@@ -159,6 +163,11 @@ ActionSet.propTypes = {
    * Sets an optional className to be added to the side panel outermost element.
    */
   className: PropTypes.string,
+
+  /**
+   * Determines if the action set buttons will receive the condensed version
+   */
+  condensed: PropTypes.bool,
 
   /**
    * Sets the size of the action set. Different button arrangements are used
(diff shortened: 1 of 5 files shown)


packages/cloud-cognitive/src/components/ActionSet/ActionSet.js
```diff
@@ -31,6 +31,7 @@ export const ActionSet = React.forwardRef(
       // The component props, in alphabetical order (for consistency).
       actions,
       className,
+      condensed,
```
dcwarwick, line 34: I gave some thought to `condensed` when creating ActionSet. I don't think this property belongs here. It's setting very specific things from the SidePanel design that don't apply more generally. For example, Tearsheet has no concept of condensed, and shouldn't get the height of either $layout-05 OR $layout-04 applied.

You're absolutely right that the height should be being set on the buttons, not on the container -- that's going to go haywire when they stack! But I think the height should be set on the buttons *by the SidePanel*. I'll try and put a suggestion in the code below as to what I mean. See what you think.

PS if we do decide to go ahead with a prop on ActionSet, we'll have to think how to avoid applying unwanted heights for Tearsheet and anything else that might go on to use ActionSet later, and we'll also need to reapply the changes after #592 is merged because that has substantially changed the structure of ActionSet.js and _action-set.scss.

packages/cloud-cognitive/src/components/SidePanel/_side-panel.scss
```diff
@@ -264,9 +264,6 @@ $max-panel-size: 75%; // set max-width on max panels to 75%
       &.#{$action-set-block-class}--lg {
         @include setActionBarSize($large-panel-size);
       }
-      &.#{$block-class}__actions-container-condensed {
-        height: $layout-04;
-      }
     }
```
dcwarwick, line 267: So, we can apply the required heights, for condensed or otherwise, to the buttons here. Something like this:

```suggestion
    }
    .#{$block-class}__actions-container .#{action-set-block-class}__action-button {
      height: $layout-05;
    }
    .#{$block-class}__actions-container.#{$block-class}__actions-container-condensed .#{action-set-block-class}__action-button {
      height: $layout-04;
    }    
```

dcwarwick: PS -- is `height` the best style to be setting here? Button heights can vary based on their contents. The carbon Button sets `min-height` rather than `height`, and I think that might be better for us to do here too.

dcwarwick: PPS -- this isn't the PR to open this issue on, but I notice we've got some SCSS nesting going on here, and I gather we should be following the carbon rule to avoid nesting of selectors. Just a heads up.

packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
```diff
@@ -91,6 +91,24 @@ export let SidePanel = React.forwardRef(
...
+      if (open && actions && actions.length && animationComplete) {
+        const sidePanelOuter = document.querySelector(`#${blockClass}-outer`);
+        const actionsContainer =
+          sidePanelOuter &&
+          sidePanelOuter.querySelector(`.${blockClass}__actions-container`);
+        let actionsHeight = actionsContainer.offsetHeight + 16; // add additional 1rem spacing to bottom padding
+        actionsHeight = `${Math.round(actionsHeight / 16)}rem`;
+        sidePanelOuter.style.setProperty(
+          `--${blockClass}--content-bottom-padding`,
+          actionsHeight
```
dcwarwick, line 104: Do we also need a resize observer to recalculate this? I'm thinking that when size='max', the panel width is 75%, which will vary as the window resizes, and so buttons may reflow as they get wider or narrower and change height, causing this calculation to change.

dcwarwick, line 104: Do we also need a resize observer to recalculate this? I'm thinking that when size='max', the panel width is 75%, which will vary as the window resizes, and so buttons may reflow as they get wider or narrower and change height, causing this calculation to change.

dcwarwick (changes requested):

packages/cloud-cognitive/src/components/ActionSet/ActionSet.js
matthewgallo: Got it, we can definitely set the height of the buttons from the SidePanel. I'll rework this a little bit.

packages/cloud-cognitive/src/components/SidePanel/_side-panel.scss
matthewgallo: Ok, yes the nesting might need to be addressed in a separate PR. But yes, you're definitely right about that ^

packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
matthewgallo: Good point, I'll add the react resize detector component to check for this

dcwarwick (approved):

packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
matthewgallo: https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/594

Issue #594, opened 2021-04-13
SidePanel: add resize detector to get current actions container height


matthewgallo:
## What will this achieve?

As @dcwarwick mentioned, in the max size side panel the max width is 75% which depending on the viewport could cause the actions container to change height. Adding the resize detector component will allow the component to know what the actual height of the actions container is at all times.

matthewgallo:
Complete

PR #596, opened 2021-04-13
fix(SidePanel): add resize detector to adapt to new actions height

matthewgallo:
Contributes to #594 

To adapt to height changes in the actions container, the resize detector component was added to make sure that the main content always has the correct amount of bottom padding.

#### What did you change?
`SidePanel.js`
`_side-panel.scss`
Storybook


lee-chase (approved):
LGTM

PR #599, opened 2021-04-13
fix(SidePanel): add transition to divider appearing during animation

matthewgallo:
Contributes to #64 

This adds an animation to the title divider line during the scroll animation. Previously this was a border bottom that just appears immediately once the user started scrolling. This was causing a slight jump and also appeared quite abruptly. This is now implemented through a pseudo element that changes opacity as the scroll animation is progressed.

#### What did you change?
`SidePanel.js`
`_side-panel.scss`
Storybook


lee-chase (approved):
LGTM

Issue #577, opened 2021-04-08
CSS conflicts integrating with existing applications and styles


SimonFinney:
#### Affected package

`@carbon/ibm-cloud-cognitive@latest`

##### Carbon global and utility styles

Importing the package CSS into an existing application can cause unintended side effects originating from global and utility styles from Carbon as a dependency - depending on the import order, this also takes precedence over any Carbon style or version that an application may already consume.

As an example, see an exploration finding from integrating the CSS into Cloud Pak for Security's application boilerplate:

| Before                                                                                                                                    | After                                                                                                                                    |
| ----------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
| <img width="742" alt="Before" src="https://user-images.githubusercontent.com/3846874/114142506-f12e5400-990a-11eb-98c4-e9a2b839689a.png"> | <img width="745" alt="After" src="https://user-images.githubusercontent.com/3846874/114142513-f2f81780-990a-11eb-8118-7274b2c841c7.png"> |

**Reduced test case -** https://codesandbox.io/s/cranky-galileo-7cqlz?file=/src/index.scss

##### Carbon component styles

There are many different solutions for removing Carbon global and utility styles, for example, toggling all global flags off as per https://github.com/carbon-design-system/ibm-cloud-cognitive/compare/SimonFinney:css-snapshot...SimonFinney:remove-carbon-utility , but there are still challenges with the package outputting styles from underlying Carbon components that conflict with the existing styles of an application.

An example is the CSS overriding the theme for the `CodeSnippet`:

| Viewport                                                                                                                                       | Code                                                                                                                                    |
| ---------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------- |
| <img width="809" alt="CodeSnippet" src="https://user-images.githubusercontent.com/3846874/114144164-f096bd00-990c-11eb-85dc-a5791e02e226.png"> | <img width="687" alt="Code" src="https://user-images.githubusercontent.com/3846874/114144156-ee346300-990c-11eb-9887-14a105f03b80.png"> |
|  |

**Reduced test case -** To come

##### Styles public API testing

As an aside, throughout the course of the exploration, we found the styles changing unexpectedly between 2 versions of this package causing side effects that added significant noise to debugging the issues outlined above.

[...]

PR #608, opened 2021-04-15
Upgrade dependency versions and package.jsons

dcwarwick:
NB this PR contains a lot of files, because it includes a refresh of the yarn.lock and yarn offline mirror. There are only a few files that have substantively changed though, and they will be found at the bottom of the page.

---

Contributes to #577 

Push dependencies into package.jsons rather than relying on top-level package.json. Upgrade versions, holding off on a few that need more work (husky 4->6, jsdom 25->26, fast-sass-loader 1->2).

Made carbon and react packages peer deps rather than runtime deps.

#### What did you change?

package.jsons, a few individual files, and yarn config and mirror.

Ran all the builds and tests I could think of, and ran up storybook.


lee-chase (approved):
LGTM

Issue #488, opened 2021-03-19
Move Page Header out of Canary


andrea-gm:
## What will this achieve?

Happy design and dev managers

## How will success be measured?

Component passes release review and testing

## Additional information

N/A

lee-chase:
Good for another review @cameroncalder.

lee-chase:
Closing in favour of #59 and #897

PR #610, opened 2021-04-15
feat: upd breadcrumb with overflow to internal

lee-chase:
Contributes to #488 

Update breadcrumb with overflow in prep for review.

#### What did you change?

- Corrected CSS class format
- Wrapped CSS in block class
- Remove from exports
- Used screen for tests

- Ran up storybook
- Ran tests


dcwarwick:
LGTM but the ci-check seems ill...

dcwarwick (dismissed):

dcwarwick:
@lee-chase So the test is failing here because the breadcrumb items are being put into the overflow menu -- the BreadcrumbWithOverflow logic concludes there is not room for them. I popped in a console.log into `checkFullyVisibleBreadcrumbItems` as follows:
image
On my own machine it comes out all zeroes:
image
But when I uploaded this as a dummy PR, I got different results when the ci-check runs:
image
It looks like the spaceAvailable (obtained from `breadcrumbItemWithOverflow.current.offsetWidth`) is coming back as `undefined`, while the breadcrumb items are each reporting themselves as 200px wide, so the logic pushes them all into the overflow. It looks like the size values that get returned are a bit unreliable, and in particular for some reason come back differently on the server than when we run the tests locally... :-/

dcwarwick (approved):

Issue #615, opened 2021-04-15
Card testing


davidmenendez:
Add testing to `Card` component

PR #616, opened 2021-04-15
Card testing

davidmenendez:
Contributes to #615 

adds tests to card component. Also adds `html` to jest coverage.


packages/cloud-cognitive/src/components/Card/Card.js
```diff
@@ -84,7 +85,13 @@ export let Card = ({
             />
           );
         return (
-          <div key={id} className={`${pkg.prefix}-card-icon`}>
+          <div
+            key={id}
+            className={`${pkg.prefix}-card-icon`}
+            onClick={onClick}
+            onKeyDown={onClick}
```
lee-chase, line 92: Buttons are clicked on return key presses (repeats) and space (key up only).

Presumably here if I press the tab or escape it triggers onClick?

lee-chase (changes requested):

packages/cloud-cognitive/src/components/Card/Card.js
davidmenendez: You raise a good point. I honestly haven't given it that much thought. I'll ask Josh Black what Carbon's been doing for keyboard controls

davidmenendez: @lee-chase i may have been overthinking this when i commented this. While you might be correct about how `onKeyDown` works, it will ultimately be up to the consumer to figure out how they want their `onKeyDown` event to fire. I went ahead an made an update to add additional keyboard support, so now `onKeyDown` has it's own prop for specific `onKeyDown` implementation.

lee-chase:
I might have been overthinking it too @davidmenendez, is there any need for us to do anything other than pass ...rest through?

davidmenendez:
@lee-chase i think just adding `onKeyDown` should be enough 👍

davidmenendez:
@lee-chase please re-review when you have a chance. thanks!

lee-chase (approved):
LGTM

PR #631, opened 2021-04-20
Ensure input fields on Tearsheets have correct colours

### LINKED ISSUE #339: Tearsheet: Input elements are not styled correctly in main content area with grey 10 theme
@carbon/ibm-cloud-cognitive-experimental

## Detailed description

We use the grey 10 theme, and are displaying a tearsheet containing a search input element and several text input elements.  With no extra styling, the input elements in the tearsheet main contain area have a grey ($ui-02) background rather than the expected white ($ui-01) background.

This appears to be due to the fact that the tearsheet inherits style and behavior from the modal component; the modal component has anomalous styling, in that it appears with white background and grey input elements in both the white (default) and grey 10 themes.  The tearsheet-specific styling appears to explicitly override the modal styling in the grey 10 theme, but seems to have just missed the input elements.

Tearsheet displayed using the grey 10 theme.

> see changed?
Expected: Inputs with white background color.
Actual: Inputs with grey background color.

Chrome and Firefox.

> What version of the Carbon Design System are you using?
carbon-components v10.28.0
carbon-components-react v7.27.0
@carbon/themes v10.26.0
@carbon/ibm-cloud-cognitive-experimental v0.19.11

>
> https://codesandbox.io/s/serene-paper-qu2g4?file=/src/index.js

Unfortunately, I'm not sure how to set the theme to grey 10 in the code sandbox.  Here is an example we've created with  a basic layout that would show the problem, were it using the grey 10 theme.

dcwarwick:
Resolves #339 

Modal styles override input fields to use $field-02 over $ui-01. However, in wide Tearsheets, the main content background is set to $ui-02 to create a gray 'layer', so we need to revert the input fields to use $field-01 over this to match.

Also, mark TearsheetNarrow as released, which was accidentally missed.

#### What did you change?

Tearsheet styles and stories, plus package-settings and the global css snapshot.

Ran build, tests, storybook.

### DIFF
--- a/packages/cloud-cognitive/src/components/Tearsheet/_tearsheet.scss
+++ b/packages/cloud-cognitive/src/components/Tearsheet/_tearsheet.scss
@@ -216,6 +216,18 @@
 
   .#{$block-class}.#{$block-class}--wide .#{$block-class}__main {
     background: $ui-02;
+
+    // revert the background colour overridden by bx--modal styles
+    .#{$prefix}--text-input,
+    .#{$prefix}--text-area,
+    .#{$prefix}--search-input,
+    .#{$prefix}--select-input,
+    .#{$prefix}--dropdown,
+    .#{$prefix}--dropdown-list,
+    .#{$prefix}--number input[type='number'],
+    .#{$prefix}--date-picker__input {
+      background-color: $field-01;
+    }
   }
 
   .#{$block-class} .#{$block-class}__buttons {
--- a/packages/cloud-cognitive/src/components/Tearsheet/_tearsheet.scss
+++ b/packages/cloud-cognitive/src/components/Tearsheet/_tearsheet.scss
@@ -215,7 +215,7 @@
   }
 
   .#{$block-class}.#{$block-class}--wide .#{$block-class}__main {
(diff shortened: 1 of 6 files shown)


matthewgallo:
Looks like this PR needs to add an updated snapshot since Tearsheets have been released @dcwarwick ^

packages/cloud-cognitive/src/components/Tearsheet/_tearsheet.scss
```diff
@@ -216,6 +216,18 @@
 
   .#{$block-class}.#{$block-class}--wide .#{$block-class}__main {
     background: $ui-02;
+
```
lee-chase, line 219: @dcwarwick Some of the input components have a 'light' version. Should we cater for that too?

dcwarwick: @lee-chase Carbon kills the 'light' prop in modals. However, it's easy enough to restore the 'light' behaviour as well as the background colour, so I've updated to do that.

lee-chase (approved):

Issue #634, opened 2021-04-20
Export testing


davidmenendez:
Add and improve testing for `ExportModal`

PR #635, opened 2021-04-20
fix: added testing for export release review

davidmenendez:
Contributes to #634

### DIFF
--- a/packages/cloud-cognitive/src/components/ExportModal/ExportModal.js
+++ b/packages/cloud-cognitive/src/components/ExportModal/ExportModal.js
@@ -86,13 +86,13 @@ export let ExportModal = ({
       modalHeading={modalHeading}
       onRequestSubmit={onSubmitHandler}
       onRequestClose={onCloseHandler}
-      className={`${pkg.prefix}-export-modal`}
+      className={`${pkg.prefix}--export-modal`}
       primaryButtonDisabled={primaryButtonDisabled}
       passiveModal={submitted}>
-      <div className={`${pkg.prefix}-export-modal-inner`}>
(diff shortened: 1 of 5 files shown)


packages/cloud-cognitive/src/components/ExportModal/ExportModal.js
```diff
@@ -86,13 +86,13 @@ export let ExportModal = ({
       modalHeading={modalHeading}
       onRequestSubmit={onSubmitHandler}
       onRequestClose={onCloseHandler}
-      className={`${pkg.prefix}-export-modal`}
+      className={`${pkg.prefix}--export-modal`}
```
matthewgallo, line 89: Some other components have been using a `blockClass` variable which would be

```js
const blockClass = `${pkg.prefix}--export-modal`;
```

then in your JSX you would just have
```JSX
<Modal
  className={cx(blockClass, `${blockClass}__another-class`)}
/>
```

packages/cloud-cognitive/src/components/ExportModal/ExportModal.test.js
```diff
@@ -16,23 +16,28 @@ const defaultProps = {
...
   invalidInputText: 'File must have valid extension .pdf',
+  successMessage: 'Success',
+  errorMessage: 'Error',
+  loadingMessage: 'Loading',
 };
 
 describe(name, () => {
-  test('should render', async () => {
-    const { click, change } = fireEvent;
+  test('default render with with extension validation', async () => {
```
matthewgallo, line 30: We're now using `it` rather than `test`

davidmenendez:
@matthewgallo fixed 👍

matthewgallo (approved):

Issue #546, opened 2021-03-30
Align tearsheet to latest design


andrea-gm:
## What will this achieve?

Alignment with design

## How will success be measured?

Design review

## Additional information

- Design - https://pages.github.ibm.com/cdai-design/pal/components/tearsheet/usage

Specifically noticed the footer section and button type are inconsistent.

dcwarwick:
The buttons need to be put in a ButtonSet rather than a plain div. This will be done via our new ActionSet component when we have made that.

dcwarwick:
The ActionSet also needs to replicate the 'full bleed' styling, which Carbon only applies if a ButtonSet is included into a modal footer.

Also, the refreshed designs call for the ability to include actions into the header area -- consider adding a prop to enable that.

PR #636, opened 2021-04-20
Fix styling on ActionSet to ensure SidePanel and Tearsheet buttons match design

dcwarwick:
Contributes to #546

The carbon ButtonSet only applies full bleed styling when placed into a modal footer, and does not have any other way to request full bleed style. So replicate the full bleed style into ActionSet to ensure the correct styles are used: heights, and alignment. Also enabled Tearsheet to request extra large buttons, and improved the stories a little.

#### What did you change?

ActionSet, also SidePanel styles (changed way button heights are set), and Tearsheet.

Ran build, ran tests, ran storybook.


lee-chase (approved):
LGTM

PR #649, opened 2021-04-23
fix: adds and improves testing to several components

davidmenendez:
Contributes to #615 #647 #634 

Adds and improves testing for `ApiKeyModal` `ExportModal` and `Card`

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "build:storybook": "yarn build:storybook-dependencies-only && yarn build:storybook-only",
     "build:storybook-dependencies-only": "yarn build $(lerna ls --scope @carbon/ibm-cloud-cognitive-core --include-dependencies | grep -v '^@carbon/ibm-cloud-cognitive-core$' | xargs -n1 echo '--scope')",
     "build:storybook-only": "cd packages/core && yarn build",
-    "cdai-coverage": "yarn test packages/cloud-cognitive/src/components --coverage --collectCoverageFrom=packages/cloud-cognitive/src/components/**/*.js --collectCoverageFrom=!packages/cloud-cognitive/src/components/**/*.stories.js",
+    "ccs-coverage": "yarn test packages/cloud-cognitive/src/components --coverage --collectCoverageFrom=packages/cloud-cognitive/src/components/**/*.js --collectCoverageFrom=!packages/cloud-cognitive/src/components/**/*.stories.js",
     "ci-check": "yarn format:diff && yarn lint && yarn test --ci && lerna run ci-check --stream",
     "clean": "yarn cache clean && lerna run clean && lerna clean --yes",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -18,6 +18,7 @@
     "build:storybook": "yarn build:storybook-dependencies-only && yarn build:storybook-only",
     "build:storybook-dependencies-only": "yarn build $(lerna ls --scope @carbon/ibm-cloud-cognitive-core --include-dependencies | grep -v '^@carbon/ibm-cloud-cognitive-core$' | xargs -n1 echo '--scope')",
     "build:storybook-only": "cd packages/core && yarn build",
+    "cdai-coverage": "yarn test packages/cloud-cognitive/src/components --coverage --collectCoverageFrom=packages/cloud-cognitive/src/components/**/*.js --collectCoverageFrom=!packages/cloud-cognitive/src/components/**/*.stories.js",
```
lee-chase, line 21:
```suggestion
    "ccs-coverage": "yarn test packages/cloud-cognitive/src/components --coverage --collectCoverageFrom=packages/cloud-cognitive/src/components/**/*.js --collectCoverageFrom=!packages/cloud-cognitive/src/components/**/*.stories.js",
```

packages/cloud-cognitive/src/components/APIKeyModal/APIKeyModal.test.js
```diff
@@ -84,11 +88,91 @@ describe(name, () => {
...
+      nextStepButtonText: 'Next step',
+      downloadable: false,
+    };
+
+    const { rerender, getByPlaceholderText, getByText } = render(
+      <APIKeyModal {...props} />
+    );
+
+    // step 1
+    expect(getByPlaceholderText('input a')).toBeVisible();
```
lee-chase, line 135: On the assumption that only render the current step is rendered you do not need `expect().toBeVisible`. screen.get... and screen.find... throw errors if they fail to locate the target.

lee-chase (changes requested):
I think cdai should be ccs as per slack channel.

Also, unless I am mistaken about the implementation or test case, I do not believe we need to use expect().visible to verify the existence of an element.

packages/cloud-cognitive/src/components/APIKeyModal/APIKeyModal.test.js
davidmenendez: true. my thinking is i'm not actually targeting the inputs to preform any action on them. the test in question has more to do with testing the flow works and not about interacting with the content, so since i'm not trying to perform any input actions using `toBeVisible` is an easy way to test that clicking the next or previous button moves the user to the correct step and displays the correct content. it also makes the tests easier to read (in my opinion).

lee-chase (approved):

PR #661, opened 2021-04-27
Split n tidy button set with overflow

lee-chase:
Contributes to #488 

Refactor ButtonSetWithoverflow and add tests

#### What did you change?
- Moved ButtonSetWithOverflow to its own folder along with TempComboButton
- Changed ButtonSetWithOverflow to use Shape
- Added tests
- Removes no longer needed PageActionItem component

- Ran tests
- Checked coverage
- Checked storybook.


packages/cloud-cognitive/src/components/ButtonSetWithOverflow/ButtonSetWithOverflow.js
```diff
@@ -0,0 +1,216 @@
...
+          />
+        )}
+      </div>
+    </ReactResizeDetector>
+  );
+};
+
+ButtonSetWithOverflow.propTypes = {
+  /**
+   * Button shape things for us to render - see Carbon button kind, label, onClick.
```
dcwarwick, line 162:
```suggestion
   * Button shape things for us to render. Each button is specified as an
   * object with an optional field 'label' to supply the button label.
   * Additional fields in the object will be passed to the Button component,
   * and these can include 'kind', 'onClick', 'className', and any other Button
   * props (but NB not a ref). Any other fields in the object will be passed
   * through to the button element as HTML attributes.
```

NB you could, if you wanted, support passing through a 'ref' in the objects, by using `forwardRef` as you map the objects to Buttons. I did that in ActionSet, in case it was ever handy. But it may not be needed.

packages/cloud-cognitive/src/components/ButtonSetWithOverflow/ButtonSetWithOverflow.js
```diff
@@ -0,0 +1,216 @@
...
+          'danger--primary',
+          'danger--ghost',
+          'danger--tertiary',
+          'tertiary',
+        ]),
+        label: PropTypes.node,
+        onClick: PropTypes.func,
+      })
+    ),
+  ]),
```
dcwarwick, line 181:
```suggestion
  buttons: PropTypes.arrayOf(
    PropTypes.shape({
      kind: PropTypes.oneOf([
        'primary',
        'secondary',
        'danger',
        'ghost',
        'danger--primary',
        'danger--ghost',
        'danger--tertiary',
        'tertiary',
      ]),
      label: PropTypes.node,
      onClick: PropTypes.func,
    })
  ),
```

The oneOfType is only needed to include a validator -- you could leave it in if you're intended to add a validator later.

packages/cloud-cognitive/src/components/ButtonSetWithOverflow/_button-set-with-overflow.scss
```diff
@@ -0,0 +1,48 @@
+//
+// Copyright IBM Corp. 2020, 2020
```
dcwarwick, line 2:
```suggestion
// Copyright IBM Corp. 2020, 2021
```

dcwarwick (changes requested):

dcwarwick (approved):

Issue #623, opened 2021-04-19
Review findings for UserProfileImage


matthewgallo:
## What will this achieve?

All outstanding items to complete release review:
- [x] Reach out to designer for final design approval
- [x] Rename `icon` prop to `kind` to match carbon precedent.
- [x] Add prop to give image `alt` text (ie `imageDescription`)
- [x] Remove default props of `size` and `icon`. This may break parts of the component, be sure to add checks that these props have values before using them.
- [x] Rename size `xl` to `xlg` to match carbon guidelines
- [x] Ln 86, replace single dash with double underscore (`${blockClass}-photo` => `${blockClass}__photo`)
- [x] Remove size className on image, add `width: 100%;` to `${blockClass}__photo` element
- [x] Use token calculation instead of hard coded `1.25rem` on `sm` size in `_user-profile-image.scss` (ie `calc(#{$spacing-05} + #{$spacing-02}),`)
- [x] Avoid nesting, specifically `.#{$block-class}-photo` inside of `.#{$block-class}`
- [x] Add license header to `_color-map.scss` file
- [x] Change scss file from camel case to kebab case (`_user-profile-image.scss`);

PR #665, opened 2021-04-27
fix(UserProfileImage): updates based on release review notes

qbi11y:
Contributes to #623
Updates to User Profile Image based on feedback from release review.

#### What did you change?
UserProfileImage.js
UserProfileImage.stories.js
_colorMap.scss 
_index.scss 
_user-profile-image.scss

All tests passed


packages/cloud-cognitive/src/components/UserProfileImage/UserProfileImage.js
```diff
@@ -39,9 +39,10 @@ export let UserProfileImage = React.forwardRef(
     {
       backgroundColor,
       className,
-      icon,
+      kind,
       initials,
       image,
+      imageDescription,
```
sydrosa, line 45: You'll need to add this to your props validation below, with some documentation/comments on how this prop is used! Additionally, because it provides an `alt` for the image, it should likely be required.

packages/cloud-cognitive/src/components/UserProfileImage/UserProfileImage.js
```diff
@@ -114,10 +115,10 @@ export let UserProfileImage = React.forwardRef(
 UserProfileImage = pkg.checkComponentEnabled(UserProfileImage, componentName);
 
 UserProfileImage.displayName = componentName;
-UserProfileImage.defaultProps = {
-  icon: 'user',
-  size: 'xl',
-};
+// UserProfileImage.defaultProps = {
```
sydrosa, line 118: Let's clean up some of the code that's commented out.

packages/cloud-cognitive/src/components/UserProfileImage/UserProfileImage.js
```diff
@@ -142,9 +143,9 @@ UserProfileImage.propTypes = {
    */
   className: PropTypes.string,
   /**
-   * When passing the icon prop, use either "user" or "group". The values match up to the Carbon Library icons.
+   * When passing the kind prop, use either "user" or "group". The values match up to the Carbon Library icons.
    */
-  icon: PropTypes.oneOf(['user', 'group']),
+  kind: PropTypes.oneOf(['user', 'group']),
```
sydrosa, line 148: Per the check, let's go ahead and sort the prop types alphabetically so this darn thing will pass 😅

packages/cloud-cognitive/src/components/UserProfileImage/UserProfileImage.stories.js
```diff
@@ -58,25 +58,30 @@ export const Default = Template.bind({});
 Default.args = {
   backgroundColor: 'light-cyan',
   theme: 'light',
+  size: 'xlg',
```
sydrosa, line 61: If you created a `defaultProps` object, you could cut down on a few lines of code that are used. Something like:

```
const defaultProps = {
  backgroundColor: 'light-cyan',
  theme: 'light',
  size: 'xlg',
};
```

and then spread them in your `Template.args`:

```
WithGroupIcon.args = {
  ...defaultProps,
  kind: 'group',
};
```

Def nitpicky and not a blocker :D

packages/cloud-cognitive/src/components/UserProfileImage/_colorMap.scss
```diff
@@ -1,3 +1,10 @@
+//
```
sydrosa, line 1: All SCSS files should use kebab case. Perhaps change this file to `_color-map.scss`?

packages/cloud-cognitive/src/components/UserProfileImage/UserProfileImage.js
```diff
@@ -55,14 +56,14 @@ export let UserProfileImage = React.forwardRef(
...
-        xl: User32,
+        xlg: User32,
       },
       group: {
         xs: Group16,
         sm: Group16,
         md: Group20,
         lg: Group24,
-        xl: Group32,
+        xlg: Group32,
```
sydrosa, line 66: Is having an `xs` icon necessary since it's the same as the small? Probably a design decision but now I'm curious 🤔

sydrosa (dismissed):
Hey James -- left some small comments for changes. I think we also discussed ensuring that (since we are no longer providing `defaultProps`), your component receives all necessary props _before_ rendering. Similar to `StatusIcon` -- check it out [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/eff78ab76cb5782d227f4f112aae07facc961cf4/packages/cloud-cognitive/src/components/StatusIcon/StatusIcon.js#L126)

Happy to discuss over Slack. 🎉

packages/cloud-cognitive/src/components/UserProfileImage/UserProfileImage.js
```diff
@@ -141,24 +135,32 @@ UserProfileImage.propTypes = {
...
-   */
-  icon: PropTypes.oneOf(['user', 'group']),
   /**
    * When passing the image prop, supply a full path to the image to be displayed.
    */
   image: PropTypes.string,
+  /**
+   * When passing the image prop use the imageDecsription prop to describe the image for screen reader.
+   */
+  imageDescription: PropTypes.string,
```
sydrosa, line 145: This should be required.

matthewgallo: @sydrosa is right, `imageDescription` needs to be marked as `required`.

packages/cloud-cognitive/src/components/UserProfileImage/UserProfileImage.js
```diff
@@ -100,24 +98,32 @@ export let UserProfileImage = React.forwardRef(
         className={cx([
           blockClass,
           className,
+          'test',
```
matthewgallo, line 101: I think this was left in from debugging @qbi11y

packages/cloud-cognitive/src/components/UserProfileImage/UserProfileImage.js
```diff
@@ -100,24 +98,32 @@ export let UserProfileImage = React.forwardRef(
...
           `${blockClass}--${size}`,
           `${blockClass}--${theme}`,
           `${blockClass}--${backgroundColor}`,
         ])}>
-        <FillItem />
+        <FillItem className="test" />
```
matthewgallo, line 106: Same with this `className="test"`

matthewgallo (changes requested):

matthewgallo (approved):

packages/cloud-cognitive/src/components/UserProfileImage/UserProfileImage.js
matthewgallo: This was addressed in @qbi11y latest commit that adds custom prop validation for `imageDescription`

PR #668, opened 2021-04-28
Use Button propTypes to creating matching prop shapes

dcwarwick:
Use Button propTypes as the basis for button-shaped array props in ActionSet, Tearsheet, ButtonSetWithOverflow. This makes a more complete set of type constraints, and is less error-prone and ensures we stay in sync with Button props if/when they change.

#### What did you change?

ActionSet, Tearsheet files, ButtonSetWithOverflow

ran build, tests, storybook

### DIFF
--- a/packages/cloud-cognitive/src/components/ActionSet/ActionSet.js
+++ b/packages/cloud-cognitive/src/components/ActionSet/ActionSet.js
@@ -55,13 +55,10 @@ const ActionSetButton = React.forwardRef(
 );
 
 ActionSetButton.propTypes = {
-  className: PropTypes.string,
-  disabled: PropTypes.bool,
+  ...Button.PropTypes,
   kind: PropTypes.oneOf(['ghost', 'secondary', 'primary']),
   label: PropTypes.string,
   loading: PropTypes.bool,
-  onClick: PropTypes.func,
-  size: Button.propTypes.size,
 };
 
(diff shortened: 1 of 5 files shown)


dcwarwick:
@lee-chase This was something that occurred to me when reviewing your ButtonSetWithOverflow PR. Take a look and see what you think.

lee-chase (commented):
Sort of what we are doing in ActionBarItem, do we need any of the extra stuff @dcwarwick 

Also what David was doing with the removeUnwantedProps.

```
// Base props on Carbon Button
const propTypes = { ...Button.propTypes };
const defaultProps = { ...Button.defaultProps };

// Remove reserved props
reservedProps.forEach((prop) => {
  delete propTypes[prop];
  delete defaultProps[prop];
});

ActionBarItem.displayName = componentName;
ActionBarItem.propTypes = {
  /**
   * The ...propTypes are copies of those from Button minus the props reserved for use by this component
   */
  ...propTypes,
  ```

dcwarwick:
> Sort of what we are doing in ActionBarItem, do we need any of the extra stuff @dcwarwick

When props are to be excluded, like in ActionBarItem, could definitely use `prepareProps`. What is "any of the extra stuff"? What extra stuff?

lee-chase (approved):

PR #684, opened 2021-04-29
Improve visual effect of stacked tearsheets

### LINKED ISSUE #517: Stacked tearsheets are resized rather than scaled
When tearsheets are stacked, the "behind" tearsheets are sized slightly narrower, and this is done with width. This means that it's possible for there to be a visible reflow as a tearsheet gets revealed and "brought forward". This could perhaps be fixed by using transform instead. Or something.

dcwarwick:
Resolves #517

Use transform rather than width to scale stacked tearsheets. This ensures that reflow does not occur as a lower tearsheet comes to the front, and also creates a nice moving-forward visual effect.

#### What did you change?

Upgraded react-resize-detector to v6, and TearsheetSheel

ran build, tests, storybook


lee-chase (approved):
LGTM but there is an opportunity to get the function triggered by a resize and maybe test.

Issue #691, opened 2021-04-30
SidePanel: animateTitle prop updates


matthewgallo:
## What will this achieve?

When this prop is set to false, the subtitle should remain fixed and the underline should show by default. This is going to assist @sydrosa in the create side panel.
## How will success be measured?

When `animateTitle={false}` allows the subtitle (and actionToolbar if provided) to remain fixed with underline.

matthewgallo:
This is complete

PR #693, opened 2021-04-30
PR for #691: side panel animate title prop updates

matthewgallo:
Contributes to #691 

This PR cleans up the side panel behavior when `animateTitle={false}` is provided to the component. The subtitle text becomes fixed, as well as the action toolbar buttons if they are provided.

#### What did you change?
`SidePanel.js`
`SidePanel.test.js`
`SidePanel.stories.js`
`_side-panel.scss`
Storybook
Verified test coverage is still at 100%


davidmenendez (approved):
LGTM

PR #695, opened 2021-04-30
Saving: Release

davidmenendez:
Contributes to #694 

* updates snapshot
* adds ref support
* updates `_index-released-only.scss` to include saving
* updatres `package-settings.js` to include saving


dcwarwick (approved):

sydrosa (changes requested):
Thanks for taking the time to go over this! 

Release review notes:

- [x] Include a `forwardRef` in your component
- [ ] Let's update the props for `onCancel` and `onSave` to `secondaryHandler` and `primaryHandler` to be consistent with Carbon + other C&C components and to match with the button names
- [x] Kebab case line 44 for `in-progress` update this in Proptypes as well
- [x] Can you rework the large if else statement into a large object that maps into description, text and icon? Lines 40 - 50. Check out this example in StatusIcon
- [ ] Buttons should be named primary and secondary per carbon
- [ ] We should remove the default props, add them as required in your propTypes, and restructure your component to ensure that the required props are being passed _before_ render. check out sTatusIcon
- [x] Add `...rest` to your component
- [ ] Need an additional test to cover `...rest` in the component (dataTestId)
- [ ] need an addtional test to cover `forwardRef`
- [ ] Add accessibility test
- [ ] Add className test
- [ ] Replace `fireEvent` -- try `userEvent`
- [ ] Line 141 - update proptypes docgen for `status`
- [ ] remove unused classNames in Storybook
- [ ] Create codesandbox in our organization

packages/cloud-cognitive/src/components/Saving/Saving.js
```diff
@@ -18,89 +18,90 @@ import PropTypes from 'prop-types';
...
+    },
+    ref
+  ) => {
+    const statusObj = {
+      default: {
+        text: defaultText,
+        iconDescription: defaultIconDescription,
+        icon: CheckmarkOutline16,
+      },
+      inProgress: {
```
sydrosa, line 48: Most strings/props are kebab case instead of camel case. `in-progress` also matches a few of our other components. Let's update this.

sydrosa (commented):
One smol thing :D

sydrosa (approved):
Welcome to the Released club 🎉

PR #697, opened 2021-05-04
fix: page header deprecations

lee-chase:
Contributes to #488 

In moving some parts of the page header, specifically pageActions, to use Button shaped props instead of propTypes.node (children etc) some test cases were missed. Simplified 

#### What did you change?

- ActionBar.actions
- ButtonSetWithOverflow tests, stories, and TempComboButton
- PageHeader use of actionBarItems and pageActions and tests
- Added propsHelper.extractShapesArray to handle non-shape values.

Test, CI-Check and Storybook


packages/cloud-cognitive/src/components/ActionBar/ActionBar.js
```diff
@@ -187,15 +192,17 @@ export let ActionBar = React.forwardRef(
 ActionBar.displayName = componentName;
 ActionBar.propTypes = {
   /**
-   * Action items to be displayed in the bar.
+   * Specifies the action bar items. Each item is specified as an object
+   * with the properties of a Carbon Button in icon only form. Button kind, size, tooltipPosition,
+   * tooltipAlignment abd type are ignored.
```
davidmenendez, line 197: i believe `abd` is a typo

packages/cloud-cognitive/src/components/ActionBar/ActionBar.js
```diff
@@ -187,15 +192,17 @@ export let ActionBar = React.forwardRef(
...
    */
   actions: PropTypes.oneOfType([
     PropTypes.arrayOf(
       PropTypes.shape({
+        ...Button.propTypes,
```
davidmenendez, line 202: my only comment here is that when you try to spread the proptypes of an imported component like this you don't get the benefit of it showing up in the storybook docs. i realize it's tedious to have to specify the same proptypes for what amounts to a wrapper. i won't say specifically that this is a bad pattern, but i'd like to say that if you think anything needs to be pointed out in the docs specifically that you'll need to type it out.

davidmenendez (commented):
2 comments

packages/cloud-cognitive/src/components/ActionBar/ActionBar.js
lee-chase: It is unfortunate, but it does protect against Carbon changes.

lee-chase: What do you think of doing this @davidmenendez 

image

dcwarwick: @lee-chase This is a great point by @davidmenendez. Unless we find a way to provide 'missing' props in storybook/docgen, I think we do need to keep some prop type and doc duplication in. On the other hand, using `...Button.propTypes` means we pull in ALL the prop type definitions for Button, which means that any other button props they include will be type-checked, which is good, but we wouldn't want ALL the Button props necessarily to dump out here, because that could be a bit unwieldy.

My thoughts:
* I like linking to the docs for the base components when we're referring to it -- I'll add the same thing for ActionSet.
* I think we should continue to pull in `...Button.propTypes` in order to get the Button props as accurately as possible.
* But I think we should duplicate a few of the props, the ones people will typically use, for clarity. For a button, I'd guess that `kind` and `onClick` are the essentials, with the label for a text button and `renderIcon` and `iconDescription` for an icon button. Duplicating those props exposes us to the (small) risk that their prop types change in Carbon and we don't notice, but will make the storybook and other DocGen consumers a bit clearer.
* To still get as much linkage as possible to the object, we could do `kind: Button.propTypes.kind` rather than copying over complex prop type definitions. Then if new carbon button kinds are added, for example, we pick them up automatically. It makes the docgen a tiny bit more opaque, but also acts as a prompt perhaps that if someone really wants to know about this prop field they should go and look at Button!

packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```diff
@@ -708,17 +705,18 @@ PageHeader = pkg.checkComponentEnabled(PageHeader, componentName);
 
 PageHeader.propTypes = {
   /**
-   * One or more ActionBarItem components, passed in as React element(s).
-   * If provided, these are rendered at the top right of the header as
-   * action icons. Optional.
+   * Specifies the action bar items. Each item is specified as an object
+   * with the properties of a Carbon Button in icon only form. Button kind, size, tooltipPosition,
+   * tooltipAlignment abd type are ignored.
```
dcwarwick, line 710:
```suggestion
   * tooltipAlignment and type are ignored.
```

dcwarwick (changes requested):

dcwarwick (approved):

packages/cloud-cognitive/src/components/ActionBar/ActionBar.js
davidmenendez: this is an issue i've been dealing with for a while. it's an unfortunate technical limitation or oversight by storybook because ultimately it would be a very nice feature. i think as long as we're pointing the consumer in the right direction it's a step forward.

PR #703, opened 2021-05-05
chore: mark ContextHeader as released

asfordmatt:
https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/702

### DIFF
--- a/packages/cloud-cognitive/src/global/js/package-settings.js
+++ b/packages/cloud-cognitive/src/global/js/package-settings.js
@@ -12,6 +12,7 @@ const defaults = {
   component: {
     // reviewed and released components:
     AboutModal: true,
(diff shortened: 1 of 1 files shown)


lee-chase (changes requested):
Outside of DSAG approval the release process also stipuldates:

- src/components/_index-released-only.scss update
- Component wrapped with React.forwardRef, ref on suitable element & test
- Add additional attributes to suitable element & test.
- Support for the user supplying a className
- SCSS format should be `prefix--Block__Element--Modifier` to match Carbon. E.g. `${pkg.prefix}--context-header__task` 
- SCSS should be use `@mixin + @include exports` pattern, see `_about-modal.scss`

Issue #716, opened 2021-05-07
PageHeader on scrolling with sticky, the tooltip-tagset stay on top


ninjasun:
PageHeader on scrolling with sticky, the tooltip-tagset stay on top and does not scroll or close
## What package(s) are you using?

- [ X] Cloud & Cognitive
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.
On scroll down the tooltip overlaps the header/tags

> Is this issue related to a specific component?
PageHeader, TagSet, Tooltip

> What did you expect to happen? What happened instead? What would you like to see changed?
Scroll the tooltip or close it but do not overlaps the tag set

> What browser are you working in?
chrome 90.0.4430.93 (Official Build) (64-bit)

> What version of the @carbon/ibm-cloud-cognitive packge are you using?
0.31
> Please create a reduced test case in CodeSandbox
https://codesandbox.io/s/shy-sun-0e2ww?file=/src/Header.js

![tooltip-overlaps-tags](https://user-images.githubusercontent.com/7422984/117421572-efae7680-af1e-11eb-8a59-41ad0d78597f.png)

lee-chase:
Fixed

PR #740, opened 2021-05-14
feat: stop tagset overflow scrolling off screen

lee-chase:
Contributes to #716

Prevent TagSet overflow scrolling off screen when used in the page header.

#### What did you change?
- Added overflowClassName to the TagSet (passing through to TagSetOverflow
- Updated page header long values story to add more tags
- Used the overflowClassName from page header to set a class on TagSet
- Added CSS variables for position and top to the document root
- Used the CSS variables with the overflowClassName to fix the position of the TagSetOverflow

NOTE: In the transition from position absolute to fixed (as the tag row sticks/unsticks from the top) there is a delay of about 0.2 of a second. It is not clear why but could be either reactivity or browser.

Storybook and ci-check


packages/cloud-cognitive/src/components/PageHeader/_page-header.scss
```diff
@@ -447,6 +447,18 @@ $raised-z-index: 99;
       }
     }
   }
+
+  :root {
+    --#{$block-class}--tagset-tooltip-offset: 0;
+  }
+
+  .#{$block-class}__tagset-tooltip.#{$block-class}__tagset-tooltip {
+    // removes the min width in Carbon
```
dcwarwick, line 456: This comment doesn't look relevant to anything here...

```suggestion
```

packages/cloud-cognitive/src/components/TagSet/TagSetOverflow.js
```diff
@@ -86,6 +87,16 @@ export const TagSetOverflow = React.forwardRef(
...
+  /**
+   * className
+   */
+  className: PropTypes.string,
+  /**
+   * fixed underneath header
+   */
+  fixed: PropTypes.shape({
+    offset: PropTypes.number,
+  }),
```
dcwarwick, line 99: This doesn't seem to do anything. Was it left over from a different approach? Or is it needed for something I didn't spot?

```suggestion
```

dcwarwick (changes requested):

dcwarwick (approved):

Issue #741, opened 2021-05-15
Clear noisy test log from IdeFilter component introduced in recent updates


dcwarwick:
@Georgegriff @vladbalanescu 
PR #720 has introduced some `console.error` noise into the CDAI test logs. Although these don't cause the tests to fail, they do appear to indicate problems in the code which should be resolved. If the code is correct as it stands, the tests need to be modified to avoid or catch the error log and mark it as expected (you're welcome to ping me to discuss how this can be done).

The errors can be seen by running `yarn jest IdeFilter`.

**First error**
![image](https://user-images.githubusercontent.com/6385315/118378896-acaf6b80-b5ce-11eb-997d-b9940d4c7a9f.png)
This appears to be caused by IdeFilter.js line 113:
![image](https://user-images.githubusercontent.com/6385315/118378923-d9fc1980-b5ce-11eb-9316-f3fedb9cfe68.png)
"filter" is not a valid value for the `type` prop of carbon-components-react `Tag` component. I wasn't sure what this code was trying to do, so couldn't see the way to fix it.

**Second error**
![image](https://user-images.githubusercontent.com/6385315/118378947-0dd73f00-b5cf-11eb-93f0-4eb57ac170f5.png)
This appears to be caused by IdeFilter.spec.js line 193:
![image](https://user-images.githubusercontent.com/6385315/118378969-2e06fe00-b5cf-11eb-83c9-628006b5503d.png)
The `searchIcon` prop expects a React element, as set in IdeFilter.js line 253:
![image](https://user-images.githubusercontent.com/6385315/118379009-660e4100-b5cf-11eb-90a6-004a0ebfc71c.png)
but a function that takes a className and returns an element is being supplied. Again, I wasn't sure enough what this code was trying to do to see the way to fix it.

It's currently very easy to overlook console errors in the tests that don't cause the tests to fail, and several other similar errors are being logged from other components -- as soon as they are all fixed I will drop in an update to the jest config to cause tests to fail if they produce unexpected console errors and warnings, which will make this easier to avoid in the future.

Georgegriff:
[...]

PR #743, opened 2021-05-15
Ensure unexpected console errors and warnings cause tests to fail

dcwarwick:
NB **this PR currently fails ci-check** -- this is expected, because there are some console errors in the tests! When they are eliminated, the ci-check should pass and this PR can then be merged, and will then take effect for all future PRs.

It's currently very easy to overlook console errors and warnings in the tests that don't actually cause the tests to fail. Sometimes these errors and warnings point to real problems that should be fixed, and even if they don't they make the tests difficult to run and see clearly what is happening because of the extra noise. This PR means that an unexpected console error or warning will always cause the test to fail. This also helps make it clear which test was actually causing the console activity, which is not currently easy to spot!

#### What did you change?

Added a global jest replacement of console `error` and `warn` methods.

Ran tests.


dcwarwick:
Once approved, this PR requires the following to be satisfied before it can be merged:

- [ ] PR #748 approved and merged.
- [ ] issue #741 resolved.
- [ ] rerun CI/check and ensure success.

dcwarwick:
This should be mergeable once #782 and #783 are approved and merged.

PR #747, opened 2021-05-17
Fix background of pagination inside wide tearsheet

nertzy:
This is directly related to #339.

#### What did you change?

We added two missing classes to the list of selectors that need to be "un-overridden" to get back to the correct look inside the wide tearsheet main area. 

Copied the list of selectors from https://github.com/carbon-design-system/carbon/blob/ce0662896cb8919f1f11b70a1f4351750b6d7e07/packages/components/src/components/modal/_modal.scss#L49-L50 to add the missing classes.

### DIFF
--- a/packages/cloud-cognitive/src/components/Tearsheet/_tearsheet.scss
+++ b/packages/cloud-cognitive/src/components/Tearsheet/_tearsheet.scss
@@ -226,6 +226,8 @@
     background: $ui-background;
 
     // revert the background colour overridden by bx--modal styles
+    .#{$prefix}--pagination,
+    .#{$prefix}--pagination__control-buttons,
     .#{$prefix}--text-input,
     .#{$prefix}--text-area,
     .#{$prefix}--search-input,
--- a/packages/cloud-cognitive/src/__tests__/__snapshots__/styles.test.js.snap
+++ b/packages/cloud-cognitive/src/__tests__/__snapshots__/styles.test.js.snap
@@ -1332,6 +1332,8 @@ path:nth-of-type(1),
   background: var(--cds-ui-background, #ffffff);
 }
 
+.exp--tearsheet.exp--tearsheet--wide .exp--tearsheet__main .bx--pagination,
+.exp--tearsheet.exp--tearsheet--wide .exp--tearsheet__main .bx--pagination__control-buttons,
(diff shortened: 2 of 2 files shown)


packages/cloud-cognitive/src/components/Tearsheet/_tearsheet.scss
```diff
@@ -238,6 +240,8 @@
     }
 
     // and restore the 'light' prop in case light fields are wanted
+    .#{$prefix}--pagination--light,
+    .#{$prefix}--pagination__control-buttons--light,
```
dcwarwick, line 244: I think these two classes are not in fact used, because, rather inconsistently, Pagination does not have a 'light' prop.

```suggestion
```

dcwarwick (changes requested):

dcwarwick:
Thank you @nertzy for the PR. I copied this exact list of cases from Carbon, so I guess pagination has been added since I did that! Good catch. I have suggested one change, because oddly Pagination doesn't seem to have a `'light` prop so the two `--light` variant classes are not currently used by Carbon. If you agree and commit that change (you can do that easily here within github) I'll be happy to approve and merge! Thank you for your contribution.

dcwarwick:
@nertzy Also, as this PR changes the styles for a released component, there's an extra step needed to enable it to pass the CI check and be merged: we need to refresh the released styles snapshot. If you have a clone of ibm-cloud-cognitive where you can run builds etc, you'll need to run `yarn jest styles -u` in this branch, commit the updated snapshot file, and push it to update this PR.

PS alternatively I'll be very happy to do this step for you if you like, but you'll need to allow me -- if you look at the bottom right side of this PR and tick the checkbox
image
then that enables us to adjust PRs for you. It's normally a good idea to leave that checkbox ticked as it can speed things up.

nertzy:
Thanks for the feedback! I'm having trouble getting node-sass compiled with my latest version of NodeJS, so I'll defer to you to make the requested changes.

I don't see the checkbox you mentioned, so I'm going to try putting this PR back into draft mode to see if that allows me to make that change.

nertzy:
I set up nvm on my machine real quick and was able to get the snapshot generated. Let me know if there are any other edits I should make. Thanks!

Issue #751, opened 2021-05-18
Remove Delete Modal: Release Review


sydrosa:
### Purpose

To prepare the `RemoveDeleteModal` for release. 

### To-do

- [x] Per the design, the `primaryButton` on the modal should be red.
- [x] Per the design, the input field in the modal should span 100% of the container, while text inputs have a 20% margin-right
- [x] I believe the size for this modal should be "small', similar to `CreateModal` -- can you confirm with design?
- [x] Could `resourceName` or `fileName` work as a prop in place of `resource`? I was initially confused by what a `resource` was or meant, and the doc gen didn't clear it up. Perhaps there is a way to be more specific?
- [x] I'm a bit unsure of the component name as `RemoveDeleteModal` -- we discussed this a bit over Slack. I would be open to hearing your thoughts.
- [x] Add copy-write header in the following files:
     - [x] `_remove-delete.scss`
     - [x] `RemoveDeleteModal.js`
- [x] Update copy-write header to read "2020, 2021" in the following files:
     - [x] `_storybook-styles.scss`
     - [x] `index.js`
     - [x] `RemoveDeleteModal.stories.js`
     - [x] `RemoveDeleteModal.test.js`
- [x] Update the test file to include a test for each prop (`resource`, `iconDescription`, etc..) and it's supposed outcome -- discussed over Slack
- [x] Let's set the propType of `iconDescription` to `isRequired`
- [x] Line 86 of on `RemoveDeleteModal.js` has a spelling error 
- [x] To match the other modal component structure in Storybook, can you remove the "with State Manager" story?
- [x] Create a CodeSandbox [here](https://codesandbox.io/dashboard/all/?workspace=7d09c446-5d37-4748-92d5-3d510c6572d4) with a functioning component

sydrosa:
@davidmenendez 

Finally, to release this component, please create a PR with the following: 

- [ ]  the flags for the components in `package-settings.js` should be changed to true
- [ ]  the component SCSS should be included in `/src/components/_index-released-only.scss.`
- [ ]  run the tests, recreating snapshots (using `yarn jest -u`), and check in the updated public CSS snapshot as part of the PR

PR #757, opened 2021-05-18
fix: remove modal release feedback updates

davidmenendez:
Contributes to #751 

- [x] Per the design, the `primaryButton` on the modal should be red.
- [x] Per the design, the input field in the modal should span 100% of the container, while text inputs have a 20% margin-right
- [x] I believe the size for this modal should be "small', similar to `CreateModal` -- can you confirm with design?
- [x] Could `resourceName` or `fileName` work as a prop in place of `resource`? I was initially confused by what a `resource` was or meant, and the doc gen didn't clear it up. Perhaps there is a way to be more specific?
- [x] I'm a bit unsure of the component name as `RemoveDeleteModal` -- we discussed this a bit over Slack. I would be open to hearing your thoughts.
- [x] Add copy-write header in the following files:
     - [ ] `_remove-delete.scss`
     - [ ] `RemoveDeleteModal.js`
- [x] Update copy-write header to read "2020, 2021" in the following files:
     - [ ] `_storybook-styles.scss`
     - [ ] `index.js`
     - [ ] `RemoveDeleteModal.stories.js`
     - [ ] `RemoveDeleteModal.test.js`
- [x] Update the test file to include a test for each prop (`resource`, `iconDescription`, etc..) and it's supposed outcome -- discussed over Slack
- [x] Let's set the propType of `iconDescription` to `isRequired`
- [x] Line 86 of on `RemoveDeleteModal.js` has a spelling error 
- [x] To match the other modal component structure in Storybook, can you remove the "with State Manager" story?
- [ ] Create a CodeSandbox [here](https://codesandbox.io/dashboard/all/?workspace=7d09c446-5d37-4748-92d5-3d510c6572d4) with a functioning component
renames `RemoveDeleteModal` to `RemoveModal` as well as addressing all the points mentioned in the issue


davidmenendez:
sandbox to be created when this is merged since there's a component name change

lee-chase (approved):

sydrosa (approved):

Issue #611, opened 2021-04-15
Pageheader title should have a loading state


AlexanderMelox:
## What will this achieve?

In our designs, some of the pageheader titles are dynamic and come from an api. We want to utilize skeletons until it loads but currently the Pageheader title prop is of string.

## How will success be measured?

If we can somehow pass a node OR a string to the title prop. Or indicated a prop if the title is loading.

andrea-gm:
I wonder if @cameroncalder has any thoughts about the loading/skeleton state for PageHeader, and how we can best go about building this into the component for consistency.

lee-chase:
Closed by #761 @cameroncalder reviewed the addition of skeleton and approved of the change.

PR #761, opened 2021-05-19
Issue657 b

lee-chase:
Contributes to #611 

Adds a loading state to the title

#### What did you change?

PageHeader, tests and stories

Storybook and CI-Check


packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```diff
@@ -630,12 +661,18 @@ export let PageHeader = React.forwardRef(
...
                       })}>
-                      {TitleIcon ? (
+                      {TitleIcon && !titleLoading ? (
                         <TitleIcon className={`${blockClass}__title-icon`} />
-                      ) : (
-                        ''
-                      )}
-                      <span title={title}>{title}</span>
+                      ) : null}
+                      <span title={titleText}>
```
matthewgallo, line 667: It seems that in most other places, you check that you either have the `titleText` before rendering it _or_ that  `titleLoading` is false before rendering `titleText`. Should that be the same in this spot @lee-chase?

matthewgallo: Maybe something like this

```jsx
<span title={!titleLoading ? titleText : null}>
```

lee-chase: Agreed.

matthewgallo (approved):

PR #762, opened 2021-05-19
chore: remove official release

davidmenendez:
- [x]  the flags for the components in `package-settings.js` should be changed to true
- [x]  the component SCSS should be included in `/src/components/_index-released-only.scss.`
- [x]  run the tests, recreating snapshots (using `yarn jest -u`), and check in the updated public CSS snapshot as part of the PR
sandbox link https://codesandbox.io/s/remove-modal-3wcmu


davidmenendez:
Screen Shot 2021-05-19 at 11 05 19 AM

includes additional fix to use `Button` instead of the button props from `ModalFooter` to address styling problems

packages/cloud-cognitive/src/components/RemoveModal/RemoveModal.js
```diff
@@ -78,14 +79,18 @@ export let RemoveModal = forwardRef(
...
-        <ModalFooter
-          danger
-          primaryButtonText={primaryButtonText}
-          secondaryButtonText={secondaryButtonText}
-          primaryButtonDisabled={primaryButtonDisabled}
-          onRequestClose={onRequestClose}
-          onRequestSubmit={onRequestSubmit}
-        />
+        <ModalFooter>
+          <Button type="button" kind="secondary" onClick={onClose}>
```
sydrosa, line 83: I still think the prop should be called `onRequestClose` to match with the props across all our components.

sydrosa (changes requested):
Just want a small change -- every other component that uses a close function has a prop called `onRequestClose` and we should stay consistent with that. Is there a reason why `onRequestClose` was changed to `onClose` ?

packages/cloud-cognitive/src/components/RemoveModal/RemoveModal.js
dcwarwick: There are two different semantics here:
* `onClose` is for components that manage their own open/closed state, and which change their state automatically in response to a user action. `onClose` is a notification that the component is closing, and this can be cancelled by returning `false` from the `onClose` handler but if the handler doesn't return false (or no handler is provided) the component will go ahead and close itself.
* `onRequestClose` means that the user has done something that suggests the component should now be closed, eg clicking on an 'x' button or 'Cancel' or whatever, but it is the client code that controls whether the component is open or closed, so this is the notification that they should now do that. There's no return value from `onRequestClose`: if the client code doesn't want to close the component it just doesn't do anything.

So, we should use whichever of these behaviours matches what this component is doing.
* `AboutModal` closes itself automatically, like a ComposedModal, so it has `onClose` which is cancellable by returning `false`.
* `Tearsheet` doesn't handle its own open/close state -- that's left to the client, so it has `onRequestClose` to notify the client that the 'x' has been clicked, or Escape key has been hit, or the user has clicked outside, or whatever.

Hope that helps.

sydrosa: @dcwarwick this brings up an additional point then, which is, should these modals handle being closed on their own or not? I'm trying to be consistent with `CreateModal` since they are essentially the same component.

davidmenendez: @sydrosa @dcwarwick here's my rational- if we look at the documentation (https://react.carbondesignsystem.com/?path=/docs/components-composedmodal--default#component-api) `onRequestClose` is specifically used in `ModalFooter` where `onClose` is used in `ComposedModal`. So the two things I want to point out is

1) if we're already completely ignoring the `ModalFooter` API then i think it makes sense to use the pre-existing `onClose` API in `ComposedModal`.

2) `onClose` is used when the user clicks the X in the header, which comes from `ComposedModal`, so this way we can use `onClose` in two places and it's using the correct API name for `ComposedModal`

I think whatever is decided this should be used to set precedence going forward for which API to use for any modals we work with.

davidmenendez: and just to point out i had to add another commit right now to add `onClose` on line 64 to `ComposedModal`

sydrosa (approved):
Dope, thanks for your patience 🎉

Issue #749, opened 2021-05-18
Remove ComboButton component


dcwarwick:
## What will this achieve?

A new ButtonMenu component has been added which more closely aligns with the current button menu design pattern. The ComboButton is a legacy component contributed from the security package and can now be removed in favour of ButtonMenu. This will also enable us to simplify the package structure as we can remove the 'security' package from the dependency network.

dcwarwick:
There is some button pattern work about to be kicked off by Laura Marshall, so we should hold this until the outcome of that, as there may be a combo button pattern that this component can continue to provide or be integrated into the ButtonMenu component (which does not currently directly support the combo button presentation).

dcwarwick:
stale bot can go boil its head

lee-chase:
Combo button is now part of new button pattern.

PR #769, opened 2021-05-20
Remove dependency on security package to simplify package structure

### LINKED ISSUE #681: Remove dependence on the discontinued carbon bundler
It's gorn. Gorn orff.

dcwarwick:
Contributes to #749
Resolves #681 

Move ComboButton into our package (possible temporarily) pending outcome of the button patterns work -- we'll then work out what we need (ButtonMenu, ComboButton, others...?) going forward. In the mean time, this simplifies our structure by removing the complexity of the dependency on the security pkg.

Ran build, tests, storybook. All seems unchanged. This shouldn't affect (or even be noticeable) to downstream users.


lee-chase (commented):
Should the component be canary and/or have the deprecate warning?

dcwarwick:
> Should the component be canary and/or have the deprecate warning?

The ComboButton component has been in our package non-canary up until now. This probably should change -- in fact, there are a number of additional things that should be done to this component:
* I think it should be canary, like other non-released public components.
* The class prefix should be switched to $pkg.prefix rather than 'security'.
* The file conventions should be brought into line with our code standards and practices.

However, I suggest these are done as follow-on updates, as they will affect downstream users. This PR is intended to be invisible to our downstream users, and just cleans up our package structure. It's simpler to check that we haven't broken anything if we take things step by step.

lee-chase (approved):

Issue #760, opened 2021-05-19
EmptyState: review findings


matthewgallo:
## What will this achieve?

- [x] Change `customIllustrationAltText` to simply `illustrationDescription`
- [x] Change `linkUrl` and `linkHref` props into a single `link` prop that is a shape of 
```js
link: PropTypes.shape({
    href: PropTypes.string.required,
    text: PropTypes.string.required,
  }),
```
- [x] Change all action props into a single `action` prop that is a shape of 
```js
action: PropTypes.shape({
    text: PropTypes.string,
    renderIcon: PropTypes.string,
    kind: PropTypes.string,
   onClick: PropTypes.func,
  }),
```
- [x] Remove console log in `EmptyState.stories.js`
- [x] Use spacing-10 and spacing-11 tokens for illustration sizes in `_empty-state.scss`
- [x] `title` and `subtitle` props in EmptyStateContent should be refactored to remove typeof string check
- [x] `EmptyState.test.js` copyright header should have 2021 in date

matthewgallo:
This is complete and the empty state components have been released.

PR #772, opened 2021-05-20
Fix storybook previews for components in folders

dcwarwick:
Contributes to #760

Prefixing the component name with a folder slug meant the component was not recognised so was categorised as 'internal'.

#### What did you change?

Added a folder slug parameter to `getStorybookSlug` and update all usages to pass null, or the appropriate folder slug for EmptyStates, Tearsheets, and Cards.

Ran storybook.

### DIFF
--- a/packages/cloud-cognitive/config.js
+++ b/packages/cloud-cognitive/config.js
@@ -17,7 +17,20 @@ export const getStorybookPrefix = (pkg, componentName) => {
     : storybookPrefixInternal;
 };
 
-export const getStorybookSlug = (pkg, componentName, scenario) => {
+/**
+ * A helper function to return the slug (structure path name reduced to lower
+ * case text and hyphens) which identifies story instances.
+ * @param {string} pkg The package settings instance.
+ * @param {string} folders The names of any sub-folders being used, or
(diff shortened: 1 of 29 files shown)


packages/cloud-cognitive/config.js
```diff
@@ -17,13 +17,26 @@ export const getStorybookPrefix = (pkg, componentName) => {
...
+ * @param {string} pkg The package settings instance.
+ * @param {string} folders The names of any sub-folders being used, or
+ * null/undefined if the component appears immediately within the category. The
+ * folders should be specified as a slug (all lower case, with hyphens rather
+ * than / to separate multiple folder levels).
+ * @param {string} componentName The name of the component.
+ * @param {string} scenario The scenario name, also as a slug.
+ * @returns
+ */
+export const getStorybookSlug = (pkg, folders, componentName, scenario) => {
```
matthewgallo, line 32: If we set a default parameter value for `folders` to `null` and have it be the last parameter of this function, we won't have to update all of these other files that use `getStorybookSlug`.

Something like...
```js
export const getStorybookSlug = (pkg, componentName, scenario, folders = null) => { ... }`
```

matthewgallo (approved):

Issue #32, opened 2020-07-28
Additional tag features


SimonFinney:
[View additional tags implementation](https://ibm-cloud-cognitive.netlify.app/?path=/story/experimental-tagset--tag-array)

To do
- [x] show all feature
- [x] Align the tooltip (https://github.com/carbon-design-system/carbon/pull/7309)

carrenelloyd:
**[Link to design](https://pages.github.ibm.com/cdai-design/pal/components/additional-tag-features/)**

Maintainer: Marcie Coral

Catalog number: 1

SimonFinney:
@carrenelloyd Would you know if there's redlines or existing code available for this?

lee-chase:
@marciecorral initial dev review.
1. In essence, the overflow part is a tooltip. The only difference is the change to Gray90
2. Gray90 is not a Carbon theme token and should NOT be used see https://www.carbondesignsystem.com/guidelines/color/usage
$interactive-02: Gray 80 is a better choice as the current tooltip background and the high contrast tooltip.
3. Hover is a poor interaction which does not work on well for interactive elements such as this one. Carbon defines the clickable interactive tooltip for this scenario.
4. What happens after the modal is closed?
4.1. A keyboard user may expect the focus to return to the +tag or list.
4.2. Is the list still open?
5. Should it support the tooltip positional settings?
6. The modal is ugly, would it not be better to present the tags as tags in this list. These could be presented many to a line.
7. If there are enough to cause the modal to scroll should it be filterable?
8. How narrow should it go? Should we assume it may not even fit the first tag https://codesandbox.io/s/brave-joliot-expkd
9. It is not true that accessibility is unchanged we need to ensure we deal with hidden information.
10. $link-01 is blue 60 and this color should be used for the link as it is the default carbon color.

Developer notes: The Carbon tabs now implements a feature where it assesses whether enough space is available. The same technique may be applicable here.

marciecorral:
This is an item that was designed in 2019 by a designer who is no longer here at IBM. This was never implemented in CP4MCM. We can discuss details on the call tomorrow.

Levinson1:
InVision prototype
https://ibm.invisionapp.com/share/ANO0H9RHEGK#/screens/319708255_AdditionalTagFeature_1

lee-chase:
@Levinson1 the questions asked above need discussing before this is ready for production.

CourtneyBanning:
@lee-chase 

[...]

PR #774, opened 2021-05-21
feat: tagset modal design review updates

lee-chase:
Contributes to #32

Design review updates, mainly spacing and sizes as a design review follow up.

#### What did you change?
- Corrected hidden measuring SCSS and moved to mixin for breadrumb, button set and tag set.
- Wrapped tag set css with mixin as per standards
- Styled all tags modal as per design, including fixing height of modal and adding scroll to remove search resize.
- Updated tooltip to add space between link and tags
- Restricted tooltip to 10 additional tags.
- Added a hundreds of tags story
- Removed filter tags from examples as not design intent
- Switched to composed modal


packages/cloud-cognitive/src/components/TagSet/TagSetModal.js
```diff
@@ -53,25 +58,31 @@ export const TagSetModal = ({
...
-        labelText={searchLabel}
-        placeholder={searchPlaceholder}
-        onChange={handleSearch}
-        size="lg"
-      />
-      <div className={`${blockClass}__show-all-tags-content`}>
-        {filteredModalTags}
-      </div>
-    </Modal>
+      onClose={onClose}>
```
davidmenendez, line 66: as a convention we should try to use spread for props with the same name. for example `...{ onClose, open }`

lee-chase: Done

PR #782, opened 2021-05-24
Fix custom validator-and-shape prop-type checking

dcwarwick:
We were not using the correct pattern for combining a custom validator and array-of-shape (or other PropTypes pattern). The validator should return an error, not throw it, and the prop needs to combine the validator with array-of-shape using an AND not a ONEOF.

#### What did you change?

Added a `allPropTypes` function that insists that all validators pass, and update relevant components to use it. Also added some documentation and tests to the deprecate prop helpers, after finding a bug which some tests might have shown up.

Ran tests, build, storybook.


matthewgallo (approved):
This looks great @dcwarwick!

Issue #840, opened 2021-06-01
Action buttons do not receive correct styling without `isExpressive` prop


matthewgallo:
## What package(s) are you using?

- [X] Cloud & Cognitive
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.
In the latest Carbon version, our action buttons (SidePanel, Tearsheets) need the `isExpressive` prop added otherwise they do not receive the full width that is expected. Instead, they get a max-width applied which breaks the design in multiple components.

> Is this issue related to a specific component?
SidePanel
CreateTearsheet
Tearsheet
> What did you expect to happen? What happened instead? What would you like to
> see changed?
Action buttons should not get a max width applied to them.
> What browser are you working in?
Chrome
> What version of the @carbon/ibm-cloud-cognitive packge are you using?
Latest

## Steps to reproduce the issue

1. Go to storybook
2. Look at CreateTearsheet action buttons

matthewgallo:
https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/841

PR #841, opened 2021-06-01
PR for #840: action button styling issue

matthewgallo:
Contributes to #840 

This fixes issues we've seen lately with the action buttons not receiving the expected width. In the latest carbon version, if you add `isExpressive` the max-width will not get applied which is what we need.

#### What did you change?
`SidePanel.js` 
`TearsheetShell.js`
`_action-set.scss`
Storybook

### DIFF
--- a/packages/cloud-cognitive/src/components/ActionSet/_action-set.scss
+++ b/packages/cloud-cognitive/src/components/ActionSet/_action-set.scss
@@ -26,6 +26,8 @@
   }
 
   .#{$block-class} .#{$block-class}__action-button {
+    @include carbon--type-style('body-short-01');
+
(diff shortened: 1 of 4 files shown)


dcwarwick (commented):
Since the ActionSet exists precisely to present the buttons in that layout, I wonder if the ActionSet should just put `isExpressive: true` on the buttons it creates, rather than SidePanel and TearsheetShell having to clone the structures to add that prop in?

matthewgallo:
Hey @dcwarwick, I made the changes you requested. I think we can close #843 since there is also some typography changes that this PR fixes as well.

dcwarwick (approved):

Issue #736, opened 2021-05-13
Create Side Panel: 100% test coverage


sydrosa:
# Name of component or pattern

Create: Side Panel

## Brief description

Adding 100% coverage to testing

PR #842, opened 2021-06-01
Issue 736/create side panel testing

sydrosa:
Contributes to #736 

- Adding testing to reach 100% coverage
- update `SidePanel.js` to include `hasIconOnly` for close button
- added aria label to `SidePanel` div and made `title` prop required.


~
```diff
@@ -0,0 +1,6 @@
+Merge remote-tracking branch 'upstream/main' into issue-736/create-side-panel-testing
```
matthewgallo, line 1: Looks like this was committed accidentally

matthewgallo (changes requested):
Now that `title` is required in `SidePanel.js`, the tests in `SidePanel.test.js` need to pass a default title so that they can pass again. See [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/runs/2720382764#step:7:152)

~
```diff
@@ -0,0 +1,6 @@
...
+# Please enter a commit message to explain why this merge is necessary,
+# especially if it merges an updated upstream into a topic branch.
+#
+# Lines starting with '#' will be ignored, and an empty message aborts
+# the commit.
```
davidmenendez, line 6: what file is this?

davidmenendez (changes requested):

~
sydrosa: Literally zero clue what this is -- will delet ⚡

matthewgallo: Looks like VIM output after merging branches

davidmenendez (approved):

Issue #849, opened 2021-06-02
Side Panel: Release review notes


sydrosa:
### Release review findings:

- [x] Complete a design review with Kacie
- [x] Let's rename `pageContentSelector` to `selectorPageContent` (breaking change for `CreateSidePanel` -- please deprecate if we _do_ have existing users
- [x] Open up a separate enhancement issue for mobile-friendliness
- [x] Back button for multi-step side panel should have the entire div become a button as opposed to just the surrounding area -- reach out with any questions
- [x] add aria-label to `back` button in multi-step panel
- [x] Add actions to the multi-step panel story template for consistency (and to remove the console error)
- [x] update _index-scss, constants.js, index.js,  copywrite to `2020, 2021` 
- [x] update collapsable heading to an h2 as opposed to the current h5
- [x] add expected behavior to test on line 185: `should test overlay exit animation`
- [x] Update codesandbox with component changes

Once one or more components have been reviewed for readiness, the following steps are needed:

- [x] the flags for the components in package-settings.js should be changed to true.
- [x] the component SCSS should be included in /src/components/_index-released-only.scss.
- [x] run the tests, recreating snapshots (using -u), and check in the updated public CSS snapshot. NB it is sufficient to run yarn jest styles -u to complete the snapshot updates.

matthewgallo:
https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/851

matthewgallo:
All items from review have been addressed, waiting for design approval

matthewgallo:
## Design review guidelines

### Standards

- [ ] All pattern updates/changes/iterations have been discussed with the component developer
- [ ] Patterns have been approved by either DSAG or another approving entity

### Pattern and behavior

- [ ] The component behaves according to the guidelines set by the pattern maintainer
- [ ] The component’s UI matches the pattern specifications set by the pattern maintainer
- [ ] The component’s motion matches the specifications set by the pattern maintainer(s)
- [ ] The UI produced is responsive, cross-browser, and responds to the currently set Carbon theme.
- [ ] Colors, themes, sizes and additional props are true and correct, aligning with Carbon set tokens (unless otherwise specified by the pattern)
- [ ] Paddings/margins/spacings are true and correct, in both desktop and mobile views

### Storybook

- [ ] A functioning component renders in Storybook
- [ ] The Storybook displays multiple scenarios that show how the component can be used
- [ ] Changing props in the Storybook updates the component

[...]

PR #851, opened 2021-06-02
fix(SidePanel): make review fixes for release

matthewgallo:
Contributes to #849 

Addresses the items found during the SidePanel component release review.

#### What did you change?
`SidePanel/SidePanel.js`
`SidePanel/SidePanel.stories.js`
`SidePanel/SidePanel.test.js`
`SidePanel/_index.scss`
`SidePanel/_side-panel.scss`
`SidePanel/constants.js`
`SidePanel/index.js`
`CreateSidePanel.js`
`CreateSidePanel.stories.js`
`CreateSidePanel.test.js`
Storybook and ran tests

### DIFF
--- a/packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
+++ b/packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
@@ -462,14 +462,11 @@ export let SidePanel = React.forwardRef(
                     </h5>
                   )}
                   <Button
-                    hasIconOnly
+                    aria-label={closeIconDescription}
                     kind="ghost"
                     size="small"
-                    disabled={false}
                     renderIcon={Close20}
                     iconDescription={closeIconDescription}
-                    tooltipPosition="bottom"
-                    tooltipAlignment="center"
                     className={`${blockClass}__close-button`}
                     onClick={onRequestClose}
                     ref={sidePanelCloseRef}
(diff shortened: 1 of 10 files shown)


packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
```diff
@@ -555,10 +576,10 @@ SidePanel = pkg.checkComponentEnabled(SidePanel, componentName);
 
 SidePanel.validatePageContentSelector =
```
sydrosa, line 577: Since we've updated the prop to be able to use either `selectorPageContent` _or_ `pageContentSelector`, should this be updated as well to include both?

sydrosa (changes requested):
There were some errors in the CI check -- complaints about the prop deprecation.

packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
matthewgallo: I did think about this, but since `pageContentSelector` is being deprecated I think it's best to only validate `selectorPageContent` because we don't want people using `pageContentSelector` anymore. Adding the `deprecateProps` helper function displays a warning message in the console, if `pageContentSelector` is being used, saying this prop is being deprecated and to use `selectorPageContent` instead.

image

matthewgallo:
CI errors are fixed now, forgot that I needed to update `CreateSidePanel` so that it's tests would stop failing after deprecating `pageContentSelector`, in favor of `selectorPageContent`.

sydrosa (approved):
🎉

Issue #848, opened 2021-06-02
[CDAI] Your focus-trap must have at least one container with at least one tabbable node in it at all times


asfordmatt:
We have been broken by v1.2.6 of CDAI components.
The break can be seen from the Storybook: IdeSlideOverPanel > With state manager

## What package(s) are you using?

- [ ] Cloud & Cognitive
- [x] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

```
Error: Your focus-trap must have at least one container with at least one tabbable node in it at all times
    at updateTabbableNodes (https://ibm-cloud-cognitive.netlify.app/vendors~main.438ed8b9.iframe.bundle.js:2:10558993)
    at Object.activate (https://ibm-cloud-cognitive.netlify.app/vendors~main.438ed8b9.iframe.bundle.js:2:10562913)
    at FocusTrap.componentDidUpdate (https://ibm-cloud-cognitive.netlify.app/vendors~main.438ed8b9.iframe.bundle.js:2:9461187)
    at Ji (https://ibm-cloud-cognitive.netlify.app/vendors~main.438ed8b9.iframe.bundle.js:2:9951122)
    at Sj (https://ibm-cloud-cognitive.netlify.app/vendors~main.438ed8b9.iframe.bundle.js:2:9968775)
    at exports.unstable_runWithPriority (https://ibm-cloud-cognitive.netlify.app/vendors~main.438ed8b9.iframe.bundle.js:2:9988338)
    at cg (https://ibm-cloud-cognitive.netlify.app/vendors~main.438ed8b9.iframe.bundle.js:2:9912816)
    at Jj (https://ibm-cloud-cognitive.netlify.app/vendors~main.438ed8b9.iframe.bundle.js:2:9965277)
    at yj (https://ibm-cloud-cognitive.netlify.app/vendors~main.438ed8b9.iframe.bundle.js:2:9961489)
    at https://ibm-cloud-cognitive.netlify.app/vendors~main.438ed8b9.iframe.bundle.js:2:9913107

```

asfordmatt:
Probably related to the version bump of `focus-trap-react` here:
https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/831/files

matthewgallo:
Hey @asfordmatt, I opened a [PR](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/853) that fixes this issue for you.

matthewgallo:
[...]

PR #853, opened 2021-06-03
fix: add fallback selector for focus trap (CDAI Legacy IdeSlideOverPanel)

matthewgallo:
Contributes to #848

This fixes the CDAI `IdeSlideOverPanel.js` component. In the recent package updates, the `fallbackFocus` value needs to be provided. Previously, the story named `With state manager` was breaking, with this changed everything works as expected again.

#### What did you change?
`IdeSlideOverPanel.js`
Storybook


asfordmatt (approved):
LGTM

Issue #858, opened 2021-06-04
ActionSet: expressive buttons have max width in latest carbon update


matthewgallo:
## What package(s) are you using?

- [X] Cloud & Cognitive
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.
In the latest carbon version, the expressive buttons in the ActionSet have a max-width set on them which causes layout issues when used in larger scenarios (Tearsheet, CreateTearsheet, max size side panels).
> Is this issue related to a specific component?
ActionSet
> What did you expect to happen? What happened instead? What would you like to
> see changed?
The ActionSet buttons should not have a max-width so that they can always take up the available space they fit in.
> What browser are you working in?
Chrome
> What version of the @carbon/ibm-cloud-cognitive packge are you using?
Latest

## Steps to reproduce the issue

1. You can see this issue in the code sandbox for [CreateTearsheet](https://codesandbox.io/s/createtearsheet-example-jzj7y?file=/src/MultiStepExample.js)

> Please create a reduced test case in CodeSandbox
>
https://codesandbox.io/s/createtearsheet-example-jzj7y?file=/src/MultiStepExample.js)

## Additional information

- Screenshots or code
- 
![image](https://user-images.githubusercontent.com/10215203/120805108-67b68f00-c513-11eb-954a-1435b6b4dcbe.png)

- Notes

matthewgallo:
https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/859

matthewgallo:
Complete

Issue #773, opened 2021-05-20
Tearsheet alignment to design guidelines


mitchellbernstein:
## Detailed description
Updates to tearsheet component are required to ensure consistency between the design guidelines and the implementation with code. There are a number of instances where things differ (probably due to my lack of specific guidance here and there). This issue will list as many as can be found.

## Issues

- [x] 1. Clicking on background overlay for large and narrow transactional/danger tearsheets should not close dialog.
- [x] 2. Navigation buttons should be always right aligned even when a right-hand-side-panel is present.
- [x] 3. Navigation buttons should use a max width of 232px for wide tearsheets, 50% for narrow tearsheets.
- [x] 4. Wide tearsheet with no side panels should give the Cancel button an extra 16px of padding to the left.
- [ ] 5. Tearsheet grid should be adjusted to fit new DSAG guidance.

---------

### Issue 1
Clicking on background overlay for large and narrow transactional/danger tearsheets should not close dialog.

#### Description
This will help prevent user errs from occurring if users are editing sensitive information and accidentally click off. We should use this to encourage users to click an action instead of an ambiguous area - they wouldn't know if the data is saved, erased, or applied if they click the background. It is noted as `preventCloseOnClickOutside` but should only apply as mandatory as false to transactional tearsheets and mandatory as true for passive tearsheets.

#### Reproduce
1. Open a tearsheet component.
2. Click on the background overlay
3. Tearsheet closes

#### Expected behavior
Expected to not allow for interactions if transactional tearsheet is open (i.e has action buttons). 

#### Image
_No image required_

---------

### Issue 2
Navigation buttons should be always right aligned even when a right-hand-side-panel is present.

#### Description
Side panels on the left influence how the navigation buttons align. The cancel button get's inside by the width of the side panel. However, on the right side, this behavior should not occur. The Primary and secondary buttons are required to sit "outside" of the content body area to acknowledge they operate the entire tearsheet not just a portion of it.

#### Expected behavior
Influencer panels on the left influence navigation button placement. Influencer panels on the right should sit above the navigation buttons.

#### Image

Current:
![image](https://user-images.githubusercontent.com/748263/119047528-a7ae3b80-b983-11eb-8422-e7d6c12eedbc.png)
Expected:
![image](https://user-images.githubusercontent.com/748263/119047358-6d449e80-b983-11eb-9b2f-5696fb180106.png)

---------

### Issue 3
[...]

PR #859, opened 2021-06-04
fix(ActionSet): ensure action set buttons have max width none

matthewgallo:
Contributes to #858 

This makes sure that the ActionSet buttons do not receive a `max-width` as this will cause layout issues in larger scenarios, such as Tearsheets, CreateTearsheets, and max size side panels.

#### What did you change?
`_action-set.scss`
Storybook


dcwarwick (approved):
@matthewgallo So there is actually a requirement for a max-width on the buttons -- see #773  -- but it will need applying carefully to ensure the layout doesn't go haywire. I'm happy to approve this PR as it basically fixes the specificity issue and applies our current behaviour -- I'll look at #773 in due course.

PR #876, opened 2021-06-09
Update eslint-config-carbon

dcwarwick:
Carbon have updated eslint-config-carbon with a fix to the breakage, so this adopts the rc of this version, which I have tried and tested and looks good. The linter settings have changed slightly so there are also some minor file updates to conform. I've also updated the scripts to add eslint to our commit checks, as previously only prettier was being applied to js files.

#### What did you change?

package.json, yarn, and a bunch of js files with linter fixes.

ran build, tests.


package.json
```diff
@@ -23,8 +23,8 @@
     "clean:lerna": "lerna clean --yes",
     "clean:files": "rimraf coverage node_modules results",
     "copy-common-files": "sh ./scripts/copy-common-files.sh",
-    "format": "yarn format:prettier --write",
-    "format:prettier": "prettier '**/*.{js,md,mdx,scss,ts}' '!**/{build,es,lib,storybook,ts,umd,coverage}/**' '!README.md'",
+    "format": "run-s 'lint:es --fix' format:prettier",
```
lee-chase, line 26: Is the --write not needed in this command?

dcwarwick: Yes -- I moved it to the end of `format:prettier`, on the next line vvv.  There's no use ever using prettier without the --write, cos it just dumps stuff on the console when you do that.

lee-chase (approved):

PR #880, opened 2021-06-10
fix: breadcrumb overflow in page header

lee-chase:
Contributes to #488 

The breadcrumb was not collapsing in the correct order in the page header or showing an overflow as a result. 

#### What did you change?

- How the title is retrieved for the breadcrumb.
- CSS affecting the measurement of the last item
- Added a story to demonstrate the last item being the current page.

Storybook


packages/cloud-cognitive/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -51,6 +51,20 @@ export let BreadcrumbWithOverflow = ({
   const internalId = useRef(uuidv4());
   const [childArray, setChildArray] = useState([]);
 
+  const getItemProps = (item) => {
```
dcwarwick, line 54: I'm not sure this is the best approach. It's quite a lot of work for very little use. Have you considered `innerText` for extracting a title? Perhaps we could discuss.

dcwarwick (changes requested):

packages/cloud-cognitive/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
lee-chase: How about using ReactDomServer.renderToString

dcwarwick (approved):

PR #881, opened 2021-06-10
fix: page header styling on scroll

lee-chase:
Contributes to #488 

On scroll, the breadcrumb was left with too little space underneath causing a number of visual glitches. This was due to a previous change that lifted the title row. 
In addition, the scrolling off of the breadcrumb in some scenarios was a little jerky, addressed by swapping the sticky state for a position relative.

#### What did you change?

- Undid previous change to lift title row and replaced with transform and margin-bottom.
- Corrected how button menu size was calculated
- Moved setting of CSS props to a direct DOM call to try and improve performance
- Added some breakpoint adjustments to fix small sizes
- Adjusted position of the example title icon as otherwise it could get cut off at the top.

Storybook


packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```diff
@@ -90,10 +90,15 @@ export let PageHeader = React.forwardRef(
...
-        dynamicRefs
+        setMetrics
       );
     };
+    // const setComponentCssCustomProps = useCallback(
+    //   (kvPairs) => {
+    //     return utilSetCustomCSSProps(headerRef, kvPairs);
+    //   },
+    //   [headerRef]
+    // );
```
dcwarwick, line 101: This looks like it was a temporary filler, and may not be needed now.

```suggestion
```

packages/cloud-cognitive/src/components/PageHeader/PageHeaderUtils.js
```diff
@@ -155,3 +155,12 @@ export const utilCalcSpacingBelowTitle = (
   }
   return belowTitleSpace;
 };
+
+export const utilSetCustomCSSProps = (targetRef, kvPairs) => {
```
dcwarwick, line 159:
```suggestion
// Set css style values directly onto a node. Note that this is effective
// immediately, rather than using the React `style` prop which goes through
// the React DOM update optimisation.
export const utilSetCustomCSSProps = (targetRef, kvPairs) => {
```

packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```diff
@@ -238,63 +242,73 @@ export let PageHeader = React.forwardRef(
...
     }, [
       actionBarColumnWidth,
       actionBarMaxWidth,
       actionBarMinWidth,
       pageActionInBreadcrumbMaxWidth,
       pageActionInBreadcrumbMinWidth,
+      headerRef,
     ]);
 
+    const doUpdateBreadcrumbPosition = (scrollY) => {
```
dcwarwick, line 258:
```suggestion
    // Update the breadcrumb position in response to vertical scroll.
    // This can include flipping between sticky and relative position.
    const doUpdateBreadcrumbPosition = (scrollY) => {
```

dcwarwick (dismissed):

dcwarwick (changes requested):
Three suggestion changes for your consideration. All commentary.

dcwarwick (approved):

PR #893, opened 2021-06-11
Rough and ready (untested) Carbon tokens in js

lee-chase:
Make some Carbon tokens available from our settings package.

#### What did you change?

settings.js

Haven't done yet other than build

### DIFF
--- a/packages/cloud-cognitive/src/global/js/utils/getCarbonColorTokens.js
+++ b/packages/cloud-cognitive/src/global/js/utils/getCarbonColorTokens.js
@@ -1,0 +1,32 @@
+/**
+ * Copyright IBM Corp. 2016, 2020
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { colors } from '@carbon/colors';
+// colors comes in as object depth 2
+// keys are color names, values are objects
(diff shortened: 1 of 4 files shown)


dcwarwick (commented):
This is cool -- really interesting to see how the values are exported from Carbon packages. There's more available than I'd realised.

I think we'd want to give a bit more thought to this, though. Our needs would be a bit different from what you were doing in the linter. There, you needed to know all the possible allowable token names people should be allowed to use. For our purposes, we'd only need to expose a single set of the most recent 'approved' tokens, not deprecated names. Also, we need the *values*, not just the names, and the more I think about it the more troublesome that is -- the *values* for the layout tokens, for example, are things like "2rem", as used to set them in CSS, whereas what we need to access to control animations etc are the *pixel values*.

So, I'm thinking a different approach might work better if we need Carbon token values in JS code. For example, we could set a custom CSS variable in our component CSS to the appropriate values we need, and then read those in from JS with getComputedStyle. That would have the advantage of giving us pixel values for the corresponding tokens used, could include using things like calc(), and could allow for theme overrides etc.

I suggest we hold off for now, and discuss a bit.

lee-chase:
Ah yes token values, my bad. Closing this in favour of something much simpler #894 which could be put on hold if needed.

Issue #801, opened 2021-05-27
Release TagSet 


lee-chase:
New production dependencies
If a PR introduces new dependencies in package.json, paste the following checklist into the PR as a comment and check the items off before approving.

Tip You should be able to copy and paste the section below directly into a GitHub comment, then check where criteria passed, strike out where not applicable.

## Review of dependencies

- [ ] Verify the dependency has previously been pedigree reviewed
- [ ] Verify the dependency doesn't contain any vulnerabilities
- [ ] Verify the dependency bundle size is acceptable
- [ ] Verify the dependency is actively maintained
- [ ] Verify license conditions are met e.g. forwarding of copyright if necessary.

Component reviews for readiness
Reviews will be carried out by one or more members of the core contribution team. They will include as a minimum the following checks, which establish the "definition of done" for a component.

Paste the following checklist into the epic or issue under which the component is being made ready, then check where criteria passed, and add notes to clarify how an item is complete or what remains to be done: for anything more than minor/simple items, open issues to track them.

Tip You should be able to copy and paste the section below directly into a GitHub comment, then check where criteria passed, strike out where not applicable, and add notes to each item where applicable.

Once one or more components have been reviewed for readiness, the following steps are needed:

the flags for the components in package-settings.js should be changed to true.
the component SCSS should be included in /src/components/_index-released-only.scss.
run the tests, recreating snapshots (using -u), and check in the updated public CSS snapshot. NB it is sufficient to run yarn jest styles -u to complete the snapshot updates.
### Readiness

- [x] One or more scenarios for a design pattern have been identified as a useful unit of functionality to publish.
- [x] A functioning component or components delivering those scenarios have been delivered and merged to master.
- [x] Design maintainer has approved the implementation for those scenarios (via a comment on the relevant issue/epic).

### Engineering review

- [x] Breaking changes have only been introduced with prior approval, discussion and documented in release notes. Ideally deprecation notices in the prior major version must have been added in a timely fashion.
- [x] The implementation takes into account, and does not impair, remaining and anticipated design scenarios.
- [ ] Public component features (names, props, etc) are consistent with Carbon-defined conventions and are consistent internally. Where there isn't an existing convention to apply, ensure robust precedents are being established.
- [ ] The UI produced is accessible, responsive, translatable, cross-browser, and responds to the currently set Carbon theme.
- [x] Components are functional components using hooks.
- [x] Public components which produce DOM structures support className.
[...]

PR #913, opened 2021-06-18
Issue801 tagset review3

lee-chase:
Contributes to #801 

Change TagSet from accepting children that are Carbon Tag to accepting an array of objects with a Carbon Tag like shape.

#### What did you change?

TagSet and related files as well as Pageheader which uses the TagSet. Also changed tests for both.

Storybook and CI-Check.


packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```diff
@@ -806,10 +803,9 @@ PageHeader.propTypes = {
    */
   subtitle: PropTypes.string,
   /**
-   * One or more tags, specified as a Carbon Tags component.
-   * Optional.
+   * An array of Carbon objects containing Carbon Tag properties plus 'label'. NOTE: 'filter' is not ignored.
    */
-  tags: PropTypes.arrayOf(PropTypes.element),
+  tags: TagSet.propTypes.tags,
```
dcwarwick, line 808: This is nice, BUT sadly doesn't show up well in Storybook etc. What we did with `actions` on Tearsheet and SidePanel was duplicate the proptypes definition from ActionSet -- messier, but it means the type shows up in the controls panel nicely.

packages/cloud-cognitive/src/components/TagSet/TagSet.js
```diff
@@ -304,6 +297,15 @@ TagSet.propTypes = {
    * label for the overflow show all tags link. **Note: Required if more than 10 tags**
    */
   showAllTagsLabel: string_required_if_more_than_10_tags,
+  /**
+   * An array of Carbon objects containing Carbon Tag properties plus 'label'. NOTE: 'filter' is not ignored.
```
dcwarwick, line 301: I'm not sure that a "Carbon object" makes sense, and I think 'label' needs just a soupçon more explanation. I've adapted the following from the docs for 'actions' on ActionSet. I didn't include the note about 'filter' because I didn't understand it: see next comment.

```suggestion
   * The tags to be shown in the TagSet. Each tag is specified as an object
   * with optional fields: 'label' to supply the tag label, and additional
   * fields in the object will be passed to the Carbon Tag component: these
   * can include 'type', 'filter', 'disabled', 'ref', 'className', and any
   * other Tag props. Any other fields in the object will be passed through
   * to the HTML element as HTML attributes.
   *
   * See https://react.carbondesignsystem.com/?path=/docs/components-tag--default
```

packages/cloud-cognitive/src/components/TagSet/TagSet.js
```diff
@@ -304,6 +297,15 @@ TagSet.propTypes = {
...
+   */
+  tags: PropTypes.arrayOf(
+    PropTypes.shape({
+      ...prepareProps(Tag.propTypes, 'filter'),
```
dcwarwick, line 305: Why are we removing 'filter' from the shape? I saw the comment about how 'filter' is NOT ignored -- if it's not ignored, are people meant to be able to supply it? And if they are meant to supply it, why not include it in the shape? Removing it in the shape doesn't stop them supplying it, of course, but it does remove the type checking from it, and I'm not sure why we'd want to do that. If this is doing something important, perhaps add some clearer comment explanation as to why we're excluding it from the shape and then using it anyway.

```suggestion
      ...Tag.propTypes,
```

dcwarwick (changes requested):

packages/cloud-cognitive/src/components/TagSet/TagSet.js
lee-chase: Ooops. one too many instances of 'not', namely the one the NOTE is about.

matthewgallo (approved):
Looks great, aligns a lot better with how we handle items passed to `ActionSet` and other components.

dcwarwick (approved):

Issue #948, opened 2021-06-25
Card refactor


davidmenendez:
the `Card` component is a fairly large and complicated component that was done pretty early on. i'd like to try to refactor it into multiple pieces to make it a little easier to work with.

PR #949, opened 2021-06-25
fix: card refactor to multiple exports

davidmenendez:
contributes to #948 

takes `Card` and puts existing code into two additional exports- `CardHeader` and `CardFooter`. the body component wasn't really worth it as it's just a pretty straight forward `<div>{children}</div>` implementation. most of the logic remains in `Card` and is passed down to the other components. aside from that updates were made to the class names to be more in line with current standards.


packages/cloud-cognitive/src/components/Card/Card.js
```diff
@@ -101,114 +101,90 @@ export let Card = ({
...
-    const hasClickEvent = !!onClick || !!onKeyDown;
     const clickable =
       (hasClickEvent && !productive) ||
       (hasClickEvent && productive && clickZone === 'one');
     const cardProps = {
-      className: cx(`${pkg.prefix}-card`, {
-        [`${pkg.prefix}-card--productive`]: productive,
-        [`${pkg.prefix}-card--clickable`]: clickable,
-        [`${pkg.prefix}-card--media-left`]: mediaPosition === 'left',
+      className: cx(`${blockClass}`, {
```
matthewgallo, line 109: This could just be

```jsx
className: cx(blockClass, {
```

packages/cloud-cognitive/src/components/Card/CardFooter.js
```diff
@@ -0,0 +1,76 @@
...
+  onSecondaryButtonClick,
+  primaryButtonKind,
+  primaryButtonText,
+  productive,
+  secondaryButtonKind,
+  secondaryButtonText,
+}) => {
+  const blockClass = `${pkg.prefix}--card`;
+  const footerClass = `${pkg.prefix}--card__footer`;
+  const footerClasses = cx(`${footerClass}`, {
```
matthewgallo, line 29: Same here, this could be abbreviated to just
```jsx
const footerClasses = cx(footerClass, {
```

packages/cloud-cognitive/src/components/Card/CardHeader.js
```diff
@@ -0,0 +1,65 @@
...
+  caption,
+  hasActions,
+  label,
+  productive,
+  title,
+  titleSize,
+}) => {
+  const blockClass = `${pkg.prefix}--card`;
+  const headerClass = `${blockClass}__header`;
+  const headerClasses = cx(`${headerClass}`, {
```
matthewgallo, line 25: Same comment as [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/949/files#r660859029)

packages/cloud-cognitive/src/components/ProductiveCard/ProductiveCard.stories.js
```diff
@@ -56,14 +56,14 @@ const defaultProps = {
       id: '1',
       icon: Edit16,
       onClick: () => {},
-      onkeydown: () => {},
+      onKeyDown: () => {},
```
matthewgallo, line 59: Might be worth having these use `action('Edit action event')` to ensure these events are working as expected

packages/cloud-cognitive/src/components/Card/Card.js
davidmenendez: doh! very dumb mistake thanks

davidmenendez:
@matthewgallo changes made 👍

matthewgallo (approved):

Issue #921, opened 2021-06-21
CreateTearsheet: view all toggle enhancements


matthewgallo:
## What will this achieve?

During the initial review of the View all toggle PR, @sydrosa caught a few things that need to be added / updated:

1. ~`CreateTearsheetSection` needs correct styling.~
2. ~`CreateTearsheetSection` needs underline~
3. ~Console warning if there are no Section components inside of a Step component and the toggle is on~
4. If toggle is on, the active side nav item should update as you scroll
5. ~Influencer padding should be `$spacing-06`, _not_ `$spacing-05`~
6. ~When toggle is on, the primary actions should match the left version of the screenshot below (as opposed to the right, which is what we currently have implemented)~
![image (75)](https://user-images.githubusercontent.com/10215203/122807550-ba28d700-d299-11eb-862d-3ec195888420.png)

## Additional information

- Designs
- Existing code
- etc

matthewgallo:
![Screen Shot 2021-06-24 at 11 12 32 AM](https://user-images.githubusercontent.com/10215203/123289820-c6ea3c80-d4de-11eb-8dc2-2e1bdf497f34.png)

With the current implementation it isn't possible to add secondary label text to the progress steps, this should be allowed.

matthewgallo:
Last remaining item to close this issue
- If toggle is on, the active side nav item should update as you scroll

matthewgallo:
All items here are complete

PR #963, opened 2021-07-01
chore(CreateTearsheet): mark side nav items as active during scrolling

matthewgallo:
Contributes to #921 

Adds functionality to CreateTearsheet that will (when view all toggle is turned on), mark SideNav items as active as you scroll the tearsheet main content area. This was achieved through the use of IntersectionObserver, as opposed to adding a scroll event listener, which I've found to be much more performant.

I also noticed a console error coming from UIShell SideNav, saying it was missing a required aria-label, this was added as part of this PR as well.

#### What did you change?
`CreateTearsheet.js`
`MultiStepWithSectionsTearsheet.js`
Storybook


packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.js
```diff
@@ -578,8 +578,45 @@ export let CreateTearsheet = forwardRef(
...
+          `.${pkg.prefix}--tearsheet__content`
+        );
+        let options = {
+          root: tearsheetMainContent,
+          rootMargin: '0px',
+          threshold: 0,
+        };
+        const viewAllSections = document.querySelectorAll(
+          `.${pkg.prefix}--tearsheet-create__section.${pkg.prefix}--tearsheet-create__step--visible-section`
+        );
```
dcwarwick, line 595: Just a minor point, but I think it might make better sense to convert the nodelist to an array once, here, rather than converting it every time an intersection observer fires.

```suggestion
        const viewAllSections = Array.from(document.querySelectorAll(
          `.${pkg.prefix}--tearsheet-create__section.${pkg.prefix}--tearsheet-create__step--visible-section`
        ));
```

packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.js
```diff
@@ -578,8 +578,45 @@ export let CreateTearsheet = forwardRef(
...
+          // isIntersecting is true when element and viewport/options.root are overlapping
+          // isIntersecting is false when element and viewport/options.root don't overlap
+          if (entries[0].isIntersecting) {
+            // DOM element that is intersecting
+            const visibleTarget = entries[0].target;
+            // Convert NodeList to array so we can find the index
+            // of the section that should be marked as `active`.
+            const arrayOfSectionElements = Array.from(viewAllSections);
+            // Get visible element index
+            const visibleTargetIndex = arrayOfSectionElements.findIndex(
```
dcwarwick, line 606:
```suggestion
            // NodeList was converted to array so we can find the index
            // of the section that should be marked as `active`.
            const visibleTargetIndex = viewAllSections.findIndex(
```

dcwarwick (changes requested):

packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.js
matthewgallo: Yeah good idea!

matthewgallo: Updated

packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.js
matthewgallo: Updated

dcwarwick (approved):

PR #965, opened 2021-07-02
Fix proptypes one-of-type for Card

dcwarwick:
Card prop `action` has a prop-types declaration that looks wrong and generates warnings. I think what was meant might be `PropTypes.oneOfType([PropTypes.array, PropTypes.node])`. @davidmenendez fyi

#### What did you change?

prop-types declaration for Card prop `action`.

Ran storybook.


davidmenendez (approved):
doh! thank you

PR #975, opened 2021-07-07
Improve AboutModal scroll fade

### LINKED ISSUE #775: AboutModal scroll fading is a bit of a bodge
Sorry Simon!

The AboutModal currently uses a cunning trick to avoid having to detect whether or not scrolling is required: the body content extends right to the bottom of the body with no padding, and the 3 rem padding area is instead covered with a gradient fill that applies a gradient to the top 45% and solid white to the lower 55%. However, this has a couple of slightly odd effects:
* the fade area is only 21.6px tall, whereas in Carbon dialogs it is normally 32px, which gives a more satisfactory gradual fade effect.
* the scroll bar is not well aligned to the content, which creates a slightly odd visual effect.

It would be better to restore a proper 48px padding on the body, and install a ResizeDetector which adds a 32px fading when scrolling is needed and hides it otherwise. In fact, the ModalBody `hasScrollingContent` should ideally be used for this, rather than rolling our own fade.

dcwarwick:
closes #775

Add a resize detector to switch on and off the Carbon modal scroll effect, instead of rolling our own.

#### What did you change?

AboutModal

Ran storybook and tests


matthewgallo (approved):
I like using the `hasScrollingContent` prop instead here 👍 . However, I did notice that the modal container element has an aria label of `aria-label="[object Object]"`. I checked our deployed storybook and I see it there too so it's not from this PR but we should probably look into why that is.

Screen Shot 2021-07-08 at 10 52 37 AM

Issue #954, opened 2021-06-30
APIKeyModal release review findings 


lee-chase:
Review findings for #910 

- [ ] Scrollbar on APIKey display panel
![image](https://user-images.githubusercontent.com/15086604/123942365-fa234480-d992-11eb-80d6-bdc269c1a7a3.png)
- [ ]

davidmenendez:
> Question: Do we need issues to track creation of other scenarios shown in the design? e.g. Edit API Key.

this appears to be a more recent additional. i'll have to go back and add a story for this.

> Rendering glitch: there is a horizontal scrollbar that appears on the APIKey display panel. You might need to turn on the option to always show scrollbars to see this, or just try side-scrolling on the key display panel to trigger the scrollbar to appear. It seems to be caused by the spaced reserved for the tooltip (and used by it, when it displays).

i'm unable to replicate this. can you explain how you were able to get this behavior?

> Story glitch: in the custom step story, checking All resources does not enable the generate button. Also, the user can still edit Which resources when All resources is checked. Checking All resources should enable the generate button. Also, either Which resources should be disabled when All resources is checked or typing something in the Which resources should uncheck All resources.

> In stories, action('submitted') should actually be action('submitted')(), in order to call the action as well as create it.

> Remove apiKeyInputId prop and use internalId approach as in ExportModal.

>  Rename apiKeyVisibility to hasApiKeyVisibilityToggle, and downloadable to hasDownloadLink, for consistency with other similar props e.g. hasCloseIcon on Tearsheet.

will fix

> TBC -- Functions inside the render function like isPrimaryButtonDisabled() and getHeader() don't just use prop values but also use stateful values like currentStep -- React question: should they therefore also be useState functions, so that React knows when the results of those functions will have changed and can decide to re-render accordingly?

i hadn't really considered this before to be honest, but yes you're right this is a great use case for `useEffect` and `useState`. i'll refactor to use those.

> Rename apiKeyVisibility to hasApiKeyVisibilityToggle, and downloadable to hasDownloadLink, for consistency with other similar props e.g. hasCloseIcon on Tearsheet.

> TBC -- Functions inside the render function like isPrimaryButtonDisabled() and getHeader() don't just use prop values but also use stateful values like currentStep -- React question: should they therefore also be useState functions, so that React knows when the results of those functions will have changed and can decide to re-render accordingly?

[...]

PR #985, opened 2021-07-12
Apikey review updates

davidmenendez:
Contributes to #954 

address the issues brought up in the review of ApiKeyModal with two major inclusions including adding an edit mode and refactoring certain functions to use `useEffect` instead. i also removed all the prop comments, which i will fix in another PR as well as finishing testing for editing.


davidmenendez:
cc @dcwarwick @lee-chase

packages/cloud-cognitive/src/components/APIKeyModal/APIKeyModal.js
```diff
@@ -59,66 +66,105 @@ export let APIKeyModal = forwardRef(
       onRequestSubmit,
       open,
       previousStepButtonText,
-      secondaryButtonText,
-      successBody,
-      successHeader,
-      stepHeaders,
+      secondaryButtonText: closeButtonText,
+      showPasswordLabel,
+      visibilityToggle,
```
lee-chase, line 71: Review requested this was renamed to `hasAPIKeyVisibilityToggle`

packages/cloud-cognitive/src/components/APIKeyModal/APIKeyModal.js
```diff
@@ -236,174 +282,83 @@ export let APIKeyModal = forwardRef(
...
+            disabled={isPrimaryButtonDisabled}>
+            {primaryButtonText}
           </Button>
         </ModalFooter>
       </ComposedModal>
     );
   }
 );
 
+const customStepsRequiredProps = isRequiredIf(
```
lee-chase, line 301: Using PropTypes.string.isRequired.if directly in propTypes will add an * to the field in storybook documentation. Also worth adding a comment/note to the comment in propTypes e.g. NOTE: This prop is only required if customSteps are supplied.

packages/cloud-cognitive/src/components/APIKeyModal/APIKeyModal.js
```diff
@@ -236,174 +282,83 @@ export let APIKeyModal = forwardRef(
...
+  PropTypes.string,
+  ({ customSteps }) => customSteps.length > 1
```
lee-chase, line 303:
```suggestion
  ({ customSteps }) => customSteps && customSteps.length > 1
```

lee-chase (changes requested):

davidmenendez:
@lee-chase @dcwarwick changes made

davidmenendez:
@lee-chase @dcwarwick i was giving these changes some additional thought in regards to the use of `useEffect` and i have concluded that the original code was optimal as it was. i did some further research and using `useEffect` here to derive conditionals from props doesn't have any additional benefit and in reality causes additional unnecessary rerenders. any props that are going to effect the state of the modal are already going to cause a rerender, so we don't need to keep track of it in our own state. it's perfectly fine to just use standard variables here.

lee-chase (approved):

PR #990, opened 2021-07-13
switch ResizeDetector in CreateTearsheet to hook pattern

dcwarwick:
The hook pattern (`useResizeDetector`) is lighter weight than the full class object pattern (`<ResizeDetector>`) and fits better with our general hook-based component approach.

@matthewgallo  please check that this still works as intended

#### What did you change?

CreateTearsheet

Storybook


matthewgallo (approved):
Nice, yeah this works as expected still. I was missing the `handleWidth: true` part. Thanks @dcwarwick!

Issue #976, opened 2021-07-07
ButtonSetWithOverflow release review findings


matthewgallo:
- [x] Custom component usage in `pageActions` prop? Currently, you have to pass an array of objects that get turned into buttons, what if I want to use my own custom component?
- [x] Add some comments how the hidden button set is measuring whatever it is measuring
- [x] `pageActionsLabel` remove default string prop, mark as required, and pass this prop instead.
- [x] Do we need to add the rest props to this component, since it is internal?
- [x] When would this component use `children` vs the `buttons` prop? Specifically we were looking at line 93 in `ButtonSetWithOverflow.js` and wondering when we'll ever be inside of that `else` statement.
- [x] Add tests to check that `size` and `rightAlign` props function as expected,

lee-chase:
- The design does not currently support pageActions that are not buttons
- Rest not needed on internal components.
- The children prop is deprecated and should be removed when convenient to do so. I'll submit this as a PR by itself.
- Removed size as not used internally
- Added rightAlign test

lee-chase:
@matthewgallo 989 addresses most issues, children will be moved in a PR dependant on this one.

dcwarwick:
issues fixed in PRs above

PR #992, opened 2021-07-13
feat: remove child support from button set with overflow

lee-chase:
Contributes to #976 

#### What did you change?

- Remove use of children 
- Enforce key in buttons shape
- Adjust PageHeader as a result.

CI-Check, test and storybook.


lee-chase:
@matthewgallo think this completes the updates to ButtonSetOverflowMenu

matthewgallo (approved):
Looks good @lee-chase. One thought, it might be nice later on to extract the Button kinds into a separate file like Carbon does [here](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/prop-types/types.js)

lee-chase:
> Looks good @lee-chase. One thought, it might be nice later on to extract the Button kinds into a separate file like Carbon does [here](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/prop-types/types.js)

Good plan will submit a seperate PR.

PR #998, opened 2021-07-14
Code spell checker full on

lee-chase:
Added cspelll with a more comprehensive command line config and script.

Ran script from the command line and checked cspell working in vscode


cspell.json
```diff
@@ -0,0 +1,62 @@
...
+        "mdx",
+        "plaintext",
+        "restructuredtext",
+        "sass",
+        "scss",
+        "typescript",
+        "typescriptreact",
+        "yaml",
+        "yml"
+    ],
```
dcwarwick, line 21: These do not seem to be required, and do not seem to be doing anything.

```suggestion
```

cspell.json
```diff
@@ -0,0 +1,62 @@
...
+    "ignorePaths": [
+        "cdai",
+        "coverage",
+        "css",
+        "es",
+        "lib",
+        "node_modules", // this will ignore anything the node_modules directory
```
dcwarwick, line 28: There isn't a comment syntax for JSON. Also, this is how all the ignorePaths work.

```suggestion
        "node_modules",
```

cspell.json
```diff
@@ -0,0 +1,62 @@
...
+        "**/node_modules", // the same for this one
+        "**/node_modules/**", // the same for this one
```
dcwarwick, line 30: These are not required: the `node_modules` pattern matches node_modules appearing anywhere in the path, so these possibilities are already excluded.

```suggestion
```

package.json
```diff
@@ -34,6 +34,7 @@
     "test": "run-s test:all 'test:jest {@}' --",
     "test:all": "yarn run-all --no-sort test",
     "test:jest": "jest",
+    "spellcheck": "cspell packages/**/*.*",
```
dcwarwick, line 37: The glob needs to be in quotation marks to work properly. I don't think we need to limit the spell-check to subfolders of packages -- let's set it loose on everything, and use the `ignorePaths` to turn off anything we don't want checked. I think the `--show-context` makes the output more useful too. The `--relative` shortens the file paths a bit, and once the glob was fixed and we have hundreds of files it became a bit tedious to have all the passing files listed to, so `--no-progress` suppresses that and we just see the errors.

```suggestion
    "spellcheck": "cspell lint --relative --no-progress --show-context '**/*'",
```

packages/cloud-cognitive/src/components/PageHeader/PageHeaderDemo.stories.js
```diff
@@ -788,6 +790,7 @@ const TemplateDemo = () => {
             breadcrumbOverflowLabel="Open and close additional breadcrumb item list."
             breadcrumbItems={
               <>
+                {/* cSpell:disable-next-line */}
```
dcwarwick, line 793: Wrong. Try again. ("homepage", once correctly spelled, is accepted by cSpell. Unfortunately I can't suggest making that change as a review comment suggestion, because it's not a line you'd already proposed changing :-/)

```suggestion
```

dcwarwick (changes requested):

dcwarwick:
I think this is a good start -- let's merge it as you've got it as a new command `yarn spellcheck` not currently included in any lint or ci-check passes. There are then a few more words we can add, plus we can start whittling down the error list by fixing typos and adding ignore lines. If we get to a point where there are no errors, we might then think about adding this to the commit lint checks.

PS sorry, I've merged the dependency updates, and that's buggered up your `yarn.lock` -- probably simplest just to delete `yarn.lock` and run `yarn` again to get a new one sorted.

cspell.json
lee-chase: It does have an effect, but not necessarily the desired effect. It adds mdx files to the list that are checked in VSCode but not when running cspell from the command line.

dcwarwick: k

dcwarwick (approved):

Issue #1000, opened 2021-07-14
CreateTearsheet (narrow): initial component


matthewgallo:
# CreateTearsheet (narrow)

## Brief description

The designs for this component can be found [here](https://www.carbondesignsystem.com/community/patterns/create-flows/#narrow-tearsheet).

**When to use**
- Use a narrow tearsheet as an alternative to a modal when there is scrolling.
- Use when the form fields can be broken down into sections using section headers.

**When not to use**
- Don’t use if the flow has high complexity and the user would benefit from having a visible progress indicator.

## Details

[Link to Design](https://www.carbondesignsystem.com/community/patterns/create-flows/#narrow-tearsheet)

![NarrowTearsheetAnatomy](https://user-images.githubusercontent.com/10215203/125665913-1520af1b-35bc-4af8-8c2d-2148532e2f56.jpg)

matthewgallo:
See https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1014

matthewgallo:
Complete

PR #1014, opened 2021-07-15
feat(CreateTearsheetNarrow): initial component structure

matthewgallo:
Contributes to #1000 

This PR has the initial code for the `CreateTearsheetNarrow` component. 

Create Narrow tearsheet

#### What did you change?
Added new files for `CreateTearsheetNarrow`
Storybook
And ran tests `yarn test /CreateTearsheetNarrow/`


packages/cloud-cognitive/src/components/CreateTearsheetNarrow/CreateTearsheetNarrow.js
```diff
@@ -0,0 +1,192 @@
...
+        actions={actions}
+        open={open}
+        ref={ref}
+        onClose={onRequestClose}
+        label={label}
+        selectorPrimaryFocus={selectorPrimaryFocus}
+        verticalPosition={verticalPosition}
+        role="presentation">
+        <h3
+          className={`${blockClass}__form-title-text ${blockClass}__content-text`}>
```
davidmenendez, line 81: personally, when i have multiple classes that are like `${blockClass}__element-text ${blockClass}__element-container ${blockClass}__element-header` etc i usually just add an additional class variable so it ends up just being `const elementClass = ${blockClass}__element; ... ${elementClass}-text ${elementClass}-container ${elementClass}-element`. not a big deal. maybe just a personal preference thing, but just putting that out there.

davidmenendez (dismissed):
Looks good. I have a single small comment, but it's not a breaking thing. just something for your consideration.

packages/cloud-cognitive/src/components/CreateTearsheetNarrow/CreateTearsheetNarrow.js
matthewgallo: Yeah, I like that better too. I'll make a separate PR to add this in so we can get an initial version of this component out there

davidmenendez (approved):

PR #1033, opened 2021-07-20
fix: remove deprecated props from stories

lee-chase:
Uses getDeprecatedArgTypes to hide deprecated props from the stories.

#### What did you change?

Stories for ActionBar, CreateSidePanel, SidePanel, Tearsheet, TearsheetNarrow and TearsheetShell.

Storybook checked the deprecated properties did not show.


matthewgallo (changes requested):
I'm seeing this TypeError now coming from TearsheetNarrow. But it looks like in the other places where `getDeprecatedArgTypes` util is being used, it's working great.

Screen Shot 2021-07-20 at 2 02 44 PM

packages/cloud-cognitive/src/components/Tearsheet/TearsheetNarrow.stories.js
```diff
@@ -13,7 +13,9 @@ import { pkg } from '../../settings';
 
 import { Button, Form, FormGroup, TextInput } from 'carbon-components-react';
 
-import { Tearsheet, TearsheetNarrow } from '.';
+import { Tearsheet } from '.';
+import { TearsheetNarrow, deprecatedProps } from '.';
```
matthewgallo, line 17: This seems like the problem. Should be imported like this

```js
import { TearsheetNarrow, deprecatedProps } from './TearsheetNarrow';
```

lee-chase: Ah, unchecked in change. Good catch.

matthewgallo (approved):

PR #1046, opened 2021-07-21
chore(PageHeader): remove unnecessary pageHeaderOffset prop

dcwarwick:
I have been looking at the `pageHeaderOffset` prop, and the various notes and comments about it. It's quite a complicated prop to explain and document, and what struck me particularly was
1) The user must also arrange to move the page header down by the width of whatever else they have above the page header (eg Carbon header);
2) The user must provide something fixed to cover the area above the page header offset, otherwise random bits of scrolling page header show through and look ugly;
3) If the user moves the page header down by one amount but sets the `pageHeaderOffset` to a different amount, exotic and probably unplanned scroll behaviours result;

This means we end up needing code something like this:
image
It's not very DRY. So I thought of two possibilities to address this:
1) We could have the PageHeader itself create the margin or spacer DIV to move itself down by the `pageHeaderOffset` amount that is specified. This would eliminate the need for the first "48px" in the above code.
2) We could have the PageHeader detect how far down the page it is sitting, and use that as the offset. This would eliminate the need for the second "48" in the above code.

Option (2) seemed like an appealing one, since anyone using something like the Carbon header must already be aware of the need to provide a 48px margin or spacer DIV to shift their page content down below it, so this would enable the PageHeader to drop in seamlessly. It will also be a backwards-compatible change requiring no update to users' code (though they can simply delete their setting of the `pageHeaderOffset` prop when they wish).

The PageHeader can detect how far down the page it is. For most cases, `headerRef.current.offsetTop` would do the trick, as this reports the position of the main SECTION element on the parent. However, there are some edge cases with margins and padding or multiple containment. The reliable way to obtain the actual definite position of an element in the document is to use `getBoundingClientRect()`, which gives the position in the viewport, add on `scrollTop`, to give the position in the document, and subtract off document element `clientTop` just in case there are margins or borders or strange box models going on at the document root level. See https://github.com/timoxley/offset.

#### What did you change?

PageHeader

Ran tests and storybook. Looks fine to me. Check this out and see if it meets all the cases you know of. Do ping me for a chat if you're not sure this is the right way to address this.


lee-chase:
Hmm, I'm sure I tried something similar but found it depended on the global header CSS being applied first.

I'll have a think in the morning.

lee-chase (approved):

Issue #1028, opened 2021-07-16
CreateTearsheet render all the steps


ninjasun:
> Describe in detail the issue you're having.
 If I have many forms in different step and I menage them from a parent state every step rerender even if is not visible
Also if any step contains data fetching on the first render it will need extra condition

> What did you expect to happen? What happened instead? What would you like to
> see changed?
A step is attached on the DOM only when is current

> What browser are you working in?
chrome
> What version of the @carbon/ibm-cloud-cognitive packge are you using?
0.43 carbon 7.39

PR #1050, opened 2021-07-22
feat(CreateTearsheet): #1028 create tearsheet step add onmount prop

matthewgallo:
Contributes to #1028 

This PR adds a new prop (`onMount`) to the CreateTearsheetStep component. @ninjasun pointed out that there wasn't a way to make an api request to get data that is needed on an individual step basis. With the addition of `onMount` on the CreateTearsheetStep component, this is now possible. An example has been added to Storybook that fetches a list of users from an API _only_ on the first step and that list is presented in a dropdown.

#### What did you change?
`CreateTearsheet.js`
`CreateTearsheetStep.js`
`MultiStepWithSectionsTearsheet.js`
`_create-tearsheet.scss`
Storybook

fyi @sydrosa, would be worth adding this to CreateFullPage as well


packages/cloud-cognitive/src/components/CreateTearsheet/preview-components/MultiStepWithSectionsTearsheet.js
```diff
@@ -80,6 +87,18 @@ export const MultiStepWithSectionsTearsheet = () => {
         viewAllToggleOnLabelText="On"
         includeViewAllToggle>
         <CreateTearsheetStep
+          onMount={async () => {
+            try {
+              let data = await fetch('https://randomuser.me/api/?results=5');
```
davidmenendez, line 92: i know this is just for storybook, but 2 things- we really should never use let and just to be safe and avoid possible unknown errors you should be doing a check for `data.ok`. i'd personally refactor to
```
const data = await fetch('https://randomuser.me/api/?results=5');
const json = await data.json();
if (!data.ok) throw new Error('received non 200 response');
setAllTopicOwners(data.results);
```

davidmenendez (approved):
looks fine. had a single comment. not a deal breaker, just something to consider.

davidmenendez (approved):

Issue #996, opened 2021-07-14
Review findings for PageHeader


dcwarwick:
Items found during release review of PageHeader:

- [x] #1008
- [x] Rename `actionBarOverflowLabel` to `actionBarOverflowAriaLabel`, `breadcrumbOverflowLabel` to `breadcrumbOverflowAriaLabel`, and rename `collapseHeaderLabel` to `collapseHeaderIconDescription` and `expandHeaderLabel` to `expandHeaderIconDescription`.
- [x] Is the prop `availableSpace` actually required? Could we not pass children of the PageHeader through to this area? The children of PageHeader doesn't currently seem to be used for anything.
- [x] Rename `background` to `hasBackground` and `collapseHeaderToggleWanted` to `hasCollapseHeaderToggle`.
- [x] Prop `breadcrumbItems` takes element rather than array of shape, while all the others (page actions, actions, tags, etc) take array of shape. Is this how it needs to be?
- [x] The assignment to attribute `data-collapse` in line 670 -- is that actually used?
- [x] The commentary for prop `pageHeaderOffset` is really unclear. When it says "page header", should that actually be "global header"? And how would you actually use this? Do you also need to force a top margin? Also, could this value not be detected automatically by the PageHeader on mounting by querying its own bounding box?
- [x] Rename `preCollapseTitleRow` to `collapseTitle`.
- [x] Rename `preventBreadcrumbScroll` to `disableBreadcrumbScroll`.
- [x] Remove defaults for `collapseHeaderIconDescription` and `expandHeaderIconDescription` and make them required (if `hasCollapseHeaderToggle` is true). Remove unnecessary defaults for `className`, `preventBreadcrumbScroll` and `preCollapseTitleRow`.
- [x] Add props for TagSet `allTagsModalTitle`, `allTagsModalSearchLabel`, `allTagsModalSearchPlaceholderText` and `showAllTagsLabel`.
- [x] When the title goes into the breadcrumb, its BreadcrumbItem has `href="#"`. This should be removed: it causes the breadcrumb item to reload the page on click. Or, if it really needs to be left in, the breadcrumb item produced should not trigger a page reload on click.
- [x] Ensure that all stories pass keys to action bar items so that warnings about missing keys and missing required fields are not produced in the storybook console.
- [x] Merge PageHeaderDemo.stories.js into PageHeader.stories.js in order to eliminate the duplicate title warnings in storybook
[...]

PR #1053, opened 2021-07-22
feat(PageHeader): update for change to BreadcrumbWithOverflow props

lee-chase:
Contributes to #996 

Change page header to use the updated form of BreadcrumbWithOverflow.

#### What did you change?
- Replace breadcrumbItems with breadcrumbs in page header. 
- Updated stories and tests relating to breadcrumbs in page header
- Updated deprecated prop tests in page header.

Ran tests and storybook.


packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```diff
@@ -912,22 +900,52 @@ PageHeader.propTypes = {
...
+       */
+      key: PropTypes.string.isRequired,
+
+      /**
+       * Pass in content that will be inside of the BreadcrumbItem
+       */
+      label: PropTypes.node,
+
+      /**
+       * A string based alternative to the children, required only if children is not of type string.
```
dcwarwick, line 938:
```suggestion
       * A text version of the `label` for display, required if `label` is not a string.
```

packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```diff
@@ -912,22 +900,52 @@ PageHeader.propTypes = {
...
+        ({ label }) => typeof label !== 'string'
+      ),
+    })
   ),
   /**
    * A zone for placing high-level, client content above the page tabs.
    * Accepts arbitrary renderable content as a React node. Optional.
-   */
-  children: PropTypes.node,
+   */ children: PropTypes.node,
```
dcwarwick, line 948:
```suggestion
   */
   children: PropTypes.node,
```

packages/cloud-cognitive/src/components/PageHeader/PageHeader.test.js
```diff
@@ -595,13 +607,14 @@ describe('PageHeader', () => {
     jest.spyOn(console, 'error').mockRestore();
   });
 
-  test('Works, for now, with deprecated props', () => {
+  test.only('Works, for now, with deprecated props', () => {
```
dcwarwick, line 610: We shouldn't check in a `test.only`

```suggestion
  test('Works, for now, with deprecated props', () => {
```

packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```diff
@@ -93,6 +94,7 @@ export let PageHeader = React.forwardRef(
     hasCollapseHeaderToggle ??= deprecated_collapseHeaderToggleWanted;
     collapseTitle ??= deprecated_preCollapseTitleRow;
     disableBreadcrumbScroll ??= deprecated_preventBreadcrumbScroll;
+    const breadcrumbs = breadcrumbsIn ?? deprecated_breadcrumbItems;
```
dcwarwick, line 97: Is there a particular reason to handle this one slightly differently from the other deprecated props?

dcwarwick (changes requested):

packages/cloud-cognitive/src/components/PageHeader/PageHeader.test.js
lee-chase: Already resolved locally

packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
lee-chase: I was experiencing infinite renders when trying to re-use `breadcrumbs` directly. It is possible I had issues elsewhere at the time.

packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```diff
@@ -912,22 +899,53 @@ PageHeader.propTypes = {
...
    */
   breadcrumbOverflowAriaLabel: PropTypes.string.isRequired.if(
-    ({ breadcrumbItems }) => breadcrumbItems && breadcrumbItems.length > 0
+    ({ breadcrumbs, breadcrumbItems }) =>
+      (breadcrumbs && breadcrumbs.length > 0) ||
+      (breadcrumbItems && breadcrumbItems.length > 0)
+  ),
+  /**
+   * One or more Carbon BreadcrumbItem components, passed in as React element(s).
+   * If provided, these are rendered at the top before other header content.
```
dcwarwick, line 912:
```suggestion
   * Specifies the breadcrumb components to be shown in the breadcrumb area of
   * the page header. Each item is specified as an object with optional fields
   * 'label' to supply the breadcrumb label, 'href' to supply the link location,
   * and 'isCurrentPage' to specify whether this breadcrumb component represents
   * the current page. Each item should also include a unique 'key' field to
   * enable efficient rendering, and if the label is not a string then a 'title'
   * field is required to provide a text alternative for display. Any other
   * fields in the object will be passed through to the breadcrumb element as
   * HTML attributes.
```

dcwarwick (changes requested):

dcwarwick (approved):

PR #1082, opened 2021-07-27
Update Reviews Guidelines

sydrosa:
This PR seeks to include requirements for motion -- ensuring that we check for reduced-motion queries if motion is included in the component design.


andrea-gm:
would it be worth mentioning and linking to one of our own examples? like mention something like “check out how we did it in create tearsheet or whatever”

davidmenendez (approved):

PR #1089, opened 2021-07-29
feat(deps): add Carbon packages as peer dependencies

SimonFinney:
Resolves https://ibm-casdesign.slack.com/archives/C013ZTX0N6B/p1627494496145700

#### Proposed change

Add remaining Carbon packages as peer dependencies.

### DIFF
--- a/packages/cloud-cognitive/package.json
+++ b/packages/cloud-cognitive/package.json
@@ -65,6 +65,7 @@
   "peerDependencies": {
     "@carbon/icons-react": "^10.36.0",
     "@carbon/import-once": "^10.5.0",
+    "@carbon/colors": "^10.29.0",
     "@carbon/layout": "^10.28.0",
     "@carbon/motion": "^10.21.0",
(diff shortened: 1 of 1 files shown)


SimonFinney:
@carbon-design-system/ibm-cloud-cognitive-development I suspect we should be declaring the remaining Carbon dependencies here too like layout, themes, and type — are there any objections?

dcwarwick:
Good catch. Any package we rely on but expect our users to obtain for themselves (rather than wanting to bundling it tightly with our package) should be called out as a peer dependency. A monorepo makes it quite tricky to spot missing peer dependencies, because many packages will be fetched for something, and are then available for us. But we should try to complete the list to avoid giving our users problems.

I think the following packages also might need adding to `peerDependencies`:
- @carbon/colors
- @carbon/layout
- @carbon/themes
- @carbon/type

dcwarwick (approved):

PR #1113, opened 2021-08-04
Issue 919/create full page toggle enhancement

sydrosa:
Contributes to #919 

This PR adds toggle functionality to the `CreateFullPage` component, as well as documentation, cleaner stories in Storybook, and some other small but necessary refactoring.


davidmenendez (approved):
Looks good. I'm a bit concerned that this file is getting a little out of hand though in terms of length. obviously this is a complex pattern. it might be worth it in the future to consider refactoring this a bit to increase readability. for example- extracting some of the hooks into their own file could be one option https://reactjs.org/docs/hooks-custom.html

PR #1131, opened 2021-08-10
fix: review feedback for cards

davidmenendez:
Contributes to https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/1086


matthewgallo (commented):
image
The mediaRatio console error is gone now, however there is now a warning

packages/cloud-cognitive/src/global/js/package-settings.js
```diff
@@ -13,10 +13,12 @@ const defaults = {
     // reviewed and released components:
     AboutModal: true,
     APIKeyModal: true,
+    Card: true,
```
dcwarwick, line 16: I don't think Card should be listed here -- it's not a component our end-users will use directly. It's internal, like TearsheetShell.

```suggestion
```

dcwarwick (dismissed):

matthewgallo:
It looks like the snapshot test is failing, I just had this happen on one of my PRs. If you run `yarn` on your local branch to update the dependencies, and then run `yarn jest styles -u` and commit the new snapshot, you should be good.

matthewgallo (approved):
mediaRatio warning in storybook is still there, but everything else looks good!

packages/cloud-cognitive/src/global/js/package-settings.js
matthewgallo: This has been addressed in one of @davidmenendez's latest commits

PR #1132, opened 2021-08-11
chore: create common-settings package

dcwarwick:
There are a few things that are not specific to c&cs but may apply to other packages that come into the repo. Some of these are already formalised as 'config' packages. Create a 'common-settings' package in config for a few more bits and pieces that don't belong inside a particular package. Move the imported-carbon-modules file there, and add a test to ensure that the imported carbon modules list is checked against current carbon so that we will be alerted if it changes.

#### What did you change?

move _imported-carbon-modules.scss into a new common-settings package and added a test.

Ran build and tests


packages/cloud-cognitive/src/index-without-carbon-released-only.scss
```diff
@@ -5,8 +5,7 @@
 // LICENSE file in the root directory of this source tree.
 //
 
-@import './global/styles/imported-carbon-modules';
```
SimonFinney, line 8: @dbrugger Would you know if this import path is still being referenced in Cloud Pak for Security? If so, I'm assuming removing it may cause Sass to fail to compile.

packages/cloud-cognitive/src/index-without-carbon.scss
```diff
@@ -5,25 +5,9 @@
 // LICENSE file in the root directory of this source tree.
 //
 
-@import './global/styles/imported-carbon-modules';
-
+@import 'common-settings/scss/imported-carbon-modules';
```
SimonFinney, line 8: I believe this may break in a production Sass environments because the `common-settings` package is currently both private and listed in `devDependencies`.

dcwarwick: Ah, good point. I may have headed down the wrong route here...

PR #1145, opened 2021-08-11
DO NOT MERGE - Animated progress indicator

kristastarr:
#### What did you change?
Added animations to the ProgressIndicator component inside the CreateTearsheet.  In the preview-components folder, I added files with the proposed changes that would be made to the Carbon ProgressIndicator component.  In the CreateTearsheet component, I modified the renderProgressSteps function to accept props required to render the modified ProgressIndicator component.  I added a file with test data that is used as props.  

Viewed it in Storybook


packages/cloud-cognitive/src/components/CreateTearsheet/preview-components/stepDataTest.js
```diff
@@ -0,0 +1,53 @@
...
+        Lorem ipsum dolor, sit amet consectetur adipisicing elit. Animi
+        consequuntur hic ratione aliquid cupiditate, nesciunt saepe iste
+        blanditiis cumque maxime tenetur veniam est illo deserunt sint quae
+        pariatur. Laboriosam, consequatur.
+      </p>
+    </Tooltip>
+  );
+};
+
+const stepDataTest = [
```
matthewgallo, line 24: I think we would need this step data created with the `CreateTearsheet` component dynamically, otherwise, these will be the steps that render for anyone using this component.

matthewgallo: The logic that renders that component is now within `CreateInfluencer` and something like this would allow you to get the shape you need

```js
let stepData = [];
createComponents.steps.forEach(step => {
    stepData = [...stepData, {
      label: step.props?.title,
      description: step.props?.title,
      secondaryLabel: step.props?.secondaryLabel
    }]
});
```

kristastarr:
@matthewgallo I pulled and merged latest changes- still having the problem where the Carbon stylesheets are being pulled in in addition to the desired stylesheet in `CreateTearsheet/preview-components/`

kristastarr:
Animated progress indicator is now working properly in the CreateTearsheet component.  I will share the deploy preview with designers so we can see how all the animations look in conjunction.  I will continue to check with Carbon team periodically on the status of the PR I submitted to add the animations to the Carbon ProgressIndicator component.  If it will not be possible for Carbon to merge those changes in the near future, I will look into merging them here instead.

Issue #1136, opened 2021-08-11
CodeSandbox crash errors


SimonFinney:
https://codesandbox.io/s/github/carbon-design-system/ibm-cloud-cognitive/tree/main/examples/cloud-cognitive/codesandbox

dcwarwick:
I see this on multiple codesandboxes:
![image](https://user-images.githubusercontent.com/6385315/129062579-13130200-2985-4283-bb83-8873f72897ff.png)

lee-chase:
Seen on ExampleComponent, rectified by removing ~ from the start of CSS import https://codesandbox.io/s/examplecomponent-8189x?file=/src/_ibm-cloud-and-cognitive.scss

![image](https://user-images.githubusercontent.com/15086604/129070359-c66de39d-6366-484e-9ddf-31160abc86bf.png)

dcwarwick:
OK, the errors in Simon's original example (https://codesandbox.io/s/github/carbon-design-system/ibm-cloud-cognitive/tree/main/examples/cloud-cognitive/codesandbox) seem to be simply that there are missing dependencies from `package.json` in our example projects -- they run locally because of the global monorepo node_modules folder having lots of goodies in it. Running `yarn clean` in the monorepo then changing down to an example folder and trying to build/run it should reproduce this error locally.

This is probably unrelated to the issues we've had with codesandbox generally being buggered recently.

dcwarwick:
I'm now seeing the "Path must be a string. Received null" message from codesandboxes that were previously giving me the "NoSuchMethodError" message.

dcwarwick:
Looks like removing the ~ from these paths fixes it:
![image](https://user-images.githubusercontent.com/6385315/129075777-73b64197-fa70-4cbd-b6c8-51bda46fc649.png)

PR #1148, opened 2021-08-12
fix: code sandbox examples

lee-chase:
Contributes to #1136

The crash errors largely appear to have improved (different error) and now just require the removal of '~' at the beginning of SCSS import statements.

#### What did you change?

- Updated package.json for the three example projects to pull in the latest version of our components.
- Updated the test so it passes for examples/cloud-cognitive/ccs-base-react-16 and 17
- Added React import to examples/cloud-cognitive/ccs-base-react-16
- Added canary instructions to examples/cloud-cognitive/codesandbox

Ran the projects locally having renamed root node_modules.
Ran the test for 16 & 17


examples/cloud-cognitive/ccs-base-react-17/src/App.test.js
```diff
@@ -3,6 +3,5 @@ import App from './App';
 
 test('renders learn react link', () => {
   render(<App />);
-  const linkElement = screen.getByText(/learn react/i);
-  expect(linkElement).toBeInTheDocument();
+  screen.getByText(/Show AboutModal/i);
```
dcwarwick, line 6: A bit of a nitpick, but I think the name of this test should also be changed. Also, it's our normal convention to go for `it` rather than `test`.

dcwarwick (changes requested):

examples/cloud-cognitive/ccs-base-react-17/src/App.test.js
dcwarwick: @lee-chase would be good to get this done and merged if there isn't a particular reason to hold it

lee-chase: Done @dcwarwick

dcwarwick (approved):

Issue #1062, opened 2021-07-23
Queries raised following Page Header evaluation


lee-chase:
Queries originally raised via email (Alan Coffey/Ireland/IBM)

## What package(s) are you using?

- [x] Cloud & Cognitive
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## What did you want to know?

I am currently evaluating designs (Cameron mentioned you as a contact) planned for release in the 4th quarter, the design utilises the PageHeader component. 

I fully understand this is currently tagged canary and some findings listed below may be working as designed, never planned or I may have made a mistake my side - I'd like to share and get your opinion nonetheless. 

**Breadcrumbs**

1. CLX Return to Application not possible with Breadcrumb items prop.  (possible candidate for composition ?) 

2. Long title prop collapses all breadcrumbs, even when space 
![image](https://user-images.githubusercontent.com/15086604/126764307-3880852f-33bd-464a-9e3d-7e122e9bbdd4.png)

**On Scroll **
3. Action buttons partially obscured regardless of pageHeaderOffset props value
![image](https://user-images.githubusercontent.com/15086604/126764326-ca787144-8674-415d-9636-798b56ca57df.png)

**Tabs**
4. No overflow arrows rendered on page load 
![image](https://user-images.githubusercontent.com/15086604/126764337-e148932b-d89b-4eaf-9be7-aa79629934c9.png)

5. rendered after window resize 
![image](https://user-images.githubusercontent.com/15086604/126764362-2816a056-b781-46b3-a993-a3855971ecd3.png)

6. Displayed when no content to show 
![image](https://user-images.githubusercontent.com/15086604/126764371-8896ffe3-8ac3-4a3f-95d1-7e03cf6fc8b9.png)

**Tags**

7. 1 long tag name hides all tags 
8. Count tag looks like more Tabs 
9. No tooltip on tag count  
![image](https://user-images.githubusercontent.com/15086604/126764385-a089ca58-d3aa-4588-8e88-e64f1975ae63.png)

**Proposed design unsupported features**

10. Not Possible to add content to the left of the title. 
![image](https://user-images.githubusercontent.com/15086604/126764409-3ccb5104-eb88-485d-bb12-2113192a6338.png)

lee-chase:
FYI @cameroncalder and Alan Coffey.

I have numbered the items above for clarity. Feel free to set up a meeting to discuss, here are my initial thoughts.

[...]

PR #1151, opened 2021-08-13
feat: custom button set with overflow straw man

lee-chase:
Contributes to #1062 

Adds the option of providing a custom component to the ButtonSetWithOverflow. 

- Warnings raised about the `closeMenu` and `handleOverflowMenuItemFocus` not implemented
- Presume above causes a lack of focus in a narrow version
- How to document/support different custom versions. For example, different behaviours when horizontal or vertical space is limited. NOTE: We could accept multiple scenario-based custom props or document the host classes the user will need to check for.

#### What did you change?

ButtonSetWithOverflow and story.

Storybook

### DIFF
--- a/packages/cloud-cognitive/src/components/ButtonSetWithOverflow/_storybook-styles.scss
+++ b/packages/cloud-cognitive/src/components/ButtonSetWithOverflow/_storybook-styles.scss
@@ -1,0 +1,14 @@
+@import '../../global/styles/carbon-settings';
+
+.custom-node.custom-node {
+  min-width: 160px;
+}
+
+.custom-node.custom-node .custom-node__trigger {
+  width: 100%;
+  padding: 0 $spacing-05;
+}
+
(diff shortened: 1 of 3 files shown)


dcwarwick:
Where do you see warnings? I didn't spot any.

It looks like the OverflowMenu would need to cancel the click handling on it in order to work properly in a drop-down.

packages/cloud-cognitive/src/components/ButtonSetWithOverflow/_storybook-styles.scss
```diff
@@ -0,0 +1,14 @@
+@import '../../global/styles/carbon-settings';
```
dcwarwick, line 1:
```suggestion
@import '../../global/styles/project-settings';
```

dcwarwick (changes requested):

Issue #1123, opened 2021-08-06
CreateTearsheet/FullPage: extract some useEffect logic into custom hooks


matthewgallo:
## What will this achieve?

The CreateTearsheet and CreateFullPage components are getting a bit hefty in terms of file size/length. #1107 will help with extracting some shared logic in regards to the influencer section. @davidmenendez pointed out that it would be a good idea to look into creating some custom hooks to help keep the size of these components as small as we can.

## How will success be measured?

Potential items that could be moved out of each component and into custom hooks: 
1. useEffect that tracks multi step logic
2. useEffect that focuses first focusable element on step change
3. usePreviousValue can move out of the component and into location where is can be used by both (reduce repeated code across these components)
4. useEffect that logs warnings to console when an incorrect configuration is provided
5. IntersectionObserver useEffect

matthewgallo:
Some of this has been completed. We still need to abstract the multi step useEffect into it's own custom hook that is shared by both CreateFullPage and CreateTearsheet components. That one contains a lot of logic and adding this will help a lot to reduce the duplicate logic.

matthewgallo:
Last item left here once https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1159 is merged is the useEffect that logs warnings if an invalid combination of CreateComponents are used together

PR #1164, opened 2021-08-17
chore(CreateFullPage): use new custom hook

matthewgallo:
Contributes to #1123 

Adds the new custom hook `useCreateComponentStepChange` to the CreateFullPage component. This reduces the amount of code needed within that component and more closely ties the logic of CreateTearsheet and CreateFullPage.

#### What did you change?
`useCreateComponentStepChange.js`
`CreateFullPage.js`


davidmenendez (approved):
reusing code- love to see it

PR #1170, opened 2021-08-18
fix(CreateTearsheet): remove grid margin

matthewgallo:
The main content of the create tearsheet should remain left aligned, this PR removes the margin from the grid component that was previously preventing this.

#### What did you change?
`_create-tearsheet.scss

Storybook


SimonFinney (approved):
LGTM — thank you 💯

Issue #911, opened 2021-06-17
Cascading Card Wrapper


skcele:
# Name of component or pattern
Cascading Card Wrapper

This component is part of the C&CS Micro-interactions project with Andrew Smith.

## Brief description

This wrapper will add cascading card motion (fade in one at a time) for cards or tiles placed within the wrapper. Ideally we will also make it available for those within a grid.

## Details
Design example: https://ibm.box.com/shared/static/9mo3tk3d5k9oqagvraw5jisms4gecljx.gif

Prototype with no grid: https://codepen.io/stephaniecree/pen/abJBzEJ
This works best if you resize the window to see them on separate rows too.

Existing motion within the Grid Layout in Cloud PAL, which we will pull from: https://pages.github.ibm.com/ibmcloud/pal/layouts/grid-layout/code#dashboard-with-motion
You can type spaces into the codebox below the demo to see the motion.

## Design Maintainer(s)

Stephanie Cree is main design maintainer, and Lisa Kaiser is a second contact.

We are working with @davidmenendez on this project!

## Tasks

Before starting work on this epic, please review and complete the following.

- [x] Initial review of design/existing code

Review has been done with the C&CS Microinteractions team, happy to do another if needed.

### Working in Cloud and Cognitive package

- [x] Have you reviewed our
      [contribution guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/CONTRIBUTING.md)?
- [x] Have you noted our
      [reviewer guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/REVIEWERS_GUIDELINES.md)?
- [ ] Did you use the CLI to scaffold your new component? -  David and I will be working on this next.

davidmenendez:
Thanks for opening an issue 👍

davidmenendez:
ok i have a basic implementation on my end that's set up based on the demo code you provided. this could possibly be expanded to integrate grid functionality, but i'll need to think the implementation through a bit more. i'll post a link shortly.

PR #1171, opened 2021-08-18
feat: cascade component

davidmenendez:
Contributes to #911 

initial component structure for cascade

### DIFF
--- a/packages/cloud-cognitive/src/components/Cascade/_cascade.scss
+++ b/packages/cloud-cognitive/src/components/Cascade/_cascade.scss
@@ -1,0 +1,50 @@
+//
+// Copyright IBM Corp. 2020, 2021
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+// Standard imports.
+@import '@carbon/import-once/scss/import-once';
+@import '../../global/styles/carbon-settings';
+@import '../../global/styles/project-settings';
+
+@mixin cascade {
+  $block-class: #{$pkg-prefix}--cascade;
+  $animationProps: 300ms cubic-bezier(0.2, 0, 0.38, 0.9) 0s 1;
+  $fadeStart: translateY(3rem);
+  $fadeEnd: translateY(0);
+
+  @keyframes fade {
+    0% {
+      opacity: 0;
+      transform: $fadeStart;
+    }
+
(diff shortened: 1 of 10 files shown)


packages/cloud-cognitive/src/components/Cascade/_cascade.scss
```diff
@@ -0,0 +1,50 @@
...
+//
+
+// Standard imports.
+@import '@carbon/import-once/scss/import-once';
+@import '../../global/styles/carbon-settings';
+@import '../../global/styles/project-settings';
+
+@mixin cascade {
+  $block-class: #{$pkg-prefix}--cascade;
+  $animationProps: 300ms cubic-bezier(0.2, 0, 0.38, 0.9) 0s 1;
```
SimonFinney, line 15: Does it make sense to reference some members from `@carbon/motion`'s Sass here?

SimonFinney (approved):
LGTM 🥇  I had one thought on leveraging the `@carbon/motion` package — thank you!

packages/cloud-cognitive/src/components/Cascade/_cascade.scss
davidmenendez: yeah i think so. sorry i should have marked this as a draft because i'm still getting input from the original developer 😅 

i'll probably be making some additional changes to this to get it in line

packages/cloud-cognitive/src/global/js/utils/story-helper.js
```diff
@@ -19,6 +19,7 @@ const storybookStructure = {
   BreadcrumbWithOverflow: `${lib}/$rci/$comp`,
   ButtonMenu: `${lib}/$rci/$comp`,
   ButtonSetWithOverflow: `${lib}/$rci/$comp`,
+  Cascade: `${lib}/$rci/$comp`,
```
dcwarwick, line 22: Oops -- this mechanism has been changed. The storybook structure is now in `/packages/core/storybook-structure.js`. Also, we will need to figure out where it should go in the new structure! @andrea-gm FYI. Is this currently described in the PAL site? For now, maybe you could add it into the 'Internal' section while we figure it out -- something like this:
image

dcwarwick (changes requested):

dcwarwick (approved):

Issue #67, opened 2020-08-05
Text bar


carrenelloyd:
Catalog # 25

[Link to Design 
](https://pages.github.ibm.com/cdai-design/pal/components/text-bar/)

**Maintainer**
Myriam Battelli

lee-chase:
Blocked as the status of the design it was set to 'status: refining' by @carrenelloyd 

Invited Myriam to join the conversation here.

m-battelli:
@carrenelloyd @lee-chase 
I have finalized this pattern with Carbon team: https://www.carbondesignsystem.com/patterns/text-toolbar-pattern/
We are looking for volunteers to implement it in code version.

lee-chase:
Thankyou @m-battelli as this is an open-source project, in theory, any dev could pick it up. Please ask any volunteer you have to join the conversation on github and provide any prototype links you may have.

lee-chase:
So looking at this component the following are not available form Carbon

- A toolbar button - a simple adjustment to a ghost button
- A toolbar separator/group - simple enough to create
- A dropdown that works within the confines of a toolbar. WARNING we may need to adjust to make it work within the confines of overflow hiding. Otherwise a slight wrap of the carbon component
- A composed dropdown e.g. for color, icon buttons, URL and search does not exist.
- A dropdown capable of loading and displaying system fonts does not exist in Carbon.
- A color picker does not exist
- A dropdown icon bar does not exist.
- A dropdown with the style and functionality of the link button does not exist.

lee-chase:
The component is a significant task requiring various drop downs and information about the current state of the text editor. In addition, it seems somewhat odd to have such a functional text bar component without an editor.

lee-chase:
@carrenelloyd is there a rich text editor we have in mind for this?

andrea-gm:
Perhaps the question above is for @m-battelli

ConradSchmidt:
If you planning to use [CKEditor](http://ckeditor.com/) (most IBM products use it) I did a prototype/PoC to see if it would work: https://pages.github.ibm.com/conrad-schmidt/ckeditor-carbon-test/
This is **not** ready to use, but it can be a starting point

Main Editor code here: https://github.ibm.com/conrad-schmidt/ckeditor-carbon-test/blob/master/src/components/Editor/Editor.js

![image](https://user-images.githubusercontent.com/3808948/106799947-2646cd80-6660-11eb-97e6-e90713de28d2.png)

[...]

PR #1173, opened 2021-08-19
feat(toolbar): add vertical variant

### LINKED ISSUE #806: Toolbar — orientation
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-orientation_attribute

SimonFinney:
#### Affected issues

- Contributes to #67 
- Resolves #806 

#### Proposed changes

- Add toolbar vertical variant logic, stories, styles, and test
- Update default toolbar story to better reflect [guidance](https://pages.github.ibm.com/cdai-design/pal/patterns/toolbars/#responsiveness)
- Refactor common toolbar tests to better accommodate individual component tests


packages/cloud-cognitive/src/components/Toolbar/_toolbar.scss
```diff
@@ -5,38 +5,55 @@
...
 @include exports($name: cloud-cognitive--toolbar) {
   $block-class: #{$pkg-prefix}--toolbar;
+
   $border: carbon--rem(
       $px: 1px,
     )
     solid $ui-03;
 
+  $dimensions: carbon--rem(
+    $px: 40px,
```
matthewgallo, line 21: I believe you could just use `$spacing-08` here

matthewgallo:
https://user-images.githubusercontent.com/10215203/130460308-af865b1d-ed6c-4840-96d0-a49abc9c2549.mov
For the vertical toolbar, it might be easier to use if the tooltip has the option to be placed from the left or right. You cannot easily hover over the item above the current toolbar item because the tooltip is on top of it.

packages/cloud-cognitive/src/components/Toolbar/_toolbar.scss
SimonFinney: Thanks — would we typically use spacing tokens for properties other than [margin and padding](https://carbondesignsystem.com/guidelines/spacing/overview/#applying-spacing)? As I understand it, Carbon mostly favors [`rem` for height and width dimensions](https://github.com/search?q=%27width%3A%27+path%3Apackages%2Fstyles+repo%3Acarbon-design-system%2Fcarbon&type=Code). What's the current thinking here?

matthewgallo: You’re totally right! Nevermind then, what you had is good

SimonFinney:
> https://user-images.githubusercontent.com/10215203/130460308-af865b1d-ed6c-4840-96d0-a49abc9c2549.mov
> For the vertical toolbar, it might be easier to use if the tooltip has the option to be placed from the left or right. You cannot easily hover over the item above the current toolbar item because the tooltip is on top of it.

@matthewgallo a quick update on this — the design recommendation is that the default tooltip position should be `top` for both orientations. My gut here is to create an issue and reduced test case in Carbon to investigate further, as it feels like a defect that may not affect just us.

matthewgallo (approved):

Issue #1180, opened 2021-08-19
Tearsheet (narrow): stories should follow design guidance


matthewgallo:
## What package(s) are you using?

- [X] Cloud & Cognitive
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

~The tearsheet (narrow) story should only use an ActionSet with two buttons.~ Narrow tearsheet can allow 3 buttons (just not CreateTearsheetNarrow). And the story named "Narrow tearsheet with all header items" is not using any action buttons, these should be included by default.
> Is this issue related to a specific component?

Tearsheet (narrow)

matthewgallo:
Complete

PR #1182, opened 2021-08-19
chore(Tearsheet): set default action set for narrow story

matthewgallo:
Contributes to #1180 

Sets a default action set to use for the story titled `Narrow tearsheet with all header items`.

#### What did you change?
`TearsheetNarrow.stories.js`
Storybook


SimonFinney (approved):
LGTM — ta 💪

PR #1186, opened 2021-08-20
chore: improve storybook build

dcwarwick:
The storybook build has been slowing down with several recent changes, and is now painfully slow to reflect dynamic CSS changes at dev time and is intermittently timing out with the netlify preview builds.

This PR:
* enables persistent cache for Webpack, so that after the first build restarting/rebuilding storybook is much faster (this helps local systems, but makes no difference of course for things like netlify which are building from scratch every time);
* pre-builds the Carbon CSS with our package feature flag requirements set, so that this is then imported into the storybook index.scss as CSS rather than SCSS, which removes a chunk of the sass compilation -- this change on its own reduced the CSS dynamic refresh time by about 25%;
* re-applies the fast-sass-loader. We had to remove this to ensure that the Carbon styles all loaded after being initially disabled through the import-once mechanism, because the fast-sass-loader optimisation was only processing the file on first import and ignoring it afterwards, but now that we are pre-building the Carbon CSS this is no longer a problem so we can go back to fast-sass-loader. This makes a huge difference to both build/start speed and to dynamic update speed.

On my own system, a "cold" storybook build-and-start (`yarn storybook`) was taking about 5.5 minutes. The persistent cache reduced this to a few seconds for a restart or for a dynamic JS update, but still 3.5 minutes for a dynamic CSS update. The pre-built Carbon CSS reduced this to 2.5 minutes for a dynamic CSS update. After reinstating fast-sass-loader a "cold" build-and-start is just under 1 minute, a dynamic CSS update is <10 seconds, and a restart (with primed cache) takes about 20 seconds (about 10 seconds for the CSS pre-build, plus 10 seconds for a cache-enabled storybook start).

#### What did you change?

Updated storybook config, scss, and build scripts in package.json.

Ran innumerable storybook builds. This PR only affects the storybook.


matthewgallo (approved):
🎉 this will help us all out a lot!

Issue #1178, opened 2021-08-19
Add `ibm-cloud-cognitive` to Carbon Devtools


dcwarwick:
The Carbon devtools () are able to identify components in pages and provide useful information about them. This is not limited to Carbon components themselves, but can also handle components from other related packages (https://github.com/carbon-design-system/devtools/blob/base/packages/web-extension/src/globals/componentList.js), but this does not currently include the C&CS components.

Add the C&CS components to the Carbon devtools, as a useful functional aid for IBM product development teams.

As part of this, we need to figure out the best way to identify C&CS components live in a page. The Carbon, security and Cloud PAL components are identified just by a block class on the main element. Is that sufficient/satisfactory? If so, perhaps we should address the rather overdue question of our own package prefix, which is currently the rather anomalous "exp--", before adding our components to the devtools! Or should we use a different approach, such as some custom tag on the main element?

SimonFinney:
- [ ] Add `ibm-cloud-cognitive` package to [Carbon Devtools repo](https://github.com/carbon-design-system/devtools)
- [ ] Deliver stable identification mechanism for browser extensions
- [ ] Understand and align on maintenance strategy with [Devtools contributors](https://github.com/carbon-design-system/devtools/graphs/contributors)

PR #1187, opened 2021-08-23
feat(devtools): add draft concept

SimonFinney:
#### Affected issue

Contributes to #1178 

#### Proposed changes

- Add mechanism to support Devtools attributes and browser extension identification
- Apply to `AboutModal`

##### Outstanding

 - Apply mechanism to CLI and remaining components
 - Unit tests to support the mechanism
 - Finalize [PR to support package in Devtools](https://github.com/carbon-design-system/devtools/compare/base...SimonFinney:cloud-cognitive)
 - ...


packages/cloud-cognitive/src/internal/devtools.js
```diff
@@ -0,0 +1,22 @@
...
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { paramCase } from 'param-case';
+
+const devtoolsAttribute = 'data-carbon-devtools-id';
+
+function getDevtoolsId(componentName) {
+  return paramCase(componentName);
```
dcwarwick, line 13: I'm a fan of reusable packages, but adding param-case imposes this as a dependency for all users of the library. As we have a fairly simple case here -- we only need to convert our own component names (SomeKindOfThing or SomeKINDOfThing) -- we can perhaps eliminate the dependency at the cost of adding in a piece of slightly less general code. Something like:
```suggestion
  // convert to param-case (OneTwoTHREEFour -> one-two-three-four)
  componentName
    .replace(/([a-z])([A-Z])/g, '$1-$2') // OneTwoTHREE -> One-Two-THREE
    .replace(/([A-Z])([A-Z][a-z])/g, '$1-$2') // THREEFour -> THREE-Four
    .toLowerCase(); // One-Two-THREE-Four -> one-two-three-four
```

packages/cloud-cognitive/src/internal/devtools.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright IBM Corp. 2021, 2021
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { paramCase } from 'param-case';
+
+const devtoolsAttribute = 'data-carbon-devtools-id';
```
dcwarwick, line 10: A thought: 'data-carbon-devtools-id' sounds fairly generic -- it doesn't specifically say C&CS, for example. So other libraries might conceivably adopt the same approach, and might wish to use the same attribute name.So perhaps we should include a prefix in the devtoolsId values we compute, i.e. xxx-about-modal rather than just about-modal. That would ensure that where libraries have components with the same name we can tell which is which. The 'xxx' could be our prefix, eg 'cip' or whatever we choose.

packages/cloud-cognitive/src/components/AboutModal/AboutModal.js
```diff
@@ -12,8 +12,10 @@ import { useResizeDetector } from 'react-resize-detector';
 // Other standard imports.
 import PropTypes from 'prop-types';
 import cx from 'classnames';
+
 import { pkg } from '../../settings';
 import uuidv4 from '../../global/js/utils/uuidv4';
+import { getDevtoolsProps, getDevtoolsId } from '../../internal/devtools';
```
dcwarwick, line 18: Our common JS helpers have up to now gone into `../../global/js/utils`. `../../internal` matches what Carbon do in their `react` package, so we could certainly start reorganising where our shared/common functions go, but I wasn't sure whether you meant to be suggesting that.

dcwarwick:
I like this, and I think I do like using an attribute as a slightly tighter mechanism than just relying on a block-level class name to appear in the DOM. I've made a couple of comments/questions.

A few comments also on https://github.com/carbon-design-system/devtools/compare/base...SimonFinney:cloud-cognitive, since github doesn't let me comment directly on a compare:
- I really like the generating the set of components from our exports rather than adding yet another long maintenance-risky list.
- This is a minor point, more philosophical than anything, but in lines 78 and 82 (and 84) you're using `name` which is the export name of the component in our package. Now, the export names are the same as our component names (maybe we should add a test to ensure that is always true), but what we are actually using to generate the devtoolsId values is the component name, rather than the export name, so I'd suggest that in lines 78 and 82 we use `Component.displayName`. This actually also means that we could just do `cloudCognitiveComponents.reduce(...)` rather than `Object.entries(cloudCognitiveComponents).reduce(...)`, since we don't actually need the export names inside the reduce function.

packages/cloud-cognitive/src/components/AboutModal/AboutModal.js
SimonFinney: Good point! I'll revert to `global/js/utils` and if we decide to reorganize, that can happen independently of this.

packages/cloud-cognitive/src/internal/devtools.js
SimonFinney: I agree — I'm starting to wonder whether it's even necessary to change the case here, or if a `prefix` + `componentName` or `displayName` approach is sufficient.

Issue #1177, opened 2021-08-19
[SidePanel]: Add tooltip for toolbar buttons & add className and/or id to buttons


lily-peng:
## What will this achieve?

The SOAR team suggests the following enhancements to the SidePanel component to facilitate our designs:

1. Add the ability for tooltips on hover to the buttons in the SidePanel toolbar
2. Add the field(s) `className` and/or `id` to the toolbar buttons

## Additional information

The image below shows our design using the SidePanel. We'd like a tooltip to display upon hovering a toolbar button, and the ability to set a `className` and/or `id` to each button can achieve the red background

![sidepanel-tooltip-bg](https://user-images.githubusercontent.com/3118961/130112017-8af51411-f3ca-4236-825f-f98962ef9971.png)

matthewgallo:
This is now complete @lily-peng, just a heads up that the click handler for the actionToolbarButtons has been renamed to simply `onClick`.

PR #1189, opened 2021-08-23
feat(SidePanel): allow actionToolbarButtons to be a bit more flexible

matthewgallo:
Contributes to #1177 

Adds some flexibility to the action toolbar buttons in the side panel and properly displays the tooltip via `hasIconOnly` prop on buttons.

#### What did you change?
`SidePanel.js`
`SidePanel.test.js`
`SidePanel.stories.js`
Storybook and updated tests

### DIFF
--- a/packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
+++ b/packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
@@ -525,7 +525,13 @@ export let SidePanel = React.forwardRef(
                       action.leading,
                   },
                 ])}
-                onClick={() => action.onClick()}>
+                onClick={(event) =>
+                  action.onClick
+                    ? action.onClick(event)
+                    : action.onActionToolbarButtonClick
+                    ? action.onActionToolbarButtonClick(event)
+                    : null
+                }>
                 {action.leading && action.label}
               </Button>
             ))}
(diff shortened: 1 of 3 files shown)


packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
```diff
@@ -505,23 +507,25 @@ export let SidePanel = React.forwardRef(
...
                     [`${blockClass}__action-toolbar-icon-only-button`]:
-                      action.icon,
+                      action.icon && !action.leading,
                     [`${blockClass}__action-toolbar-leading-button`]:
-                      !action.icon,
+                      action.leading,
                   },
                 ])}
-                onClick={() => action.onActionToolbarButtonClick()}>
+                onClick={() => action.onClick()}>
```
SimonFinney, line 528: This makes sense for both consistency and predictability — will it break existing consumers?

One other thought, and I likely need a history lesson on this, but is there a rationale for preventing the event parameter to be passed to the callback?

SimonFinney (commented):
Thanks @matthewgallo — this makes absolute sense to me but I wanted to clarify something first. 🙏

packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
matthewgallo: Updated @SimonFinney

matthewgallo: Screen Shot 2021-08-23 at 6 07 40 PM

Used the `deprecateProp` helper that will give a console warning if `onActionToolbarButtonClick` is used.

SimonFinney (approved):
LGTM — thank you! 💯

PR #1193, opened 2021-08-24
fix(SidePanel): fix animating state of panel

matthewgallo:
I have noticed a visual defect when looking at the static title side panel stories (which this also includes the CreateSidePanel as it uses this variation of the SidePanel). The title, subtitle, and if provided action toolbar buttons jump into place after the animation of the side panel first opening. They should remain in the same place the entire time which this PR fixes.

For example, see this [story](https://ibm-cloud-cognitive.netlify.app/?path=/story/cloud-cognitive-components-side-panel-sidepanel--with-static-title-and-action-toolbar) to view the problem.

#### What did you change?
`SidePanel.js`
`_side-panel.scss`
Storybook

### DIFF
--- a/packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
+++ b/packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
@@ -12,6 +12,7 @@ import React, { useState, useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import { useResizeDetector } from 'react-resize-detector';
+import { moderate02 } from '@carbon/motion';
 import wrapFocus from '../../global/js/utils/wrapFocus';
 import { pkg } from '../../settings';
 import { allPropTypes } from '../../global/js/utils/props-helper';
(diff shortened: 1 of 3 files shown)


packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
```diff
@@ -583,11 +585,11 @@ export let SidePanel = React.forwardRef(
...
                   ? placement === 'right'
-                    ? 'sidePanelEntranceRight 250ms'
-                    : 'sidePanelEntranceLeft 250ms'
+                    ? `sidePanelEntranceRight ${moderate02}`
+                    : `sidePanelEntranceLeft ${moderate02}`
                   : placement === 'right'
-                  ? 'sidePanelExitRight 250ms'
-                  : 'sidePanelExitLeft 250ms'
+                  ? `sidePanelExitRight ${moderate02}`
+                  : `sidePanelExitLeft ${moderate02}`
```
dcwarwick, line 592: Should the animations 'sidePanelOverlayEntrance 250ms' and 'sidePanelOverlayExit 250ms' (forty lines further down) also be changed to use ${moderate02}?

dcwarwick (changes requested):

dcwarwick:
@matthewgallo There is a pronounced visual jump that occurs when the side panel is open as you mouse-over the icon buttons. This didn't happen previously. There is also a brief jump that occurs when you close the side panel, and I think that that *did* happen previously.

packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
matthewgallo: Yes, they should. Thanks for catching that.

matthewgallo:
I had noticed this but thought it got fixed. I've made this PR a draft while I take another look at it

matthewgallo:
For some reason, the `onAnimationStart` synthetic event was being called when you hovered on the action toolbar buttons. So I added a conditional check to make sure that the `event.target` is actually the panel container element. This fixed the jumpy behavior that was happening before @dcwarwick

packages/cloud-cognitive/src/components/SidePanel/SidePanel.js
matthewgallo: Updated

Issue #1195, opened 2021-08-24
CreateTearsheet: Add tests for initialStep prop


matthewgallo:
## What will this achieve?

A test should be added to check the new prop, `initialStep`, that has been added recently.

## How will success be measured?

When the custom hook that contains this logic receives test coverage.

![image (78)](https://user-images.githubusercontent.com/10215203/130669138-cbb120c2-a9d9-4df8-a07f-66df0a175489.png)

PR #1196, opened 2021-08-24
chore(CreateTearsheet): add tests to cover custom hooks

matthewgallo:
Contributes to #1195 

Adds tests for the `initialStep` prop and the `onMount` prop that can be optionally called on a per step basis.

#### What did you change?
`CreateTearsheet.test.js`
`useCreateComponentFocus.js`
Ran tests

### DIFF
--- a/packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.js
+++ b/packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.js
@@ -32,7 +32,7 @@ import {
 } from '../../global/js/hooks';
 import { hasValidChildType } from '../../global/js/utils/hasValidChildType';
 
-const componentName = 'CreateTearsheet';
+export const componentName = 'CreateTearsheet';
 const blockClass = `${pkg.prefix}--tearsheet-create`;
 
 export let CreateTearsheet = forwardRef(
--- a/packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.test.js
+++ b/packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.test.js
@@ -11,7 +11,7 @@ import { act, render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { moderate02 } from '@carbon/motion';
 import { pkg, carbon } from '../../settings';
-import { CreateTearsheet } from './CreateTearsheet';
+import { CreateTearsheet, componentName } from './CreateTearsheet';
 import { CreateTearsheetStep } from './CreateTearsheetStep';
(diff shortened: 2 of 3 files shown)


packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.js
```diff
@@ -32,7 +32,7 @@ import {
 } from '../../global/js/hooks';
 import { hasValidChildType } from '../../global/js/utils/hasValidChildType';
 
-const componentName = 'CreateTearsheet';
+export const componentName = 'CreateTearsheet';
```
dcwarwick, line 35: Nothing wrong with this per se, but we don't actually need to do this as the component name is attached to the component too.

```suggestion
```

packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.test.js
```diff
@@ -11,7 +11,7 @@ import { act, render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { moderate02 } from '@carbon/motion';
 import { pkg, carbon } from '../../settings';
-import { CreateTearsheet } from './CreateTearsheet';
+import { CreateTearsheet, componentName } from './CreateTearsheet';
```
dcwarwick, line 14:
```suggestion
import { CreateTearsheet } from './CreateTearsheet';
```

packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.test.js
```diff
@@ -153,6 +155,45 @@ describe(CreateTearsheet.displayName, () => {
...
+    const createTearsheetSteps = container.querySelector(
+      `.${createTearsheetBlockClass}__content .${carbon.prefix}--form`
+    ).children;
+    expect(
+      createTearsheetSteps[0].classList.contains(
+        `.${createTearsheetBlockClass}__step--visible-section`
+      )
+    );
+    expect(warn).toBeCalledWith(
+      `${componentName}: An invalid \`initialStep\` prop was supplied. The \`initialStep\` prop should be a number that is greater than 0 or less than or equal to the number of steps your ${componentName} has.`
```
dcwarwick, line 190: The component name is attached to the component render function as `displayName`, so we can use that to build the expected warning message.

```suggestion
      `${CreateTearsheet.displayName}: An invalid \`initialStep\` prop was supplied. The \`initialStep\` prop should be a number that is greater than 0 or less than or equal to the number of steps your ${CreateTearsheet.displayName} has.`
```

dcwarwick (changes requested):

matthewgallo:
Ah ok, I see! I'll update to use the display name

PR #1198, opened 2021-08-24
fix: better warnings when deprecated-usage props have invalid values

### LINKED ISSUE #1197: Confusing/incorrect warning messages produced by prop-types when a prop has an alternative form that is deprecated
If a prop has an alternative form that has been deprecated, like prop `pageActions` on `PageHeader`, we are coding the prop-types like this:
```
pageActions: PropTypes.oneOfType([
    /* new type */
    ,
    deprecatePropUsage(
      /* old type */,
      'Expects an array of objects with the following properties: label and onClick.'
    ),
  ]),
```

However, if the prop-types definition for the new type *fails*, then the `oneOfType` processing tries validating the supplied value with the old type, and the warnings that are produced are not appropriate and not helpful.

For example, if the value of `pageActions` is set to something like this:
```
            pageActions={[
              {
                key: 'primary',
                kind: 'primary',
                label: 'Wooo',
                onClick: () => {},
                renderIcon: Add16,
              },
            ]}>
```
then this will cause a prop-types validation warning because `renderIcon` has been supplied but `iconDescription` has not also been supplied, and `iconDescription` is required whenever `renderIcon` is supplied according to the Button prop-types. However, currently the following rather bewildering set of warnings is produced:
image
This happens because the validation warning emitted by the Button prop-types is discarded by the `oneOfType` processing, which simply continues to the second possible type in the list. That rather vague "invalid prop supplied" message is what `oneOfType` produces when all the props fail to validate. The deprecation warning is issued whenever the first type fails validation, even if the actual type wouldn't have matched the old pre-deprecation type either. This is not ideal.

If the `oneOfType` is temporarily removed, so that the prop-types looks like the following:
```
  pageActions: PropTypes.arrayOf(
    PropTypes.shape({
      ...Button.propTypes,
      key: PropTypes.string.isRequired,
      kind: Button.propTypes.kind,
      label: PropTypes.node,
      onClick: PropTypes.func,
    })
  ),
```
then a much more useful result is obtained:
image

We should find a way to provide a deprecated alternative prop-type that has the following characteristics:
* The deprecation warning should only be issued if the value supplied doesn't match the new type but WOULD have matched the old/deprecated type.
* If the old/deprecated type does not match, then the new type should be treated as the principal validator and any warnings it issues should be passed on to the user and not replace with an unhelpful "invalid prop supplied" message.

dcwarwick:
Closes #1197

Change the way `deprecatePropUsage` works so that the deprecated usage warning is only emitted if the supplied value matches the old type, and otherwise the new type is used to validate the value, with all warnings being emitted, because they are helpful in figuring out why a value isn't valid.

#### What did you change?

`deprecatePropUsage` in props-helper.js

Added tests, and ran all tests, and ran storybook.

### DIFF
--- a/packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
+++ b/packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
@@ -879,7 +879,7 @@ PageHeader.propTypes = {
    * Each item is specified as an object with the properties of a Carbon Button in icon only form.
    * Button kind, size, tooltipPosition, tooltipAlignment and type are ignored.
    */
-  actionBarItems: PropTypes.oneOfType([
+  actionBarItems: deprecatePropUsage(
     PropTypes.arrayOf(
       PropTypes.shape({
         ...prepareProps(Button.propTypes, [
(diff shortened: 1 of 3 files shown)


matthewgallo (approved):
Looks to be working great! I tested this with the same scenario that Dishant ran into earlier and got a much more helpful warning

Screen Shot 2021-08-24 at 10 35 24 PM

PR #1225, opened 2021-08-31
feat: page header props to influence grid width

lee-chase:
#### What did you change?

The PageHeader is hosted in a grid but was missing properties influencing the width of the grid "narrow" and "fullWidth". This PR adds the props "narrowGrid" and "fullWidthGrid".

Storybook and additional tests for grid width classes.

### DIFF
--- a/packages/cloud-cognitive/src/components/PageHeader/PageHeader.test.js
+++ b/packages/cloud-cognitive/src/components/PageHeader/PageHeader.test.js
@@ -725,4 +725,15 @@ describe('PageHeader', () => {
 
     warn.mockRestore(); // Remove mock
   });
+
+  test.only('PageHeader grid settings narrow and fullWidth', () => {
+    const dataTestId = uuidv4();
+    const { container } = render(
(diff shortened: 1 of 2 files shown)


packages/cloud-cognitive/src/components/PageHeader/PageHeader.test.js
```diff
@@ -725,4 +725,15 @@ describe('PageHeader', () => {
 
     warn.mockRestore(); // Remove mock
   });
+
+  test.only('PageHeader grid settings narrow and fullWidth', () => {
```
matthewgallo, line 729: I'm not sure we want to keep the `.only` here. This will cause `PageHeader.test.js` to [only run this one test](https://jestjs.io/docs/api#testonlyname-fn-timeout), right?

lee-chase: Ooops thanks @matthewgallo

matthewgallo (approved):

Issue #1230, opened 2021-09-01
Create Full Page - design exploration


sydrosa:
# Name of component or pattern
CreateFullPage

## Brief description
Toggle enhancement - exploring the UI and possibilities once the toggle is turned 'on' in `CreateFullPage` component.

## Design Maintainer(s)

- @andrea-gm

PR #1231, opened 2021-09-01
Create Full Page - Prototype

sydrosa:
Contributes to #1230


matthewgallo:
Closing since this work is outdated now, we can revisit if/when this design is requested

PR #1249, opened 2021-09-09
fix(CreateFullPage): explicitly set bg color for page content

matthewgallo:
In working on adding theme switching capabilities to the code sandboxes, I noticed that the CreateFullPage was not setting the background color of the main content area so when I changed themes the background remained white.

#### What did you change?
`_create-full-page.scss`
Update snapshot
Storybook

_Unexpected background color behavior in code sandbox_
Screen Shot 2021-09-09 at 1 23 22 PM


SimonFinney (approved):
LGTM — thanks 💯

Issue #1250, opened 2021-09-11
Specify sideEffects in package.json required for correct tree shaking with Webpack


szaflarskim:
## What will this achieve?
Correct tree shaking when importing components as `import { SidePanel } from '@carbon/ibm-cloud-cognitive'`.

Currently tree shaking does not work with Webpack because sideEffects are not defined in package.json and import of a component bundles carbon-components-react. 
Workaround is to import each component in a form `import { SidePanel } from '@carbon/ibm-cloud-cognitive/es/components/SidePanel';`

szaflarskim:
Verified correct tree shaking with `@carbon/ibm-cloud-cognitive@0.85.1`.

dcwarwick:
Excellent! Thanks @szaflarskim

PR #1251, opened 2021-09-11
fix: added sideEffects field to package.json

szaflarskim:
Contributes to #1250

With `sideEffects` specified in package.json of dependency, Webpack can easily determine which code can be excluded from the output. See https://webpack.js.org/guides/tree-shaking/

#### What did you change?

Added sideEffects field to package.json, so that tree shaking works correctly with Webpack. 
I looked through the code, and as far as I can see only CSS and SCSS files have sideEffect.

1. I've build packages 
2. Imported local version of 'ibm-cloud-cognitive' to project using Webpack.
3. Imported SidePanel with `import { SidePanel } from '@carbon/ibm-cloud-cognitive'`
2. Build project in `production` mode to enable tree shaking with enabled BundleAnalyzer plugin. 
3. Created bundle did not bundle unnecessary code from carbon-components-react.


dcwarwick (changes requested):

dcwarwick:
@szaflarskim Thank you for looking into this and proposing this update. There is one adjustment I think is needed, and I've requested it as a change -- could you take a look, and if you think it looks right (I'm not a webpack expert!) then accept the change and I'll be happy to approve and merge this. Thanks!

szaflarskim:
Thank you @dcwarwick . I must have missed that file.

dcwarwick (approved):

PR #1273, opened 2021-09-20
feat: ensure prettier linting causes ci-check to fail

dcwarwick:
The ci-check script invokes prettier on all the source files in the repo, but includes the `--write` option which means that (a) the files are modified in situ and then appear as uncommitted and (b) the linter task doesn't return an error code so the ci-check appears to pass. This means that if there are files that fail prettier linting they appear to pass ci-check but then cause the release task to fail by generating uncommitted changed files. Normally, our commit check will prevent files that fail prettier linting from being committed, but if the prettier rules themselves change then this can cause files that previously passed to now fail even though they haven't been changed and are therefore not checked at commit time. This has recently happened, so releases are now failing.

#### What did you change?

Updated the package.json prettier scripts so that the `--write` option is only included when carrying out a prettier format and NOT when carrying out a prettier lint. Also switched to the `--check` option rather than `--list-different`, just because this gives more useful output (it otherwise does the same thing).

Also applied the updated prettier formatting using `yarn format`.

Created a PR containing just the package.json script updates, and reviewed the ci-check output of the PR to see that the wrongly formatted files now caused ci-check to fail.

Then created this PR which includes the package.json script updates AND updates source files to the new prettier format, and reviewed the ci-check output of the PR to see that ci-check now passes.

This PR should be merged as a "feat:" so as to roll up missed fix and feat PRs whose release tasks failed. Its release task ought to complete ok, but this will need to be checked manually.


matthewgallo (approved):
Thank you @dcwarwick for digging into this!

Issue #1234, opened 2021-09-02
Web terminal: Add create new deployment button


AlexanderMelox:
## What will this achieve?
This will add the `create new deployment` actions on the Web terminal header bar.

![Screen Shot 2021-09-02 at 10 49 04 AM](https://user-images.githubusercontent.com/12755042/131865619-00a97b62-1579-49d9-b7af-30aa075dbd35.png)
*The code icon is to create a new deployment*

## How will success be measured?
The component should be able to render a create new deployment action button and expose a onClick handler when the action is clicked

## Additional information

- Designs
https://pages.github.ibm.com/cdai-design/pal/patterns/web-terminal
- Existing code
https://ibm-cloud-cognitive.netlify.app/?path=/story/cloud-cognitive-canary-webterminal--default

AlexanderMelox:
Added in #1279

PR #1279, opened 2021-09-21
feat: adds actions prop to web terminal

AlexanderMelox:
Contributes to #1234 

This adds the ability for the user to pass an array of actions to be added to the web terminal header bar. 

#### What did you change?

1. Added the actions property, this allows the user to pass an array of objects with the desired `renderIcon`, `onClick`, and `iconDescription` for each action.  

Screen Shot 2021-09-21 at 9 54 47 AM

2. Talked to the designer about utilizing carbons Icon Buttons instead for the header bar, so some of the components got swapped with those. The documentation dropdown button and the close button are now an icon button.

3. Updated the block class variable in the JSX and CSS.
4. Removed unused CSS styles from the stylesheets.

Checked to see if the user can successfully pass actions into the web terminal.

### DIFF
--- a/packages/cloud-cognitive/src/components/WebTerminal/WebTerminal.js
+++ b/packages/cloud-cognitive/src/components/WebTerminal/WebTerminal.js
@@ -15,14 +15,23 @@ import { pkg } from '../../settings';
 
 // Carbon and package components we use.
 import { Close16 as Close, Help16 as Help } from '@carbon/icons-react';
+import { Button } from 'carbon-components-react';
 
 // The block part of our conventional BEM class names (blockClass__E--M).
 const componentName = 'WebTerminal';
-const blockClass = `${pkg.prefix}-web-terminal`;
+const blockClass = `${pkg.prefix}--web-terminal`;
 
 export let WebTerminal = React.forwardRef(
   (
-    { children, className, closeTerminal, documentationLinks, open, ...rest },
+    {
+      children,
+      className,
+      closeTerminal,
+      documentationLinks,
+      open,
+      actions = [],
+      ...rest
+    },
     ref
   ) => {
     const [shouldRender, setRender] = useState(open);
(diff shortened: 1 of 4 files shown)


packages/cloud-cognitive/src/components/WebTerminal/WebTerminal.js
```diff
@@ -79,10 +88,12 @@ export let WebTerminal = React.forwardRef(
         <header className={`${blockClass}__bar`}>
           <div className={`${blockClass}__actions`}>
             {showDocumentationLinks && (
-              <button
+              <Button
+                hasIconOnly
+                kind="ghost"
                 type="button"
+                renderIcon={Help}
```
matthewgallo, line 95: This is going to need an `iconDescription` since it uses `hasIconOnly`

packages/cloud-cognitive/src/components/WebTerminal/WebTerminal.stories.js
```diff
@@ -51,13 +53,30 @@ const Template = (args) => {
...
 export const WithDocumentationLinks = prepareStory(Template, {
   args: { documentationLinks },
 });
 
+export const WithActions = prepareStory(Template, {
+  args: {
+    actions: [
+      {
+        renderIcon: Code,
+        onClick: () => alert('clicked on action'),
```
matthewgallo, line 68: For consistency, we should use storybook action here

```jsx
import { action } from '@storybook/addon-actions';
...
...
actions: [
      {
        renderIcon: Code,
        onClick: action('Clicked action')
       }
],
...
...

```

AlexanderMelox: Weird, I didn't realize you don't pass it as a callback. Thanks for pointing this out

AlexanderMelox:
I am stuck [on the lint check](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1279/checks?check_run_id=3735640107), I ran the formatter and nothing changed, why am I getting this linting error?

dcwarwick:
@AlexanderMelox There was a change in the recent version of `prettier`, to do with where it places the close-angle-bracket in complex tags in JSX. I suspect you were still running the older version, which formatted the files the old way, but the ci-check is running the newer version, which expects them to be formatted the new way! If you run `yarn` to bring your local packages all up-to-date, it should sort the problem out.

I've run a `yarn format` and added the file changes to your branch -- the lint check now passes.

I saw that you also had a few of the WebTerminal tests failing. Three of the failures were caused by the fact you've changed the `blockClass` value in the WebTerminal.js -- which is great -- but you just needed to make the same change in the tests to ensure that they are looking for the correct class names. The fourth failure was caused by the fact that while updating the 'Close' button you dropped the class name from it, but one of the tests was looking for the 'Close' button via that class name: I changed the test to look for the 'Close' button by label text instead, which is a better way to test anyway. I've also added those fixes to the tests to your branch -- the tests now all pass.

If you need to make further changes in your branch, make sure you do a `git pull` to grab those changes first.

I hope this helps :-)

PR #1299, opened 2021-09-29
Silence warning in prop-types

nertzy:
PropTypes.oneOf takes an array, not a string.


dcwarwick:
Good catch. Thanks Grant.

dcwarwick (approved):

PR #1303, opened 2021-09-30
fix(CreateSidePanel): scoping issue for create side panel

matthewgallo:
I noticed that the `_create-side-panel.scss` sets a left value on the actions, this is causing layout issues because of a scoping issue.

_Before_
Screen Shot 2021-09-30 at 11 46 22 AM
_After_
Screen Shot 2021-09-30 at 11 50 11 AM
#### What did you change?
`_create-side-panel.scss`
Storybook


SimonFinney (approved):
LGTM — thank you! 💯

Issue #1335, opened 2021-10-14
add button icon options for cards


davidmenendez:
the primary and secondary buttons in cards should support an icon option

PR #1336, opened 2021-10-14
fix: added button icon props for cards

davidmenendez:
Contributes to #1335 and #1338

also adds href support to the buttons


davidmenendez:
thanks @dcwarwick. I thought about adding tests, but i'm of the opinion that if you're only adding support for already existing functionality of a carbon component that you don't really need to add new tests because I think it's safe to assume that carbon already tests that their components support their own props. for example- in this case of adding support for href and icons to the carbon button i think it's safe to assume carbon already tests that their buttons support href and icons.

but yes if that's a request you're making then i will open a follow up PR to handle it.

PR #1360, opened 2021-10-28
Fix(Tearsheet): avoid focus traps with stacked tearsheets

### LINKED ISSUE #1332: Focus Trap issue with Stacked Tearsheet
The focus is not coming on top most tearsheet input elements, it still remains back on the previous tearsheet, due to which i am not able to type anything in the textbox of 2nd(top most) tearsheet.
Once second tearsheet opens , the focus remains on previous tearsheet close icon.
Screenshot 2021-10-13 at 6 34 30 PM

 I have used  data-modal-primary-focus prop in teaxtInput of second tearsheet but once it is opened focus still remains on close icon, even after clicking on textInput , it is not focusing.

Sample code i am referring:

```


export const TearsheetExample = () => {
  const [isOpen, setIsOpen] = useState(true);
  const [isstackedOpen, setstackedIsOpen] = useState(false);
  const handleOpenModalClick = () => {
    setIsOpen(true);
  };
  const handleOpenstackedModalClick = () => {
    setstackedIsOpen(true);
  };
  const handleCloseModal = () => {
    setIsOpen(false);
  };
  const handlestackedCloseModal = () => {
    setstackedIsOpen(false);
  };

  return (
    <>
      <style>{`.exp-tearsheet { opacity: 0 };`}</style>
      <Button onClick={handleOpenModalClick}>Reopen Tearsheet</Button>
      <Tearsheet
        actions={[
          { kind: "secondary", label: "Cancel", onClick: handleCloseModal },
          { kind: "primary", label: "Create", onClick: handleCloseModal }
        ]}
        closeIconDescription="Close the tearsheet"
        description={
          <span>
            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
            eiusmod tempor <strong>incididunt ut labore</strong> et dolore magna
            aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco
            laboris nisi ut aliquip ex ea commodo consequat.
          </span>
        }
        label="This is the label of first tearsheet"
        onClose={handleCloseModal}
        open={isOpen}
        preventCloseOnClickOutside
        title="This is the title of the tearsheet"
      >
        <div className="tearsheet-example__dummy-content-block">
          <Button onClick={handleOpenstackedModalClick}
           onKeyDown={(e) =>   e.key === 'Enter' && handleOpenstackedModalClick}
          >
            open second Tearsheet
          </Button>
          <TextInput
            autoComplete="off"
            disabled={false}
            id="tagName"
            name="tagName"
            light={false}
            maxLength={100}
            size="md"
            type="text"
            helperText=""
            required={true}
            data-modal-primary-focus
          />
        </div>
      </Tearsheet>

      <Tearsheet
        actions={[
          {
            kind: "secondary",
            label: "Cancel",
            onClick: handlestackedCloseModal
          },
          { kind: "primary", label: "Create", onClick: handlestackedCloseModal }
        ]}
        closeIconDescription="Close the tearsheet"
        description={
          <span>
            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
            eiusmod tempor <strong>incididunt ut labore</strong> et dolore magna
            aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco
            laboris nisi ut aliquip ex ea commodo consequat.
          </span>
        }
        label="This is the label of the tearsheet"
        onClose={handlestackedCloseModal}
        open={isstackedOpen}
        preventCloseOnClickOutside
        title="This is the title of second tearsheet"
      >
        <div className="tearsheet-example__dummy-content-block">
          <TextInput
            autoComplete="off"
            disabled={false}
            id="tagName"
            name="tagName"
            light={false}
            maxLength={100}
            size="md"
            type="text"
            helperText=""
            required={true}
            data-modal-primary-focus
          />
          The second content of the Tearsheet should be placed here.
        </div>
      </Tearsheet>
    </>
  );
};

```

Screenshot 2021-10-13 at 6 28 38 PM

dcwarwick:
Closes #1332

This PR:
- adds tearsheets to the modal exit selector list so that stacking tearsheets do not "trap" the focus in the first tearsheet.
- catches focus entry into a non-topmost tearsheet and transfers focus to the topmost tearsheet.
- checks as each tearsheet becomes topmost that focus is within the tearsheet, and if not claims focus.

This should avoid the focus trap issue and also clean up the handling of focus when multiple tearsheets are being opened and closed.

#### What did you change?

Tearsheet.js, also added to the stacked tearsheet stories so the issue can be verified. Also updated the tests to expect warnings from Carbon -- these should eventually be removed (see https://github.com/carbon-design-system/carbon/issues/10000).

Ran storybook.

### DIFF
--- a/packages/cloud-cognitive/src/components/Tearsheet/Tearsheet.stories.js
+++ b/packages/cloud-cognitive/src/components/Tearsheet/Tearsheet.stories.js
@@ -241,7 +241,7 @@ const StackedTemplate = ({ actions, ...args }) => {
   return (
     <>
       <style>{`.${pkg.prefix}--tearsheet { opacity: 0 }`};</style>
-      <div
+      <ButtonSet
         style={{
           display: 'flex',
           position: 'fixed',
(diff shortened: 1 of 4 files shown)


dcwarwick:
@matthewgallo @lee-chase I looked at the approach to this that replaced the ComposedModal with plain `<div>`s except for the topmost tearsheet -- ingenious, but I couldn't fix the animation problem. Basically, when React switches the rendering from the ComposedModal render method to the fragment is rerenders each side of an animation frame, so the transitions will not trigger correctly. I did explore first replacing the modal with `<div>`s in the OLD state, then after a delay switching the classes to try to get the transitions to occur, but it was becoming very fiddly.

Fortunately I found this alternative, which uses a rather obscure prop `selectorsFloatingMenus`. So obscure, in fact, that I think no one else has ever used it, because it actually has several bugs! I've opened https://github.com/carbon-design-system/carbon/issues/10000 on Carbon to resolve those bugs. Fortunately, they only cause console warnings in non-production builds, so this is still a viable fix in the mean time. I've updated the tests to expect those warnings, and if/when Carbon is fixed we can remove those updates :-)

matthewgallo (approved):
Glad you found a work around for this! 🥇

PR #1373, opened 2021-11-04
chore: use sass for security package style snapshot test

dcwarwick:
This is a suggestion, which updates the style snapshot test to use 'sass' rather than node-sass. This has the advantage of using the same sass compiler used for builds, and removing the dependency on the now long-deprecated node-sass. Somewhat surprisingly, it even runs faster for me: a 46s run vs 74s for the node-sass version. Note that this test calls sass at the command line rather than through the API because of the huge performance penalty the sass API currently imposes on compiles that include a large number of @imports.

#### What did you change?

styles-test.js

Ran tests

### DIFF
--- a/packages/security/package.json
+++ b/packages/security/package.json
@@ -87,7 +87,6 @@
     "faker": "^5.5.3",
     "jest": "^27.3.1",
     "jest-config-ibm-cloud-cognitive": "^0.19.0",
-    "node-sass": "^6.0.0",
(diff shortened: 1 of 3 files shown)


SimonFinney:
Thanks for the reference @dcwarwick — this solution has been synced from `ibm-security` to `packages/security` in https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1409.

PR #1379, opened 2021-11-05
fix: improve tree-shaking optimisation by importing icons directly from @carbons/icons-react

asfordmatt:
as per https://github.com/carbon-design-system/carbon/issues/6602#issuecomment-729207760
we should be importing from the top level package '@carbon/icons-react' only, not via direct references to 'lib'


dcwarwick:
@asfordmatt good catch Matt -- this looks like it will help with optimising bundles. Thanks for opening this PR. I notice that ComboButton also uses the deprecated way to import icons, so I've added a fix for that too.

dcwarwick (approved):

PR #1389, opened 2021-11-08
feat(security): upgrade `carbon-components`, fix public API snapshot

SimonFinney:
#### Proposed changes

- Import package settings directly in `jest-config` to avoid side effects from `Canary`
- Align `security` Carbon version with rest of the repo
- Update `security` style snapshot


dcwarwick (approved):
Brilliant! Well tracked down! Yes, I can sort of see how this was causing that effect... I'm still bemused why it was the `carbon-components` version that worked around it, rather than the `carbon-components-react` version, but I guess just weirdness :-)

dbrugger (approved):

PR #1393, opened 2021-11-08
fix(security): reinstate minified CSS

dbrugger:
Contributes to

- index.min.css was missing for the security package

#### What did you change?

- added `build:css-min`

- local build


lee-chase (commented):
@SimonFinney @dbrugger This or #1390?

dbrugger:
@lee-chase i am open for either variant. this one would
- not any further (dev) dependency
-  replace the current build step instead of adding a new one. the idea being that I expect consumers to either want 
  1) an optimized style build in which case they should use scss per component and compile together with their app styles
  2) to have an all-in-one prebuilt css, in which case only a minified version should be needed

what do you think?

dbrugger:
i did some further comparison.
| tool | output size |
| ---- | ----------- |
| cleancss -O2 | 857K |
| sass --style=compressed | 887K |
so cleancss seems to be more efficient

PR #1404, opened 2021-11-08
chore: update gallery config

dcwarwick:
This PR checks in an updated gallery config which should stop the automated script dumping random files every time we try and do any other kind of commit...!

#### What did you change?

The files are all within the /examples/ tree and don't affect the main library at all.

I didn't. @lee-chase can test and verify these in the morning! This is an emergency fix to stop random files spewing out whenever we do a `git commit`.


lee-chase (approved):

lee-chase:
@dcwarwick these will stop the spewings (not random).

Essentially the script needs running any time the script or examples directory is updated.

Running it normally should be a no-op, but it would be good to add a test to detect new examples without config.

PR #1411, opened 2021-11-10
feat: add options-tile component

janhassel:
Adds the implementation of the [OptionsTile component](https://pages.github.ibm.com/cdai-design/pal/components/options-tile/usage)

#### What did you change?
- Added `<OptionsTile />` component, styles, stories and tests

- Manual testing with storybook
- Automated testing


lee-chase (changes requested):
Can we make the toggle toggle as well as report an action?

janhassel:
@lee-chase Just as a demo in the storybook or would you suggest to have the component control its state?

lee-chase:
> @lee-chase Just as a demo in the storybook or would you suggest to have the component control its state?

Have the story react to the change of state. Something like

```
const TemplateStatic = ({enabled, ...rest}) => {
  const [liveEnabled, setaLiveEnabled] = useState(enabled);
  const toggleAction = action('onToggle');
  const handleToggle = (ev) => {
    toggleActionS(ev);
    setLiveEnabled(ev.target.value);
  };
  return <OptionsTile onToggle={handleToggle} {...rest} enabled={liveEnabled}/>;
};
```

janhassel:
@lee-chase 👍  just pushed the update

lee-chase:
> @lee-chase 👍 just pushed the update

@janhassel  the story action call returns a function to be called by the event handler. This is why currently no output is logged.

lee-chase (approved):

Issue #1385, opened 2021-11-08
Remove deprecated props, behaviors and components


dcwarwick:
The v1 release is an opportunity to remove deprecated features where appropriate.

Here is a list of what we should remove:
 - in ActionBar:
   - use of `children` instead of `actions` (#1412).
 - in BreadcrumbWithOverflow:
   - use of `children` instead of `breadcrumbs` (#1421).
 - entire component ContextHeader (#1417).
 - in CreateSidePanel:
   - use of `pageContentSelector` instead of `selectorPageContent` (#1406).
 - in PageHeader:
   - use of `actionBarItems` with an element or an array of elements instead of an array of shape (TODO).
   - use of `actionBarOverflowLabel` instead of `actionBarOverflowAriaLabel` (TODO).
   - use of `availableSpace` instead of `children` (TODO).
   - use of `background` instead of `hasBackgroundAlways` (TODO).
   - use of `breadcrumbItems` instead of `breadcrumbs` (TODO).
   - use of `breadcrumbOverflowLabel` instead of `breadcrumbOverflowAriaLabel` (TODO).
   - use of `collapseHeaderLabel` instead of `collapseHeaderIconDescription` (TODO).
   - use of `collapseHeaderToggleWanted` instead of `hasCollapseHeaderToggle` (TODO).
   - use of `expandHeaderLabel` instead of `expandHeaderIconDescription` (TODO).
   - use of `pageHeaderOffset`, which is no longer required (TODO).
   - use of `preCollapseTitleRow` instead of `collapseTitle` (TODO).
   - use of `preventBreadcrumbScroll` instead of `disableBreadcrumbScroll` (TODO).
 - in SidePanel:
   - use of `pageContentSelector` instead of `selectorPageContent` (#1406)
   - use of `onActionToolbarButtonClick` in `actionToolbarButtons` elements instead of `onClick` (https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1414).
 - in Tearsheet, TearsheetNarrow and TearsheetShell:
   - use of `preventCloseOnClickOutside`, which is no longer required (#1416).

dcwarwick:
Fixed by merge of above PRs for v1 release

PR #1417, opened 2021-11-10
chore(ContextHeader): remove non-released component

dcwarwick:
Contributes to #1385

ContextHeader has not been released, and probably won't be because it's not a PAL pattern now. The component has been added to CDAI to allow production use of it where needed, so it's no longer required in this package.

This PR doesn't affect any released components, so can be reviewed and merged without waiting for v1.


matthewgallo (approved):

dcwarwick:
Although we don't need to wait for v1 for this, since v1 is imminent maybe hold off the merge until we build v1 just to be on the safe side.

PR #1418, opened 2021-11-10
fix: page header background over lays

lee-chase:
#### What did you change?

- Previously the page header had a background that appeared (opacity: 0 to $ui-01). This did not work when the background was turned off (on is the default). Changed this to transition between $ui-background and $ui-01.
- The Breadcrumb was only 60% of 60% of 100% width when shown with a pageAction, corrected to 60% of 100%;
- The breadcrumb was not quite wide enough to cover the row background (excluding pageActions, without actionBar) due to grid margins. Corrected.
- The pageActions when stuck to top (no action bar) did not cover the background. Updated to do so.

Storybook, lint and tests


dcwarwick (approved):
LGTM

Issue #1311, opened 2021-10-06
PageHeader action button not getting clicked on scroll


Nishi-Chandra1:
I am trying to use <PageHeader> component but on scroll when pageActions button is getting stick on top, its not getting clicked. But without scroll it’s clickable.
The problem can also be seen in  main story book :

https://ibm-cloud-cognitive.netlify.app/?path=/story/cloud-cognitive-components-page-header-pageheader--with-button
<img width="978" alt="Screenshot 2021-10-06 at 11 30 33 AM" src="https://user-images.githubusercontent.com/80027651/136220880-e4dfc5a2-93f9-4c22-a37c-bb2beaf510a6.png">
s

What is happening is that as the header starts to scroll up, the breadcrumb bar is overlaying the page actions. If you scroll just part-way up, until the title is half-way through transitioning into the breadcrumbs, you’ll see that the bottom half of the page action is still clickable but the top half is obscured by the (transparent) breadcrumb bar and can’t be clicked.

PR #1428, opened 2021-11-12
fix(PageHeader): action bar location

lee-chase:
Fix issue introduced by #1311 

#### What did you change?

Corrected CSS selector for the breadcrumb to ensure it did not use too much space, pushing the action bar items off to the right.

Storybook, updated style snapshot.


dcwarwick (approved):
LGTM. @lee-chase i initially made this draft when I saw the v1 label, but actually there is no reason to hold this back so I'll set it to merge right away.

Issue #1426, opened 2021-11-12
Clean the input field when close the RemoveModal


ninjasun:
## What package(s) are you using?

- [ x] Cloud & Cognitive 0.97
- [ ] Legacy/CDAI
- [ ] Legacy/Security

How to access the input?
In case of cancel (but even on submit) the input field stays populated and looks like there is not a prop I can set.

https://codesandbox.io/s/remove-modal-3wcmu?file=/src/components/TestComponent.js

Thanks for support

matthewgallo:
https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1433

ninjasun:
@matthewgallo this is working, however, will be possible to clean the input field after the modal is closed?

Now is being instantly clean during the closing animation. Does not seem nice to me.
What is your opinion?

PR #1433, opened 2021-11-12
fix(RemoveModal): #1426 reset remove modal text input value

matthewgallo:
Contributes to #1426 

This clears the text input of the RemoveModal when it is closed. I also added a missing `value` attribute to the TextInput so that it works as expected.

#### What did you change?
`RemoveModal.js`
Storybook and ran tests


packages/cloud-cognitive/src/components/RemoveModal/RemoveModal.js
```diff
@@ -54,6 +56,16 @@ export let RemoveModal = forwardRef(
     const primaryButtonDisabled =
       textConfirmation && userInput !== resourceName;
     const blockClass = `${pkg.prefix}--remove-modal`;
+
+    // Clear the user input this way so that if the onRequestSubmit handler fails for some reason
+    // the value of the input will still remain, we only want to empty the input value
+    // until open is false.
```
dcwarwick, line 62: Just a tiny suggested commentary change -- it took me a moment to work out what was happening here :-)

```suggestion
    // the value of the input will still remain: we only want to empty the input value
    // when open actually changes to false.
```

dcwarwick (changes requested):

dcwarwick (approved):

PR #1434, opened 2021-11-12
fix: add side effect for props-helper to package.json

### LINKED ISSUE #1432: UserProfileImage fails to load in production build
UserProfileImage fails to load in production build: Using the UserProfileImage component in built app produces an error in the console (d.a.string.isRequired.if is not a function).

- [x] Cloud & Cognitive
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

Using the UserProfileImage component produces an error in the console (d.a.string.isRequired.if is not a function). This happened only after upgrading @carbon/ibm-cloud-cognitive from 0.65.0 to 0.98.0. Note that this only happens after the component is part of a built app (npm run build). It does not produce this error on local build (npm start). See images below. Code where issue occurs was first introduced in https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1042 but version 0.65.0 (which I believe has the code) does not have the issue. I tried a number of versions and I believe it began failing roughly after version 0.75.0.

UserProfileImage component.

> What version of the @carbon/ibm-cloud-cognitive package are you using?
@carbon/ibm-cloud-cognitive 0.98.0

> should be aware of?
Workflow Designer

1. Use @carbon/ibm-cloud-cognitive 0.98.0 in package file
2. Use the UserProfileImage component
3. Build the page (npm run build) and deploy it. Page will fail to load: (d.a.string.isRequired.if is not a function)

image
image

matthewgallo:
Closes #1432 

Add sideEffects to main package.json to avoid tree shaking issue as seen in #1432 

#### What did you change?
`package.json`

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     ],
     "nohoist": []
   },
+  "sideEffects": "packages/cloud-cognitive/src/global/js/utils/props-helper.js",
   "scripts": {
     "audit": "node scripts/audit.js dependencies moderate",
     "build": "run-s build:* storybook:build:storybook",
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
     ],
     "nohoist": []
   },
-  "sideEffects": "packages/cloud-cognitive/src/global/js/utils/props-helper.js",
   "scripts": {
     "audit": "node scripts/audit.js dependencies moderate",
     "build": "run-s build:* storybook:build:storybook",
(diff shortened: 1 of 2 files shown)


matthewgallo:
@dcwarwick I've updated the cloud-cognitive package.json for now like we discussed and will do some digging around to see other examples of how people have done this.

dcwarwick (approved):

Issue #1422, opened 2021-11-12
Update package name and descriptions to ibm-products from ibm-cloud(-and)-cognitive


dcwarwick:
- [x] Publish @carbon/ibm-cloud-cognitive as @carbon/ibm-products, with identical content and matching versions. Later, we can deprecate @carbon/ibm-cloud-cognitive, and eventually remove it when nobody is using it any more.
- [x] Change documentation and titles to refer to the library as Carbon for IBM Products and remove mention of Cloud & Cognitive etc.

davidmenendez:
@dcwarwick is this complete?

dcwarwick:
I've created ongoing issues to rename the repo and the package directory. Other than that, this is complete.

PR #1437, opened 2021-11-14
fix: replace "cloud & cognitive" with "Carbon for IBM Products" in docs

dcwarwick:
Contributes to #1422

Update various docs and similar, changing the branding from Cloud & Cognitive to Carbon for IBM Products. NB this PR does not attempt to do a complete job: references to the PAL can be updated when the PAL updates, and references to the library @carbon/ibm-cloud-cognitive can be updated once we are sure we are publishing @carbon/ibm-products in sync. This PR just updates mentions of the package in our docs, storybooks, and a few mentions in constants, ids, etc. This PR shouldn't change any actual use or functioning of the components.

Ran tests and storybook.


dcwarwick:
@andrea-gm FYI

packages/cloud-cognitive/src/components/Toolbar/_toolbar.scss
```diff
@@ -8,7 +8,7 @@
 @import 'carbon-components/scss/components/button/button';
 @import '../../global/styles/project-settings';
 
-@include exports($name: cloud-cognitive--toolbar) {
+@include exports('toolbar') {
```
SimonFinney, line 11: These styles will never be exported due to [naming conflicts with Carbon's toolbar](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/main/packages/cloud-cognitive/src/global/styles/_imported-carbon-modules.scss#L23).

```suggestion
@include exports('ibm-products--toolbar') {
```

SimonFinney (commented):
I have one concern but otherwise this LGTM 🎉

davidmenendez (dismissed):

packages/cloud-cognitive/src/components/Toolbar/_toolbar.scss
dcwarwick: Thanks, good catch

Issue #1439, opened 2021-11-15
Production build with Webpack failing because of incorrect sideEffects definition


szaflarskim:
## What package(s) are you using?

- [x] Cloud & Cognitive
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

Using version `0.98.0`
Production build with Webpack offers tree shaking for used modules based on `sideEffects` property of the imported module specified in package.json.
Build is silently failing now by dropping required code, because `"src/global/js/utils/props-helper.js",` specified is not included in published package. This created runtime error when package is to be used.
props-helper.js is found under `lib/global/js/utils/props-helper.js` and `es/global/js/utils/props-helper.js`, so we can specify sideEffect for this file with glob pattern, like `"**/props-helper.js",` 

## Additional information
![image](https://user-images.githubusercontent.com/22325359/141803147-a48f826f-1072-4e05-80e8-57b2c2649f0d.png)

szaflarskim:
I missed the original issue. Closing as duplicate #1432

PR #1440, opened 2021-11-15
fix: sideEffect definition

szaflarskim:
Contributes to #1439 

Fixing side effect definition in package.json

#### What did you change?

Using glob pattern to specify sideEffects for `props-helper.js`

Run production build with the fixed sideEffect specification.


szaflarskim:
I missed the issue already reported. Closing as duplicate  https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1434

PR #1469, opened 2021-11-22
feat: add experimental `Popover` to `ToolbarButton`

SimonFinney:
Contributes to #817 — blocked pending [`@carbon/styles` migration](https://github.com/carbon-design-system/carbon/tree/main/packages/styles#usage).

Add [experimental `Popover`](https://carbon-react-next.netlify.app/?path=/story/experimental-unstable-popover--playground&args=align:top-left) to `ToolbarButton` caret variant logic, stories, styles, and tests.


packages/cloud-cognitive/src/components/Toolbar/ToolbarButton.js
```diff
@@ -6,52 +6,134 @@
  */
 
 import { Button } from 'carbon-components-react';
+
+import {
+  Popover,
+  PopoverContent,
+} from 'carbon-components-react/es/components/Popover';
```
dcwarwick, line 13: Is this the way unstable components from Carbon are meant to be used? Direct imports from /es/ or /lib/ can be problematic, because our users need both. An import from /es/ will probably work in Node environments which load ESM, eg our jest tests, but may not survive a WebPack bundling which loads the /lib/ transpiled versions...

SimonFinney: You're absolutely right, this won't play nice with CommonJS. When I started looking at this, `Popover` was unavailable from the entry point but luckily this has changed with https://github.com/carbon-design-system/carbon/pull/9965 — I'll update to use this instead.

If we come across this scenario again, there's the option of promoting https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1402 into `babel-preset-ibm-cloud-cognitive` — this ensures ES module imports don't make it into the CommonJS build and vice-versa.

davidmenendez:
@SimonFinney what's the status with this?

matthewgallo:
Closing for now as there is not currently a way to import the Popover styles in Carbon v10.

Issue #1486, opened 2021-12-01
Add href support to action icons in cards


davidmenendez:
`actionIcons` in both expressive and productive cards should support `href` in addition to `onClick`

PR #1487, opened 2021-12-01
feat: icon href support for cards

davidmenendez:
Contributes to #1486


packages/cloud-cognitive/src/components/Card/Card.js
```diff
@@ -90,9 +90,17 @@ export let Card = forwardRef(
                 size={actionsPlacement === 'top' ? 'sm' : 'field'}
                 iconDescription={iconDescription}
                 kind="ghost"
+                href={href}
               />
             );
           }
+          if (href) {
+            return (
+              <a key={id} className={`${blockClass}__icon`} href={href}>
```
dcwarwick, line 99: Should we include an `onClick` here if they happen to provide both? Like we do for the button case above.

```suggestion
              <a
                key={id}
                className={`${blockClass}__icon`}
                onClick={onClick}
                href={href}
              >
```

dcwarwick (changes requested):
LGTM, with just one suggestion

packages/cloud-cognitive/src/components/Card/Card.js
davidmenendez: yeah that makes sense to me. good call 👍

dcwarwick (approved):

Issue #1130, opened 2021-08-10
Toolbar — keyboard accessibility


SimonFinney:
- https://pages.github.ibm.com/cdai-design/pal/patterns/toolbars/#accessibility
- https://www.npmjs.com/package/react-roving-tabindex

PR #1514, opened 2022-01-04
feat(toolbar): add keyboard accessibility

SimonFinney:
Contributes to #1130

Add [keyboard accessibility](https://pages.github.ibm.com/cdai-design/pal/patterns/toolbars/#keyboard) to `Toolbar`.


packages/cloud-cognitive/src/components/Toolbar/Toolbar.js
```diff
@@ -17,14 +26,74 @@ const blockClass = `${prefix}--toolbar`;
 const ToolbarContext = createContext();
 
 /** Toolbars are a collection of action items that organize a program’s interaction patterns into a series of closely related commands. */
-let Toolbar = forwardRef(({ children, className, vertical, ...rest }, ref) => {
+let Toolbar = forwardRef(({ children, className, vertical, ...rest }, r) => {
+  const _ref = useRef();
+  const ref = r || _ref;
```
davidmenendez, line 31: i'd change this to `const ref = r || useRef()` in order to avoid instantiating `_ref` unnecessarily.

davidmenendez (approved):

packages/cloud-cognitive/src/components/Toolbar/Toolbar.js
SimonFinney: > i'd change this to `const ref = r || useRef()` in order to avoid instantiating `_ref` unnecessarily.

Good eye and I usually agree — this was throwing [conditional Hook lint errors for `react-hooks/rules-of-hooks`](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level), which is frowned upon according to the documentation.

I think the impact of ignoring the rule is minimal as it affects a ref versus side-effects or state, but I thought it safer to walk the line. What are your thoughts?

davidmenendez: oh well if it's a linter issue then that's fine with me. i don't really care to fight with the linter at this time haha

dcwarwick: Yes, it's a React limitation -- we *have* to instantiate `_ref` whether we need it or not in order that React can guarantee the reactive stability. It's not an expensive instantiation and happens in several other similar places.

dcwarwick (approved):

Issue #1492, opened 2021-12-06
Inline Edit Component


lee-chase:
# Inline Edit

## Brief description

Adds an inline/in-place edit

## Details

https://pages.github.ibm.com/cdai-design/pal/patterns/edit#inline-edit
![image](https://user-images.githubusercontent.com/15086604/144872604-cf179a5b-db8f-4be5-b074-cef0e8b67b72.png)

## Design Maintainer(s)

- @vikkipaterson 

## Tasks

Before starting work on this epic, please review and complete the following.

- [x] Initial review of design/existing code

### Working in Carbon for IBM Products package

- [x] Have you reviewed our
      [contribution guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/CONTRIBUTING.md)?
- [x] Have you noted our
      [code guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/docs/CODE_GUIDELINES.md)?
- [x] Did you use the CLI to scaffold your new component?

lee-chase:
While this component is titled `Inline edit` it should be thought of as `In-place edit` to avoid confusion with both the DOM display term 'inline' and the Carbon TextInput option 'inline'.

Carbon Search uses an aria-label for the 'Clear search input' button. 
https://carbon-components-react.netlify.app/?path=/story/components-search--default

The design uses tooltips for 'Cancel/Revert' and 'Save'

The Carbon TextInput has 'inline', 'invalidText' and 'invalid' options. However it is not possible to see the `invalidText` when using the inline option. Carbon does not add the invalid message at all into the HTML.

![image](https://user-images.githubusercontent.com/15086604/144876874-2817591f-05bb-4181-8a0f-eaf1bfde1eab.png)
![image](https://user-images.githubusercontent.com/15086604/144877130-20da426d-69fb-48f9-b4f9-c4fda46d33f6.png)

As can be seen above, Carbon does allow the help text. Assuming this is also permitted for this design.

@vikkipaterson perhaps you can assist or suggest someone to assist with the following questions

1. Should we change the name to 'In-Place edit'?
2. How should we resolve the discrepancy with Carbon Search? [Do as Carbon does, suggest a change to Carbon, note in the design but disregard discrepancy]
3. How should we resolve the discrepancy with Carbon TextInput with inline and error message [Do as Carbon does, suggest a change to Carbon, note in the design but disregard discrepancy]
[...]

PR #1533, opened 2022-01-18
feat: add initial inline component and use in page header

lee-chase:
Contributes to #1492 

The first draft of inline edit for discussion with interested parties @cameroncalder @sdignum

#### What did you change?

- Updates to CancellableEdit
  - Added option to hide label
  - Added onRevert event
  - Renamed revertLabel, saveLabel to revertDesccription, saveDescription to more closely align with iconDescription for icon only button, added size prop, added saveDisabled prop/feature.
- Added inline edit
- Added usage of inline edit as part of the page header and it's storybook.. 

Storybook, will review design before revisiting tests.


davidmenendez (approved):
Looks good to me.

Issue #1521, opened 2022-01-11
Open prop to be added to OptionsTile


seonyukim:
## What will this achieve?

- enhancement: 
   - To sets the (initial) state of the openable/closable part
   - To enable the tile to be shown as open by default, or to open it in response to a user interaction elsewhere on the page if necessary.

## How will success be measured?
To check prop is added.

## Additional information

- Designs
- Existing code
- etc

janhassel:
@dcwarwick  We might wanna standardize this concept across components with something like [Carbon's `useControllableState` hook](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/internal/useControllableState.js).

elycheea:
@janhassel Opened a new issue for your suggestion. Feel free to add more detail or contribute/submit a PR.
https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/1540

elycheea:
Forgot to close this. Included in 1.4.0 release.

PR #1534, opened 2022-01-18
feat(OptionsTile): add open prop

elycheea:
Contributes to #1521 

This PR adds an open prop to the **OptionsTile**, allowing users to set an initial open/close state to an options tile.

#### What did you change?
Renamed `open` state to `isOpen` and added `open` prop with a default of `false`.  Added another state `prevIsOpen` to prevent infinite loop (pattern based on `AccordionItem`).

Storybook and ran test on OptionsTile.


packages/cloud-cognitive/src/components/OptionsTile/OptionsTile.js
```diff
@@ -49,6 +49,7 @@ export let OptionsTile = React.forwardRef(
       locked,
       lockedText,
       onToggle,
+      open = false,
```
davidmenendez, line 52: i think by convention we should keep the default values in `defaultProps`. if you can just remove the value here and `size` and just pop the `open` default value in `defaultProps` that should be good.

davidmenendez (changes requested):
looks good. just one comment

davidmenendez (approved):

PR #1537, opened 2022-01-19
fix: Prevent id duplication in page header

lee-chase:
Prevent duplicate ids in the hidden sizing items used in the page header.

#### What did you change?

Moved ID from hidden sizing items to `data-original-id` and added tests to ensure duplication of IDs does not happen.

Test and storybook.


davidmenendez:
if i'm understanding the intent correctly, this PR removes the optional `id` prop from `ActionBarItem`, `BreadcrumbItem`, and `Tag` and puts it in the `data-original-id` attribute. just out of curiosity what's the problem here? i'm not an expert on these components, so i'm curious as to why we're trying to safeguard these specific components from the user error of providing duplicate id's.

lee-chase:
> if i'm understanding the intent correctly, this PR removes the optional `id` prop from `ActionBarItem`, `BreadcrumbItem`, and `Tag` and puts it in the `data-original-id` attribute. just out of curiosity what's the problem here? i'm not an expert on these components, so i'm curious as to why we're trying to safeguard these specific components from the user error of providing duplicate id's.

Not quite. These components monitor both the size they would like to be and the space available. To do so they clone the active elements, hide them and measure the space they would use if it were available. It is these clones that we move the 'id' to 'data-originial-id'.

PR #1545, opened 2022-01-20
Rework breadcrumb back button

lee-chase:
Rework the back link/button shown in PageHeader with breadcrumbs and the breadcrumb with overflow on narrow screens.

On narrow screens this
image
would become this
image

#### What did you change?

Reworked to include the back as a breadcrumb item producing
image

Also fixed a glitch with the title animation into the breadcrumb area which was broken at narrow widths.

Storybook.


packages/cloud-cognitive/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -281,6 +281,8 @@ export let BreadcrumbWithOverflow = ({
     targetRef: breadcrumbItemWithOverflow,
   });
 
+  console.log(backItem);
```
matthewgallo, line 284: ^^ This is probably why you're tests are failing @lee-chase  😄

packages/cloud-cognitive/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -312,6 +299,20 @@ export let BreadcrumbWithOverflow = ({
           noTrailingSlash={noTrailingSlash}
           {...other}
         >
+          {backItem?.href && (backItem?.label || backItem?.title) && (
```
davidmenendez, line 302: i don't think the `?.` operator is necessary since you're not chaining anything. you should get the same results with just `.`

davidmenendez (changes requested):
looks good. just had one comment.

packages/cloud-cognitive/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
lee-chase: Removed when fixing tests.

packages/cloud-cognitive/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
lee-chase: Probably true in 99% of cases. However, if only one breadcrumb is passed and it happens to be the current page then 271 will result in backItem being nullish.

davidmenendez: ahh ok i see. that makes sense. thanks for clarifying.

davidmenendez (approved):

Issue #1546, opened 2022-01-20
add search filter functionality to add select components


davidmenendez:
the add select components need to allow the user to either pass in their own filter function or nothing and use a default filter.

PR #1547, opened 2022-01-20
feat: add search filter to add select

davidmenendez:
Contributes to #1546


matthewgallo (approved):
Don't forget the doc gen comments for your prop types. The filter work looks good!

PR #1556, opened 2022-01-24
chore: add component playground to storybook

dcwarwick:
Add a component playground into the storybook. This shows multiple components working together.

#### What did you change?

Added a story to the storybook

Ran storybook


packages/core/src/component-playground/ComponentPlayground.stories.js
```diff
@@ -0,0 +1,18 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { getPathForComponent } from '../../story-structure';
+
+import App from './ComponentPlayground';
+
+export default {
+  title: getPathForComponent('o', 'Examples'),
```
lee-chase, line 14: Change layout to full screen 

`
  parameters: { layout: 'fullscreen' },
`

lee-chase (changes requested):
Looks good, 1 change request to make fullscreen otherwise the header looks odd.

- It would also be nice if there was no content jump on showing the side panel. 
- Personally prefer the scroll bar to be under the global header as per in-context example.

packages/core/src/component-playground/ComponentPlayground.stories.js
dcwarwick: Good catch -- will check that in. @matthewgallo This fixes the extra space we were seeing around the PageHeader, including the content showing through.

matthewgallo: I jumped in to help with some things, including adding this piece ^

Issue #1508, opened 2021-12-23
Action button truncated in the Header component for SM dimensions


aplachka:
## What package(s) are you using?

- [x ] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.
The actions button got overlapped with a white element:

![image](https://user-images.githubusercontent.com/65154889/147278421-71028272-ba71-47d7-88de-bf832a977173.png)

> Is this issue related to a specific component?
PageHeader
> What did you expect to happen? What happened instead? What would you like to
> see changed?
The actions button should render without glitch/overlap
> What browser are you working in?
MS Edge
> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?
1.2.2
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue
https://carbon-for-ibm-products.netlify.app/?path=/story/carbon-for-ibm-products-components-page-header-pageheader--with-tags
when viewed from SM viewport

lee-chase:
There are narrower sizes that struggle to render the PageActions and provide a fixed space for the title. The simplest option would be to move page actions into the action bar, but we have to account for the PageActions being user defined.

PR #1557, opened 2022-01-24
fix(PageHeader): ensure page actions are not truncated

lee-chase:
Contributes to #1508 

When the page header drops down to the most narrow format the title and page actions are on separate lines taking up 100% width. This results in the pageActions not being in the right state to sit alongside the breadcrumb on scroll. 

#### What did you change?

This update changes the narrow layout to behave as if an actionBar is always provided. This results in the pageActions jumping into the page header instead of sticking to the top.

Storybook and tests.


dcwarwick (approved):
LGTM

PR #1562, opened 2022-01-25
fix(PageHeader): title without breadcrumb and action bar remains visible on scroll

### LINKED ISSUE #1566: PageHeader with action bar items but no breadcrumb loses title on scroll
There's an odd behavior with the title of the PageHeader in the case where there are action bar items (so there is a breadcrumb row) but *no* breadcrumbs and the page starts to scroll.

When there are breadcrumbs, the title starts to scroll with the page and then fades out to appear at the end of the breadcrumbs. When there is no breadcrumb row at all, the title just scrolls off with the page. But when there are action bar items but no breadcrumbs, the title starts to scroll with the page and then fades out, but doesn't appear anywhere. This looks a bit odd, and is probably not the intended behavior.

Title scrolling when there is no breadcrumb row -- this works OK:
no-a-and-no-b-ok

Title scrolling when there are action bar items and breadcrumbs:
a-and-b-ok

Title scrolling when there are action bar items but *no* breadcrumbs -- the title vanishes after it starts scrolling, but does NOT appear in the breadcrumb row:
a-and-no-b-wrong

I think that in this case, when there is a breadcrumb row (because of there being action bar items) but there happen to be no breadcrumbs, the title should fade when it starts to scroll and appear in the breadcrumb row exactly as if there is a set of zero breadcrumbs at the start of the row. This would seem more consistent with the other configurations, and with the design as described.

lee-chase:
Closes #1566 

Addresses concern @dcwarwick raised regarding usage with no breadcrumbs, but with title, page actions and an action bar. 

#### What did you change?

On page scroll the title appears as a breadcrumb.

Storybook but will request design review.


lee-chase:
Hi, @cameroncalder a query was raised about a specific use case in which the title fades but is not replaced in the breadcrumb row.

E.g. As per the following link with no breadcrumbs.
https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-page-header-pageheader--demo&args=breadcrumbs:0

This PR changes the behaviour (link below) but does not feel satisfactory. I'm tempted to suggest the line between breadcrumb row and title should not be there in this scenario and the title not be breadcrumb sized.

Could you take a look at this and advise.

dcwarwick:
@lee-chase @cameroncalder Lee, those links you've posted don't work for me. I don't know why, and I don't know whether they'll work for Cameron. Also, you've not really given much context, and you don't say why you think this fix doesn't "feel satisfactory" -- it looks fine to me, but maybe I've not noticed whatever it was you thought felt wrong. I'd certainly be happy to hear what Cameron thinks about it. A PR thread is not an ideal place to have a discussion and capture a decision. I've opened issue #1566, which this PR fixes, and maybe you could add some more detail there about what the concerns are that you'd like Cameron to comment on?

packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```diff
@@ -839,7 +839,7 @@ PageHeader.propTypes = {
    * It is used in an overflow menu when there is insufficient space to display all breadcrumbs inline.
    */
   breadcrumbOverflowAriaLabel: PropTypes.string.isRequired.if(
-    ({ breadcrumbs }) => breadcrumbs && breadcrumbs.length > 0
+    ({ breadcrumbs }) => breadcrumbs && breadcrumbs.length > 1 // always shows the title
```
dcwarwick, line 842: I don't think I get this. If the 'breadcrumbs' prop had one or more items in it, we were already moving the title up into the breadcrumb bar on scroll, so the behaviour in that case hasn't changed with this PR. Why does the logic for when we require an aria label for the breadcrumb overflow button change?

packages/cloud-cognitive/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -372,7 +372,9 @@ BreadcrumbWithOverflow.propTypes = {
   /**
    * overflowAriaLabel label for open close button overflow used for action bar items that do nto fit.
```
dcwarwick, line 373:
```suggestion
   * overflowAriaLabel label for open close button overflow used for breadcrumb items that do not fit.
```

packages/cloud-cognitive/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -372,7 +372,9 @@ BreadcrumbWithOverflow.propTypes = {
...
    */
-  overflowAriaLabel: PropTypes.string.isRequired,
+  overflowAriaLabel: PropTypes.string.isRequired.if(
+    ({ breadcrumbs }) => breadcrumbs.length > 1
```
dcwarwick, line 376: I don't think I get this. The BreadcrumbOverflow component hasn't changed, so how come the logic for when an overflow aria-label is required has changed? Was it incorrect before? The comment 'title always shown' doesn't really seem to apply here, because that's a detail of the PageHeader implementation and not part of the BreadcrumbWithOverflow component logic.

packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
dcwarwick: If anything, does the fact that we might move the title up into the breadcrumb bar on scroll even when there aren't any other breadcrumbs mean that we need a breadcrumb overflow aria-label all the time and not just when there are breadcrumbs?

dcwarwick (changes requested):

packages/cloud-cognitive/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
lee-chase: I think the logic was incorrect. 

When there is only one item supplied, the overflow is never shown.

packages/cloud-cognitive/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -370,9 +370,11 @@ BreadcrumbWithOverflow.propTypes = {
    */
   noTrailingSlash: PropTypes.bool,
   /**
-   * overflowAriaLabel label for open close button overflow used for action bar items that do nto fit.
+   * overflowAriaLabel label for open close button overflow used for breadcrumb items that do not fit.
    */
-  overflowAriaLabel: PropTypes.string.isRequired,
+  overflowAriaLabel: PropTypes.string.isRequired.if(
+    ({ breadcrumbs }) => breadcrumbs.length > 1
+  ), // title always shown
```
dcwarwick, line 377:
```suggestion
  ),
```

dcwarwick (changes requested):

dcwarwick (approved):
LGTM thanks

Issue #1568, opened 2022-01-25
Allow danger buttons in modal-like patterns (Tearsheet, Side Panel, etc)


dcwarwick:
A question was raised by @lily-peng concerning buttons in modal-like patterns (e.g. Tearsheets, Side Panels). The designs for these show/describe the navigation buttons as being primary, secondary or ghost. However, in “danger” situations, the primary button can often be replaced by a danger button. We should be supporting the use of a danger button in place of a primary button for those “danger” situations, as confirmed by Mitchell Bernstein: "Yes we should be supporting the replacement of primary blue with danger red and same for outline and ghost replacements in any scenario".

The prop-types for the affected components should accept danger buttons, and we should also update the ordering logic in ActionSet if necessary so that danger buttons appear in the right place.

matthewgallo:
I think we can close this one now 🎉

PR #1572, opened 2022-01-27
fix(Tearsheet,SidePanel): enable danger button kinds as actions

dcwarwick:
Contributes to #1568

Allow 'danger' in place of 'primary' and 'danger--ghost' in place of 'ghost' button kinds. This PR also updates the logic in ActionSet which ensures the buttons appear in the correct order.

#### What did you change?

Updated ActionSet, prop-types for Tearsheet and SidePanel, and tests and stories.

Ran tests and storybook.


matthewgallo (approved):
Looks great! 🎉 Probably a good idea to update side panel stories with the complete list of actions accepted, similar to the tearsheet story, but we can do that separately.

Issue #1582, opened 2022-02-01
refactor AddSelect custom filter


davidmenendez:
the current implementation for custom filtering is off and needs to be redone. currently it requires the user to maintain the state, but they should only need to provide the component with a filtering function.

PR #1583, opened 2022-02-01
fix(AddSelect): refactor custom filter and init navigation

davidmenendez:
Contributes to #1582 

* the code now checks for a `children` attribute in the items array to distinguish the hierarchy of the list and adds a chevron to indicate that you can drill down further. just the start of the implementation as this PR shifted focus while I was working on this.
* i refactored the way the search filter work and how custom filtering works. now it appropriately uses a hook to filter the data and now the user only needs to provide a filter function instead of maintaining the state. clearly the initial implementation was done hastily on my part. this should represent a more polished version.

### DIFF
--- a/packages/cloud-cognitive/src/components/AddSelect/AddSelect.js
+++ b/packages/cloud-cognitive/src/components/AddSelect/AddSelect.js
@@ -5,7 +5,7 @@
 // LICENSE file in the root directory of this source tree.
 //
 
-import React, { forwardRef, useState, useEffect } from 'react';
+import React, { forwardRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import {
@@ -56,25 +56,9 @@ export let AddSelect = forwardRef(
     };
 
     // hooks
-    const [filteredItems, setFilteredItems] = useState([]);
     const [singleSelection, setSingleSelection] = useState('');
     const [multiSelection, setMultiSelection] = useState([]);
     const [searchTerm, setSearchTerm] = useState('');
-
-    // filter items as a search term is entered
-    useEffect(() => {
-      const results = searchTerm
-        ? items.filter((item) => {
-            // if user provides their own filter function use that
-            if (onSearchFilter) {
-              return onSearchFilter(item, searchTerm);
-            }
-            // otherwise use the default label filter
-            return item.label.includes(searchTerm);
-          })
-        : items;
-      setFilteredItems(results);
-    }, [items, onSearchFilter, searchTerm]);
 
     // handlers
     const handleSearch = (e) => {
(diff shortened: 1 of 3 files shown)


packages/cloud-cognitive/src/components/AddSelect/AddSelect.js
```diff
@@ -55,17 +56,32 @@ export let AddSelect = forwardRef(
...
+    const [filteredItems, setFilteredItems] = useState([]);
     const [singleSelection, setSingleSelection] = useState('');
     const [multiSelection, setMultiSelection] = useState([]);
     const [searchTerm, setSearchTerm] = useState('');
 
+    // save items locally
+    useEffect(() => {
+      setFilteredItems(items);
+    }, [items]);
+
```
dcwarwick, line 68: I'm a bit wary of having two useEffects both of which call 'setFilteredItems'. They are both dependent on `items`, and if `items` changes I don't think we can be certain which order the hooks will run in. This could mean that when there is a filter in effect, programmatically changing the set of items causes the filtering to be temporarily "lost" if the second useEffect triggers first (applying filtered items) but then this one triggers second, applying the unfiltered items list.

I think it might be safer to just have the second useEffect, and make sure that if there is no `searchTerm` set then results is just set equal to the unfiltered items. Then that hook is responsible for making sure the list is set, filtered or unfiltered.

packages/cloud-cognitive/src/components/AddSelect/AddSelect.js
```diff
@@ -55,17 +56,32 @@ export let AddSelect = forwardRef(
...
+    // filter items as a search term is entered
+    useEffect(() => {
+      const results = items.filter((item) => {
```
dcwarwick, line 71: So something along the lines of 
```
      const results = searchTerm ? items.filter((item) => { ... }) : items;
```

packages/cloud-cognitive/src/components/AddSelect/AddSelect.js
davidmenendez: ahh yes i see what you mean. yes that does sound like a good solution.

davidmenendez:
@dcwarwick changes made

dcwarwick (approved):
LGTM, thanks

davidmenendez:
sorry @dcwarwick i made additional changes. after looking at this for a bit i realized i made mistake with this implementation. i thought i needed to keep the items in state, but then i realized that that was completely unnecessary and is react anti-pattern. since the user provides the items we don't need to track it and only need to worry about filtering it, so i removed the extraneous state management and now the filtering will happen normally during render. you can see these changes in the last commit https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1583/commits/def7ac94322a08e27ab803c215474e798367ea71

matthewgallo (approved):

PR #1585, opened 2022-02-02
chore(InlineEdit): use content-editable element for input

lee-chase:
Contributes to #1492 

#### What did you change?

- Replaces inline edit use of an input box with 'contentEditable'.
- No longer requires the CancellableEdit which can be removed.
- Used in page header, needs tweaking based on title height/location

Storybook and test.


packages/cloud-cognitive/src/components/InlineEdit/_inline-edit.scss
```diff
@@ -18,24 +18,196 @@
...
 // the Carbon import-once mechanism.
 @mixin inline-edit {
   // The block part of our conventional BEM class names (blockClass__E--M).
   $block-class: #{$pkg-prefix}--inline-edit;
 
-  .#{$block-class} {
+  @include block-wrap($block-class, 3) {
+    --size: #{$spacing-08};
+    --icon-size: #{$spacing-05};
+    --background-color: #{$ui-01};
```
dcwarwick, line 51: I think CSS variable names need to be scoped, as they are global once on the DOM. Our usual pattern here is to include the block-class, which seems a good plan, _i.e._
- `--#{$block-class}--size`
- `--#{$block-class}--icon-size`
- `--#{$block-class}--background-color`

packages/cloud-cognitive/src/components/InlineEdit/_inline-edit.scss
```diff
@@ -18,24 +18,196 @@
...
+}
+
 // Define all component styles in a mixin which is then exported using
 // the Carbon import-once mechanism.
 @mixin inline-edit {
   // The block part of our conventional BEM class names (blockClass__E--M).
   $block-class: #{$pkg-prefix}--inline-edit;
 
-  .#{$block-class} {
+  @include block-wrap($block-class, 3) {
```
dcwarwick, line 48: Woah there horsey...  I like the ingenuity here, but this feels like a big hammer. Depth 3? This is going to bloat the CSS considerably, with three copies of the whole block class on every CSS rule. I'm guessing you were fighting a specificity war, but one problem with this approach is that it passes the whole battle on to our users: if anyone wants to override any of the CSS styles, they are going to have to out-specify us, which means all those rules will need specificity 4 or more. It's going to be very hard to override any of our styles!

I think we should be a lot more selective -- if there are some rules that need extra specificity, we can add repeated classes there, rather than swamping our whole CSS in a three-level-deep specificity wrapper. One advantage of popping `&.#{$block-class}.#{$block-class} .#{$block-class}--entry-field {` on each affected rule is that it's a clear and maintainable declaration that we needed to boost the specificity here.

packages/cloud-cognitive/src/global/styles/_mixins.scss
```diff
@@ -5,8 +5,13 @@
...
 
-@mixin block-wrap($block-class) {
-  .#{$block-class} {
+@mixin block-wrap($block-class, $deep: 1) {
+  $selector: '';
+
+  @for $i from 1 through $deep {
+    $selector: $selector + '.' + $block-class;
+  }
+  #{$selector} {
```
dcwarwick, line 14: I don't think I like this. Well, I quite like it :-) but I'm not sure we should do it. It encourages the wrapping of great swathes of CSS in multi-class specificity wrappers that may be over the top and will bloat the CSS because these repeated classes will appear on every single CSS rule generated.

We could make this more efficient. For example,
```
@mixin block-wrap($block-class, $deep: 1) {
  $selector: $block-class;

  @for $i from 2 through $deep {
    $selector: $selector + ':not(\9)';
  }
  #{$selector} {
    @content;
  }
}
```
will I think achieve the same specificity with rather less bloat. However, I think it's actually better not to do this at all -- we should include the specificity boost on each rule that needs it, because that way we don't apply more specificity-busting than we need, and it is clear in the SCSS source which rules needed the specificity fixes.

dcwarwick (changes requested):

dcwarwick (approved):
LGTM. Bonus points for comments!

Issue #1590, opened 2022-02-04
Add hierarchy support to AddSelect


davidmenendez:
`AddSelect` needs to support nested lists.

PR #1591, opened 2022-02-04
fix(AddSelect): adds nested lists

davidmenendez:
Contributes to #1590 

adds nested list support to `AddSelect`. it works by keeping track of the levels in state `path`, which might look something like this

```
path = [
  { id: '1', label: 'the value used to display in the breadcrumb' },
  { id: '2', label: 'breadcrumb 2' },
];
```

every time the user click the button to go down another level an entry is added to `path`. then on render the `getFilteredItems` function is fired to filter the results. if a path is detected `path` is looped over to traverse the `items` array. the `id` in the `path` object corresponds to the id of the item, so while `path` is being looped over it's using `array.find` to locate the object in the array and then returns the list of `children` items to be displayed.

this is a tad messy, but i wanted to get the functionality working. my next step is to clean up and separate the code into their own files like the breadcrumbs.


dcwarwick (approved):
LGTM. Don't forget to award yourself a 2022 in the copyright headings at some point :-)

Issue #1589, opened 2022-02-04
Notifications panel: add support for prefers-reduced-motion


matthewgallo:
## What will this achieve?

We need to be able to opt out of the entrance/exit animations for the NotificationsPanel when prefers-reduced-motion is turned on.

## How will success be measured?

When the notifications panel does not animate if the user's reduced motion setting is turned on.

matthewgallo:
Complete

PR #1598, opened 2022-02-07
feat(NotificationsPanel): support reduced motion

matthewgallo:
Contributes to #1589 

Prevents the entrance/exit animation when the user's preferences are set to "reduce motion".

#### What did you change?
`NotificationsPanel.js`
Storybook


dcwarwick (approved):
LGTM. Set to merge, because I've got a PR coming that touches the same file so I'll sort the conflict after this merges.

Issue #1618, opened 2022-02-10
add utility to MultiAddSelect to find and displayed selected items


davidmenendez:
the sidebar in `MultiAddSelect` needs an easy way to find and display items regardless of how nested the item is. my solution is to take the list of items and flatten it into a single array of items so that it can easily use `array.find`

so we turn something like this
```
const items = [{
  id: 1,
  value: item a,
  children: [{
    id: 2,
    value: item a-a,
  }],
}];
```

into
```
const items = [{
  id: 1,
  value: item a,
}, {
  id: 2,
  value: item a-a,
}];
```

now it can easily use `items.find` to find and display the selected items in the sidebar

PR #1619, opened 2022-02-10
fix: add select sidebar flatten

davidmenendez:
Contributes to #1618 

reworks the sidebar to use the carbon `Accordion` to display selected items and adds a utility function to flatten the list of items to make nested items easily searchable.


packages/cloud-cognitive/src/components/AddSelect/AddSelectSidebar.js
```diff
@@ -27,23 +40,32 @@ export let AddSelectSidebar = ({
...
-      </div>
+        </div>
+      )}
     </div>
   );
 };
 
 AddSelectSidebar.propTypes = {
   influencerTitle: PropTypes.string,
+  items: PropTypes.array,
```
matthewgallo, line 68: It also might be worth extending the prop types for `items` a bit now that it looks like an `id` property is required

ie:
```jsx
items: PropTypes.arrayOf(
  PropTypes.shape({
    value: PropTypes.string,
    id: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
  })
)
```

I don't know if those are the correct prop types but giving a bit more depth as to the structure of `items` might be good.

davidmenendez: that's fair 👍

davidmenendez:
@matthewgallo updated the base `AddSelect` component to require required props in the items array

packages/cloud-cognitive/src/components/AddSelect/AddSelectSidebar.js
```diff
@@ -27,23 +40,32 @@ export let AddSelectSidebar = ({
...
-      <div className={`${blockClass}-body`}>
-        {multiSelection.length > 0 ? (
-          multiSelection.map((item) => <p key={item}>{item}</p>)
-        ) : (
+      {multiSelection.length > 0 ? (
+        <Accordion align="start">
+          {sidebarItems.map((item) => (
+            <AccordionItem title={item.value} key={item.id}>
+              {Object.keys(item).map((key) => (
+                <p key={key}>{`${key} - ${item[key]}`}</p>
```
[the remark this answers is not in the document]
davidmenendez: this is just temporary. i can make another update to this to clean it up.

matthewgallo (dismissed):

davidmenendez:
@matthewgallo also updated the sidebar string issue

PR #1622, opened 2022-02-11
fix: inline edit validation presentation

lee-chase:
Contributes to #1492 

Ensure validation and warning messages can be presented.

#### What did you change?

- Changed the structure to allow the validation icon to appear in the correct place.
- Added validation icon and text styles
- Adjusted the animation on starting to type text to account for the structure chagne.

Storbyook.


packages/cloud-cognitive/src/components/InlineEdit/InlineEdit.js
```diff
@@ -83,6 +83,11 @@ export let InlineEdit = React.forwardRef(
     ) : null;
 
     const doSetEditing = (value) => {
+      if (value === false) {
+        console.log(refInput.current.scrollWidth);
+        // move scroll to start
```
dcwarwick, line 88:
```suggestion
      if (value === false) {
        // move scroll to start
```

packages/cloud-cognitive/src/components/InlineEdit/InlineEdit.js
```diff
@@ -264,21 +272,24 @@ export let InlineEdit = React.forwardRef(
...
-              (refInput.current && value !== internalValue),
+              !invalid &&
+              !saveDisabled &&
+              refInput.current &&
+              value !== internalValue,
           })}
+          // tabindex -1 fixes blur target test when clicking on controls background
+          tabIndex="-1"
         >
+          {!editing && <div className={`${blockClass}__ellipsis`}>...</div>}
```
dcwarwick, line 287:
```suggestion
          {!editing && <div className={`${blockClass}__ellipsis`}>&hellip;</div>}
```

packages/cloud-cognitive/src/components/InlineEdit/_inline-edit.scss
```diff
@@ -162,8 +184,11 @@
       position: absolute;
       top: 0;
       right: 0;
-      width: calc(2 * var(--#{$block-class}--size));
+      display: flex;
+      width: calc(3 * var(--#{$block-class}--size));
       height: 100%;
+      justify-content: flex-end;
+      // width: room for validation, and two buttons
```
dcwarwick, line 191:
```suggestion
```

packages/cloud-cognitive/src/components/InlineEdit/_inline-edit.scss
```diff
@@ -162,8 +184,11 @@
       position: absolute;
       top: 0;
       right: 0;
-      width: calc(2 * var(--#{$block-class}--size));
+      display: flex;
+      width: calc(3 * var(--#{$block-class}--size));
```
dcwarwick, line 188: This comment seems to have become detached from the line it refers to...

```suggestion
      // width: room for validation, and two buttons
      width: calc(3 * var(--#{$block-class}--size));
```

packages/cloud-cognitive/src/components/InlineEdit/_inline-edit.scss
```diff
@@ -162,8 +184,11 @@
...
       height: 100%;
+      justify-content: flex-end;
```
dcwarwick, line 190: `flex-end` doesn't seem quite right here: it stacks the ellipsis over to the right, miles away from the truncated text. I think the ellipsis should be right up at the left edge. With just an ellipsis and one button this can be accomplished with `space-between`, but actually I wonder whether the ellipsis really belongs inside the controls div at all? It isn't, after all, a control, and however many controls there are they probably want to stack from the right while the ellipsis wants to be at the extreme left abutting to the truncated text. Maybe the ellipsis and the controls div should be siblings in a wrapping div that occupies the blank space next to the text.

```suggestion
      justify-content: space-between;
```

packages/cloud-cognitive/src/components/InlineEdit/_storybook-styles.scss
```diff
@@ -5,9 +5,9 @@
 // LICENSE file in the root directory of this source tree.
 //
 
-@import './index';
-
 // TODO: add any additional styles used by InlineEdit.stories.js
+@import '../../global/styles/project-settings';
+@import '../../global/styles/mixins';
```
dcwarwick, line 10: Not sure why you import mixins here -- you don't seem to be using it.

```suggestion
@import '../../global/styles/project-settings';
```

dcwarwick (changes requested):

packages/cloud-cognitive/src/components/InlineEdit/_inline-edit.scss
lee-chase: Moved up to correct location

packages/cloud-cognitive/src/components/InlineEdit/_inline-edit.scss
lee-chase: @dcwarwick had a bit of a rework here as suggested now have

__right-input-elements containing ellipsis and toolbar, and __toolbar containing the validation icon and buttons.

dcwarwick (approved):
LGTM thx

PR #1641, opened 2022-02-15
feat(InlineEdit): add validation

lee-chase:
Contributes to #1492 

Adds validation as a strawman to inline edit

#### What did you change?

- Added possible responses to onChange and onSave
  - returning false or { invalid: true } sets the internal state to invalid
  - returning { warn: true } sets the internal state to warn. 
- Added internal state for warn and invalid states to handle onSave and onChange responses

Storybook.


dcwarwick (approved):
LGTM

PR #1654, opened 2022-02-16
chore: remove use of defaultProps (issue #1644)

### LINKED ISSUE #1644: Remove use of defaultProps on functional components
defaultProps should be replaced with default values on functional components

related #facebook/react/pull/16210

dcwarwick:
Closes #1644

Use of `defaultProps` in functional components is now being deprecated. This PR replaces `defaultsProps` with ES6 default values on restructured prop parameters. This PR establishes a convention of storing the default values to be used in a `defaults` object, which ensures that all the values are initialised before they are used as parameter values, and so helps avoid the anti-pattern of instantiating a non-primitive value in the parameter default value expression which can cause React to re-render unnecessarily on every call, as well as being a clear and helpful way to set out the values being used.

NB a number of the default values we were providing weren't actually needed -- mostly falsy -- so I took the opportunity to drop those where possible.

#### What did you change?

default props for many components

ran tests, build, and checked storybook.


packages/cloud-cognitive/src/components/APIKeyModal/APIKeyModal.js
```diff
@@ -33,19 +33,26 @@ import uuidv4 from '../../global/js/utils/uuidv4';
 
 const componentName = 'APIKeyModal';
 
+// Default values for props
+const defaults = {
```
lee-chase, line 37: Are you suggesting we do not use `...defaults` as it is clearer?

lee-chase (dismissed):

packages/cloud-cognitive/src/components/OptionsTile/OptionsTile.js
```diff
@@ -32,13 +32,19 @@ const componentName = 'OptionsTile';
 
 // NOTE: the component SCSS is not imported here: it is rolled up separately.
 
+// Default values for props
+const defaults = {
+  size: 'xl',
```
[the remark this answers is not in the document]
dcwarwick: I didn't just drop falsy values -- each one I dropped I went through the code and checked how it was being used, and dropped it if I thought that false, null, undefined would all be handled the same so the default was unnecessary. So a few `false` default values do remain.

packages/cloud-cognitive/src/components/APIKeyModal/APIKeyModal.js
dcwarwick: I couldn't find a neater way :-)

lee-chase (approved):

Issue #1642, opened 2022-02-15
Add Select multi select hierarchy support


davidmenendez:
the add select pattern needs to support multi select hierarchy with these series of columns

<img width="795" alt="Screen Shot 2022-02-15 at 4 07 55 PM" src="https://user-images.githubusercontent.com/6370760/154157815-a0b473c9-10d4-4474-adac-af279145bdb7.png">

PR #1659, opened 2022-02-17
fix: adds mutli select hierarchy support

davidmenendez:
Contributes to #1642

adds multi select hierarchy support. there's a lot going on here, so i'll try to explain everything that's being done.

the biggest hurdle was figuring out how to properly display "pages" or "columns" of children. i did some reading and decided that the best way to handle this was to normalize the items into a single object with everything at the top level. the benefit here is that there's no need to do extensive digging through objects and arrays to locate a deeply nested child. the data schema that we would expect from a user should look like this

```
{
   "items":[
      {
         "id":"1",
         "value":"1",
         "title":"item 1",
         "subtitle":"item 1",
         "children":[
            {
               "id":"1-1",
               "value":"1-1",
               "title":"item 1-1",
               "children":[
                  {
                     "id":"1-1-1",
                     "value":"1-1-1",
                     "title":"item 1-1-1"
                  },
                  {
                     "id":"1-1-2",
                     "value":"1-1-2",
                     "title":"item 1-1-2"
                  }
               ]
            },
            {
               "id":"1-2",
               "value":"1-2",
               "title":"item 1-2",
               "children":[
                  {
                     "id":"1-2-1",
                     "value":"1-2-1",
                     "title":"item 1-2-1"
                  }
               ]
            }
         ]
      },
      {
         "id":"2",
         "value":"2",
         "title":"item 2",
         "subtitle":"item 2"
      },
      {
         "id":"3",
         "value":"3",
         "title":"item 3",
         "subtitle":"item 3",
         "children":[
            {
               "id":"3-1",
               "value":"3-1",
               "title":"item 3-1"
            }
         ]
      }
   ]
}
```

but trying to navigate a structure like this would be difficult. instead, now a `normalize` function is added to transform that structure into this structure.

```
[
   [
      {
         "id":"1",
         "value":"1",
         "title":"item 1",
         "subtitle":"item 1",
         "children":[
            "1-1",
            "1-2"
         ]
      },
      {
         "id":"2",
         "value":"2",
         "title":"item 2",
         "subtitle":"item 2"
      },
      {
         "id":"3",
         "value":"3",
         "title":"item 3",
         "subtitle":"item 3",
         "children":[
            "3-1"
         ]
      }
   ]
]
```

once normalized the new data schema is put in state. this structure represents the "page" or "column" structure where each array represents a list of items. when rendered it looks like this

Screen Shot 2022-02-17 at 2 08 07 PM

now all the ui has to do is loop over this set of results. we can easily locate the list of direct children that each item has. if we drill down into `item 1` the new normalized data looks like this

```
[
   [
      {
         "id":"1",
         "value":"1",
         "title":"item 1",
         "subtitle":"item 1",
         "children":[
            "1-1",
            "1-2"
         ]
      },
      {
         "id":"2",
         "value":"2",
         "title":"item 2",
         "subtitle":"item 2"
      },
      {
         "id":"3",
         "value":"3",
         "title":"item 3",
         "subtitle":"item 3",
         "children":[
            "3-1"
         ]
      }
   ],
   [
      {
         "id":"1-1",
         "value":"1-1",
         "title":"item 1-1",
         "parent":"1",
         "children":[
            "1-1-1",
            "1-1-2"
         ]
      },
      {
         "id":"1-2",
         "value":"1-2",
         "title":"item 1-2",
         "parent":"1",
         "children":[
            "1-2-1"
         ]
      }
   ]
]
```

and looks like this

Screen Shot 2022-02-17 at 2 15 36 PM

we still have the top level items at the start of the array and then we append the next set of results, which are the children of `item 1`. when the drill down arrow is clicked the ui uses the id of the current id and any existing parent id to determine the correct path to take and what the results should look like.

finally if we drill down once again we see this

Screen Shot 2022-02-17 at 2 14 57 PM

and the data looks like this

```
[
   [
      {
         "id":"1",
         "value":"1",
         "title":"item 1",
         "subtitle":"item 1",
         "children":[
            "1-1",
            "1-2"
         ]
      },
      {
         "id":"2",
         "value":"2",
         "title":"item 2",
         "subtitle":"item 2"
      },
      {
         "id":"3",
         "value":"3",
         "title":"item 3",
         "subtitle":"item 3",
         "children":[
            "3-1"
         ]
      }
   ],
   [
      {
         "id":"1-1",
         "value":"1-1",
         "title":"item 1-1",
         "parent":"1",
         "children":[
            "1-1-1",
            "1-1-2"
         ]
      },
      {
         "id":"1-2",
         "value":"1-2",
         "title":"item 1-2",
         "parent":"1",
         "children":[
            "1-2-1"
         ]
      }
   ],
   [
      {
         "id":"1-1-1",
         "value":"1-1-1",
         "title":"item 1-1-1",
         "parent":"1-1"
      },
      {
         "id":"1-1-2",
         "value":"1-1-2",
         "title":"item 1-1-2",
         "parent":"1-1"
      }
   ]
]
```

there is still a lot more additional work to do including additional functionality and styling, but getting this part to work was a large chunk of work, so i'm doing feature as it's own PR.


lee-chase (approved):
The hierarchy looks good, but I'm not sure about the UX.

- It would seem I can only view the hierarchy using the mouse. Plus even when using the mouse it would appear I have to click on the chevron to view the sub-terms. 
- Search does not work in the MultiAddSelect story.
- The single select version does not select the radio button if I click on the row. 
- The 'Enter', 'Tab' or 'Right arrow' keys would seem a natural way to open the hierarchy. However, cursor keys are already used by radio buttons, and there is no clear winner for reversing the use of 'Enter'. 'Tab' sort of works but ordinarily would take you out of a radio group.

Issue #1652, opened 2022-02-16
Dataspreadsheet: keyboard support for active cell


matthewgallo:
## What will this achieve?

Will allow arrow key usage for moving around the active cell when the entire data spreadsheet has focus.

<img src="https://user-images.githubusercontent.com/10215203/153658630-1170f9f4-8ddf-4348-9a7a-05c37ce6f81c.png" alt="Spreadsheet with active cell displaying image">

## How will success be measured?

When you can navigate through the spreadsheet via keyboard, allowing the active cell to change based on arrow key usage.

## Additional information

- Designs
- Existing code
- etc

PR #1663, opened 2022-02-18
feat(DataSpreadsheet): #1652 spreadsheet keyboard active cell

matthewgallo:
Contributes to #1652 

Adds keyboard support allowing the user to move the active cell in the spreadsheet via the use of the arrow keys or clicking on a cell within the spreadsheet.

**Note: the keyboard behavior within the spreadsheet only works when the spreadsheet is focused on
spreadsheet_with_arrow_keys

#### What did you change?
```
packages/cloud-cognitive/src/components/DataSpreadsheet/DataSpreadsheetHeader.js
packages/cloud-cognitive/src/components/DataSpreadsheet/createActiveCellFn.js
packages/cloud-cognitive/src/components/DataSpreadsheet/getCellSize.js
packages/cloud-cognitive/src/global/js/hooks/index.js
packages/cloud-cognitive/src/global/js/hooks/useActiveElement.js
```
Storybook and ran tests


elycheea (commented):
Overall looks good but what’s the expected behavior if someone does hit tab while in the data spreadsheet? Right now, the whole table is focused. Hitting tab seems to exit the table and tab through Storybook (probably expected), however, the data table doesn’t lose its focus class/style, which it probably should.

Also, it looks like the header row in the compact size data table gets misaligned though. I can open a separate bug for that since it doesn’t seem like related to these changes.

matthewgallo:
I made sure that if the spreadsheet loses it's focus, then it's border will no longer show when tabbing away

elycheea (approved):
Looks good! Thanks for making that change. 🌟

PR #1679, opened 2022-02-21
feat(InlineEdit): add cancel on use of escape key

lee-chase:
Contributes to #1492 

#### What did you change?

Pressing escape now attempts to cancel edit.

Storybook.


packages/cloud-cognitive/src/components/InlineEdit/InlineEdit.js
```diff
@@ -197,9 +197,14 @@ export let InlineEdit = React.forwardRef(
...
-      if (ev.key === 'Enter') {
-        ev.preventDefault();
-        refInput.current.blur(); // will cause save
+      switch (ev.key) {
+        case 'Enter':
+          ev.preventDefault();
+          refInput.current.blur(); // will cause save
+          break;
+        case 'Escape':
+          refInput.current.blur(); // will cause save
```
dcwarwick, line 206: The PR title says "add cancel on use of escape key", but what you're actually doing on use of escape key is save. Just checking which of these was intended -- should escape trigger save, or should it trigger cancel? If the former, maybe fix the PR title :-) if the latter, fix the code :-)

lee-chase: Ah yes, oops @dcwarwick

Issue #1682, opened 2022-02-22
Tearsheet heading border being cut by margin-right


davidebibm:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)

## Detailed description

> Describe in detail the issue you're having.
Tearsheet heading border being cut by margin-right

> Is this issue related to a specific component?
Tearsheet

> What did you expect to happen? What happened instead? What would you like to
> see changed?
border not begin cut

> What browser are you working in?
chrome

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?
v1.7.0

> What offering/product do you work on? 
https://github.ibm.com/automation-base-pak/pm-wui-react

## Steps to reproduce the issue

1. just open Tearsheet on https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-tearsheet-tearsheet--fully-loaded

## Additional information

https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-tearsheet-tearsheet--fully-loaded
![image](https://user-images.githubusercontent.com/98974352/155150436-12bd3e19-b790-4577-bc6e-1015a4770ec3.png)

elycheea:
Released in [@carbon/ibm-products@1.9.0](https://github.com/carbon-design-system/ibm-cloud-cognitive/releases/tag/%40carbon%2Fibm-products%401.9.0).
Thanks again for both opening this issue and contributing the fix!

PR #1683, opened 2022-02-22
fix: Using padding instead of margin to prevent border to being cut

davidebibm:
Contributes to #1682 

Fix Tearsheet heading border being cut by margin-right by replace margin with padding

#### What did you change?

Replaced margin right by adding same space to padding right

Loaded storybook up locally to test

BEFORE
BEFORE

AFTER
after


elycheea (dismissed):
Lgtm! Thanks for opening and fixing this!

elycheea (approved):
Updated the styles snapshot as discussed with @matthewgallo

PR #1684, opened 2022-02-23
chore: try different extent indicator

lee-chase:
People seem confused by the border placed around components in storybook to show the extent of the component as controlled by storybook. Typically this is used in cases where shrinking the space used by a component demonstrates an alternative behaviour.

#### What did you change?

Usage of this decorator
image

to this one
image

Storybook


davidmenendez:
to be honest i find this design to be much more... jarring? could we simply implement a way to change the background color of the page for components / patterns that obfuscate the border of these components?

dcwarwick:
So a key concern from the designer was that anything with backgrounds or borders or the like risks making it unclear what is part of the component itself and what isn't. However, I agree that the initial version of this was a bit 'in your face' and in several cases dominated the component itself! Lee and I had a chat about ways to 'soften' the design and I've adjusted it to show this style:
image
One advantage of this change is that this is now all controlled by the DisplayBox component, so we can make further adjustments of the visual effect going forwards quite quickly that will apply at once to all the places it is used.

dcwarwick (approved):

PR #1710, opened 2022-03-01
fix: remove reserved space for invalid icon

lee-chase:
Contributes to #1492 

Remove the reserved space for the invalid/warn icons.

#### What did you change?

Inline edit styles adding space for the invalid/warn icon if needed.

Storybook.

### DIFF
--- a/packages/cloud-cognitive/src/components/InlineEdit/_inline-edit.scss
+++ b/packages/cloud-cognitive/src/components/InlineEdit/_inline-edit.scss
@@ -81,7 +81,9 @@
 
     position: relative;
     display: inline-block;
-    // min-width: 250px; // similar to min input box with room for toolbar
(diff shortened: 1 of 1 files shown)


dcwarwick (approved):
LGTM. Some things to consider:

- The invalid icon appears rather suddenly compared to the slinky animate-in of the save button. I wonder if it could have some kind of ease-in too? Althought it doesn't push over the other buttons like the save button does. Maybe a slide out, or even a fade in, with the same timing as the save button? Just a thought.

- I wonder if we should even be reserving two icons of space. We could just reserve one icon of space minimum, two when either invalid/warn or outstanding changes, three when both invalid/warn and outstanding changes. For example, it currently looks a bit weird if you have invalid/warn and the edit button is optional and there is no focus in the field -- big gap to right of invalid/warn icon. Again, just a thought.

image

PR #1712, opened 2022-03-01
chore: ignore builds for dependabot PRs

matthewgallo:
This will hopefully save us in preventing netlify from creating PR builds for dependabot PRs. See [here](https://www.netlify.com/blog/2020/04/27/ignore-unnecessary-builds-to-optimize-your-build-times/) for details.

#### What did you change?
```
netlify.toml
scripts/ignore_dependabot.sh
```


dcwarwick (approved):
Oh, cool! I hadn't thought of doing this. It's the last couple of weeks when we suddenly seem to have been deluged with dependabot issues -- I think there must be a couple of advisories on widely used dependencies right now -- and the builds are a bit pointless since the code hasn't actually changed and we'll probably pick up the package upgrades at our next weekly update rather than merge the dependabot PRs anyway! I don't want to turn off dependabot, because it's useful to know that there are things we need to sort out, but this will definitely help reduce our netlify burn!! Thanks.

PR #1729, opened 2022-03-03
ci: add concurrency option to CI workflow

joshblack:
Hi there! 👋 This PR adds a setting to the CI workflow to cancel existing jobs for a given PR when a new commit is added. This option helps to reduce the number of runners used across repos as the org is limited to only 20 runners.

#### What did you change?

I updated the `.github/workflows/ci.yml` workflow with the `concurrency` option from: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

We are currently using this config over on the Design System monorepo to reduce the number of runners being used by actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - main
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


matthewgallo (approved):
💯  thanks @joshblack

Issue #852, opened 2021-06-02
SidePanel (enhancement): mobile responsiveness of panels


matthewgallo:
## What will this achieve?

When the panels hit a certain viewport width, they should become full width (similar to Carbon's modal). Will discuss with Kacie to confirm this is the intended behavior.

jagathgj:
@matthewgallo Could you help me connect with designer? I would need to know more about the behaviour. Perhaps a design in Invision may help.

matthewgallo:
Complete 🎉

PR #1735, opened 2022-03-04
fix: SidePanel (enhancement): mobile responsiveness of panels

jagathgj:
Contributes to #852 

SidePanel (enhancement): mobile responsiveness of panels

#### What did you change?

Changed min-width to width and max-width set to 100% in `_sidepanel.scss`.

Storybook


jagathgj:
@matthewgallo I was able to fix this issue in changing scss only. There was a min-width which needs to be changed to width and a max-width set to 100% solved the issue. Kindly verify and let me know if this is the best way.

jagathgj:
@matthewgallo I have added style snapshot. Kindly check if this is good to go.

matthewgallo (approved):
Looks great @jagathgj 🥳

Issue #1649, opened 2022-02-16
Design review for inline-edit


lee-chase:
Hi @sdignum @cameroncalder 

The inline edit component is good for a design review. 

It would be good to look at this as a group to discuss the current behaviour, unimplemented behaviour and at least one obvious issue with PageHeader usage.

You can find the inline-edit component deployed here 
https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-patterns-edit-and-update-inlineedit-canary--inline-edit

And see a it's usage in the PageHeader here by changing the tile in the controls tab to 'Editable title'
https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-page-header-pageheader--with-title

Relates to #1492 

### Standards

- [x] All pattern updates/changes/iterations have been discussed with the component developer 
- [x] Patterns have been approved by either DSAG or another approving entity

### Pattern and behavior

- [x] The component behaves according to the guidelines set by the pattern maintainer
- [x] The component’s UI matches the pattern specifications set by the pattern maintainer
- [x] The component’s motion matches the specifications set by the pattern maintainer(s)
- [x] The UI produced is responsive, cross-browser, and responds to the currently set Carbon theme.
- [x] Colors, themes, sizes and additional props are true and correct, aligning with Carbon set tokens (unless otherwise specified by the pattern)
- [x] Paddings/margins/spacings are true and correct, in both desktop and mobile views

### Storybook

- [x] A functioning component renders in Storybook
- [x] The Storybook displays multiple scenarios that show how the component can be used
- [x] Changing props in the Storybook updates the component

lee-chase:
As part of the design review, we should discuss the following:
- The content is not static and tabbing to it currently starts an edit. This enables a screen reader to see content is editable without having to go past it to see an edit button and also means that clicking on the text places the caret at the pointer location if in the text.
- It felt very odd to have a disabled save button appear under the edit button when clicked. Changed so that only the cancel is displayed until the content is changed.
- Validation can be tested by changing the 'Validation' control. Note, that validation can take place on 'Save' or 'Change' events the latter being on every keystroke.
- Both form validation and warning are supported.
- Handling navigation away from the page is not something a component should be handling, it should instead be deferred to the application. It's also worth noting that while displaying a custom dialog is technically possible in **SOME** browsers, the most common behaviour is to defer to the browser. E.g. in Chrome
[...]

PR #1741, opened 2022-03-04
chore: design review remove warn state

lee-chase:
Contributes to #1649 

Removes "warn" state following design review.
Validation now only supports the 'invalid' state.

#### What did you change?

InlineEdit component, styles, story and tests.

Storybook and test


dcwarwick (approved):
LGTM

Issue #1743, opened 2022-03-04
Profile pictures have incorrect hover state


mitchellbernstein:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

Profile pictures should have no inherent hover state change on icons inside them. Doing so may become inaccessible and look visually incorrect.

> Is this issue related to a specific component?

Profile pictures

> What did you expect to happen? What happened instead? What would you like to see changed?

Nothing except a tooltip should appear when hovering.

> What browser are you working in?

Firefox, Safari, Google Chrome

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

Latest?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Additional information

![image](https://user-images.githubusercontent.com/748263/156798218-e7e2a599-4779-4549-b54e-12c278c4f26b.png)
![image](https://user-images.githubusercontent.com/748263/156798239-59b3a343-c60a-4a09-98a1-a06de0551220.png)

matthewgallo:
Thanks for opening this issue @mitchellbernstein, it will be fixed with https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1745

PR #1745, opened 2022-03-04
fix(UserProfileImage): fix tooltip version of user profile image

matthewgallo:
Contributes to #1743 

Fixes the hover/focus state of the tooltip version of the UserProfileImage component.

#### What did you change?
```
packages/cloud-cognitive/src/__tests__/__snapshots__/styles.test.js.snap
packages/cloud-cognitive/src/components/UserProfileImage/UserProfileImage.js
packages/cloud-cognitive/src/components/UserProfileImage/_user-profile-image.scss
```
Storybook


elycheea (approved):
Want to double check that `--container` is meant to be the modifier here — otherwise, 👍

matthewgallo:
Yes, you're right @elycheea, the `--container` is kind of unnecessary. I'll update and remove it to just be `.${blockClass}__tooltip`

Issue #1736, opened 2022-03-04
PageHeader should prevent Breadcrumb scrolling off by default


lee-chase:
- [ ] Deprecate disableBreadcrumbScroll property
- [ ] Add enableBreadcrumbScroll property with reverse logic to that of disableBreadcrumbScroll

PR #1748, opened 2022-03-07
feat(PageHeader): prevent breadcrumb scroll by default

lee-chase:
Contributes to #1736 

Keeps breadcrumb on screen by default, reversing logic (and name) of property.

#### What did you change?
- deprecated disableBreadcrumbScroll
- added enableBreadcrumbScroll
- Fixed scroll height of navigation when breadcrumb scroll was enabled and scrolled off screen.

Storybook


packages/cloud-cognitive/src/components/PageHeader/PageHeader.stories.js
```diff
@@ -815,7 +815,7 @@ export const demo = prepareStory(TemplateDemo, {
     navigation: 3,
     tags: 3,
     actionBarItems: 4,
-    disableBreadcrumbScroll: true,
+    enableBreadcrumbScroll: false,
```
dcwarwick, line 818: We don't need to provide a default when the default is already falsy.

```suggestion
```

packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```diff
@@ -779,6 +783,14 @@ const TYPES = {
 const tagTypes = Object.keys(TYPES);
 
 export const deprecatedProps = {
+  /**
+   * Standard behavior scrolls the breadcrumb off to leave just tabs. This
+   * option preserves vertical space for both the breadcrumb and tabs if they're supplied.
```
dcwarwick, line 788:
```suggestion
   * **Deprecated** see property `enableBreadcrumbScroll`
```

packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```diff
@@ -103,14 +105,16 @@ export let PageHeader = React.forwardRef(
     // state based on props only
     const hasActionBar = actionBarItems && actionBarItems.length > 0;
     const hasBreadcrumbRow = !!breadcrumbs || !!actionBarItems;
+    const enableBreadcrumbScroll =
+      in_enableBreadcrumbScroll ?? deprecated_disableBreadcrumbScroll ?? false;
```
dcwarwick, line 109: Shouldn't this be:

```suggestion
      in_enableBreadcrumbScroll ?? !deprecated_disableBreadcrumbScroll ?? false;
```

Which also suggests we should have a test for the deprecated prop while it continues to exist, as we do for the new prop?

packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```diff
@@ -12,6 +12,7 @@ import cx from 'classnames';
 import { useResizeDetector } from 'react-resize-detector';
 
 import { Grid, Column, Row, Button, Tag } from 'carbon-components-react';
+import { breakpoints } from '@carbon/layout/es';
```
dcwarwick, line 15:
```suggestion
import { breakpoints } from '@carbon/layout';
```

Was there a reason for specifically loading from /es? The problem is that jest wants the /lib while browser builds typically want the /es, so explicitly coding one of the other won't work in both places. That's why the tests are failing. There ought to be a top-level export that the package.json ensures is loaded in the form that's required so that we don't need the /es, but I've not tested it!

dcwarwick (changes requested):

packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
lee-chase: Yes, in error.

dcwarwick (approved):
LGTM

Issue #1711, opened 2022-03-01
Create a tool that produces a report on what components a codebase is using


dcwarwick:
The goal is to be able to produce a report on what components are being used by a product or project, suitable for presenting at DUX review or similar. The components listed in the report should be the components from `carbon-components-react` and `@carbon/ibm-products` and other similar packages (CDAI? security? cloud pal? etc) that are being imported by one or more source files in or below the directory in which the tool is being run.

A key piece of this tool will be the code that scans source files and compiles the list of components from various packages being imported. There may be useful dependencies we can build this with, or we may need to scan and match the source files ourselves using some heuristics. One CLI invocation of the tool will then be to use this code to scan for components and then format the result into a neat report of some kind.

This can be built first as a proof of concept, for testing on a few codebases to check if it is going to be useful.

Possible abilities to add/todo:
 - [ ] Strip out pkg or restrict based on a list of components
 - [ ] Target Carbon, Security, CloudPAL
 - [ ] Note if linter is used
 - [ ] Versions of packages
 - [ ] Provide context of component usage, omponents up to pages etc.

lee-chase:
A straw man has been published, test and announce.

lee-chase:
Meetings arranged for April 8th with @joshblack @rodet 

Placing on hold until then.

PR #1763, opened 2022-03-09
feat: make cli out of component usage report

lee-chase:
Contributes to #1711

Make work via install as per @babel/cli

#### What did you change?

Had a bit of a re-org and created a command that could be run directly in package.json

Ran using updated command.


package.json
```diff
@@ -65,6 +65,7 @@
   "devDependencies": {
     "@babel/core": "^7.17.5",
     "@babel/eslint-parser": "^7.17.0",
+    "@carbon/component-usage-report": "0.1.0",
```
matthewgallo, line 68: I think we should possibly rename this to `@carbon/ibm-products-component-usage-report`. It is a bit longer, but I can see there being confusion with the current package name, as there is no mention that it checks for `@carbon/ibm-products` components.

lee-chase: It could at some point add in Carbon components, perhaps even allow for others in the same way DevTools does.

matthewgallo (approved):

Issue #1752, opened 2022-03-08
add column filter and sort to add select


davidmenendez:
add select with multi select hierarchy support filter and sorts on individual columns.

![Screenshot 2022-03-08 at 08-39-38 Add select Pattern Preview Mode - InVision](https://user-images.githubusercontent.com/6370760/157260117-402cfc5c-3fc1-446b-b1b7-f8026a6c1af5.png)
![Screenshot 2022-03-08 at 08-39-49 Add select Pattern Preview Mode - InVision](https://user-images.githubusercontent.com/6370760/157260126-ff414019-5c2c-4d9e-b198-b48033323e51.png)

PR #1765, opened 2022-03-09
fix: adds sort and filter functionality to add select columns

davidmenendez:
Contributes to #1752 

adds the filter and sort functionality to multi select with hierarchy columns.

Screen Shot 2022-03-09 at 3 37 07 PM
Screen Shot 2022-03-09 at 3 37 13 PM
Screen Shot 2022-03-09 at 3 37 20 PM

accomplishing this required a refactor of the items schema because previously there was no way (as far as i could tell) to support dynamically loading in sort and filter options. now the schema looks more like this

```
{
   "items":{
      "sortBy":[
         "title"
      ],
      "entries":[
         {
            "id":"1",
            "value":"folder 1",
            "title":"folder 1",
            "children":{
               "sortBy":[
                  "title",
                  "size"
               ],
               "filterBy":"fileType",
               "entries":[
                  {
                     "id":"1-1",
                     "value":"file1.pdf",
                     "title":"file1.pdf",
                     "fileType":"pdf",
                     "size":"100"
                  },
                  {
                     "id":"1-2",
                     "value":"index.js",
                     "title":"index.js",
                     "fileType":"js",
                     "size":"200"
                  },
                  {
                     "id":"1-3",
                     "value":"sitemap.xml",
                     "title":"sitemap.xml",
                     "fileType":"xml",
                     "size":"10"
                  }
               ]
            }
         },
         {
            "id":"2",
            "value":"folder 2",
            "title":"folder 2"
         }
      ]
   }
}
```

most notably is the change from `items` and `children` being just an array of items to being an object that contains additional props for `sortBy` and `filterBy`. now the items will live in an `entries` prop.

the `soryBy` prop represents an array of existing props that will be used as options for the sort dropdown. `sortBy: ['title', 'size'],` in this example `title` and `size` will be used to sort the items. note- there can be multiple attributes to sort by, which is why the value should be an array.

similarly, `filterBy` represents the prop that will be used to filter items. `filterBy: 'fileType',` in this example `fileType` will be used to filter items. as per the spec you can only use a single attribute to filter by, which is why this is not an array.

when there are no values for either `sortBy` or `filterBy` the filter and sort dropdown menus should not be present.


matthewgallo:
Seems that one of your tests is failing, maybe because of the change in the `items` schema that you mentioned.

davidmenendez:
@matthewgallo you're correct. thank you. fixed.

packages/cloud-cognitive/src/components/AddSelect/AddSelect.js
```diff
@@ -303,4 +323,10 @@ AddSelect.propTypes = {
   title: PropTypes.string,
 };
 
+AddSelect.defaultProps = {
```
matthewgallo, line 326: Should we stick to using a `defaults` object? How come you added back `defaultProps`?

matthewgallo (approved):
Just had a question about defaultProps vs a defaults object which we've switched most components to use now

packages/cloud-cognitive/src/components/AddSelect/AddSelect.js
davidmenendez: i didn't realize that was the new standard. i can update that going forward

Issue #1656, opened 2022-02-17
Release Review for OptionsTile


janhassel:
#1655

matthewgallo:
## Review for release

### Readiness

- [ ] One or more scenarios for a design pattern have been identified as a useful unit of functionality to publish.
- [ ] A functioning component or components delivering those scenarios have been delivered and merged to main.
- [ ] Design maintainer has approved the implementation for those scenarios (via a comment on the relevant issue/epic).

### Engineering review

- [ ] Breaking changes have only been introduced with prior approval, discussion and documented in release notes. Ideally deprecation notices in the prior major version must have been added in a timely fashion.
- [ ] The implementation takes into account, and does not impair, remaining and anticipated design scenarios.
- [ ] Public component features (names, props, etc) are consistent with Carbon-defined conventions and are consistent internally. Where there isn't an existing convention to apply, ensure robust precedents are being established.
- [ ] The UI produced is accessible, responsive, translatable, cross-browser, and responds to the currently set Carbon theme.
- [ ] Components are functional components using hooks.
- [ ] Public components which produce DOM structures support className.
- [ ] Public components support a ref (via React.forwardRef).
- [ ] Public component supports a [Devtools attribute](https://github.com/carbon-design-system/ibm-cloud-cognitive/search?l=JavaScript&q=devtools&type=)
- [ ] All significant DOM elements have meaningful classes.
- [ ] Additional attributes that are not identified as props (such as title, aria-*, etc) are passed through to an appropriate DOM node of the component as HTML attributes.
- [ ] No warnings, errors or log messages in the console.
- [ ] Each public component JS is exported in /src/components/index.js, each public component SCSS is included in /src/components/_index.scss, and each public component has a flag in package-settings.js.
- [ ] Each public component SCSS lists all of the Carbon and C&CS components imported and used by the JavaScript code and explicitly imports the SCSS for each of these components.

### Standards

- [ ] No linter warnings or errors are produced.
- [ ] Carbon tokens (theme, layout, motion) are used unless the design specifies otherwise.
- [ ] All components utilizing motion must include reduced-motion queries for accessibility purposes - [read more here](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion).
- [ ] Code is formatted according to prettier rules (no use of //prettier-ignore).
[...]

PR #1767, opened 2022-03-10
test(options-tile): extend automated test coverage

janhassel:
Contributes to #1656

#### What did you change?
- Add new test: "it hides the summary when props.enabled = false"
- Enhance existing tests
  - "it renders open state when passed"
    - renamed to "it can be controlled by setting props.open"
    - test both cases: set to expanded when collapsed and set to collapsed when expanded
  - "it expands and collapses on click"
    - also test collapsing on click again

Automated tests

#### Additional notes
@matthewgallo The remaining "untested" parts are solely for the animation. I would assume we don't necessarily want test this to avoid additional delays in the testing suite when waiting for animations to be complete. What do you think?

### DIFF
--- a/packages/cloud-cognitive/src/components/OptionsTile/OptionsTile.test.js
+++ b/packages/cloud-cognitive/src/components/OptionsTile/OptionsTile.test.js
@@ -157,7 +157,7 @@ describe(componentName, () => {
     expect(summary.getAttribute('aria-hidden')).toBe('true');
   });
 
-  it('it can be controlled by setting props.open', () => {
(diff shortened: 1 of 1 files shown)


lee-chase (approved):

matthewgallo:
I think that all of the props have test coverage it should be ok, but react testing library does have an event that allows you to test animations within your components.

`fireEvent.animationStart(elementThatHasTheAnimation)`
`fireEvent.animationEnd(elementThatHasTheAnimation)`

PR #1778, opened 2022-03-14
chore: skip a11y test for summary card container

lee-chase:
Skip a11y test for summary card container

#### What did you change?

Added a test skip

Ran test.


matthewgallo (approved):
I think this is ok until we can address this issue separately

Issue #1770, opened 2022-03-10
ButtonMenu doesn't support ghost kind


talhaham:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description
I need to use `ghost` button menu. Based on the [documentation]( https://pages.github.ibm.com/cdai-design/pal/components/button-menu/#variations) ButtonMenu should support `ghost` and `tertiary` variation. 
However, setting kind to ghost has no affect, and looking on the code, it seems to use hard coded ${carbon.prefix}--btn--primary 

> Is this issue related to a specific component?
Yes, ButtonMenu

> What did you expect to happen? What happened instead? What would you like to
> see changed?

I expect that if I add a ButtonMenu with `kind` set to `ghost` I will get ghost button menu, but I get a primary button.

> What browser are you working in?
Chrome.

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

 @carbon/ibm-products 1.8

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

Working on IBM Orchestrate

PR #1779, opened 2022-03-15
fix(ButtonMenu): doesn't support ghost kind

Ratheeshrajan:
Contributes to #1770 

ButtonMenu doesn't support ghost & tertiary kind

#### What did you change?
- Button kind class name was hardcoded as `${carbon.prefix}--btn--primary` and it's replaced with `${carbon.prefix}--btn--${size}` to accept kind props.
- Along with this, all the three kinds of button menus(primary, tertiary and ghost) are displayed in StoryBook for reference.
Storybook.


packages/cloud-cognitive/src/components/ButtonMenu/ButtonMenu.js
```diff
@@ -116,6 +118,11 @@ ButtonMenu.propTypes = {
    */
   iconDescription: Button.propTypes.iconDescription,
 
+  /**
+   * The three types the menu button supports: primary, tertiary and ghost.
+   */
+  kind: Button.propTypes.kind,
```
elycheea, line 124:
```suggestion
  kind: PropTypes.oneOf(['primary', 'tertiary', 'ghost']),
,
```

elycheea (changes requested):
Rather than showing the three kinds of button menus in Storybook, I think it’s better to use the add them as options in Storybook controls.

image

Since we’ve hardcoded the `kind` in Storybook, the examples break when you try to modify `kind` in the controls tab. We may want to show just one button menu but allow changing of the prop. I think we have examples in SidePanel for setting the size or placement where we have known options.

packages/cloud-cognitive/src/components/ButtonMenu/ButtonMenu.js
Ratheeshrajan: @elycheea Many thanks for reviewing this. Proposed changes were implemented.

elycheea (commented):
Looks good but we’re still running into some lint errors in CI.
Should be fixed after a `yarn format` though.

Ratheeshrajan:
@elycheea fixed the lint failure.

lee-chase (approved):

elycheea (approved):

Issue #1680, opened 2022-02-21
Compact header columns in DataSpreadsheet is misaligned with rest of column


elycheea:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.
When DataSpreadsheet `cellSize` is `compact`, the header rows are shorter in width than the cells of the column. This seems to be related to scrollbars being set to on in macOS. When scrollbars are set to “when scrolling” instead of “always,” this issue no longer appears.

![image](https://user-images.githubusercontent.com/1717398/155025663-e7f40c22-7d4b-4afe-9a41-009c5254d199.png)

> Is this issue related to a specific component?
DataSpreadsheet

> What browser are you working in?
Chrome on macOS

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?
1.7.0

## Steps to reproduce the issue

0. On macOS, go to System Preferences and set scrollbars to always show
1. Go to DataSpreadsheet in Storybook
2. Set `cellSize` to `compact`

## Additional information

May be able to resolve it with the CSS property for `scrollbar-gutter` ... 👀 

Update: Adding `scrollbar-gutter` seems to fix the alignment, but causes a gap in the focus state/container instead.

[Scrollbar gutter](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) on MDN

PR #1780, opened 2022-03-15
fix(DataSpreadsheet): address header scrollbar issue

matthewgallo:
Contributes to #1680 

Currently if the spreadsheet does not contain a scrollbar, there are alignment issues with the header cells. To maintain consistency, I used the value from `scrollBarSize` since the DataSpreadsheetBody uses this to calculate overall width.

#### What did you change?
`DataSpreadsheet.js`
`DataSpreadsheetHeader.js`
Storybook


lee-chase (changes requested):
How do I verify what has improved? Things don't seem to line up when there is a scroll bar.

On initial load I see this in basic spreadsheet
image

If I go to compact things line up
image

matthewgallo:
Talked with @lee-chase, going to go with a different approach for the scrollbars where the scroll bar gets it's own pseudo column/row that is fixed

lee-chase (approved):
Gets a bit upset if I switch my Mac settings post-render, but I'm prepared to let that edge case slide 👍

Issue #1825, opened 2022-03-22
add select modifier


davidmenendez:
another feature for add select is the modifier functionality

davidmenendez:
![add-select-modifier](https://user-images.githubusercontent.com/6370760/159545582-44e0f203-632e-4a36-9251-a68b8b88f4fc.png)

PR #1826, opened 2022-03-22
chore: adds modifier to add select

davidmenendez:
Contributes to #1825 

starts the implementation for modifiers in add select. i'm still not 100% sure about how i'm going hook up the functionality, but this starts by adding the required html elements


davidmenendez:
Screen Shot 2022-03-22 at 1 19 27 PM

davidmenendez:
added an update to fix the checkbox centering

Screen Shot 2022-03-22 at 2 39 45 PM

Issue #1759, opened 2022-03-08
DataSpreadsheet: edit mode cell editor `font-size` incorrect


matthewgallo:
## What will this achieve?

The edit mode cell editor should have the same `font-size` as the rest of the cells in the spreadsheet. Currently, the cell editor receives a slightly larger `font-size`.

## How will success be measured?
When the cell editor and the cells all have the same `font-size`.

## Additional information

- Designs
- Existing code
- etc

PR #1831, opened 2022-03-24
fix(DataSpreadsheet): use type token for cells/cell editor/active cell

matthewgallo:
Contributes to #1759 

This PR adds the `body-short-01` type token to all spreadsheet cells, the active cell, and the cell editor (text area). Previously, the text area font size was slightly larger than the font size of the body cells.

I also updated the calculation of the paddingTop value of the cellEditor, it was off by 1px. Now, going into edit mode, the contents of the cell being edited does not move or jump anymore.

#### What did you change?
```
```
Storybook


lee-chase (approved):
LGTM

Issue #1846, opened 2022-03-28
DataSpreadsheet: empty spreadsheet


matthewgallo:
## What will this achieve?

A user should be able to create an empty spreadsheet and enter content in via cell editing.

## How will success be measured?

When there is a spreadsheet without any previous data, a user can render one that is empty from which data can be added.

## Additional information

- Designs
- Existing code
- etc

PR #1848, opened 2022-03-29
feat(DataSpreadsheet): support empty spreadsheets

matthewgallo:
Contributes to #1846 

Includes the support for empty spreadsheets that can be edited, with the addition of a new prop, `defaultEmptyRowCount`, that will set the number of empty rows the spreadsheet will create. I also added a story for this variation of the component.

#### What did you change?
```
packages/cloud-cognitive/src/components/DataSpreadsheet/DataSpreadsheet.stories.js
packages/cloud-cognitive/src/components/DataSpreadsheet/DataSpreadsheet.test.js
```
Storybook and ran tests


packages/cloud-cognitive/src/components/DataSpreadsheet/DataSpreadsheetBody.js
```diff
@@ -338,63 +341,88 @@ export const DataSpreadsheetBody = forwardRef(
...
+          });
+          onDataUpdate(emptyRowData);
+        };
+        buildEmptyRows();
+      }
+    }, [rows, headerGroups, defaultEmptyRowCount, onDataUpdate]);
+
+    const RenderEmptyRows = () => {
+      return <div />;
+    };
```
matthewgallo, line 365: This was only necessary because the `<FixedSizeList>` component requires that something always be rendered. However, as soon as we know that `rows` doesn't have any data, we create the empty rows via the useEffect above on line [346](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1848/files#diff-44f993b0394c37dc8c3ca1aa905880d04545f5991688d57afb30d9524f7da45dR346) which are the visible empty rows.

Issue #1884, opened 2022-04-04
DataSpreadsheet: a11y updates


matthewgallo:
## What will this achieve?

The cells in the spreadsheet need some refactoring to following best accessibility guidance on grid components following [WAI aria guidelines](https://www.w3.org/TR/wai-aria-practices/#grid).

## How will success be measured?

Cells/header cells/rows/etc all receive the appropriate aria labels and roles according to WAI aria grid guidelines.

## Additional information

- Designs
- Existing code
- etc

matthewgallo:
Complete

PR #1892, opened 2022-04-07
feat(DataSpreadsheet): a11y updates and keyboard shortcuts

matthewgallo:
Contributes to #1884 

Refactored to address accessibility issues, mainly wrapped button cells inside of a div with the role of `gridcell` (because interactive elements can't have a `role="gridcell"` or `role="cell"`).

I also refactored how multiple key presses were being tracked. As I was working on some of the keyboard shortcuts I realized that what was currently there was not correct and not reliably tracking keys. I created a custom hook, `useMultipleKeyTracking.js`, that correctly returns an array of keys being pressed together.

Keyboard shortcuts added in this PR:
`CMD+A`: Select all cells
`CTRL+SPACE`: Select current column
`SHIFT+SPACE`: Select current row
`HOME`: Move active cell to first cell in current row
`END`: Move active cell to last cell in current row

#### What did you change?
```
cspell.json
packages/cloud-cognitive/src/components/DataSpreadsheet/DataSpreadsheetHeader.js
packages/cloud-cognitive/src/components/DataSpreadsheet/hooks/index.js
packages/cloud-cognitive/src/components/DataSpreadsheet/hooks/useMultipleKeyTracking.js
packages/cloud-cognitive/src/components/DataSpreadsheet/hooks/useResetSpreadsheetFocus.js
packages/cloud-cognitive/src/components/DataSpreadsheet/hooks/useSpreadsheetOutsideClick.js
packages/cloud-cognitive/src/components/DataSpreadsheet/utils/handleMultipleKeys.js
```
Storybook/tests


elycheea:
Hoping someone else will get to a more thorough review soon, but I ran into some issues with `CTRL + SPACE`. It wasn’t working for me at all, but I do also use that shortcut for changing Input Sources (to other languages on macOS). Even when I turned off input sources, I wasn’t able to get `CTRL + SPACE` working.

If I hit `CTRL + SPACE`, I’m unable to get keyboard actions (including arrows) until after a `CMD + A` (maybe because it includes an `event.preventDefault()`??) If I turn off input modes, `CTRL + SPACE` doesn’t work ... but I don’t lose control over arrow keys.

If this is a non-issue for others, then my my suspicion is that it’s related to [`CompositionEvent`](https://developer.mozilla.org/en-US/docs/Web/API/CompositionEvent) — in which case, it might be better to open and assign a separate issue to look into it.

matthewgallo:
Hmm, was the active cell in a row header or cell header position? If so, I had noticed issues in that case. I fixed that in a separate PR (just didn’t want to make this PR any larger).

elycheea:
No, active cell was within the cell body. 😞

lee-chase:
Works for me, although based on google sheets I would expect slightly different behavioir for both Ctrl+Space and Shift+Space, where the current selection is extended to the entire column/row.

If multiple adjacent cells are selected (different columns) Ctrl+Space should select multiple columns.
If multiple adjacent cells are selected (different rows) Shift+Space should select multiple rows.

elycheea:
Update: Works after I turn off the Input Sources and then do a whole browser restart.
So maybe it’s related to there being a duplicate shortcut at all rather than IME input.

image

Since Google Sheets uses the same commands for the shortcut, I don’t think we should necessarily change the shortcut even though I think the macOS defaults include this in for Input Sources. 🤔  Losing keyboard / arrow control after running into the duplicated shortcut is probably still a bug though. In Google Sheets, the shortcut fails but I still have arrow key control so hopefully that’s easier to resolve.

matthewgallo:
@elycheea I _think_ the loss of keyboard/arrow controls after attempting to use the shortcut could be from the browser keyup event being cancelled if you use a system level keyboard shortcut. We can follow up next week and see 🤞

matthewgallo:
@lee-chase if it's ok, can I address the interaction you're describing in a separate PR?

lee-chase:
Yes @matthewgallo

elycheea:
> @elycheea I _think_ the loss of keyboard/arrow controls after attempting to use the shortcut could be from the browser keyup event being cancelled if you use a system level keyboard shortcut. We can follow up next week and see 🤞 

Yeah, we can look into it more next week. Didn't get a chance to look into it more, but don’t want this to block the PR if everything else is good to go.

Issue #1893, opened 2022-04-07
Add Select global filter


davidmenendez:
![Screenshot 2022-04-07 at 16-34-04 Add   select Pattern Preview Mode - InVision](https://user-images.githubusercontent.com/6370760/162323775-f35b7add-2d23-4b5a-949d-0bc1ec96e736.png)

Add Select should support a global filter. Filter options should be generated based on the item properties that a user passes through.

PR #1894, opened 2022-04-07
fix: add select global filter

davidmenendez:
Contributes to #1893 

Screen Shot 2022-04-07 at 4 41 15 PM

adds global filter support. now a user can specify the properties to filter the list of items and those options should be generated based on the possible values.

```
globalFilters: [
      {
        id: 'fileType', // the property to look for in the entries
        label: 'File type', // label for the dropdown
      },
];
```
when `globalFilters` is detected an array of values will be generated `[{ id: 'fileType', opts: ['pdf', 'js', 'xml']}]` and then `opts` is used as the values in the dropdown.


packages/cloud-cognitive/src/components/AddSelect/AddSelectFilter.js
```diff
@@ -0,0 +1,138 @@
...
+                  selectedItem={getSelectedItem(filterOpts.id)}
+                  label={placeholder}
+                />
+              ))}
+            </div>
+          </div>
+          <ButtonSet className={`${blockClass}-button-set`}>
+            <Button
+              kind="secondary"
+              onClick={resetFilters}
```
matthewgallo, line 105: The dropdown filters were reset when I tried this out, but the main content in the component did not get reset and I couldn't find a way to get the main content back.

matthewgallo, line 105: The dropdown filters were reset when I tried this out, but the main content in the component did not get reset and I couldn't find a way to get the main content back.

packages/cloud-cognitive/src/components/AddSelect/AddSelectFilter.js
```diff
@@ -0,0 +1,138 @@
...
+          renderIcon={Filter16}
+          hasIconOnly
+          kind="ghost"
+          onClick={() => setOpen(!open)}
+          iconDescription={iconDescription}
+          className={`${blockClass}-toggle`}
+          size="md"
+        />
+      )}
+      {open && (
```
matthewgallo, line 83: I think you should add the `useClickOutside` hook that we have here. I feel like the expected behavior here is that if you click outside the filter panel it should close.

matthewgallo, line 83: I think you should add the `useClickOutside` hook that we have here. I feel like the expected behavior here is that if you click outside the filter panel it should close.

packages/cloud-cognitive/src/components/AddSelect/AddSelectFilter.js
davidmenendez: i interpreted the reset functionality to act as like a only a reset for the filters. so you still have to click apply to apply the filters. but that's a good point. i'll double check with design on the UX 👍

packages/cloud-cognitive/src/components/AddSelect/AddSelectFilter.js
davidmenendez: i thought about that. i'll double check with design on that to verify that's the desired UX 👍

packages/cloud-cognitive/src/components/AddSelect/AddSelectFilter.js
matthewgallo: Ohh I see what you mean, ok.

packages/cloud-cognitive/src/components/AddSelect/AddSelectFilter.js
matthewgallo: Sounds good

matthewgallo (approved):
Since the things I commented on are UX related, I think we're good to approve 🎉

PR #1898, opened 2022-04-08
feat(DataSpreadsheet): #1884 spreadsheet more shortcuts

matthewgallo:
Contributes to #1884 

**_https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1892 should be reviewed first_**

Adds two more keyboard shortcuts to the spreadsheet:
`CMD+HOME`: Moves active cell to first cell in first row
`CMD+END`: Moves active cell to last cell in last row

I also fixed some other shortcuts that were broken if used when the active cell was either positioned in a column or row header.

#### What did you change?
```
packages/cloud-cognitive/src/components/DataSpreadsheet/DataSpreadsheet.js
packages/cloud-cognitive/src/components/DataSpreadsheet/utils/handleMultipleKeys.js
```
#### How did you test and verify your work?
Storybook

PR #1900, opened 2022-04-11
fix(DataSpreadsheet): updates to keys pressed custom hook

matthewgallo:
Contributes to @lee-chase's comment in https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/1898

When using the command key, keyup events from other keys are lost so I made some updates to the `useMultipleKeyTracking` hook so that all other keys are cleared out. This fixes the problem @lee-chase noticed in the review of #1898 

#### What did you change?
`useMultipleKeyTracking.js`
Storybook


lee-chase (approved):
LGTM

PR #1905, opened 2022-04-12
feat: add portal target prop to tearsheet shell

### LINKED ISSUE #1456: Tearsheet breaks horizontal & vertical scroll and is out of position if component is lower down in the DOM

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)

## Detailed description

The `Add task`  tearsheet is being rendered from a component that isn’t near the top level parent and because of this it seems to be breaking the horizontal and vertical scroll on the page when I open the `Task` sidepanel because the tearsheet is hidden, and the tearsheet is also shifted out of position when it is visible. The `Edit global task` components parent is the top level parent so it renders correctly. This behaviour with the `Add task` tearsheet didn’t happen with the ibm security tearsheet so could this be an issue with the cloud and cognitive tearsheet itself?

Yes

> see changed?
Expected tearsheet to render correctly regardless of where the component is added in the DOM

Chrome

> package are you using?
@carbon/ibm-cloud-cognitive v0.83.1

> should be aware of?
SOAR Playbook Designer

## Additional information
Attached screen recording of behaviour. The `Add task` tearsheet is lower in the DOM than the `Edit global task` tearsheet. This seems to cause the `Add task` tearsheet to not render correctly. 

https://user-images.githubusercontent.com/25666525/142445421-d818dcac-b320-4837-9d56-c067dc0fb4d5.mp4

lee-chase:
Closes #1456

Makes document.body the default location for the tearsheet modal, adds portalTarget to redirect to a specific dom node.

#### What did you change?

- Added portalTarget as a property to TearsheetShell
- Added story updates for TearsheetShell, Tearsheet and TearsheetNarrow

Storybook.


davidmenendez (approved):
LGTM. looks like a test is failing though.

lee-chase:
Fixed the test @davidmenendez

matthewgallo (approved):

PR #1919, opened 2022-04-13
fix(AboutModal): scrolling content fade visible

### LINKED ISSUE #1494: AboutModal with scrolling content doesn't have the scroll fade-out
See https://carbon-for-ibm-products.netlify.app/?path=/story/carbon-for-ibm-products-patterns-about-modal-aboutmodal--fully-loaded

image

lizisawizard:
Closes #1494 

The About Modal's fade at the bottom of its content was not appearing so I added some CSS to fix it so it does

#### What did you change?
`__about-modal.scss`

Storybook

PS I forgot to change my branch's name my apologies T-T


elycheea (approved):
Looks good to me!

Interestingly, the overflow-indicator from Carbon uses rgb of the theme variable at 0% alpha (CSS vars break in rgba function though), but `rgba(0, 0, 0, 0)` seems to work just as well across each theme. 👀

Issue #1937, opened 2022-04-20
AddSelect global filter tags


davidmenendez:
The global filter needs to support showing tags for active filters

![Screenshot 2022-04-20 at 15-59-57 Add   select Pattern Preview Mode - InVision](https://user-images.githubusercontent.com/6370760/164322061-786b9209-73d2-4386-88ae-8478176ef266.png)

PR #1938, opened 2022-04-20
fix: adds global filter tags to add select

davidmenendez:
Contributes to #1937 

adds global filter tags and fixed a visual but where the tearsheet wasn't properly applying some styles


packages/cloud-cognitive/src/components/AddSelect/AddSelectFilter.js
```diff
@@ -54,77 +65,118 @@ export let AddSelectFilter = ({
...
+              onClose={() => removeTag(filterType)}
+              filter
             >
-              {primaryButtonText}
-            </Button>
-          </ButtonSet>
+              {`${filterType}: ${appliedFilters[filterType]}`}
+            </Tag>
+          ))}
+          <span
```
matthewgallo, line 158: I think would be best to use a ghost or tertiary button here. Also with the way you have `onKeyDown` currently, all the filters are cleared out when you press any key. I'm not sure that's the intended behavior.

matthewgallo (approved):

Issue #1926, opened 2022-04-18
Build errors with node 18


jlibbyibm:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

Upon updating to node 18 the following build issue occurs. Investigation implies it's due to the react-dnd package which is under the carbon/ibm-products requires list. React-dnd v16.0.0 claims to be fixed to be compatible with node 18 so the hope is that this dependency can be updated to v16 without major issue.
https://github.com/react-dnd/react-dnd/releases/tag/v16.0.0

`Module not found: Package path ./jsx-runtime.js is not exported from package /Users/jeremylibby/Workspace/ui2/ui-management/node_modules/react (see exports field in /Users/jeremylibby/Workspace/ui2/ui-management/node_modules/react/package.json)

Import trace for requested module:
./node_modules/react-dnd/dist/esm/core/index.mjs
./node_modules/react-dnd/dist/esm/index.mjs
./node_modules/@carbon/ibm-products/es/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
./node_modules/@carbon/ibm-products/es/components/Datagrid/Datagrid/addons/CustomizeColumns/CustomizeColumnsModal.js
./node_modules/@carbon/ibm-products/es/components/Datagrid/Datagrid/addons/CustomizeColumns/ModalWrapper.js
./node_modules/@carbon/ibm-products/es/components/Datagrid/Datagrid/addons/CustomizeColumns/index.js
./node_modules/@carbon/ibm-products/es/components/Datagrid/useCustomizeColumns.js
./node_modules/@carbon/ibm-products/es/components/Datagrid/index.js
./node_modules/@carbon/ibm-products/es/components/index.js
./node_modules/@carbon/ibm-products/es/index.js
./src/pages/lsm/Permission/Permission.js`

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

v.1.14.0

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

Would likely be present at least any time a page or component is built on node 18 and v1.14.0 of ibm-products

matthewgallo:
[...]

PR #1939, opened 2022-04-21
build: upgrade react dnd to work with node 18

matthewgallo:
Contributes to #1926 

Addresses the build errors noted in #1926, stemming from `react-dnd`. An update was made to both `react-dnd` and `react-dnd-html5-backend` requiring us to add these dependencies to `transformIgnorePatterns` inside our jest config. This seems to a common problem with people using `react-dnd` v16.x.x and jest together (https://github.com/react-dnd/react-dnd/issues/1396) with react versions 16 or earlier.

With this change to the jest config I was able to get the tests working again with the new version.

#### What did you change?
Upgraded `react-dnd` and `react-dnd-html5-backend`, as well as updates to jest config
Ran tests with new package versions


elycheea (approved):
Lgtm

PR #1942, opened 2022-04-22
docs: update docu for security components and patterns

amtoussam:
Contributes to #

{{short description}}


packages/security/src/components/TrendingCard/index.js
```diff
@@ -9,7 +9,7 @@ import React, { Fragment } from 'react';
 
 import { getComponentNamespace } from '../../globals/namespace';
 
-const namespace = getComponentNamespace('trending-card');
+const namespace = getComponentNamespace('trending-card#legacy');
```
elycheea, line 12: Changing the namespace looks like it also affects (breaks) the styles in Storybook here.

| Current | Expected |
---------|-----------|
image | image

elycheea (changes requested):
The changes to namespace in TrendingCard breaks the styles in Storybook since it affects the classnames too.

Also @cameroncalder mentioned we wanted to include an inline notification ([like this](https://carbon-for-ibm-products.netlify.app/?path=/story/security-layouts-detail--default)) with a link to the replacement component. Will that be addressed in another PR?

packages/security/src/components/TrendingCard/index.js
amtoussam: @elycheea Right! I have updated the PR with the labels on the page and also updated the Trending Card component so the styles don't break. There are three components that I have not been able to update due to the nature of the component but will continue to work on it. I do not think I want to hold up the PR for those three components and will create another. But let me know if you think differently.

packages/security/src/components/Card/Card-story.js
```diff
@@ -43,6 +45,27 @@ const props = () => ({
 const searchLabelText = 'Search';
 
 storiesOf(components('Card#legacy'), module)
+  .addDecorator((Story) => (
+    <>
+      <InlineNotification
+        className="page-layouts__banner"
+        actions={
+          <NotificationActionButton
+            href="https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-cards-productivecard--default&knob-Lines%20(`lines`)=1&knob-Score%20(`score`)=3&knob-Title%20(`title`)=10.0.0.0&knob-Truncate%20(`truncate`)=true&knob-Type%20(`type`)=IP&knob-Value%20(`value`)=10.0.0.0.563409u53250u&knob-Width%20(`width`)=100px&knob-type=IP&knob-value=127.0.0.1"
```
elycheea, line 54:
```suggestion
            href="https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-cards-productivecard--default"
```

We can probably drop the URL to just the default page without these knobs since our Storybook doesn’t use knobs?

packages/security/src/components/Header/Header-story.js
```diff
@@ -22,12 +22,35 @@ import {
...
+
 const headerProps = {
   labels,
   links,
 };
 
 disableCenteredStories(storiesOf(patterns('Header#legacy'), module))
+  .addDecorator((Story) => (
+    <>
+      <InlineNotification
```
elycheea, line 35: Looks like the Header is actually covering the inline notification for all the header stories.

packages/security/src/components/PanelV2/PanelV2-story.js
```diff
@@ -64,6 +66,27 @@ const content = (
 );
 
 disableCenteredStories(storiesOf(patterns('PanelV2#legacy'), module))
+  .addDecorator((Story) => (
+    <>
+      <InlineNotification
```
elycheea, line 71: Another instances where it looks like the header in Storybook is covering the actual inline notification for the new component.

packages/security/src/components/ProfileImage/ProfileImage-story.js
```diff
@@ -13,15 +13,38 @@ import { components } from '../../../.storybook';
...
-  />
-));
+storiesOf(components('ProfileImage#legacy'), module)
+  .addDecorator((Story) => (
+    <>
+      <InlineNotification
+        className="page-layouts__banner"
+        actions={
+          <NotificationActionButton
+            href="https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-patterns-user-profile-images-userprofileimage--default"
```
elycheea, line 30:
```suggestion
            href="https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-patterns-user-profile-images-userprofileimage--default"
```

Can we change this link to the story instead of docs?

packages/security/src/components/Shell/Shell-story.js
```diff
@@ -24,13 +24,36 @@ import {
...
   toolbar,
 };
 
 disableCenteredStories(storiesOf(patterns('Shell#legacy'), module))
+  .addDecorator((Story) => (
+    <>
+      <InlineNotification
+        className="page-layouts__banner"
+        actions={
+          <NotificationActionButton
```
elycheea, line 41: This one’s also getting hidden by the Shell in Storybook.

packages/security/src/components/StatusIcon/StatusIcon-story.js
```diff
@@ -23,6 +24,27 @@ const storyProps = () => ({
 const status = STATUS[0];
 
 storiesOf(components('StatusIcon#legacy'), module)
+  .addDecorator((Story) => (
+    <>
+      <InlineNotification
+        className="page-layouts__banner"
+        actions={
+          <NotificationActionButton
+            href="https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-components-cards-productivecard--default"
```
elycheea, line 33: Can we double check this link? I think this one should probably follow Nav and just say it’s now legacy because I don’t see an updated link in the Box note @cameroncalder shared. In which case, we can remove the `NotificationActionButton` altogether.

packages/security/src/components/StatusIndicator/StatusIndicator-story.js
```diff
@@ -63,6 +65,27 @@ const { statusIndicator: statusIndicatorProps, statusSteps: statusStepsProps } =
 const { basic, retry: retryProps } = statusIndicatorProps;
 
 storiesOf(patterns('StatusIndicator#legacy'), module)
+  .addDecorator((Story) => (
+    <>
+      <InlineNotification
+        className="page-layouts__banner"
+        actions={
+          <NotificationActionButton
```
elycheea, line 73: Since this one doesn’t include a new home, I think we can remove the `NotificationActionButton`.

packages/security/src/components/TrendingCard/TrendingCard-story.js
```diff
@@ -19,6 +20,29 @@ export default {
...
     ...getDocsParameters(),
   },
+  decorators: [
+    (story) => (
+      <>
+        <InlineNotification
+          className="page-layouts__banner"
+          actions={
+            <NotificationActionButton
+              href="https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-components-cards-productivecard--default"
```
elycheea, line 30:
```suggestion
              href="https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-cards-productivecard--default"
```

packages/security/src/components/Wizard/Wizard-story.js
```diff
@@ -238,6 +240,27 @@ const markdown = (useDefault = true) =>
   `;
 
 storiesOf(patterns('Wizard#legacy'), module)
+  .addDecorator((Story) => (
+    <>
+      <InlineNotification
```
elycheea, line 245: Another one where the InlineNotification is hidden by other elements in the story.

elycheea (commented):
Builds are working now, and the TrendingCard looks right now!

Unfortunately, with the `InlineNotification`s, there’s a few stories where other elements (Shells/Headers) cover the actual notification and therefore the link to the updated component is hidden.

I also made a few notes where we’re directing to the docs page instead of story. I think it’s better to keep those consistent so left notes on those pages. Finally, a few of the legacy components don’t have actual replacements yet so suggest we take the `NotificationActionButton`s out in those stories.

elycheea (approved):

PR #1951, opened 2022-04-27
1914 reduce motion web terminal

AlexanderMelox:
Contributes to #1914 

Adds reduced motion to web terminal and adds a wrapper component for content that sits besides the web terminal


matthewgallo (changes requested):
It looks like the tests are failing because if `WebTerminalContentWrapper` is going to be a component made available to end users, you need to add it to `package-settings.js` and add the canary check within `WebTerminalContentWrapper` like so:
```js
WebTerminalContentWrapper = pkg.checkComponentEnabled(WebTerminalContentWrapper, componentName);
```
The reduced motion part works great though!

matthewgallo (approved):

PR #1963, opened 2022-05-04
fix: Feature/roll back legacy tags

amtoussam:
Contributes to #

{{short description}}

#### What did you change?
There were some components that are not yet legacy, so I have remove the tag and banner from those components. I also made the banner a bit wider and changed the copy in the banners.

I ran the work on my local storybook to verify linking as well as placement.


packages/security/src/components/Header/Header-story.js
```diff
@@ -40,15 +44,15 @@ disableCenteredStories(storiesOf(patterns('Header#legacy'), module))
...
-            More info
+            View replacement
           </NotificationActionButton>
         }
         kind="info"
-        subtitle="This component is now legacy. Please click to see the go-forward component"
+        subtitle="Pattern no longer supported. The pattern will remain available, but plan to migrate to the pattern replacement."
         title=""
         hideCloseButton
+        style={{ display: 'block' }}
```
elycheea, line 54: Are we intentionally breaking this one out into a separate line?
image

elycheea (commented):
Legacy tags now match the bottom section of the Box note except that `DataDecorator` doesn’t link to `Decorator` as a replacement. Other differences were clarified in Slack (type layout and combo button).

elycheea (approved):

PR #1966, opened 2022-05-04
docs: Features/longshorejeff/security package readme note

### LINKED ISSUE #1855: Update CODEOWNERS for Security package
The CODEOWNERS file still lists dbrugger who no longer has write access.
We should update this to @amtoussam and @jlongshore.

jlongshore:
Updated the readme within the Security package to include a note about updating

Closes #1855


CODEOWNERS
```diff
@@ -1,3 +1,4 @@
 * @carbon-design-system/ibm-cloud-cognitive-development
 
-/packages/security/ @dbrugger
+# Carbon for IBM Products > Security package
+/packages/security/ Jeff-Longshore  am-toussaint
```
elycheea, line 4: I think it should actually be this.

```suggestion
/packages/security/ @jlongshore @amtoussam
```

elycheea (changes requested):

CODEOWNERS
```diff
@@ -1,3 +1,5 @@
+# cSpell:ignore amtoussam, jlongshore
```
elycheea, line 1: So I know I suggested this originally but now that I’m looking at it again, I wonder if `disable-next-line` or `disable-line` would make more sense so that if Security team codeowners ever need to change, we only need to modify one line instead of in two places.

elycheea (approved):

Issue #1930, opened 2022-04-19
Data table: Implement table empty states


jjennings7:
# Data table extension: Empty states

## Brief description

WKC data table does not include Carbon for IBM Products PAL empty state.

## Details
See Carbon for IBM Products PAL site guidance for empty states here:
https://pages.github.ibm.com/cdai-design/pal/patterns/empty-state/usage

### Links to other materials
Carbon empty state guidelines:
https://pages.github.ibm.com/cdai-design/pal/patterns/empty-state/usage

Carbon for IBM Products empty states in Storybook:
https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-patterns-empty-state-emptystate--default

## Design Maintainer(s)

- Marion Brülls
- Johanna Koval

## Tasks

Before starting work on this epic, please review and complete the following.

- [ ] Initial review of design/existing code

### Working in Carbon for IBM Products package

- [ ] Have you reviewed our
      [contribution guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/CONTRIBUTING.md)?
- [ ] Have you noted our
      [code guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/docs/CODE_GUIDELINES.md)?
- [ ] Did you use the CLI to scaffold your new component?

matthewgallo:
I believe this is already something that we support with the current status of the Datagrid component. See [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/main/packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridEmptyBody.js). It's possible there may not be a story in storybook for this.

PR #1969, opened 2022-05-05
fix: data table empty state

Ratheeshrajan:
Contributes to #1930 

The data table to include Carbon for IBM Products PAL empty state

#### What did you change?

- Updated the `DatagridEmptyBody` component to include `EmptyState'
- New story for empty state

Storyboo.


matthewgallo (approved):
🎉

Issue #1931, opened 2022-04-19
Data table: Implement data table header


jjennings7:
# Data table extension: Table header

## Brief description

WKC data table does not include Carbon for IBM Products PAL table headers.

## Details
See Carbon for IBM Products PAL site guidance for data table headers here:
https://pages.github.ibm.com/cdai-design/pal/components/data-table/headers/usage

### Links to other materials
See also Carbon data table guidelines for data table title and description:
https://www.carbondesignsystem.com/components/data-table/usage/#formatting

## Design Maintainer(s)

- Marion Brülls
- Johanna Koval

## Tasks

Before starting work on this epic, please review and complete the following.

- [ ] Initial review of design/existing code

### Working in Carbon for IBM Products package

- [ ] Have you reviewed our
      [contribution guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/CONTRIBUTING.md)?
- [ ] Have you noted our
      [code guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/docs/CODE_GUIDELINES.md)?
- [ ] Did you use the CLI to scaffold your new component?

PR #1973, opened 2022-05-06
feat(Datatable): table header

Ratheeshrajan:
Contributes to #1931 

WKC data table does not include Carbon for IBM Products PAL table headers.

#### What did you change?

- `title` and `description` props(supplied via `datgridState`) added in `<TableContainer />` component. 
- `<DatagridToolbar />` moved inside `<TableContainer />`
- Fix overflow scroll of table container

Storybook


Ratheeshrajan:
@matthewgallo Thanks for spotting the issue in the left panel. I have fixed that one in the last commit and now it looks like the following screenshot. 
Screenshot 2022-05-13 at 3 30 40 PM

I have moved the left panel inside `TableContainer` to fix that issue.

Ratheeshrajan:
> Hey @Ratheeshrajan, the fix for the left panel story/placement of the toolbar looks great. But my only reservation now is that there seems to be double scrollbars on some cases which is something we should fix.
> 
> This screenshot is from the `Batch actions` story: image

@matthewgallo fixed the double scrollbar issue. Please have a look.

packages/cloud-cognitive/src/components/Datagrid/_storybook-styles.scss
```diff
@@ -33,3 +33,7 @@ $block-class: #{$pkg-prefix}--datagrid;
     border-bottom: none;
   }
 }
+
+.#{$block-class}__hidden--columns.#{$block-class}__datagridWrap-simple {
+  height: calc(100% - 80px);
```
elycheea, line 38: Can we clarify what this calc is for? Also we typically use `spacing` or `rem` instead of pixel values.

packages/cloud-cognitive/src/components/Datagrid/styles/datagrid.scss
```diff
@@ -264,6 +258,10 @@
   height: calc(100% + 40px);
 }
 
+.#{$block-class}__grid-container.#{$block-class}__with-pagination {
+  height: calc(100% - 40px);
```
elycheea, line 269: Again, think it’d be good to include a comment, but also we should use spacing tokens or rem here.

elycheea (commented):
For the values we’re hardcoding in calc, I think it’d be good to clarify why it’s necessary. I know much of the datagrid is already using pixels, but we should use tokens or rems for any of the changes we are making.

No need to go and update all pixel values though, we can address those in a larger release review.

elycheea:
Also just noticed that the data table header now blends into the table. Might have been a conflict when you merged from `main`?

image

packages/cloud-cognitive/src/components/Datagrid/_storybook-styles.scss
Ratheeshrajan: This is only for storybook display. There is information listed under the table about the hidden column. I'll convert the px values into rem.
Screenshot 2022-06-24 at 3 29 22 PM

packages/cloud-cognitive/src/components/Datagrid/styles/datagrid.scss
Ratheeshrajan: Ths calc is for creating room for pagination under the table body. I'll include comments and make it rem value.

davidmenendez:
@elycheea can you please review @Ratheeshrajan response?

elycheea:
@davidmenendez We discussed via Slack. Ratheesh will add the dense header variant and will follow up with design on clarification since the styles are coming from the Carbon default.

Ratheeshrajan:
@elycheea @matthewgallo For dense header, I added one data table header variant. I'm also waiting for Marion's response on the grey10 header background.

packages/cloud-cognitive/src/components/Datagrid/Datagrid/Datagrid.js
```diff
@@ -41,6 +41,9 @@ export let Datagrid = React.forwardRef(({ datagridState, ...rest }, ref) => {
     verticalAlign = 'center',
     variableRowHeight,
     className,
+    gridTitle,
+    gridDescription,
+    DatagridDenseActions,
```
elycheea, line 46: Since the dense header is more of a variant, it may make more sense to create the variant as a boolean prop rather than another set of DatagridActions.

elycheea (changes requested):
Visually, I think the changes look good.

Upon closer inspection and some discussion with Matt, I think rather than using `DatagridDenseActions`, we should use a boolean prop `dense` or `useDenseHeader` (something that matches our or Carbon’s existing conventions). From there, we can modify the markup as needed to achieve the dense header styles while using the same `DatagridActions`.

This should also be easier for the consumer to switch between dense and non-dense header styles. Thanks for your patience, Ratheesh, and I hope this makes sense.

packages/cloud-cognitive/src/components/Datagrid/Datagrid/Datagrid.js
Ratheeshrajan: @elycheea I'll make the changes requested and get back.

Ratheeshrajan: @elycheea the consumer can now configure the dense header by setting the property in 'datagridState' as follows,

`const datagridState = useDatagrid({
    columns,
    data,
    gridTitle,
    gridDescription,
    initialState: {
      pageSize: 10,
      pageSizes: [5, 10, 25, 50],
    },
    useDenseHeader: true,
    DatagridActions,
    DatagridPagination,
  });`

elycheea (dismissed):

matthewgallo:
Discussed with both @elycheea and @Ratheeshrajan over slack but there seems to be some layout issues with the pagination appearing at the bottom of the page in the deploy preview. After that is fixed, I think we'll be ready to merge this one!

elycheea (approved):
Changes look good to me but @matthewgallo may need to take another look so he can approve or dismiss his previous review.

Thanks for coordinating and communicating with us all @Ratheeshrajan! It’s not easy to do with all our timezone differences!

matthewgallo (approved):

Issue #1978, opened 2022-05-10
[A11Y] Security header profile button has an accessibility violation


jowen:
## What package(s) are you using?

- [ ] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [x] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

When running the IBM Equal Access Accessibility Checker on the "Default shell component" [storybook](http://localhost:3000/?path=/story/security-patterns-shell-legacy--default) there is an accessibility violation:

Violation: Accessible name does not match or contain the visible label text [Learn more](chrome-extension://lkcagbfjnkomcinoddgooolagloogehp/devtoolsSubpanel.html#)

> Is this issue related to a specific component?

yes, the profile button in the Shell header

> What did you expect to happen? What happened instead? What would you like to
> see changed?

There should be no accessibility violations

> What browser are you working in?

Chrome Version 101.0.4951.54

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

@carbon/ibm-security@2.11.9

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM Guardium Inisghts

## Steps to reproduce the issue

1. Go to the storybook: https://carbon-for-ibm-products.netlify.app/?path=/story/security-patterns-shell-legacy--default
2. Hide the side bar to avoid noise (too many accessibility violations on it)
3. Run the IBM Equal Access Accessibility Checker (Chrome/Firefox extension) with rule set "Latest Deployment" / accessibility guidelines "IBM Accessibility"
There is now an accessibility violation

## Additional information

- Screenshots or code
![image](https://user-images.githubusercontent.com/802501/167622218-13d545f9-195f-445b-9d33-d5c487bfe7fe.png)

<img width="482" alt="image" src="https://user-images.githubusercontent.com/802501/167621901-46939d8a-c39d-442a-87c9-640ebd857a47.png">

- Notes
See also https://github.com/carbon-design-system/ibm-security/issues/1141
https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/1789

matthewgallo:
Thanks for the contribution here @jowen! This will be included in our weekly release this Tuesday

PR #1979, opened 2022-05-10
fix: a11y issue profile-button (#1978)

jowen:
Contributes to #1978

Fix for accessibility issue with the profile button in the Security Header Shell.

#### What did you change?
Switch from using aria-label to using aria-describedby + accessibility text.  aria-label requires the aria-label matches the button text.  In this case we want the assistive text to be different than the username/initials used in the button.

- Ran IBM accessibility checker and verified no accessibility issues.
- Verified unit tests passed.


jlongshore (approved):
LGTM

Issue #1982, opened 2022-05-11
Add Select meta panel


davidmenendez:
add select needs to support an optional meta panel feature for item entires with meta data information

![Screenshot 2022-05-11 at 15-49-19 Add   select Pattern Preview Mode - InVision](https://user-images.githubusercontent.com/6370760/167945347-24ba4da5-d041-443d-900e-0f157cf67b0b.png)
![Screenshot 2022-05-11 at 15-50-50 Add   select Pattern Preview Mode - InVision](https://user-images.githubusercontent.com/6370760/167945353-109dcc8d-e2c4-4b4e-b35f-1a6f071b1501.png)

PR #1983, opened 2022-05-11
fix: add select meta panel

davidmenendez:
Contributes to #1982

Screen Shot 2022-05-11 at 3 54 12 PM

when an item has a `meta` attribute and you hover over it in the list you should now see an icon that when clicked will open up the metal data panel.

### DIFF
--- a/packages/cloud-cognitive/src/components/AddSelect/AddSelectList.js
+++ b/packages/cloud-cognitive/src/components/AddSelect/AddSelectList.js
@@ -7,6 +7,7 @@
 
 import React from 'react';
 import {
+  Button,
   Checkbox,
   RadioButton,
   StructuredListRow,
@@ -24,6 +25,7 @@ const componentName = 'AddSelectList';
 
 export let AddSelectList = ({
   filteredItems,
+  metaIconDescription,
   modifiers,
   multi,
   multiSelection,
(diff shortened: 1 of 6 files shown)


packages/cloud-cognitive/src/components/AddSelect/AddSelectList.js
```diff
@@ -166,6 +167,11 @@ export let AddSelectList = ({
                   {item.children && (
                     <ChevronRight16 onClick={() => onNavigateItem(item)} />
                   )}
+                  {item.meta && (
+                    <div className={`${blockClass}-hidden-hover`}>
+                      <View16 onClick={() => setDisplayMetaPanel(item)} />
```
matthewgallo, line 172: This should ideally be an icon only button to support keyboard interactions. Something like 
```js
<Button
  renderIcon={View16}
  iconDescription={item.meta.title} // not sure if that's correct or not but something to pass to iconDescription since it is required for iconOnly buttons
  tooltipPosition="bottom"
  tooltipAlignment="center"
  hasIconOnly
  onClick={() => setDisplayMetaPanel(item)}
/>
```

packages/cloud-cognitive/src/components/AddSelect/AddSelectMetaPanel.js
```diff
@@ -0,0 +1,46 @@
...
+  const blockClass = `${pkg.prefix}--add-select__meta-panel`;
+  const onCloseHandler = () => {
+    setDisplayMetaPanel({});
+  };
+
+  return (
+    <div className={blockClass}>
+      <div className={`${blockClass}-header`}>
+        <p className={`${blockClass}-title`}>{title}</p>
+        <Close16 className={`${blockClass}-close`} onClick={onCloseHandler} />
```
matthewgallo, line 17: This should also be an icon only button
```js
<Button
  renderIcon={Close16}
  iconDescription={closeDescriptionStringProp}
  tooltipPosition="bottom"
  tooltipAlignment="center"
  hasIconOnly
  onClick={onCloseHandler}
/>
```

matthewgallo (changes requested):
Just had some suggestions around adding icon only buttons

davidmenendez:
@matthewgallo good call! fixed

matthewgallo (approved):
🎉

PR #1988, opened 2022-05-13
fix: converts chevron in add select to button

davidmenendez:
this just converts an icon with an onclick to be a button as well as a few minor css changes to accompany the change to the html.


matthewgallo (approved):
💯

PR #2005, opened 2022-05-24
fix: add select global filter styles

davidmenendez:
small update to the add select global filter that adds a missing box shadow and now when you click apply the filter should close. also added a small adjustment to only enable the filter apply and reset buttons when any of the filter dropdowns have changed.


matthewgallo (approved):

matthewgallo (commented):
I don't think this was caused from this PR, but I was looking at the preview deploy and after selecting one item from the list, when trying to select a second, it looks like something is breaking. Following the link provided, it looks like you're accidentally rendering an object

```
Objects are not valid as a React child (found: object with keys {id, title, value}). If you meant to render a collection of children, use an array instead.
```

image

davidmenendez:
weird. thanks for catching that

Issue #2016, opened 2022-06-06
Card (productive) component caption isn't showing in storybook


johanna-koval:
## Bug description
Card component with caption variation isn't displaying the text bellow the title
<img width="800" alt="CleanShot 2022-06-02 at 08 23 18@2x" src="https://media.github.ibm.com/user/232597/files/47680000-e24d-11ec-85ec-9c377edcf24d">

## URL/Link of Page
Link here: https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-cards-productivecard--with-caption&args=caption:caption+jfsjnnfkjnsdjf;actionsPlacement:top;clickZone:one;titleSize:default

## Expected behavior
Caption should display when selected, example:

<img width="644" alt="CleanShot 2022-06-02 at 08 22 37@2x" src="https://media.github.ibm.com/user/232597/files/2ef7e580-e24d-11ec-9aed-e77e24965dc4">

PR #2026, opened 2022-06-09
chore(caption in productive card): caption fixed

jagathgj:
Contributes to #2016 

Productive Card caption/description visibility issue fixed.

image

#### What did you change?
prop changed from `caption` to `description` since `Card` component accepts `description` prop only as caption/description

storybook


elycheea (approved):
Ahh, I vaguely recall this being changed during [a release review](https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/1086#issuecomment-896275840). We decided to stick with the name `description` instead of `caption` but must have missed this in the story.

Issue #2001, opened 2022-05-20
Inconsistency narrow tearsheet design guidance and storybook component 


marion-bruells:
## Inconsistencies PAL guidance and storybook component narrow tearsheet
The narrow tearsheet title block shows a line and different spacings in the guidance than it is in the storybook.

## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description
The narrow tearsheet title block shows a line and different spacings in the guidance than it is in the storybook.

PAL guidance website and Sketch/Figma Kit:
https://pages.github.ibm.com/cdai-design/pal/components/tearsheet/style
<img width="465" alt="image" src="https://user-images.githubusercontent.com/51152537/169545152-f4bbb089-dda0-4352-8719-5a0af1872c08.png">

Coded implementation, storybook:
https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-tearsheet-tearsheetnarrow--tearsheet-narrow
<img width="506" alt="image" src="https://user-images.githubusercontent.com/51152537/169545104-d86642a0-6d1a-478a-92f3-a5aa4dfa061c.png">

> Is this issue related to a specific component?
Narrow tearsheet
https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-tearsheet-tearsheetnarrow--tearsheet-narrow

> What did you expect to happen? What happened instead? What would you like to
> see changed?
Adjust spacings, add line below title block to align with PAL guidelines.
https://pages.github.ibm.com/cdai-design/pal/components/tearsheet/style

> What browser are you working in?
Firefox, Safari

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
Watson Knowledge Catalog

elycheea:
@marion-bruells Wanted to check the spacings that needs to be adjusted.
I noticed when checking Figma that our spacing between the title and description is a bit taller than it should be.

Storybook | Figma
-----------|-------
![image](https://user-images.githubusercontent.com/1717398/173086711-29eb3869-42bf-4b5e-bcfe-d002feb7c626.png) | ![image](https://user-images.githubusercontent.com/1717398/173086777-2cd7f5cd-3ebb-476f-948e-cf52800436e4.png)
16px / `$spacing-05` | 8px / `$spacing-03`

[...]

PR #2027, opened 2022-06-09
chore(narrow tearsheet header): border bottom fix

jagathgj:
Contributes to #2001 

Inconsistencies PAL guidance and storybook component narrow tearsheet

image

#### What did you change?
Provided border bottom to the header of narrow tearsheet

storybook


elycheea (changes requested):
I believe we also need to decrease the spacing of the `.#{$block-class}--tearsheet__header-description` from 16px to 8px — so `$spacing-05` to `$spacing-03`.

I’m double checking with Marion on an inconsistency I’m seeing in Sketch and Figma but that should be all for now.

Also since you’re making changes to styles, you’ll want to update snapshots once you update the padding. You can do this with the command `yarn test:c4p styles -u` for now (@davidmenendez adds this in package.json in #2025).

jagathgj:
@elycheea  padding updated. Kindly verify

image

elycheea (approved):

Issue #1967, opened 2022-05-04
PageHeader on scroll shows breadcrumbs when using Tabs.


trumanbutter:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)

## Detailed description

> Describe in detail the issue you're having.
The PageHeader component with breadcrumbs sits 8px below the UIShell header when scrolling (screenshot attached below). 

> Is this issue related to a specific component?
`PageHeader`

> What browser are you working in?
Chrome Version 101.0.4951.54 (Official Build) (x86_64)

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?
"@carbon/ibm-products": "1.13.0"

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
Product: IBM Security Verify

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/funny-minsky-sfjohw?file=/src/Example/SamplePage/index.js

## Additional information
Screenshot with offset issue:

![image](https://user-images.githubusercontent.com/32148874/166836933-cd6e5a6d-1fa8-43e6-9cfa-a768aef4b1bd.png)

jagathgj:
@trumanbutter the prop `disableBreadcrumbScroll={false}` for pageHeader is deprecated and we are using `enableBreadcrumbScroll={false}` for making the breadcrumb sticky. I think, this may provide solution to the issue. Kindly check and let me know.

jagathgj:
@trumanbutter yes there is a miscalculation happening in header height value while enabling/disabling the prop `enableBreadcrumbScroll={false}`, which causes an 8px difference.

PR #2053, opened 2022-06-21
chore(page header): header top value fix

jagathgj:
Contributes to #1967 

breadcrumb not scrolling fully while e`nableBreadcrumbScroll={true}`

#### What did you change?

headerTopValue is changed if navigation and breadcrumb is available or not.

storybook


elycheea (commented):
From Storybook, it looks like this fixes the extra spacing in many cases but it looks like it also causes the tabs to get cut off in a few stories. At a glance, it appears to be affecting the stories with `actionBarItems`.

image

jagathgj:
@elycheea I have modified the prior logic. kindly review it and let me know your comments.

elycheea (approved):
Looks good!

Issue #1925, opened 2022-04-15
Data table: Add column alignment option


johanna-koval:
## **Data table: Column alignment**

Related epic: https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/1607

### Description
The Current WKC Data table [column alignment ](https://pages.github.ibm.com/wdp-gov/wkc-shared-ui/?path=/story/datagrid--right-aligned-columns) follows the alignment parameters for numerical and textual data. However, it does not include the option to have a center alignment for multi-part data and icon-only content.

[Link to Design](https://pages.github.ibm.com/cdai-design/pal/components/data-table/column-alignment/usage)

Multi-part center alignment
<img width="962" alt="CleanShot 2022-04-15 at 16 12 38@2x" src="https://user-images.githubusercontent.com/5809654/163627588-6a8467bc-93c3-491f-97ce-118118c8e1e9.png">

Icon-only center alignment
<img width="1000" alt="CleanShot 2022-04-15 at 16 13 32@2x" src="https://user-images.githubusercontent.com/5809654/163627653-e675ee23-0165-4870-b84f-024eae2570c4.png">

- [ ] Add center alignment option

PR #2056, opened 2022-06-22
feat: added center aligned columns component and centered new columns

faraznaseem019:
Contributes to #1925

{{short description}}
For this issue, a new Datagrid component was added, that allows users to center-align the content in their columns (headers and table data). The other Datagrid components have columns that are right-aligned or left-aligned. However, the 'Center Aligned Columns' is a perfect example of displaying center-aligned columns, because the last 2 columns are center-aligned.

Screenshot 2022-06-22 at 2 23 17 PM

Screenshot 2022-06-22 at 2 23 20 PM

#### What did you change?
I created a new Datagrid component, where the last 2 columns contain content that's center-aligned. I did this by creating a react hook called, useColumnCenterAlign.js, and I added CSS styling to a SCSS file I created called, useColumnCenterAlign.scss.

I tested and verified my work by ensuring that the columns that had the centerAlignedColumn flag set to True, had the table content displayed in the center of the table.
Screenshot 2022-06-22 at 2 43 36 PM

Screenshot 2022-06-22 at 2 43 44 PM

Screenshot 2022-06-22 at 2 44 03 PM

Screenshot 2022-06-22 at 2 44 10 PM


packages/cloud-cognitive/src/components/Datagrid/Datagrid.stories.js
```diff
@@ -36,6 +36,9 @@ import {
   useCustomizeColumns,
   useSelectAllWithToggle,
 } from '.';
+
+import useColumnCenterAlign from './useColumnCenterAlign';
```
matthewgallo, line 40: You should import this hook with all the other hooks above (not from it's own import).

packages/cloud-cognitive/src/components/Datagrid/styles/useColumnCenterAlign.scss
```diff
@@ -0,0 +1,23 @@
...
+*/
+
+@import './variables';
+
+.#{$block-class}__center-align-header {
+  width: 100%;
+  text-align: center;
+}
+
+.#{$block-class}__center-align-header button {
```
matthewgallo, line 16: If possible, I would use a class to target the button element instead of just applying this style to all buttons inside of the `.#{$block-class}__center-align-header` class

packages/cloud-cognitive/src/components/Datagrid/styles/useColumnRightAlign.scss
```diff
@@ -25,5 +25,6 @@
 }
 
 .#{$block-class}__right-align-cell-renderer.sortDisabled {
-  padding-right: 0;
+  /* stylelint-disable-next-line carbon/layout-token-use */
+  padding-right: 12px;
```
matthewgallo, line 29: I don't think there should be any changes to this file at all. The right aligned columns was already working, right?

matthewgallo (changes requested):
One last comment, with the creation of the new hook `useColumnCenterAlign`, you need to export it so it will be available to use. This is done in `Datagrid/index.js`, you'll see all of the other hooks provided by this component, just add `useColumnCenterAlign`.

Great job overall @faraznaseem019! Just a few things to tidy up

packages/cloud-cognitive/src/components/Datagrid/styles/useColumnCenterAlign.scss
faraznaseem019: It seems like I don't need this line. So, I'll just remove it from the stylesheet.

packages/cloud-cognitive/src/components/Datagrid/styles/useColumnRightAlign.scss
faraznaseem019: Removed the lines that I added, and added the single, 'padding-right: 0;' line.

packages/cloud-cognitive/src/components/Datagrid/Datagrid.stories.js
```diff
@@ -35,9 +35,10 @@ import {
   useDisableSelectRows,
   useCustomizeColumns,
   useSelectAllWithToggle,
+  useColumnCenterAlign,
 } from '.';
 
-import useColumnCenterAlign from './useColumnCenterAlign';
+// import useColumnCenterAlign from './useColumnCenterAlign';
```
matthewgallo, line 41: Can you remove this line entirely (now that it is imported with the other hooks above)?

matthewgallo (approved):
Great work @faraznaseem019, the new `useColumnCenterAlign` hook is exported and storybook example looks good. 🎉

Issue #1981, opened 2022-05-10
[Customize columns] Content issues in Data table column customization data table extension


Hayley-l-williams:
## Content issues and considerations for the Customize columns data table extension

Pattern: https://pages.github.ibm.com/cdai-design/pal/components/data-table/column-customization/usage/

**Content bug:**

In the instructions, it reads: "Click and drag the white box to reorder the columns."

This is confusing because there is no white box. **Consider updating the instructions on click & drag to reflect the visuals of the UI.**

**Considerations:**

In the instructions, it reads: "Deselect columns to hide them."
**Consider adjusting the instructions to include the opposite: "Select columns to display them," either by replacing the original sentence or as an alternative.**

While this can make sense if the majority of columns are selected and displayed by default, this is not intuitive if a minority are selected. For example, where we want to use this in Data Virtualization, we want four columns displayed on default, with the option to display another seven. I would think for any user accessing the Customizing columns, the task for them is choosing more columns to display, not choosing which of the few already displayed to hide.

------

Also, I don't know if this is something that needs to be considered in the pattern instructions but I will write anyway: the term "Manage columns" is misleading in context of data virtualization because the concept of managing columns exists, though carrying a different meaning. 
Currently in the pattern, the header in the modal is called "Manage columns." The hover state over the icon that invokes this feature also uses "Manage columns."
We are implementing this pattern in Data virtualization / Watson Query but will need to use "Customize columns" to prevent confusion, despite being inconsistent with the pattern.

<img width="931" alt="Screen Shot 2022-05-10 at 12 07 49 PM" src="https://user-images.githubusercontent.com/102319049/167703834-bf834f6d-91f1-4704-9106-5193ec18ecf6.png">

elycheea:
Looks like this has already been resolved on the PAL.

elycheea:
Reopening after discussion with @Hayley-l-williams. Hayley is opening an issue in the PAL repo, but we should also update our example to be close to the examples in the PAL guidance.

Regardless of the changes to the PAL site, we should remove the “click and drag the white box” example content from the Storybook example.

Comparing again against the PAL guidelines, I’ll open another issue for column customization @kai-liu. Looks like we’re still using the older styles from WKC.

elycheea:
Reopening after discussion with @Hayley-l-williams. Hayley is opening an issue in the PAL repo, but we should also update our example to be close to the examples in the PAL guidance.

[...]

PR #2062, opened 2022-06-23

kai-liu:
Contributes to #1613, #1981

#### What did you change?
within the `Datagrid` directory:

`DraggableElement.js` Added `selected` attribute to help with css selectors
`addons/CustomizeColumns/Actions.js` Removed search button, as required by design. (the button has only been commented out, in case future design changes requires the button again)
`CustomizeColumns/CustomizeColumnsModal.js` Added logic to show number of visible columns vs total number of columns
`/styles/addons/CustomizeColumnsModal.scss` Style changes

storybook

### DIFF
--- a/packages/cloud-cognitive/src/components/Datagrid/Datagrid.stories.js
+++ b/packages/cloud-cognitive/src/components/Datagrid/Datagrid.stories.js
@@ -7,13 +7,12 @@
  */
 
 import React, { useState, useEffect } from 'react';
-// TODO: import action to handle events if required.
-// import { action } from '@storybook/addon-actions';
 import { useColumnOrder } from 'react-table';
 import { range, makeData, newPersonWithTwoLines } from './utils/makeData';
 
 import { getStoryTitle } from '../../global/js/utils/story-helper';
 
+import { action } from '@storybook/addon-actions';
 import {
   Activity16,
   Restart16,
@@ -58,10 +57,6 @@ import { SidePanel } from '../SidePanel';
 export default {
   title: getStoryTitle(Datagrid.displayName),
   component: Datagrid,
-  // TODO: Define argTypes for props not represented by standard JS types.
-  // argTypes: {
-  //   egProp: { control: 'color' },
-  // },
   parameters: {
     styles,
     docs: {
@@ -678,6 +673,7 @@ const DatagridActions = (datagridState) => {
       bottom: '-37px',
     },
   };
+
   return (
     isNothingSelected &&
     (useDenseHeader && useDenseHeader ? (
@@ -811,7 +807,7 @@ export const SelectItemsInAllPages = () => {
     <>
       <Datagrid datagridState={{ ...datagridState }} />
       <h3>Doc in Notes...</h3>
-      <p>{`Are all selected across all pages? - ${areAllSelected}`}</p>
+      <p>{`Are all entries selected across all pages? - ${areAllSelected}`}</p>
     </>
   );
 };
@@ -940,6 +936,7 @@ const DatagridBatchActions = (datagridState) => {
   const totalSelected = selectedFlatRows && selectedFlatRows.length;
   const onBatchAction = () => alert('Batch action');
   const actionName = 'Action';
+
   return (
     <TableBatchActions
       shouldShowBatchActions={totalSelected > 0}
@@ -953,6 +950,44 @@ const DatagridBatchActions = (datagridState) => {
   );
 };
 
+const getBatchActions = () => {
+  return [
+    {
+      label: 'Duplicate',
+      renderIcon: Add16,
+      onClick: action('Clicked batch action button'),
+    },
+    {
+      label: 'Add',
+      renderIcon: Add16,
+      onClick: action('Clicked batch action button'),
+    },
+    {
+      label: 'Select all',
+      renderIcon: Add16,
+      onClick: action('Clicked batch action button'),
+      type: 'select_all',
+    },
+    {
+      label: 'Publish to catalog',
+      renderIcon: Add16,
+      onClick: action('Clicked batch action button'),
+    },
+    {
+      label: 'Download',
+      renderIcon: Add16,
+      onClick: action('Clicked batch action button'),
+    },
+    {
+      label: 'Delete',
+      renderIcon: Add16,
+      onClick: action('Clicked batch action button'),
+      hasDivider: true,
+      kind: 'danger',
+    },
+  ];
+};
+
 export const BatchActions = () => {
   const columns = React.useMemo(() => defaultHeader, []);
   const [data] = useState(makeData(10));
@@ -960,10 +995,13 @@ export const BatchActions = () => {
     {
       columns,
       data,
+      batchActions: true,
+      toolbarBatchActions: getBatchActions(),
       DatagridActions,
       DatagridBatchActions,
     },
-    useSelectRows
+    useSelectRows,
+    useSelectAllWithToggle
   );
 
   return <Datagrid datagridState={{ ...datagridState }} />;
@@ -987,8 +1025,6 @@ export const DisableSelectRow = () => {
 
   return <Datagrid datagridState={{ ...datagridState }} />;
 };
-
-// export { StickyActionsColumn };
 
 NestedRows.story = {
   parameters: {
--- a/packages/cloud-cognitive/src/components/Datagrid/Datagrid.test.js
+++ b/packages/cloud-cognitive/src/components/Datagrid/Datagrid.test.js
@@ -879,6 +879,16 @@ describe(componentName, () => {
     jest.spyOn(global.console, 'warn').mockImplementation(() => {});
     jest.useFakeTimers();
     jest.spyOn(global, 'setTimeout');
+    window.ResizeObserver = jest.fn().mockImplementation(() => ({
+      observe: jest.fn(),
+      unobserve: jest.fn(),
+      disconnect: jest.fn(),
+    }));
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+    window.ResizeObserver = ResizeObserver;
   });
 
   it('renders a basic data grid component with devTools attribute', () => {
--- a/packages/cloud-cognitive/src/components/Datagrid/Datagrid/Datagrid.js
+++ b/packages/cloud-cognitive/src/components/Datagrid/Datagrid/Datagrid.js
@@ -29,6 +29,7 @@ export let Datagrid = React.forwardRef(({ datagridState, ...rest }, ref) => {
     );
     return null;
   }
+
   const {
     getTableProps = () => {},
     withVirtualScroll,
--- a/packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridToolbar.js
+++ b/packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridToolbar.js
@@ -1,29 +1,173 @@
-/*
- * Licensed Materials - Property of IBM
- * 5724-Q36
- * (c) Copyright IBM Corp. 2020
- * US Government Users Restricted Rights - Use, duplication or disclosure
- * restricted by GSA ADP Schedule Contract with IBM Corp.
+/**
+ * Copyright IBM Corp. 2022, 2022
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
  */
-import React from 'react';
-import { DataTable } from 'carbon-components-react';
-import { pkg } from '../../../settings';
+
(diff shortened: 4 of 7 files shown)


packages/cloud-cognitive/src/components/Datagrid/Datagrid/DraggableElement.js
```diff
@@ -12,6 +12,7 @@ import { Draggable16 } from '@carbon/icons-react';
 import { useDrag, useDrop } from 'react-dnd';
 import cx from 'classnames';
 import { pkg } from '../../../settings';
+// import { TableSelectRow } from 'carbon-components-react';
```
markkulube, line 15: if the  `TableSelectRow` no longer used this line should be deleted.

(the less comments and less commented out code, the easier it is to parse the module)

packages/cloud-cognitive/src/components/Datagrid/Datagrid/DraggableElement.js
```diff
@@ -129,10 +132,10 @@ const DraggableElement = ({
...
         }
       }}
-      onBlur={(e) => {
-        // handle when focus move to inner elements
-        setIsFocusedOnItem(e.currentTarget === e.target);
-      }}
+      // onBlur={(e) => {
+      //   // handle when focus move to inner elements
+      //   //setIsFocusedOnItem(e.currentTarget === e.target);
+      // }}
```
markkulube, line 138: If this handler is no longer used it should removed.

(the less comments and less commented out code, the easier it is to parse the module)

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Actions.js
```diff
@@ -7,24 +7,25 @@
...
 const blockClass = `${pkg.prefix}--datagrid`;
 
 const Actions = ({
   searchText,
-  columns,
-  originalColumnDefinitions,
-  setColumnsObject,
+  // columns,
+  // originalColumnDefinitions,
+  // setColumnsObject,
```
markkulube, line 25: if these arguments are no longer used they should be removed

(the less comments and less commented out code, the easier it is to parse the module)

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Actions.js
```diff
@@ -53,18 +54,18 @@ const Actions = ({
         test-id={`${blockClass}__customize-columns-modal--reset`}
       >
         {resetToDefaultLabel}
-      </Button>
+      </Button> */}
```
markkulube, line 57: commented out code should really be cleaned out (there are of course exceptions).

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Actions.js
```diff
@@ -53,18 +54,18 @@ const Actions = ({
...
-    isVisible: isColumnVisible(colDef),
-  }));
-};
+// const resetToOriginal = (columnDefinitions, originalColumnDefinitions) => {
+//   const keyedDefs = keyBy(columnDefinitions, 'id');
+//   return originalColumnDefinitions.map((colDef) => ({
+//     ...keyedDefs[colDef.id],
+//     isVisible: isColumnVisible(colDef),
+//   }));
+// };
```
markkulube, line 68: commented out code should really be cleaned out (there are of course exceptions).

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -16,6 +16,10 @@ import update from 'immutability-helper';
 import { pkg } from '../../../../../settings';
 import DraggableElement from '../../DraggableElement';
 import { isColumnVisible } from './common';
+// import cx from 'classnames';
+// import { Draggable16 } from '@carbon/icons-react';
+// import { RowSizeDropdown } from '../RowSize';
+// import { elementOrParentIsFloatingMenu } from '../../../../../global/js/utils/wrapFocus';
```
markkulube, line 22: commented out code should really be cleaned out (there are of course exceptions).

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -31,13 +35,30 @@ const getNextIndex = (array, currentIndex, key) => {
 };
 
 const Columns = ({
+  setVisibleColumns,
+  getVisibleColumnsCount,
```
markkulube, line 39: commented out code should really be cleaned out (there are of course exceptions).

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -31,13 +35,30 @@ const getNextIndex = (array, currentIndex, key) => {
...
+  //     if (columns[i].isVisible == true) {
+  //       result++;
+  //     }
+  //   }
+  //   return result;
+  // };
+
+  // const setVisibleColumnsChild = () => {
+  //   this.setVisibleColumns(getVisibleColumnRatio());
+  // };
```
markkulube, line 61: commented out code should really be cleaned out (there are of course exceptions).

markkulube (changes requested):
@kai-liu 

It seems like there is a lot more _commented out code_ than one would except. Yes, there are exceptional situations wherein _lines of commented out code_ are kept in production deployments. Though nothing comes immediately because I generally rely on version history should there be a need to restore a previous code snippet. 

My guess is the commented out code is being left in case future design requirements warrants a revert. I would let the developer of tomorrow worry about code needed for future enhancements and/or feature functions.

There is no absolute rule of thumb for commenting, but generally comments  should reserved to explain code patterns aren't easily understandable or the reason for doing something unconventional.

Ideally though, code should be self documenting, i.e. to write code that is self explanatory.

I will leave it to the maintainers to override this change request should the commenting style be compliant with repo standards.

Otherwise before moving forward to other refactor I would start by removing all unnecessary comments and/or dead code!

cc: @matthewgallo @elycheea @hoiwingh

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
kai-liu: This code is not commented out and therefore should not be removed.

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -54,6 +57,73 @@ const Columns = ({
...
+      `li.${blockClass}__draggable-handleHolder`
+    );
+    for (let i = 0; i < list.length; i++) {
+      if (columns[i].id.includes(id)) {
+        if (isVisible) {
+          list[i].setAttribute('selected', 'true');
+        } else {
+          list[i].setAttribute('selected', 'false');
+        }
+      }
```
markkulube, line 72: Perhaps try using array functions. This could end up looking something like

```
list.forEach((li, i) => {
       if(!columns[i].id.includes(id)) continue
       li[i].setAttribute('selected', isVisible)
})
```

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -54,6 +57,73 @@ const Columns = ({
...
+    }
+  };
+
+  const initializeList = () => {
+    for (let i = 0; i < columns.length; i++) {
+      setColumnStatus(columns[i].id, columns[i].isVisible);
+    }
+  };
```
markkulube, line 80: @kai-liu 

A `forEach `loop could be used here as well.

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -54,6 +57,73 @@ const Columns = ({
...
+
+  const setSelectAllColumnClassName = () => {
+    if (getVisibleColumnsCount() > 0) {
+      return `${blockClass}__draggable-handleHolder -selected`;
+    }
+    return `${blockClass}__draggable-handleHolder`;
```
markkulube, line 86: @kai-liu 

You could appeal to the tenary operator and reduce to one line:
```
return return `${blockClass}__draggable-handleHolder${getVisibleColumnsCount() > 0 ? ' -selected'} : ''`;
```

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -54,6 +57,73 @@ const Columns = ({
...
+  };
+
+  const setSelectAllColumnSelected = () => {
+    let target = document.querySelectorAll('#customize-columns-select-all');
+    if (getVisibleColumnsCount() > 0) {
+      target[0].setAttribute('selected', 'true');
+    } else {
+      target[0].setAttribute('selected', 'false');
+    }
```
markkulube, line 95: @kai-liu 

`if-else` block should be reducible to a one liner

```
target[0].setAttribute('selected', JSON.stringify(getVisibleColumnsCount() > 0))
```

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -54,6 +57,73 @@ const Columns = ({
...
+        setSelectAllColumnSelected();
+      }
+    } else {
+      selectAll = true;
+      for (let i = 1; i < checkboxes.length; i++) {
+        checkboxes[i].checked = true;
+        checkBoxChangeHandler(columns[i - 1], selectAll);
+        setSelectAllColumnSelected();
+      }
+    }
```
markkulube, line 117: @kai-liu 

There is duplicate code here. So I would try to abstract out what's repeated.

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/CustomizeColumnsModal.js
```diff
@@ -76,12 +81,30 @@ const CustomizeColumnsModal = ({
     }
   };
 
+  const getVisibleColumnsCount = useCallback(() => {
+    let result = 0;
+    for (let i = 0; i < columnObjects.length; i++) {
+      if (columnObjects[i].isVisible == true) {
+        result++;
+      }
+    }
```
markkulube, line 90: @kai-liu 

Probably another opportunity to use array functions / forEach loop.

markkulube (changes requested):
Hello @kai-liu 

Overall looks good.

There some opportunities for to refactor` for-loop`s into `forEach` loops. Same thing but in the Javascript world we love array functions where they can be used. They tend to simplify the code (`forEach`, `filter`, `reduce`, `map`).
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#instance_methods

If you do decide to transition into array functions,  do so in a granular fashion. This will make it easier to ensure changes to each snippet does not result in a functional regression.

Should you deem that is unnecessary and/or better to leave as is, that's okay too. Just let respond back to the change request with your rationale.

cc: @hoiwingh @matthewgallo @markkulube

holmansze (commented):
@kai-liu thanks for the enhancement. A few comments:
1. The convention is to use single quotes for string literals and package names, double quotes for React props or element attributes. Are the changes made by prettifier? Can you revert to single quotes (follow convention of existing code) and then do `npm run lint` to verify?
2. Why are we taking out "reset to default"? I think the request is to add a select all / unselect all checkbox in the customize column modal. It doesn't seem to mention that we need to remove the "reset to default" feature. Is this an explicit request from design?
3. Can we further simplify the select / unselect all logic? On a high level, when the "select/unselect all" checkbox is clicked, there should be a for-loop that selects / unselects each column checkbox by calling the corresponding on change callback. The existing on change callback doesn't need to be changed. Also, we can probably implement a function which walks the column checkbox states and return a value that indicates if all/some/none column checkboxes are checked. This can be used to populate the select/unselect all checkbox state.

Please ping me if you need clarification and we can discuss in details.

Issue #1891, opened 2022-04-07
Datagrid: Left panel story


matthewgallo:
## What will this achieve?

The styles for the story labeled "Left panel" seem to be broken. The left panel does not appear to be position fixed which is causing the data table to render below the panel currently.

![2022-04-27 at 10 15 14 - Datagrid Left Panel](https://user-images.githubusercontent.com/1717398/165539451-1241bdb4-5a65-417d-930c-fb1f9f332f92.gif)

## How will success be measured?

![image](https://user-images.githubusercontent.com/1717398/165539516-8c6afe2a-0f9a-49d1-bacb-c69ee8aaf9ac.png)

## Additional information

- Designs
- Existing code
- etc

PR #2064, opened 2022-06-24
feat: merged left panel content and datagrid on same horizontal level

faraznaseem019:
Contributes to #1891 

{{short description}}
The datagrid table (on the right) and the panel content (on the left)  are now horizontally aligned on the same level, instead of being vertically stacked on each other (in our case, the panel content was on top of the datagrid table).

#### What did you change?
The change that I made is, I changed the div with the classes <b>c4p--datagrid__grid-container</b> and <b>c4p--datagrid__displayFlex</b>, to have a display value of inherit.

I tested and verified my work by ensuring that I received the expected result (panel content and datagrid table next to each other) when running storybook.

Screenshot 2022-06-23 at 9 31 31 AM

### DIFF
--- a/packages/cloud-cognitive/src/components/Datagrid/Datagrid.stories.js
+++ b/packages/cloud-cognitive/src/components/Datagrid/Datagrid.stories.js
@@ -776,14 +776,22 @@ export const LeftPanel = () => {
   const columns = React.useMemo(() => defaultHeader, []);
   const [data] = useState(makeData(10));
   const datagridState = useDatagrid({
+    leftPanel: {
+      isOpen: true, // this toggling will happen from datagridActions.
+      panelContent: (
+        <div style={{ display: 'inline' }}>
+          Panel content will go here along with any button interactions
+        </div>
+      ),
+    },
     columns,
     data,
-    leftPanel: {
+    /*leftPanel: {
       isOpen: true, // this toggling will happen from datagridActions.
       panelContent: (
         <div>Panel content will go here along with any button interactions</div>
       ),
-    },
+    },*/
     DatagridActions,
     DatagridBatchActions,
   });
--- a/packages/cloud-cognitive/src/components/Datagrid/Datagrid/Datagrid.js
+++ b/packages/cloud-cognitive/src/components/Datagrid/Datagrid/Datagrid.js
@@ -98,6 +98,7 @@ export let Datagrid = React.forwardRef(({ datagridState, ...rest }, ref) => {
       {leftPanel && (
         <div
           className={`${blockClass}__grid-container ${blockClass}__displayFlex`}
+          style={{ display: 'inherit' }}
         >
           {leftPanel && leftPanel.isOpen && (
             <div className={`${blockClass}__datagridLeftPanel`}>
--- a/packages/cloud-cognitive/src/components/AddSelect/AddSelectSidebar.js
+++ b/packages/cloud-cognitive/src/components/AddSelect/AddSelectSidebar.js
@@ -38,7 +38,7 @@ export let AddSelectSidebar = ({
     const selectedItem = items.find((item) => item.id === cur);
     // certain properties should not be displayed in the sidebar
     // eslint-disable-next-line no-unused-vars
-    const { icon, avatar, ...newItem } = selectedItem;
+    const { meta, icon, avatar, ...newItem } = selectedItem;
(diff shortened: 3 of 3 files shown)


packages/cloud-cognitive/src/components/Datagrid/Datagrid.stories.js
```diff
@@ -776,14 +776,22 @@ export const LeftPanel = () => {
   const columns = React.useMemo(() => defaultHeader, []);
   const [data] = useState(makeData(10));
   const datagridState = useDatagrid({
+    leftPanel: {
+      isOpen: true, // this toggling will happen from datagridActions.
+      panelContent: (
+        <div style={{ display: 'inline' }}>
```
markkulube, line 782: @faraznaseem019 

Maybe let's move this to a `*.scss` file.

While this works, inline styling is more commonly used when style is dynamically updated based on a state variable. For example, we may want to adjust the `background-color `based on minute changes in screen size - here probably want to declare the selector inline as we could end up with an infinite number of classes to cater for the possible RGB values.

I do note this is for a story, and if it's consistent with how styles are being applied for stories it could be okay to leave as is.


cc: @holmansze @matthewgallo

packages/cloud-cognitive/src/components/Datagrid/Datagrid.stories.js
```diff
@@ -776,14 +776,22 @@ export const LeftPanel = () => {
...
     columns,
     data,
-    leftPanel: {
+    /*leftPanel: {
       isOpen: true, // this toggling will happen from datagridActions.
       panelContent: (
         <div>Panel content will go here along with any button interactions</div>
       ),
-    },
+    },*/
```
markkulube, line 794: @faraznaseem019 

Maybe add a comment explaining why this snippet is be commented out and not removed completely.

packages/cloud-cognitive/src/components/Datagrid/Datagrid/Datagrid.js
```diff
@@ -98,6 +98,7 @@ export let Datagrid = React.forwardRef(({ datagridState, ...rest }, ref) => {
       {leftPanel && (
         <div
           className={`${blockClass}__grid-container ${blockClass}__displayFlex`}
+          style={{ display: 'inherit' }}
```
markkulube, line 101: @faraznaseem019 

For the same reasons as above, this inline styling should definitely be declared in an `*.scss` unless if there is an exceptional case for having it here?

markkulube (changes requested):
@faraznaseem019 

Some minor tweaks are in order - see comments!

packages/cloud-cognitive/src/components/Datagrid/Datagrid.stories.js
```diff
@@ -46,6 +46,10 @@ import mdx from './Datagrid.mdx';
 
 import styles from './_storybook-styles.scss';
 
+// import pkg from '../../settings';
+
+// const blockClass = `${pkg.prefix}--datagrid`;
```
markkulube, line 51: @faraznaseem019 

Let's clean this up. Unless the commented out is required for reason, it should be removed.

Though I'm almost certain there no longer needed as this code has been moved to lines 779-781

markkulube (dismissed):
@faraznaseem019 

LGTM barring the left  over commented code in the story.

Final validation/approval will have to wait for @matthewgallo / @holmansze / @elycheea

markkulube:
@faraznaseem019 you may want to update your feature branch with `main`!

packages/cloud-cognitive/src/components/Datagrid/styles/datagrid.scss
```diff
@@ -334,7 +334,7 @@
 
 .#{block-class}__displayFlex {
   position: relative;
-  display: flex;
+  display: inherit;
```
matthewgallo, line 337: I'm not 100% this should change. This is for providing a way to have a data grid that uses display flex. Possibly there's a different approach we could take that allows us to keep `display: flex;`

faraznaseem019: Fixed.

faraznaseem019 (commented):
Made changes to datagrid.scss

matthewgallo (approved):

Issue #2078, opened 2022-07-07
TagSet multiline attribute


ConradSchmidt:
# TagSet multiline attribute

## Brief description

TagSet should have a multiline option

## Details

Used in IBM products already:
https://github.ibm.com/wdp-gov/tracker/issues/58287

### Links to other materials

![image](https://user-images.githubusercontent.com/3808948/177771294-c2caa0eb-aeae-4319-bf39-84f2881385d2.png)
![image](https://user-images.githubusercontent.com/3808948/177771360-ff509fe2-a393-4517-ba72-93d0b0ee70bf.png)

## Design Maintainer(s)

- Marion Bruells

## Tasks

Before starting work on this epic, please review and complete the following.

- [x] Initial review of design/existing code
[design](https://pages.github.ibm.com/cdai-design/pal/components/tag-set/usage/)  
[code](https://github.com/carbon-design-system/ibm-cloud-cognitive/tree/main/packages/cloud-cognitive/src/components/TagSet)

### Working in Carbon for IBM Products package

- [x] Have you reviewed our
      [contribution guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/CONTRIBUTING.md)?
- [x] Have you noted our
      [code guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/docs/CODE_GUIDELINES.md)?
- [x] Did you use the CLI to scaffold your new component?
Its not a new component

PR #2079, opened 2022-07-07
feat: add multiline attribute to tagset

ConradSchmidt:
Contributes to #2078

TagSet with multiline attribute

#### What did you change?
Add a multiline attribute to TagSet

Open storybook and open the TagSet > multiline story

Result:
image


matthewgallo:
This is great @ConradSchmidt! Just one thing, because this is a released component you need to update the styles snapshot with this command, `yarn test:c4p:snapshot`, from the root.

matthewgallo (approved):

Issue #2017, opened 2022-06-06
Card (productive) overflow menu should match field size


johanna-koval:
## Bug description
The overflow menu inside the bottom action bar in the productive card component should be 40px.
Right now the overflow menu when in the bottom bar is 48px which causes this bar to be bigger than defined. This can create inconsistencies in the layout with cards using bottom bars without overflow

<img width="658" alt="CleanShot 2022-06-02 at 08 36 27@2x" src="https://media.github.ibm.com/user/232597/files/1ee10580-e24f-11ec-84e3-5d0785846d2d">

## URL/Link of Page
Link here: https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-cards-productivecard--with-overflow&args=actionsPlacement:bottom

## Expected behavior
Overflow in bottom bar should use the 40px button size matching the rest of the button used in this space:
<img width="654" alt="CleanShot 2022-06-02 at 08 33 18@2x" src="https://media.github.ibm.com/user/232597/files/48019600-e24f-11ec-83b1-f8777c72afda">
<img width="799" alt="CleanShot 2022-06-02 at 08 33 01@2x" src="https://media.github.ibm.com/user/232597/files/49cb5980-e24f-11ec-9b84-a353b54a959d">

PR #2081, opened 2022-07-08
chore(card overflow): height issue fixed

jagathgj:
Contributes to #2017 

Overflow menu icon needs to have a height of 40px. Currently it is 48px.

#### What did you change?
Added extra styles for height in card.scss.

storybook

### DIFF
--- a/packages/cloud-cognitive/src/components/Card/Card.js
+++ b/packages/cloud-cognitive/src/components/Card/Card.js
@@ -113,7 +113,7 @@ export let Card = forwardRef(
                 renderIcon={Icon}
                 hasIconOnly
                 onClick={onClick}
-                size={actionsPlacement === 'top' ? 'sm' : 'field'}
+                size="sm"
                 iconDescription={iconDescription}
                 kind="ghost"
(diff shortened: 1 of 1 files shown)


elycheea (changes requested):
I looked into `actionsPlacement` some, it looks like we change the button size to `field` based on the position of the overflow? If the size should be `sm` in both cases, we may be better off removing `field` than overriding the CSS.

Going to tag @davidmenendez in case I’m missing some other use case or need for the `field` size.

davidmenendez:
@elycheea thanks for tagging me.

@jagathgj you're updating the wrong bit of code here. line 88 is where the size for the overflow menu is set. inside `getActions`. if you just update `xl` to `md` i think that should fix the problem.

jagathgj:
@davidmenendez Thanks David for pointing it out. Kindly take a look at the suggested changes I have committed.

davidmenendez (approved):

davidmenendez (approved):

elycheea (approved):

Issue #1869, opened 2022-04-01
Carbon V11 Support


lee-chase:
With Carbon V11 released on March 31st 2022 there are a number of questions to answer with regards to how it affects @carbon/ibm-products.

Break down of tasks:(Only convert to issue, assigned to epic, when ready to be worked on). 
- [x] #1870
- [ ] Find out when our consumers are wanting to move to V11 and discuss findings with the team.

asfordmatt:
Just noting my interest in this. We will want to move up to V11 at some point, but don't have a timeframe as yet.

Mikadv:
We are asked by our designers to use certain components of carbon 11 and want to utilize the build time improvements and code cleanup possibilities. Hence, we need this library on carbon 11 or our apps break. Now this means that `ibm-cloud-cognitive` is blocking us from upgrading to carbon 11 and to move forward.

Affected products: wkc-glossary, wkc-workflow, wkc-reference-data.

PR #2082, opened 2022-07-08
fix: carbon 11 token linter

lee-chase:
Contributes to #1869 

 - Updates the Carbon token linter to 2.0.0-beta.7 which supports V11 and V10. 
 - Does not lint CDAI and Security. Would need to move linter commands down to each project similar to build.
 
#### What did you change?

Carbon token linter version and config. Updated a number of lines to not generate linter errors.

Ran the style linter.


matthewgallo:
These changes have been merged in the Carbon v11 branch from the main branch

matthewgallo (commented):
As we discussed earlier today, maybe there could be a way the linter could appropriately lint v10 vs v11 together. If not, we can keep what you have here!

lee-chase:
@matthewgallo I have tried doing this by moving the linter commands below into each package and running them from the root. 

```
    "lint:style": "yarn lint:style:files './src/**/*.scss'",
    "lint:style:files": "stylelint --report-needless-disables --report-invalid-scope-disables",
```

Unfortunately this does not result in the Carbon modules being read from the package folder. 

In effect the linter needs to use the package version of node_modules for the Carbon dependencies. I've not figured out how to do this. You're welcome to have a go, but as the CDAI and Security code is linted elsewhere I would not worry about it.

matthewgallo (approved):

PR #2086, opened 2022-07-11
chore: update release schedule

matthewgallo:
The release action still appears to be run by Dave, I think because he was the last to update the schedule of the release action. This small change to the cron schedule should fix this. See screenshot below for release run from last week:
image

#### What did you change?
`release.yml`


AlexanderMelox (dismissed):
Nothing fancy happening here

Issue #1415, opened 2021-11-10
WebTerminal: release review


matthewgallo:
## What will this achieve?

A release review needs to happen in order for us to release the `WebTerminal` component.

The release review guidelines can be found [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/main/docs/reviews/RELEASE_REVIEW_GUIDELINES.md)

### Readiness

- [x] One or more scenarios for a design pattern have been identified as a useful unit of functionality to publish.
- [x] A functioning component or components delivering those scenarios have been delivered and merged to main.
- [x] Design maintainer has approved the implementation for those scenarios (via a comment on the relevant issue/epic).

### Engineering review

- [x] Breaking changes have only been introduced with prior approval, discussion and documented in release notes. Ideally deprecation notices in the prior major version must have been added in a timely fashion.
- [x] The implementation takes into account, and does not impair, remaining and anticipated design scenarios.
- [ ] Public component features (names, props, etc) are consistent with Carbon-defined conventions and are consistent internally. Where there isn't an existing convention to apply, ensure robust precedents are being established.
- [ ] The UI produced is accessible, responsive, translatable, cross-browser, and responds to the currently set Carbon theme.
- [x] Components are functional components using hooks.
- [x] Public components which produce DOM structures support className.
- [x] Public components support a ref (via React.forwardRef).
- [ ] Public component supports a Devtools attribute 
- [x] All significant DOM elements have meaningful classes.
- [x] Additional attributes that are not identified as props (such as title, aria-*, etc) are passed through to an appropriate DOM node of the component as HTML attributes.
- [x] No warnings, errors or log messages in the console.
- [x] Each public component JS is exported in /src/components/index.js, each public component SCSS is included in /src/components/_index.scss, and each public component has a flag in package-settings.js.
- [x] Each public component SCSS lists all of the Carbon and C&CS components imported and used by the JavaScript code and explicitly imports the SCSS for each of these components.

### Standards

- [x] No linter warnings or errors are produced.
- [x] Carbon tokens (theme, layout, motion) are used unless the design specifies otherwise.
- [x] All components utilizing motion must include reduced-motion queries for accessibility purposes - read more here.
- [x] Code is formatted according to prettier rules (no use of //prettier-ignore).
- [x] Code is clear, maintainable and follows standard React practices and the code guidelines.
- [x] Copyright header in every source file (js, css, scss etc.) with the appropriate years.

### Testing

[...]

PR #2097, opened 2022-07-14
feat(WebTerminal): refactors code to useContext and custom hooks 

AlexanderMelox:
Contributes to #1415 

Refactored web terminal to use a custom hook and context to share state between the various web terminal components. This PR is in response to the release review question "Can we change `closeTerminal` to `onClose` or `onCloseTerminal` instead for consistency.".

This change altogether allows the user to now worry about passing an `onClose` or `open` prop to the web terminal and gives the user the liberty to do what they please when they call the hooks `closeWebTerminal, openWebTerminal, toggleWebTerminal`


packages/cloud-cognitive/src/components/WebTerminal/WebTerminal.js
```diff
@@ -146,7 +146,7 @@ export let WebTerminal = React.forwardRef(
             hasIconOnly
             renderIcon={Close}
             kind="ghost"
-            iconDescription={closeIconDescription}
+            iconDescription="Close web terminal"
```
davidmenendez, line 149: why did you remove the `closeIconDescription` prop?

davidmenendez (commented):
looks good to me. i just had one comment about the close icon label.

packages/cloud-cognitive/src/components/WebTerminal/WebTerminal.js
AlexanderMelox: Thinking about it, what else can this be called? The literal action is closing the web terminal, I don't think designers would really want to change this to something else? Thoughts?

AlexanderMelox:
Some new changes with the last two commits. 

1. Update props
Added a new prop `isInitiallyOpen` this allows the user to make the WebTerminal open on load. This prop also updates the custom hook to reflect the state change.

2. Updated the tests and added new ones to test the functionality of the `useWebTerminal()` hook.
3. Added `@testing-library/react-hooks` to be able to test react custom hooks in jest.
We need this library to use hooks since we cannot use hooks directly in the test function since it's not a react component.

AlexanderMelox:
New updates: 

1. Added the WebTerminalProvider to the canary flag enabled test. 
2. Added WebTerminalProvider to package settings

matthewgallo:
Just posting an update here from our discussion around this from earlier today, I think this approach will be ready to merge once changes are made in `index-all-enabled.test.js` to scope the `WebTerminalProvider` to only wrap around the `WebTerminal` component.

matthewgallo (approved):

PR #2104, opened 2022-07-18
fix: inline edit overflow to right

lee-chase:
Contributes to #

This should prevent inline-edit from overflowing to the right when near a scroll container boundary. 

NOTE: it the user needs to choose a scroll safe position for the button tooltips too.

#### What did you change?

How the save button is rendered when not active.

Storybook.


davidmenendez:
looks like the tests are failing. please take a look.

matthewgallo (approved):

Issue #1617, opened 2022-02-10
Data table extension: clickable row items


matthewgallo:
# Data table extension: clickable row items

## Brief description

This issue will deliver the clickable row items data table extension.

## Details

[Link to Design](https://pages.github.ibm.com/cdai-design/pal/components/data-table/clickable-row-items/usage/)

### Links to other materials

e.g. sketch files, icons, images, prototypes

## Design Maintainer(s)

- Marion Brülls
- Mitchell Bernstein

## Tasks

Before starting work on this epic, please review and complete the following.

- [ ] Initial review of design/existing code

### Working in Carbon for IBM Products package

- [ ] Have you reviewed our
      [contribution guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/CONTRIBUTING.md)?
- [ ] Have you noted our
      [code guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/docs/CODE_GUIDELINES.md)?
- [ ] Did you use the CLI to scaffold your new component?

PR #2105, opened 2022-07-18
chore(Data table): clickable row items

Ratheeshrajan:
Contributes to #1617

 Data table extension: clickable row items data table extension with side panel.

#### What did you change?

- 'onRowClick' trigger the side panel open and show the data of the selected row in the side panel.
- Toggle selected row style
- Indication of the selected row.

Storybook


matthewgallo (changes requested):
I don't _think_ that the SidePanel functionality should be embedded inside of the Datagrid. Seems like it would be good to give the consumers of Datagrid the flexibility to implement it how they want (slide in vs slide over side panel, the size of the panel, etc.).

We can keep your implementation as an example, but I think all of the SidePanel handling should be done in Datagrid.stories.js.

packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridRow.js
```diff
@@ -23,6 +23,7 @@ const DatagridRow = (datagridState) => {
     <TableRow
       className={cx(`${blockClass}__carbon-row`, {
         [`${blockClass}__carbon-row-expanded`]: row.isExpanded,
+        [`bx--data-table--selected`]: row.isSelected,
```
matthewgallo, line 26: This line should import the carbon prefix from our settings like the following:

```jsx
import { carbon } from '../../settings';

...
...
[`${carbon.prefix}--data-table--selected`]: row.isSelected,
...
...
```

Ratheeshrajan: @matthewgallo `${carbon.prefix}` class updated.

matthewgallo (approved):
🎉 looks great @Ratheeshrajan

Issue #2071, opened 2022-06-29
Carbon v11: update component tests


AlexanderMelox:
## What will this achieve?

With v11 some of the DOM elements, class names, and props have changed and have broken some of the tests for our components. 

When updating a component's API to use Carbon v11 make sure to check all of the imported Carbon components within the component file itself, as well as any Carbon components used in the stories.

- [x] APIKeyModal
- [x] APIKeyDownloader
- [x] AddSelect
- [x] AddSelectBreadcrumbs
- [x] AddSelectSort
- [x] AddSelectMetaPanel
- [x] AboutModal (has a11y violations all other tests pass)
- [x] ActionBar
- [x] ActionBarItem
- [x] ActionSet
- [x] BreadcrumbWithOverflow
- [x] ButtonMenu
- [x] ButtonSetWithOverflow
- [x] Card
- [x] CardHeader
- [x] CardFooter
- [x] Cascade
- [x] ComboButton
- [x] CreateFullPage
- [x] CreateInfluencer
- [x] CreateModal
- [x] CreateSidePanel
- [x] CreateTearsheet
- [x] CreateTearsheetNarrow
- [x] DataSpreadsheet
- [x] Datagrid (in progress @matthewgallo)
- [x] EditSidePanel
- [x] EmptyStates
- [x] ExampleComponent
- [x] ExportModal
- [x] ExpressiveCard
- [x] HTTPErrors
- [x] ImportModal
- [x] InlineEdit
- [ ] ModifiedTabs (has no tests)
- [x] MultiAddSelect
- [x] NotificationsPanel
- [x] OptionsTile
- [x] PageHeader
- [x] ProductiveCard
- [x] RemoveModal (@elycheea)
- [x] Saving
- [x] SidePanel
- [x] SingleAddSelect
- [x] StatusIcon
- [x] TagSet (@elycheea)
- [x] Tearsheet
- [x] Toolbar 
- [x] UserProfileImage 
- [x] WebTerminal
- [x] Wrap
- [x] `src/__tests__/styles.test.js` (will be complete when https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2121 is reviewed/merged)
- [x] `src/__tests__/scss.test.js`

## How will success be measured?'

In your terminal run `yarn run jest /[Component folder name]/` and all tests should pass.

AlexanderMelox:
(July 6) Went through all the components and they are updated to reflect the tests that have passed and failed

[...]

PR #2115, opened 2022-07-20
chore: update RemoveModal accessibility

elycheea:
Contributes to #2071

Updates aria label and skips accessibility test

#### What did you change?
- Adds aria label for RemoveModal
- Skips test due to Carbon changes

Ran tests


AlexanderMelox (approved):

AlexanderMelox:
Are we creating new issues for the components we decide to skip violations?

elycheea:
@AlexanderMelox We have not been yet but we’ll should. The modals largely all run into the same violation so those should be a single issue. For the others, may be better to address them as individual issues than one story so it’s easier for people to take on.

Issue #2117, opened 2022-07-20
AddSelectBody testing


davidmenendez:
`AddSelectBody` needs testing

PR #2118, opened 2022-07-20
chore: add select body testing

davidmenendez:
Contributes to #2117 

updates the tests for `AddSelectBody`

Screen Shot 2022-07-20 at 4 09 17 PM

one important thing to note about this PR is that it removes the remove icon that appeared next to selections in the sidebar. tests could not pass because `AccordianItem` is a button and a `Button` cannot exist inside a button. so we will have to work with design to find a new solution post release.


davidmenendez:
added an additional commit to increase the sidebar coverage to 100%

Screen Shot 2022-07-21 at 3 51 02 PM

AlexanderMelox (approved):

Issue #2065, opened 2022-06-27
An unexpected scrollbar is showing up when using the PageHeader component in some browsers


Danilo-Ribeiro-IBM:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

We have come across a problem where the PageHeader component is behaving differently between browsers, on Chrome the PageHeader works normally, but on Firefox it looks like it ignores the overflow property which causes an unwanted scrollbar to appear in some cases.

> Is this issue related to a specific component?

PageHeader on @carbon/ibm-products

> What did you expect to happen? What happened instead? What would you like to
> see changed?

I'd like to see matching behaviour between all the browsers, PageHeader works fine on chrome, but it seems that on Safari and Firefox the aforementioned bug appears.

> What browser are you working in?

The bug can be found on Firefox and Safari, Chrome behaves as expected.

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

1.20.3

## Steps to reproduce the issue

Can be reproduced on the [playbook](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-page-header-pageheader--with-title)

1. Open the playbook on the PageHeader page with Firefox
2. Position the mouse on top of the Page Title
3. Scroll with the mouse, or click with the MMB to try to scroll the page down, you should see the title bounce.

> Please create a reduced test case in CodeSandbox
>

https://codesandbox.io/s/u3hix?file=/src/index.js

## Additional information

- Screenshots or code
![image](https://user-images.githubusercontent.com/108276493/175948709-82a7dbb9-03d1-407c-877a-07efcc125acf.png)

punnoosewilson:
Hi @Danilo-Ribeiro-IBM 
The unexpected scrollbar issue is fixed. you can check and close this issue.

Danilo-Ribeiro-IBM:
> Hi @Danilo-Ribeiro-IBM The unexpected scrollbar issue is fixed. you can check and close this issue.

Thanks a lot for the quick response @punnoosewilson! I'll close the issue.

PR #2119, opened 2022-07-21
fix(PageHeader): solved unexpected scrollbar issue

punnoosewilson:
Contributes to #2065

An unexpected scrollbar is showing up when using the PageHeader component in some browsers

#### What did you change?
- Gave same min-height to the child-element

- Storybook (manual cross browser testing)


elycheea (approved):
🎉

PR #2121, opened 2022-07-21
fix: Sass refactor on `carbon-v11` branch to address duplicate Carbon styles in built css

matthewgallo:
When we generated a new snapshot on the `carbon-v11` it highlighted some sass issues that needed to be investigated around duplicate Carbon styles. This was because each components `index.scss` file was still using `@import`, changing this to `@use` removed the duplicate styles.

There was another issue around providing built styles of each one of our components, but _without_ any Carbon styles included (because consumers may be already importing Carbon styles). To still be able to provide this, two new sass files have been created for each of our components (`_carbon-imports.scss` and `_index-with-carbon.scss`). This allows us to provide styles in two ways:
1. Only our component styles
2. Our component styles, plus any Carbon components used within that component

This change means that any _new_ components from this point forward will need to follow this new structure. We can update the `generate` component script in a separate PR to help with this.

#### What did you change?
Component sass files, added `_carbon-imports.scss` and `_index-with-carbon.scss` to each of our components. Then `index-released-with-carbon.scss` and `index-with-carbon.scss` import (`@use`) each components `_index-with-carbon.scss` file.

I ran the build scripts for our package and was able to confirm that the generated css files were being built with a similar output as on our `main` branch. Also updated the snapshot and it went down from ~50k lines to ~7.7k lines.

FYI @lee-chase, this includes the approach we discussed yesterday


davidmenendez (approved):
> 6000+ line changes

> 150+ file changes

i'm assuming this is fine 😅

matthewgallo:
Haha, I know. In order for us to emulate the same build output as `v10` or our `main`, branch we had to restructure all of the component sass files. Happy to discuss this further next week before we decide on merging this.

elycheea:
Let’s discuss on Monday and open the corresponding tickets we’ll need to create as a result of these changes.

elycheea:
Hoping a close/reopen will restart the checks here. 🙃

packages/cloud-cognitive/src/components/MultiAddSelect/_index-with-carbon.scss
```diff
@@ -0,0 +1,8 @@
+//
+// Copyright IBM Corp. 2022, 2022
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+@use '../AddSelect/index-with-carbon';
```
elycheea, line 8: Any reason why we’re doing this as opposed to 

```scss
@use './carbon-imports';
@use './multi-add-select';
```

like we do for other components, including SingleAddSelect?

elycheea (commented):
Noticed one small difference in the import for MultiAddSelect from the rest of the components. Otherwise looks good to me.

packages/cloud-cognitive/src/components/MultiAddSelect/_index-with-carbon.scss
matthewgallo: Good question, initially it was because of the fact that `MultiAddSelect` and `SingleAddSelect` both simply use the styles from `AddSelect`, but I agree with you about changing it, both for consistency and in the event these components get their own styles the pattern will already be in place.

elycheea: Heh, I thought that might be the case but then thought to ask when I noticed SingleAddSelect was following the same patterns as the others.

elycheea (approved):

PR #2128, opened 2022-07-25
chore: update ModifiedTabs component to use new v11 Tabs api

matthewgallo:
Contributes to https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/2020

Makes necessary updates to ModifiedTabs component so that it renders and interacts as it did when using Carbon v10.

#### What did you change?
```
packages/cloud-cognitive/src/components/ModifiedTabs/ModifiedTabLabelNew.js
packages/cloud-cognitive/src/components/ModifiedTabs/ModifiedTabLabelWithClose.js
packages/cloud-cognitive/src/components/ModifiedTabs/ModifiedTabs.js
packages/cloud-cognitive/src/components/ModifiedTabs/ModifiedTabs.stories.js
packages/cloud-cognitive/src/components/ModifiedTabs/_modified-tabs.scss
```
Storybook


packages/cloud-cognitive/src/components/ModifiedTabs/ModifiedTabs.js
```diff
@@ -5,93 +5,107 @@
...
-      >
-        <div className="some-content">{newTabContent}</div>
-      </Tab>
-    </Tabs>
-  );
-};
+    return (
+      <div className={cx(blockClass, className)} ref={localRef}>
+        <Tabs className="modified-tabs">
+          <TabList aria-label="Modified tab list temp aria label">
```
elycheea, line 65: Was this left in by mistake? Expecting we want an actual aria-label here?

packages/cloud-cognitive/src/components/ModifiedTabs/ModifiedTabs.js
```diff
@@ -5,93 +5,107 @@
...
+              role="button"
+              className={`${blockClass}__tab-new`}
+            >
+              <ModifiedTabLabelNew label={newTabLabel} />
+            </Tab>
+          </TabList>
+          <TabPanels>
+            {tabs.map((tab) => (
+              <TabPanel key={tab.id} id={tab.id}>
+                <div className="some-content">{tab.content?.()}</div>
```
elycheea, line 88: Since this is in the actual component and not Storybook, I feel like we should replace this with an actual classname. But if we left this intentionally to minimize potential conflicts until we resolve in v10, then I think that’s probably okay.

elycheea (commented):
Some minor differences from v10 but looks like the v11 changes are actually more inline with the guidance. 👍 

Looks like we still have a few test classes/labels in the component but want to check if intentional before I approve.

packages/cloud-cognitive/src/components/ModifiedTabs/ModifiedTabs.js
matthewgallo: Agreed, I forgot to update this class.

packages/cloud-cognitive/src/components/ModifiedTabs/ModifiedTabs.js
matthewgallo: Yes, left by mistake. I'll update this so it's passed as a prop

elycheea (approved):

Issue #2075, opened 2022-07-06
Combo Button


qbi11y:
Make updates to the combo button component per designer design updates. This breaks 1430 into smaller issues:[https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/1430](url)

qbi11y:
Questions to be answered:
* .bx--btn is a Carbon class set to 63px
* Is it truly 160px min-width
* Is adding rules a Carbon ask?
* how do sizes get added?

elycheea:
Based on Popover component, which is v11. For current v10 version, we’ll need to use Overflow but can update to Popover for v11 version.

elycheea:
https://react.carbondesignsystem.com/?path=/story/experimental-unstable-combobutton--default

PR #2130, opened 2022-07-25
feat(ComboButton): move to Canary

qbi11y:
Contributes to #2075

Moving Combo button to Canary

#### What did you change?
Updated carbon.yml, story-structure.js, package-settings.js.


packages/cloud-cognitive/src/components/ComboButton/_combo-button.scss
```diff
@@ -14,52 +14,60 @@
   /// Sets the prefix for all IBM Security-specific classes.
   /// @access private
   /// @type String
-  $security--prefix: security;
+  //$security--prefix: cfp;
```
elycheea, line 17: Can we remove this comment?

packages/core/story-structure.js
```diff
@@ -42,6 +42,7 @@ const s = [
           { n: 'Tearsheet', s: ['c/Tearsheet', 'c/TearsheetNarrow'] },
           { n: 'DataSpreadsheet', s: ['c/DataSpreadsheet'] },
           { n: 'Datagrid', s: ['c/Datagrid'] },
+          { n: 'ComboButton', s: ['c/ComboButton']}
```
elycheea, line 45: Our order looks loosely alphabetical ... Maybe we should move this after `Card`?

elycheea (changes requested):
I’m seeing some differences in the [styling](https://pages.github.ibm.com/cdai-design/pal/components/button-guidance/menu-and-combo-buttons#styling) compared to the guidance. Some of it may be differences in theming so feel free to reach out to Laura with any questions.

It looks like there’s also some more examples we could include based on the [modifiers](https://pages.github.ibm.com/cdai-design/pal/components/button-guidance/menu-and-combo-buttons#modifiers).

ComboButton should probably also include a prop for `size` based on the [variations](https://pages.github.ibm.com/cdai-design/pal/components/button-guidance/menu-and-combo-buttons#combo-button) listed. These should be included in the controls for the stories.
Also, I think it’d be nice if we could make the controls more interactive. I think PageHeader could be a good example of how to show examples with varying states (danger, dividers, disabled) through the `children`.

packages/cloud-cognitive/src/components/ComboButton/_combo-button.scss
qbi11y: Hmmm that's odd... I switched over to $blockClass. I will take a look at it.

packages/core/story-structure.js
qbi11y: Yes, I thought alphabetical would be best but it wasn't presented that way so I figured it was to be added at the bottom. I can make the list alphabetical if that's the way we want to go.

qbi11y:
@elycheea I had a conversation with Laura in regards to the styling. Her guidance is based on the popover component, this has been built using the overflow menu. I informed her the styling would come more in line when we flipped to v11. She was ok with that. If you have other suggestions, I am all ears.

qbi11y:
Looking at the Carbon OverflowMenu the storybook behavior is identical to the storybook behavior on our end. The issue would not present in production as it only presents when the menu is open and you change the size of the button in storybook. Can we agree there will be no storybook fix on our end?

The large button seems the aligning text to baseline and not center. When center align is added it seems to be slightly off center from chevron. Could be optical illusion will need more eyes.

elycheea:
@qbi11y Don’t forget to resolve conflicts!

packages/cloud-cognitive/src/components/ComboButton/ComboButton.stories.js
```diff
@@ -17,19 +17,60 @@ import { ComboButton, ComboButtonItem } from '..';
...
-));
+const Template = (args) => {
+  return (
+    <ComboButton {...args}>
+      <ComboButtonItem>ComboButtonItem 1</ComboButtonItem>
+      <ComboButtonItem renderIcon={args.icon && args.icon}>
+        ComboButtonItem 2
+      </ComboButtonItem>
+      <ComboButtonItem>ComboButtonItem 3</ComboButtonItem>
+      <ComboButtonItem hasDivider={args.hasDivider}>
```
elycheea, line 49: It might make more sense to use something like `<ComboButtonDivider/>`, similar to how Carbon does in its SideNav.

Example from Carbon:
image

While I have seen `hasDivider`, I think usually, that’s for dividers in a fixed position. In the case of the combo button, the consumer/should have more control over where example the divider is placed, in which case the divider subcomponent might be the better option.

elycheea (commented):
Hi! I think we’re close — there’s a comment we should remove and then I made a suggestion to move `hasDivider` to a `ComboButtonDivider` subcomponent.

packages/cloud-cognitive/src/components/ComboButton/_combo-button.scss
```diff
@@ -14,52 +14,105 @@
...
-  .#{$security--prefix}--combo-button__action {
+  .#{$blockClass}__action {
     @include text-overflow;
   }
 
-  .#{$security--prefix}--combo-button
-    .#{$security--prefix}--combo-button__overflow-menu {
-    width: carbon--mini-units($count: 6);
+  .#{$blockClass}__btn--sm.bx--btn {
+    min-width: 128px;
```
elycheea, line 29: These widths should all use `rem`s — `rem(128px)` should work.

packages/cloud-cognitive/src/components/ComboButton/ComboButton.js
```diff
@@ -13,99 +13,119 @@ import {
...
+  /** Provide an optional flag to disable the ComboButton */
+  disabled: bool,
+
   /** Provide the [props of the `OverflowMenu`](https://react.carbondesignsystem.com/?path=/docs/overflowmenu) */
   overflowMenu: shape(OverflowMenu.propTypes),
+
+  /**
+   * Set the size of the combo button
+   */
+  size: oneOf(['sm', 'md', 'lg']),
```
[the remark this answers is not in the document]
elycheea: Or ... we could potentially allow consumers to use `lg` now and then map this to `default` in v10 so we don’t cause a breaking change when we move to v11? 🤔 

We can discuss this on Slack.

elycheea (changes requested):
`ComboButtonItem`s are 40px instead of 48px height when `default` / `lg` is selected. Also some more thoughts on how to handle the `default` and `lg` size differences.

Based on our discussion with Laura, I believe we wanted to keep these in one simple story — see comments below.

Then a few other smaller changes for conventions/standards to update units and prefixes.

elycheea:
@qbi11y Looks like your latest change only included whitespace changes. Can you check again? Also, as discussed on scrum a few days ago, let’s move `Datagrid` and `DataSpreadsheet` back to the end of the `story-structure.js` since Matt’s been continuing to open new PRs that will create new conflicts here.

packages/cloud-cognitive/src/components/ComboButton/ComboButton.js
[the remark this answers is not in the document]
matthewgallo: Seems that they changed `default` to `lg` in v10 as well!
https://v7-react.carbondesignsystem.com/?path=/docs/components-overflowmenu--default

packages/cloud-cognitive/src/components/ComboButton/ComboButton.stories.js
```diff
@@ -38,49 +34,67 @@ export default {
 };
 
 const Template = (args) => {
+  const myRef = useRef();
```
elycheea, line 37: I think usually when we use `ref` in our stories, we just name it `ref` rather than `myRef`.

packages/cloud-cognitive/src/components/ComboButton/_combo-button.scss
```diff
@@ -98,6 +98,20 @@
     display: none;
   }
 
+  .#{$blockClass}__divider {
+    width: 100%;
+    height: 1px;
+    background-color: $border-subtle;
+  }
+
+  .#{$blockClass}__overflow-menu__divider::before {
```
elycheea, line 107: Is this still in use? May be leftover from an older implementation or is this intended as a divider when there’s also a danger button after? Either way, I don’t see this reflected in the current implementation.

packages/cloud-cognitive/src/components/ComboButton/_combo-button.scss
```diff
@@ -98,6 +98,20 @@
...
+    width: 100%;
+    height: 1px;
+    padding: 0;
+    border-top: 1px solid $button-danger-primary;
+    content: "'hello'";
```
elycheea, line 112:
```suggestion
    content: "''";
```

elycheea (changes requested):
There’s a few more things that still need to be addressed.

qbi11y:
I have addressed the quick fixes. Will revert back to hasDivider and condense the story.

packages/cloud-cognitive/src/components/ComboButton/ComboButton.js
```diff
@@ -146,15 +122,27 @@ export let ComboButton = React.forwardRef(
 ComboButton = pkg.checkComponentEnabled(ComboButton, componentName);
 ComboButton.displayName = componentName;
 ComboButton.propTypes = {
+  /** Provide an optional flag to disable entire ComboButton */
+  buttonDisabled: bool,
```
elycheea, line 126: Do we need three disabled states here? I’m seeing `buttonDisabled`, `disabled`, and `itemDisabled`, but only `disabled` is used in ComboButton. If these are meant for just the stories, they should be set there instead.

packages/cloud-cognitive/src/components/ComboButton/ComboButton.js
```diff
@@ -146,15 +122,27 @@ export let ComboButton = React.forwardRef(
...
 
+  /** Provide an optional flag to disable a divider between ComboButtonItems */
+  hasDivider: bool,
+
+  /** Provide an optional flag to convert a ComboButtonItem to a delete item */
+  isDelete: bool,
+
+  /** Provide an optional flag to disable selected ComboButtonItem */
+  itemDisabled: bool,
+
```
elycheea, line 145: I think these props should all be set on `ComboButtonItem` and not `ComboButton`.

elycheea (changes requested):
Submitting partial comments — I think things are looking much better, but we have a lot of code commented out and a lot of code that was remove (including classes related to conditionals, but didn’t see styles removed alongside so we should make sure if those styles are no longer necessary, we remove them).

We’ve added props for `ComboButtonItem` on the `ComboButton` itself. Is there a need to duplicate these here? If no, we should move them to just the `ComboButtonItem`. If this was for the story, we can work on the updating the stories next.

packages/cloud-cognitive/src/components/ComboButton/ComboButtonItem/index.js
```diff
@@ -17,15 +17,15 @@ ComboButtonItem.propTypes = {
...
-  danger: bool,
-
   /** Specify whether the `ComboButton` should be disabled, or not */
   disabled: bool,
 
   /** Provide an optional `href` for the `ComboButtonItem` to become an `a` element */
   href: string,
 
+  /** Provide an optional flag to convert `ComboButtonItem` to a delete button */
+  isDelete: bool,
```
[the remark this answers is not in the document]
qbi11y: OverflowMenuItem in Carbon accepts isDelete to render 'danger' button.

packages/cloud-cognitive/src/components/ComboButton/ComboButton.js
qbi11y: I placed these here to make sure they would show up in the story. If there is a better way to make that happen I am all ears!

packages/cloud-cognitive/src/components/ComboButton/ComboButton.js
```diff
@@ -12,100 +12,154 @@ import {
...
+  buttonDisabled: bool,
+
   /** Provide the contents of the `ComboButton` */
   children: node.isRequired,
 
   /** Provide an optional class to be applied to the containing node */
   className: string,
 
+  /** Provide an optional flag to disable the ComboButton or a selected ComboButtonItem */
+  disabled: bool,
```
elycheea, line 135: This does only disables the `ComboButton` (and the overflow menu) and not its children. Let’s update this for accuracy.

packages/cloud-cognitive/src/components/ComboButton/ComboButton.stories.js
```diff
@@ -17,19 +16,112 @@ import { ComboButton, ComboButtonItem } from '..';
...
 import styles from './_combo-button.scss';
 
+const modifiers = {
+  'Is Delete': { isDelete: true },
+  'No Delete': { isDelete: false },
+  'Has Divider': { hasDivider: true },
+  'No Divider': { hasDivider: false },
+  'Item Disabled': { itemDisabled: true },
+  'Item Active': { itemDisabled: false },
+};
```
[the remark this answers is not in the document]
elycheea: This treats the “modifiers” as if they can never stack, but one ComboButtonItem could fit more than one of these conditions, e.g. an option with a divider that is also a delete action.

This is why I recommend replacing this with 

```js
const children = {
  'Three options': [
      <ComboButtonItem key="option-1" renderIcon={Edit16}>
        Option 1
      </ComboButtonItem>,
      <ComboButtonItem key="option-2">
        Option 2
      </ComboButtonItem>,
      <ComboButtonItem key="option-3">
        Option 3
      </ComboButtonItem>,
  ],
  ...
```

to show the different variations.

You could do something like
- Three options (or whatever number — but the guidance uses 3 children so probably a good starting point)
- Options with divider
- Options with disabled
- Options with danger
- Options with [???/all modifiers]

^ or combinations of them. The main thing is that one `ComboButton` could have children with more than one of these states.

packages/cloud-cognitive/src/components/ComboButton/ComboButton.stories.js
```diff
@@ -17,19 +16,112 @@ import { ComboButton, ComboButtonItem } from '..';
...
+      options: ['sm', 'md', 'lg'],
+    },
+    // }
+    modifiers: {
+      control: {
+        type: 'select',
+        labels: Object.keys(modifiers),
+      },
+      options: Object.values(modifiers).map((_k, i) => i),
+      mapping: Object.values(modifiers),
```
elycheea, line 44: If you update with the format I suggest, then this should also be replaced with `children`.

packages/cloud-cognitive/src/components/ComboButton/ComboButton.stories.js
```diff
@@ -17,19 +16,112 @@ import { ComboButton, ComboButtonItem } from '..';
...
 
-export const Default = prepareStory(() => (
-  <ComboButton>
-    <ComboButtonItem>ComboButtonItem 1</ComboButtonItem>
-    <ComboButtonItem renderIcon={CloudApp16}>ComboButtonItem 2</ComboButtonItem>
-    <ComboButtonItem>ComboButtonItem 3</ComboButtonItem>
-  </ComboButton>
-));
+const Template = (args) => {
+  console.log(args);
```
elycheea, line 54: I think we can remove this?

packages/cloud-cognitive/src/components/ComboButton/ComboButton.stories.js
```diff
@@ -17,19 +16,112 @@ import { ComboButton, ComboButtonItem } from '..';
...
+//     <ComboButtonItemDivider />
+//     <ComboButtonItem danger={args.danger}>ComboButtonItem 5</ComboButtonItem>
+//   </ComboButton>
+// );
+
+export const Default = prepareStory(Template, {
+  args: {
+    //hasDivider: false,
+    //isDelete: false,
+    buttonDisabled: false,
```
elycheea, line 106: We should use consistent naming and just use `disabled` here. There’s not really any reason I can see for both so let replace instances of `buttonDisabled` with `disabled`. I believe their intention is the same.

image

elycheea: Also make sure to add `children: 0` here if you update with the suggested changes. — `children: 0` just means that we are using the first of the options we set.

packages/cloud-cognitive/src/components/ComboButton/ComboButton.stories.js
```diff
@@ -17,19 +16,112 @@ import { ComboButton, ComboButtonItem } from '..';
...
+      </ComboButtonItem>
+      <ComboButtonItem disabled={args?.modifiers?.itemDisabled}>
+        ComboButtonItem 3
+      </ComboButtonItem>
+      <ComboButtonItem hasDivider={args?.modifiers?.hasDivider}>
+        ComboButtonItem 4
+      </ComboButtonItem>
+      <ComboButtonItem isDelete={args?.modifiers?.isDelete}>
+        ComboButtonItem 5
+      </ComboButtonItem>
```
elycheea, line 69: If you use `children` as suggested, you will want to replace these as well.

packages/cloud-cognitive/src/components/ComboButton/ComboButton.stories.js
```diff
@@ -17,19 +16,112 @@ import { ComboButton, ComboButtonItem } from '..';
...
-    <ComboButtonItem>ComboButtonItem 1</ComboButtonItem>
-    <ComboButtonItem renderIcon={CloudApp16}>ComboButtonItem 2</ComboButtonItem>
-    <ComboButtonItem>ComboButtonItem 3</ComboButtonItem>
-  </ComboButton>
-));
+const Template = (args) => {
+  console.log(args);
+  return (
+    <ComboButton disabled={args.buttonDisabled} size={args.size}>
+      <ComboButtonItem>ComboButtonItem 1</ComboButtonItem>
```
elycheea, line 57: This one’s already in the Storybook, but I actually recommend we rename this to “Primary button” for more consistency with the way we usually refer to actions.

packages/cloud-cognitive/src/components/ComboButton/_combo-button.scss
```diff
@@ -11,55 +11,118 @@
...
+    width: carbon--mini-units($count: 4);
+  }
+
+  .#{$blockClass}__overflow-menu--md.#{$blockClass}__overflow-menu {
+    width: carbon--mini-units($count: 5);
+  }
+
+  .#{$blockClass}__overflow-menu--lg.#{$blockClass}__overflow-menu {
+    width: carbon--mini-units($count: 6);
+  }
```
elycheea, line 69: Again, mini-units are getting removed. Let’s stick to spacing or rem here.

https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2130#discussion_r953059175

elycheea (changes requested):
I’ll try to summarize requested changes here, but please see commented lines for more detail.

- [ ] Remove conflict markers and fix status
- [ ] Remove unnecessary use of `classnames` package
- [ ] Re-add standard comments (and spacing)
- [ ] Remove unused props (`modifiers`, `buttonDisabled`)

What you’re doing with `modifiers` in `ComboButton.stories.js` is closer to what I had in mind, but also gets redundant. See comments for suggested changes.

- [ ] Rename combo button content
    - This is a nice to have, but looking at examples, we usually name the first one `Primary button`. We can name the `ComboButtonItem`s options to follow the guidance.
- [ ] Remove use of mini-units. Again, [these are deprecated in v11](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2130#discussion_r953059175).
- [ ] Remove `ComboButtonItemDivider`
- [ ] Fix conflict in `story-structure.js`

**Please removed unused code. We have an excessive amount of commented code.**

elycheea:
Replaced by #2075

PR #2138, opened 2022-07-27
fix: add v11 branch to list of branches for ci workflow to run against

matthewgallo:
This should cause the ci workflow to run against any future PRs with `carbon-v11` as the base branch (including this one!). This does highlight some changes it might be worth us adding to the ci workflow on the v11 branch, such as only running checks against the `@carbon/ibm-products` components.

UPDATE:
With the latest changes, we now skip the `cdai` and `security` jobs on the `carbon-v11` branch.

#### What did you change?
`ci.yml`


matthewgallo:
The remaining failing tests are going to be covered in separate PRs so this should be ready for a review now

Issue #2134, opened 2022-07-27
Add workflow for merging main to carbon-v11 branch


elycheea:
## What will this achieve?

Ensures that Carbon 11 branch regularly merges from `main`.

## How will success be measured?

Should create a PR to merge `main` to `carbon-v11` branch on a weekly basis

## Additional information

- Designs
- Existing code
- etc

matthewgallo:
I _think_ the workflow file may need to be added to the `main` branch in order for github actions to pick up on it

elycheea:
@matthewgallo Oooh. Gah, I’ll open another PR for it. Thanks for checking! How did I not notice? 🙃

PR #2151, opened 2022-08-03
chore: add action to sync v11 with main

elycheea:
Contributes to #2134 

Creates (hopefully) a PR from `main` to `carbon-v11` branch — currently set for Mondays 4:30am but open to changing the timing if we prefer not to do this before releases on Tuesdays.

#### What did you change?
Adds GitHub action for opening a PR to merge `main` to `carbon-v11`

Partly via [Act](https://github.com/nektos/act) and partly on forked repo. This will be the real test though. 🤞

Note: Examples didn’t always show needing a token for checkout action, but testing locally via Act required a personal access token. Used the lerna token here since usage seemed similar but can create a new token for this if it doesn’t work / we prefer.

### DIFF
--- a/.github/workflows/sync-main.yml
+++ b/.github/workflows/sync-main.yml
@@ -15,14 +15,14 @@ jobs:
           ref: carbon-v11
           token: ${{secrets.GH_TOKEN_LERNA}}
 
-      - name: Fetch main branch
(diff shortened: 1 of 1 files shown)


matthewgallo (approved):
🎉 Looks great! That's pretty neat you can test actions locally with [Act](https://github.com/nektos/act)!

PR #2152, opened 2022-08-03
fix: use latest Carbon react to fix selectorsFloatingMenus prop type issue on v11 branch

matthewgallo:
Contributes to #2071

This PR upgrades `@carbon/react` to the latest version to prevent the issues we were experiencing with the ComposedModal's `selectorsFloatingMenus` prop, which was changed accidentally from an array of strings to just a string. This caused most of our tearsheet tests to fail.

I ran the tests again after this upgrade and everything effected by this issue is now fixed.

#### What did you change?
```
packages/cloud-cognitive/package.json
packages/core/package.json
yarn.lock
```
Ran tests locally and confirmed the tests that were failing previously are all passing again


davidmenendez (approved):
LGTM but tests are failing

matthewgallo:
I think it's just the AddSelect tests now, then they should all be passing.

PR #2172, opened 2022-08-11

kai-liu:
Contributes to
[https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/1613](url)
[https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/1981](url)

What did you change?
within the Datagrid directory:

`DraggableElement.js` Added selected attribute to help with css selectors
addons/CustomizeColumns/Actions.js Removed search button, as required by design. 

`CustomizeColumns/CustomizeColumnsModal.js` Added logic to show number of visible columns vs total number of columns. Removed elements in accordance to the carbon design document [https://pages.github.ibm.com/cdai-design/pal/components/data-table/column-customization/usage/](url)

`/styles/addons/CustomizeColumnsModal.scss` Style changes. 

storybook


elycheea:
@kai-liu FYI looks like there’s a conflict!

kai-liu:
Comments from a [previous PR](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2062) :

The convention is to use single quotes for string literals and package names, double quotes for React props or element attributes. Are the changes made by prettifier? Can you revert to single quotes (follow convention of existing code) and then do npm run lint to verify?

Seems like I was previously using the a different lint config that has double quotes by default. I have changed the lint config and used `npm run lint` to verify.  

Why are we taking out "reset to default"? I think the request is to add a select all / unselect all checkbox in the customize column modal. It doesn't seem to mention that we need to remove the "reset to default" feature. Is this an explicit request from design?

The changes are made to meet the requirements as illustrated in the [design document ](https://pages.github.ibm.com/cdai-design/pal/components/data-table/column-customization/usage/). It was not an explicit request. I have discussed this with the PAL team, and was advised to follow the document as much as possible.

Can we further simplify the select / unselect all logic? On a high level, when the "select/unselect all" checkbox is clicked, there should be a for-loop that selects / unselects each column checkbox by calling the corresponding on change callback. The existing on change callback doesn't need to be changed. Also, we can probably implement a function which walks the column checkbox states and return a value that indicates if all/some/none column checkboxes are checked. This can be used to populate the select/unselect all checkbox state.

I have simplified the select/ unselect all logic, as well as reverted the change callback to the original callback. The logic uses a for-loop to check/uncheck each checkboxes. It makes the decision by traversing all checkboxes and collect their states. If all checkboxes are 'checked', then for loop will uncheck them, and vice versa. If some checkboxes are checked but not all, then the for loop will check them.

PR #2178, opened 2022-08-15

kai-liu:
Contributes to
[https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/1613](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/url)
[https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/1981](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/url)

What did you change?
within the Datagrid directory:

`DraggableElement.js` Added selected attribute to help with css selectors
`addons/CustomizeColumns/Actions.js` Removed search button, as required by design.

`CustomizeColumns/CustomizeColumnsModal.js` Added logic to show number of visible columns vs total number of columns. Removed elements in accordance to the carbon design document [https://pages.github.ibm.com/cdai-design/pal/components/data-table/column-customization/usage/](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/url)

`/styles/addons/CustomizeColumnsModal.scss` Style changes.

storybook

### DIFF
--- a/config/stylelint-config-ibm-products/package.json
+++ b/config/stylelint-config-ibm-products/package.json
@@ -23,6 +23,6 @@
     "stylelint-config-standard": "^27.0.0",
     "stylelint-config-standard-scss": "^5.0.0",
     "stylelint-no-unsupported-browser-features": "^5.0.3",
-    "stylelint-plugin-carbon-tokens": "2.0.0-beta.14"
+    "stylelint-plugin-carbon-tokens": "2.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -156,7 +156,7 @@
     "**/*": "yarn spellcheck:files"
   },
   "dependencies": {
-    "stylelint-plugin-carbon-tokens": "2.0.0-beta.14"
+    "stylelint-plugin-carbon-tokens": "2.0.0"
   },
   "packageManager": "yarn@3.2.2"
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -15039,7 +15039,7 @@ __metadata:
     rimraf: ^3.0.2
     stylelint: ^14.10.0
     stylelint-config-ibm-products: "*"
-    stylelint-plugin-carbon-tokens: 2.0.0-beta.14
+    stylelint-plugin-carbon-tokens: 2.0.0
     webpack: ^5.74.0
   languageName: unknown
   linkType: soft
@@ -24542,7 +24542,7 @@ __metadata:
     stylelint-config-standard: ^27.0.0
     stylelint-config-standard-scss: ^5.0.0
     stylelint-no-unsupported-browser-features: ^5.0.3
-    stylelint-plugin-carbon-tokens: 2.0.0-beta.14
+    stylelint-plugin-carbon-tokens: 2.0.0
   peerDependencies:
     stylelint: ^14.10.0
   languageName: unknown
@@ -24719,9 +24719,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"stylelint-plugin-carbon-tokens@npm:2.0.0-beta.14":
-  version: 2.0.0-beta.14
-  resolution: "stylelint-plugin-carbon-tokens@npm:2.0.0-beta.14"
+"stylelint-plugin-carbon-tokens@npm:2.0.0":
+  version: 2.0.0
+  resolution: "stylelint-plugin-carbon-tokens@npm:2.0.0"
   dependencies:
     lodash: ^4.17.21
     postcss-scss: ^4.0.4
@@ -24732,7 +24732,7 @@ __metadata:
     "@carbon/motion": ">=10 <= 11"
     "@carbon/themes": ">=10 <= 11"
     "@carbon/type": ">=10 <= 11"
-  checksum: fdd3b54799d7a1befb92e4c523081bfacc8644a00944983ced1babc61adad57d698bd92c1e4bb60d0bb8b3c48c9dc866d4de1ec71c617024ab1cd09d1bc3a4ee
+  checksum: d0a6c9ab0f4792e96ecec96181677f996242b8b3c4c876c15785db1cd284c29610972735a6d98097d5834836a4668db7d5426657f9fa236a643c8b2da707d667
   languageName: node
   linkType: hard
 


packages/cloud-cognitive/src/components/Datagrid/styles/addons/_CustomizeColumnsModal.scss
```diff
@@ -13,30 +13,45 @@
...
     flex-flow: column;
-    /* stylelint-disable-next-line declaration-no-important */
-    padding-top: 0 !important;
+    padding: 0;
+    margin: 0;
   }
 }
 
-.#{$block-class}__customize-columns-checkbox {
+.bx--form-item.bx--checkbox-wrapper.#{$block-class}__customize-columns-checkbox-wrapper {
```
matthewgallo, line 21: Can you add the carbon prefix variable?
```scss
.#{$block-class} .#{$carbon-prefix}--form-item.#{$carbon-prefix}--checkbox-wrapper.#{$block-class}__customize-columns-checkbox-wrapper
```

packages/cloud-cognitive/src/components/Datagrid/styles/addons/_CustomizeColumnsModal.scss
```diff
@@ -13,30 +13,45 @@
...
+}
+
+.#{$block-class}__customize-columns-modal--actions input[role='searchbox'] {
+  height: $spacing-09;
+  padding-left: $spacing-08;
+  border-bottom: none;
+}
+div.#{$block-class}__customize-columns-modal--actions,
+div.bx--search-magnifier,
+.bx--search-magnifier-icon {
```
matthewgallo, line 43: These also need to use the `$carbon-prefix` variable as well

packages/cloud-cognitive/src/components/Datagrid/styles/addons/_CustomizeColumnsModal.scss
```diff
@@ -46,3 +61,22 @@
   position: relative; // needed for the react-dnd, otherwise the drag preview will not work correctly
   overflow: auto;
 }
+
+#customize-columns-select-all {
+  padding-left: $spacing-08;
+}
+
+.bx--checkbox-label {
```
matthewgallo, line 69: Use `$carbon-prefix` and also this rule should be scoped to the datagrid component, ie
```scss
.#{$block-class} .#{$carbon-prefix}--checkbox-label {
...
}
```

packages/cloud-cognitive/src/components/Datagrid/styles/addons/_CustomizeColumnsModal.scss
```diff
@@ -46,3 +61,22 @@
...
+    padding-left: $spacing-07;
+  }
+
+  &.#{$block-class}__customize-columns-select-all {
+    padding-left: $spacing-07;
+    @include carbon--font-weight('semibold');
+  }
+}
+
+#bx--modal-body--modal-2 {
```
matthewgallo, line 80: This id changes each time the modal is rendered so it's not a reliable selector here. You could potentially use something like:
```scss
.#{$block-class} .#{$block-class}__customize-columns-modal .#{$carbon-prefix}--modal-content {
...
}
```

packages/cloud-cognitive/src/components/Datagrid/styles/addons/_CustomizeColumnsModal.scss
kai-liu: Removed the `form-item` and `checkbox-wrapper` selectors as they are unecessary

packages/cloud-cognitive/src/components/Datagrid/styles/addons/_CustomizeColumnsModal.scss
kai-liu: added `$carbon-prefix`

packages/cloud-cognitive/src/components/Datagrid/styles/addons/_CustomizeColumnsModal.scss
kai-liu: replaced `.bx--checkbox-label` with `.#{$block-class} .#{$carbon-prefix}--checkbox-label`

kai-liu: Also moved the rule to `datagrid.scss`

packages/cloud-cognitive/src/components/Datagrid/styles/addons/_CustomizeColumnsModal.scss
kai-liu: replaced `#bx--modal-body--modal-2` with `.#{$block-class} .#{$block-class}__customize-columns-modal .#{$carbon-prefix}--modal-content`

kai-liu:
Comments from  https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2062 :

The convention is to use single quotes for string literals and package names, double quotes for React props or element attributes. Are the changes made by prettifier? Can you revert to single quotes (follow convention of existing code) and then do npm run lint to verify?

Seems like I was previously using the a different lint config that has double quotes by default. I have changed the lint config and used npm run lint to verify.

Why are we taking out "reset to default"? I think the request is to add a select all / unselect all checkbox in the customize column modal. It doesn't seem to mention that we need to remove the "reset to default" feature. Is this an explicit request from design?

The changes are made to meet the requirements as illustrated in the [design document ](https://pages.github.ibm.com/cdai-design/pal/components/data-table/column-customization/usage/). It was not an explicit request. I have discussed this with the PAL team, and was advised to follow the document as much as possible.

Can we further simplify the select / unselect all logic? On a high level, when the "select/unselect all" checkbox is clicked, there should be a for-loop that selects / unselects each column checkbox by calling the corresponding on change callback. The existing on change callback doesn't need to be changed. Also, we can probably implement a function which walks the column checkbox states and return a value that indicates if all/some/none column checkboxes are checked. This can be used to populate the select/unselect all checkbox state.

I have simplified the select/ unselect all logic, as well as reverted the change callback to the original callback. The logic uses a for-loop to check/uncheck each checkboxes. It makes the decision by traversing all checkboxes and collect their states. If all checkboxes are 'checked', then for loop will uncheck them, and vice versa. If some checkboxes are checked but not all, then the for loop will check them.

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/CustomizeColumnsModal.js
```diff
@@ -76,12 +80,29 @@ const CustomizeColumnsModal = ({
...
+
+    columnObjects.forEach((col) => {
+      if (col.isVisible) {
+        result++;
+      }
+    });
+    return result;
+  }, [columnObjects]);
+
   const string = searchText.trim().toLowerCase();
```
holmansze, line 94: You may simplify this by something like:

```
columnObjects.filter((col) => col.isVisible).length
```

holmansze (changes requested):

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -90,6 +136,25 @@ const Columns = ({
               ? instructionsLabel
               : disabledInstructionsLabel}
           </span>
+          <div
+            id={'customize-columns-select-all'}
+            className={setSelectAllColumnClassName()}
```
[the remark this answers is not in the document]
kai-liu: Replaced `className={setSelectAllColumnClassName()}` with `<div className={
   classNames({
      [`${blockClass}__draggable-handleHolder`]: getVisibleColumnsCount() === 0,
      [`${blockClass}__draggable-handleHolder-selected`]: getVisibleColumnsCount() > 0,
   })`

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -54,6 +56,50 @@ const Columns = ({
...
+      `li.${blockClass}__draggable-handleHolder`
+    );
+
+    list.forEach((li, i) => {
+      if (!columns[i].id.includes(id)) {
+        return;
+      }
+      li.setAttribute('selected', isVisible);
+    });
+  };
```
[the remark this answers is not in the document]
kai-liu: Removed `setColumnStatus` and `initializeList`. Revised logic to handle attribute values by passing a `selected` prop to the `DraggableElement` component during render. This reduces code clunkiness a lot. Thanks for the advice!

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/CustomizeColumnsModal.js
kai-liu: Thanks for the suggestion. Replaced the code block with `columnObjects.filter((col) => col.isVisible).length`

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -54,6 +57,19 @@ const Columns = ({
...
+      `.${blockClass}__customize-columns-modal .bx--checkbox`
+    );
+    let selectAll = getVisibleColumnsCount() != columns.length;
+
+    for (let i = 1; i < checkboxes.length; i++) {
+      checkboxes[i].checked = selectAll;
+      onSelectColumn(columns[i - 1], selectAll);
+    }
+  };
+
```
holmansze, line 72: Instead of setting the `checked` attributes explicitly, update the `columns` array. Once the `columns` array is updated, React will re-render the checkboxes with the updated state.

The following code from `CustomizeColumnsModal.js` may be a useful reference. Take a look and see a similar technique can be applied here:
https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/94f3ba91fefcc8acdcb8dc12d3db6d2287e0bc81/packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/CustomizeColumnsModal.js#L65

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -90,6 +106,27 @@ const Columns = ({
...
+            id={'customize-columns-select-all'}
+            className={classNames(`${blockClass}__draggable-handleHolder`, {
+              ['-selected']: getVisibleColumnsCount() > 0,
+            })}
+            selected={getVisibleColumnsCount() > 0}
+          >
+            <Checkbox
+              wrapperClassName={`${blockClass}__customize-columns-checkbox-wrapper`}
+              className={`${blockClass}__customize-columns-select-all`}
+              checked={getVisibleColumnsCount() == columns.length}
```
holmansze, line 119: In general, when doing this kind of comparison, use `===` instead. `==` performs type conversion, `===` is a strict comparison which takes data type into account.

You may try the following experiment:
- 0 == false (evaluates to true)
- 0 === false (evaluates to false)

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -54,6 +57,19 @@ const Columns = ({
     },
     [columns, setColumnsObject]
   );
+
+  const setSelectAllCheckBox = () => {
+    let checkboxes = document.querySelectorAll(
+      `.${blockClass}__customize-columns-modal .bx--checkbox`
+    );
+    let selectAll = getVisibleColumnsCount() != columns.length;
```
holmansze, line 65: Use `!==` instead of `!=`.

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/CustomizeColumnsModal.js
```diff
@@ -62,7 +65,8 @@ const CustomizeColumnsModal = ({
   const onCheckboxCheck = (col, value) => {
     const changedDefinitions = columnObjects.map((definition) => {
       if (definition.id === col.id) {
-        return { ...definition, isVisible: value };
+        definition.isVisible = value;
+        return definition;
       }
```
holmansze, line 70: The original code should work and is a better approach, because it creates a new object with the new `isVisible` value (as opposed to mutating the `definition` object).

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/CustomizeColumnsModal.js
```diff
@@ -106,6 +120,8 @@ const CustomizeColumnsModal = ({
       />
       {isOpen && (
         <Columns
+          setVisibleColumns={setVisibleColumns}
```
holmansze, line 123: Do we need to introduce `setVisibleColumns`? Does the existing `setColumnsObject` satisfy our needs?

holmansze (changes requested):

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
kai-liu: Removed `checkboxes[i].checked = selectAll;` since it is redundant

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
kai-liu: Replaced `==` with `===`

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/CustomizeColumnsModal.js
kai-liu: `setVisibleColumns` is not needed in the context of `Columns`. I have removed it.

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/CustomizeColumnsModal.js
kai-liu: Reverted to `return { ...definition, isVisible: value };`

kai-liu: Implemented logic that differentiates array parameters and object parameters.

Issue #2163, opened 2022-08-11
v11: add new workflow file to handle v11 releases


matthewgallo:
## What will this achieve?

To support the new release cadence with support for v10 and v11 support, we need to add another workflow file to make releases against the `carbon-v11` branch.

For future reference, here are some of the publish commands we'll need to use that I've tested against a sample project

# Publish commands
### v1 publish command
When we still want Carbon v10 to be the latest dist tag/default installed version by npm/yarn:
`yarn lerna publish --conventional-commits --create-release github --yes`
When we want Carbon v10 to have a 1.x dist tag (ie only when v2 has official release, non release candidate):
`yarn lerna publish --dist-tag=1.x --conventional-commits --create-release github --yes`

### ~~Initial v2 release candidate publish command (only to be run ONCE):~~
~~`yarn lerna publish premajor --conventional-commits --preid rc --dist-tag next --ignore-changes 'config/**/*' 'packages/security/**/*' 'packages/cdai/**/*' --yes`~~

### Subsequent v2 release candidate publish command:
`yarn lerna publish --preid rc --dist-tag next --create-release github --yes`

### Initial v2 publish of prerelease package to major version (non-prerelease, only to be run ONCE):
`yarn lerna version --conventional-commits --conventional-graduate=ibm-products`

### Subsequent v2 publish command:
`yarn lerna publish --conventional-commits --create-release github --yes`

- [x] Create initial 2.x.x release candidate, https://www.npmjs.com/package/@carbon/ibm-products/v/2.0.0-rc.0
- [x] Update publish script in v11 workflow file, https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2180

## How will success be measured?
When we're able to have a release from `main` and `carbon-v11` branches.

## Additional information

- Designs
- Existing code
- etc

matthewgallo:
We also need to make some updates to our readme, as there is mentions of installing `carbon-components-react` which is only for Carbon v10 and prior.

PR #2191, opened 2022-08-17
chore(v11): add weekly release for v11

matthewgallo:
Contributes to #2163 

Now that we've made the first release candidate and the first release candidate version bump, I think it makes sense for us to now be able to run this release action on a weekly basis. It will run at the same time as the other version updates (Tues 4:30am).

#### What did you change?
`.github/workflows/release-v11.yml`
Uses the same cron scheduling as our `main` branch releases


elycheea:
@allcontributors add @matthewgallo for infra

elycheea (approved):

Issue #2174, opened 2022-08-11
Data table: Scrolling scrollbar styles


elycheea:
# Data table extension: Scrolling

## Brief description

This issue covers the scrollbar styling and behavior in the data table. This issue is broken out from #1977

## Details

[Link to design](https://pages.github.ibm.com/cdai-design/pal/components/data-table/scrolling/usage/#styling)

Our implementation most likely would follow the browser default scrollbar although WKC does use a custom scrollbar in the [Infinite Scroll](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary--infinite-scroll) and [Ten Thousand Entries](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary--ten-thousand-entries) stories.

> ### **Scroll bar**
> 
> WKC scrollbar is always present when horizontal scroll is enabled and styling is strange - the bars cut off the content that they normally float above:
> ![Screen Shot 2022-05-09 at 5 11 38 PM](https://user-images.githubusercontent.com/48838320/167518308-63386035-98e6-4bf1-b33c-2dc394434a18.png)
> _Note: Example is from Chrome_
> 
> PAL scrollbar appears on hover, and floats above the content without cutting it off (standard browser behavior):
> <img width="796" alt="Screen Shot 2022-05-09 at 5 15 38 PM" src="https://user-images.githubusercontent.com/48838320/167518446-6033607f-da18-44ef-b3ee-ec6b60531f52.png">

### Links to other materials

e.g. sketch files, icons, images, prototypes

## Design Maintainer(s)

- Marion Brülls
- Johanna Koval

## Tasks

Before starting work on this epic, please review and complete the following.

- [ ] Initial review of design/existing code

### Working in Carbon for IBM Products package

- [ ] Have you reviewed our
      [contribution guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/CONTRIBUTING.md)?
- [ ] Have you noted our
      [code guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/docs/CODE_GUIDELINES.md)?
- [ ] Did you use the CLI to scaffold your new component?

elycheea:
Also note that the Infinite Scroll and Ten Thousand Entries stories include an empty Table Toolbar.

[...]

PR #2193, opened 2022-08-18
chore(Datagrid): scrollbar styling

Ratheeshrajan:
Contributes to #2174 

Covers the scrollbar styling and behavior in the data table

#### What did you change?
- Nawwow scrollbar styling for WebKit browser 
- Datagrid horizontal scrollbar appears on hover.

Storybook

### DIFF
--- a/packages/cloud-cognitive/src/components/Datagrid/styles/_datagrid.scss
+++ b/packages/cloud-cognitive/src/components/Datagrid/styles/_datagrid.scss
@@ -150,7 +150,23 @@
   .#{$carbon-prefix}--data-table-content {
     width: 100%;
     height: 100%;
+    overflow-x: hidden;
+  }
+
+  .#{$carbon-prefix}--data-table-content:hover {
     overflow-x: auto;
+  }
+
+  .#{$carbon-prefix}--data-table-content:hover::-webkit-scrollbar {
+    height: 8px;
+    background-color: transparent;
+    opacity: 0;
+  }
+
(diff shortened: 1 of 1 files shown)


elycheea (changes requested):
While I do think this is a clever approach to getting the scrollbar to appear on hover only, since this is also OS specific behavior, it may be better for us to clarify with the designers who wrote the guidance on this. It’s probably better to keep with the OS settings rather than creating custom behavior so that this remains in the user’s control. Also, when the overflow style changes, we also can see a shift in the DOM as the scrollbar appears. This is more noticeable on the pagination story. 

That said, we should still investigate whether or not we need custom scrollbar styling as opposed to the browser default here too. In the [Ten Thousand Entries](https://deploy-preview-2193--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary--ten-thousand-entries) story, we do add a custom “virtual scrollbar” so perhaps our custom scrollbar is to keep vertical/horizontal styles consistent?

elycheea:
@Ratheeshrajan See comments from earlier when you get the chance, but can you also resolve the conflict in datagrid.scss?

Ratheeshrajan:
@elycheea The changes in the current PR have been removed and merged in [2171](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2171/files). Are we still required to keep the changes?

elycheea:
Good point. Let’s close this for now. Since the changes themselves were fairly small, we can make them again if still necessary.

Issue #2184, opened 2022-08-16
Multi Add Select design review


OliverKauselmann:
## Design review

Component epic #

### Standards

- [ ] All pattern updates/changes/iterations have been discussed with the
      component developer
- [ ] Patterns have been approved by either DSAG or another approving entity

### Pattern and behavior

- [ ] The component behaves according to the guidelines set by the pattern
      maintainer
- [ ] The component’s UI matches the pattern specifications set by the pattern
      maintainer
- [ ] The component’s motion matches the specifications set by the pattern
      maintainer(s)
- [ ] The UI produced is responsive, cross-browser, and responds to the
      currently set Carbon theme.
- [ ] Colors, themes, sizes and additional props are true and correct, aligning
      with Carbon set tokens (unless otherwise specified by the pattern)
- [ ] Paddings/margins/spacings are true and correct, in both desktop and mobile
      views

### Storybook

- [ ] A functioning component renders in Storybook
- [ ] The Storybook displays multiple scenarios that show how the component can
      be used
- [ ] Changing props in the Storybook updates the component

OliverKauselmann:
Hi @davidmenendez, here is a list of findings reviewing the implementation:

**1. Spacing left**
Please adjust the 32px indentation from the left, with 16px of searchfield and lines
Compared to the layout

**2. Searchfield**
Use large component and remove label on searchfield 
(use searchfield component)

**3. RHS Panel general**
Do not scale the RHS sidepanel when resizing the window. Keep fixed width of 360 px.
Empty state should have a spacing of 48px to the top
Please adjust typo tokens and shorten line length

**4. Table row spacing and interactions** 
Please correct row spacing from border, checkbox to label
Row separator lines missing when selected
Please implement all interactive table row states (see also [single select comments](https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/2014#issuecomment-1155246087))

Carbon guidance on clickable data table row items
https://pages.github.ibm.com/cdai-design/pal/components/data-table/clickable-row-items/usage/#style

**5. RHS Selection items**
Please compare spacing and adjust where needed
If items are collapsed arrows point to the right
Add a remove icon and functionality on hover
Horizontal title line doubles to 2px line when selection is done (overlays empty state?)

**6. RHS Info panel**
Trigger panel with click on eye and also in parallel by click on complete row
The eye-icon should only be visible on hover
Show blue line indicator in the row that refers to the info panel content

[...]

PR #2195, opened 2022-08-18
fix: Add select design review updates

davidmenendez:
Contributes to #2184 

there are still some follow up questions i have for design, so this doesn't address every single point, but should handle basically everything else.


packages/cloud-cognitive/src/components/AddSelect/AddSelectList.js
```diff
@@ -85,31 +90,51 @@ export let AddSelectList = ({
...
-                          onChange={(checked) =>
-                            handleMultiSelection(item.id, checked)
-                          }
-                          labelText={item.title}
-                          id={item.id}
-                          checked={isSelected(item.id)}
-                          className={`${blockClass}-checkbox-wrapper`}
-                        />
+                        {
+                          // hacky way to prevent checkbox from triggering the meta onclick handler
```
matthewgallo, line 124: Can you call `event.stopPropagation()` in the checkbox onChange event? Or does that not prevent the click from bubbling up?

matthewgallo: Maybe something like this:
```jsx
onChange={(value, id, event) => {
    event.stopPropagation();
    handleMultiSelection(value, id);
}}
```

matthewgallo (approved):

Issue #2173, opened 2022-08-11
Data table: Scrolling truncation and text wrapping


elycheea:
# Data table extension: Scrolling

## Brief description

This issue covers the truncation and text wrapping behaviors of the data table scrolling pattern. This issue is broken out from #1977 

## Details

[Link to design](https://pages.github.ibm.com/cdai-design/pal/components/data-table/scrolling/usage/#truncation-and-text-wrap)

Our current implementation of truncated text uses the browser default title attribute, but guidelines suggest we should use Carbon tooltip.
We’ll also need to add scenarios for wrapping text to two lines when space is narrow.

> ### **Tooltip on truncation**
> 
> WKC table uses browser tooltips on truncation:
> ![Screen Shot 2022-05-09 at 5 06 21 PM](https://user-images.githubusercontent.com/48838320/167517694-f662a54a-250e-4721-8cab-99b5d471542f.png)
> 
> PAL uses Carbon tooltips:
> 
> > Text should be truncated optically, not in the underlying code (technical truncation). This ensures that data is preserved and accessible to screen readers. On hover, truncation should display a Carbon tooltip (Carbon tooltips afford more control over delay and placement). 
> <img width="375" alt="Screen Shot 2022-05-09 at 5 08 52 PM" src="https://user-images.githubusercontent.com/48838320/167517837-55d74577-7449-48b8-97a2-8ad2fd5a988a.png">

### Links to other materials

e.g. sketch files, icons, images, prototypes

## Design Maintainer(s)

- Marion Brülls
- Johanna Koval

## Tasks

Before starting work on this epic, please review and complete the following.

- [ ] Initial review of design/existing code

### Working in Carbon for IBM Products package

- [ ] Have you reviewed our
      [contribution guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/CONTRIBUTING.md)?
- [ ] Have you noted our
      [code guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/docs/CODE_GUIDELINES.md)?
- [ ] Did you use the CLI to scaffold your new component?

elycheea:
@marion-bruells When we were working on our v11 migration, @matthewgallo saw that tooltips are now meant more for interactive child elements (e.g. a button). For consistency between v10 and v11, we were thinking it may be better to stick the browser default title attribute/tooltip in this case.

[...]

PR #2196, opened 2022-08-18
feat(Datagrid): add multi line wrap option

matthewgallo:
Contributes to #2173 

Adds the option to specify `multiLineWrap` on a per column basis. There is an example added to the basic usage story on the last column.
Screen Shot 2022-08-18 at 3 23 02 PM

#### What did you change?

```
packages/cloud-cognitive/src/components/Datagrid/useDefaultStringRenderer.js
```
Storybook

### DIFF
--- a/packages/cloud-cognitive/src/components/Datagrid/styles/_datagrid.scss
+++ b/packages/cloud-cognitive/src/components/Datagrid/styles/_datagrid.scss
@@ -195,6 +195,13 @@
     white-space: nowrap;
   }
 
+  .#{$block-class}__defaultStringRenderer.#{$block-class}__defaultStringRenderer-multiline {
+    display: -webkit-box;
+    -webkit-box-orient: vertical;
+    -webkit-line-clamp: 2;
+    white-space: initial;
+  }
+
   .#{$block-class}__expanded-row {
     display: flex;
     overflow: hidden;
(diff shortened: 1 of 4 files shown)


packages/cloud-cognitive/src/components/Datagrid/styles/_datagrid.scss
```diff
@@ -195,6 +195,13 @@
     white-space: nowrap;
   }
 
+  .#{$block-class}__defaultStringRenderer.#{$block-class}__defaultStringRenderer-multiline {
```
elycheea, line 198: I think for BEM, `multiline` should be a modifier so can we rename this to `...__defaultStringRenderer--multiline`?

elycheea (changes requested):
Minor change — I think it should be `--multiline` instead of `-multiline` since I consider it a modifier?

Also might be good to include an example of the text wrapping in one of our stories although we can address this later when we do the story updates as well.

packages/cloud-cognitive/src/components/Datagrid/styles/_datagrid.scss
matthewgallo: Fixed [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2196/commits/6e2150100777970185d6485f71b4da188a4c0564)

elycheea (approved):
🙌

Issue #2077, opened 2022-07-07
MultiAddSelect throws error when changing items.entries


zachhardesty7:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)

## Detailed description

`MultiAddSelect` throws an error when removing an entry from `items.entries` after interacting with it. For example, if you don't permit "adding" duplicate entries and you're fully controlling the component, when you try to filter out duplicates in the `onSubmit` handler the component throws an error. The CodeSandbox makes it pretty clear. This shouldn't throw an error and instead just not render the removed entries.

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

`latest`

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

ISV

## Steps to reproduce the issue

1. Load CodeSandbox...
2. Press the big blue button: "Show MultiAddSelect"
3. Add 1-3 of the entries
4. Press the next big blue button: "Add"
5. Error

> Please create a reduced test case in CodeSandbox

https://codesandbox.io/s/pedantic-sky-okbfoi?file=/src/Example/Example.js

## Additional information

- This can technically be worked around by wrapping the component with a boolean conditional to only render when `showTearsheet` is `true`, but then there's no exit animation 🙁 
- Let me know if anything else is needed from me 🙂

davidmenendez:
thanks for opening this up. this appears to be a problem where the internal state isn't being reset when you close the component. i will investigate this and try to find a solution.

PR #2219, opened 2022-08-23
fix: reset state on submit or close

davidmenendez:
Contributes to #2077

fixes a bug that was being caused when the user submitted or closed the add select component. basically the problem was being caused by the internal state not resetting when the component closed.


AlexanderMelox (approved):
lgtm 🚀

Issue #2107, opened 2022-07-18
InlineEdit bug, placeholder text not loading properly


David-Yuen:
I'm seeing a bug with InlineEdit, where the placeholder text doesn't load properly. I think it has something to do with render not executing or animation isn't executed.. (i noticed a little bit animation for the placeholder text to fade in?)
Generally, if I use a condition to load `InlineEdit`, for example if we're loading/processing the values in the first few react cycles, and we're going to show the Skeleton loader while waiting.. it doesnt work.

## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.
Generally, if I use a condition to load `InlineEdit`, for example if we're loading/processing the values in the first few react cycles, and we're going to show the Skeleton loader while waiting.. it doesnt work.

> Is this issue related to a specific component?
InlineEdit

> What did you expect to happen? What happened instead? What would you like to
> see changed?
I expected the placeholder text to load

> What browser are you working in?
Chrome

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?
`1.23.0`

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
IBM Cloud

## Steps to reproduce the issue

1st screenshot: The placeholder text is missing from the InlineEdit
<img width="416" alt="Screen Shot 2022-07-18 at 3 10 33 PM" src="https://user-images.githubusercontent.com/37594705/179596050-47b8ea76-e4bd-4f9d-8091-2f019a08ac47.png">

2nd screenshot: I clicked (activated) on the InlineEdit, the placeholder text fades in
<img width="437" alt="Screen Shot 2022-07-18 at 3 11 23 PM" src="https://user-images.githubusercontent.com/37594705/179597147-79f80186-2b83-4453-9f86-b467a99afadd.png">

3rd screenshot: Without entering text, I removed focus from the InlineEdit (click somewhere else on the page). Placeholder text is visible (as it should on component load)
<img width="434" alt="Screen Shot 2022-07-18 at 3 12 00 PM" src="https://user-images.githubusercontent.com/37594705/179597996-725b2e1a-6dd1-48f9-a1df-f59782ea3c94.png">

Here is a code snippet to reproduce the issue:
```
const MyComponent = () => {
[...]

PR #2222, opened 2022-08-23
fix: Inline edit 2107

davidmenendez:
Contributes to #2107 

fixes an issue where if no value was supplied to inline edit it causes the placeholder text to not show.

i'll be honest, i'm not as familiar with the need for so much need of `ref`'s in this component. the problem appeared to be related to the fact that `refInput` wasn't being set until the user clicked inside the input when no value prop was supplied. this fix changes that check to use `value` instead of `refInput`. @lee-chase i know you have a lot of experience with this component. perhaps you should review this change and let me know if this works or not since it appears to be a different approach.

you can verify this behavior by going to `InlineEdit.stories.js` and in the template props setting `value: ''` and adding a `placeholder` prop (assuming you're on the main branch and not this one).

also, i noticed this on line 292 `data-placeholder={placeholder ?? labelText}` i don't think this is necessary. the user either wants a placeholder or not.

### DIFF
--- a/packages/cloud-cognitive/src/components/InlineEdit/InlineEdit.js
+++ b/packages/cloud-cognitive/src/components/InlineEdit/InlineEdit.js
@@ -72,7 +72,7 @@ export let InlineEdit = React.forwardRef(
     },
     refIn
   ) => {
-    const refInput = useRef(null);
+    const refInput = useRef({ innerText: value });
     const localRef = useRef(null);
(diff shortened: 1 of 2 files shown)


davidmenendez:
thanks lee. i'll keep working on it

lee-chase (commented):
LGTM

lee-chase (approved):

Issue #1924, opened 2022-04-15
Data table: Implement inline editing 


johanna-koval:
# Data table extension: Inline editing

## Description 

WKC data table does not include any of the features and behaviors of the PAL Inline editing pattern:
https://pages.github.ibm.com/cdai-design/pal/components/data-table/inline-editing/usage

There is an initial prototype of some of the edit behaviors in this storybook:
https://pages.github.ibm.com/Jan-Hassel/spss-prototype-components/?path=/story/components-variableview--default

Additional details from #1616 

> ## Details
> [Link to Design](https://pages.github.ibm.com/cdai-design/pal/components/data-table/inline-editing)
> 
> ## Design Maintainer(s)
> * Marion Brülls
> * Virginia Nicholson
> * Mitchell Bernstein
> 
> ## Tasks
> Before starting work on this epic, please review and complete the following.
> 
> * [ ]  Initial review of design/existing code
> 
> ### Working in Carbon for IBM Products package
> * [ ]  Have you reviewed our
>   [contribution guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/CONTRIBUTING.md)?
> * [ ]  Have you noted our
>   [code guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/docs/CODE_GUIDELINES.md)?
> * [ ]  Did you use the CLI to scaffold your new component?

David-Yuen:
So i've manually implemented using the InlineEdit, found quite a few issues and I hope you guys resolve them. Once your implementation here is ready, I hope I can just replace my current InlineEdit component with this

janhassel:
@matthewgallo design review based on PR #2269 deploy preview:

| # | Screenshot | Description |
| :- | :- | :- |
| 1 | ![Screen Recording 2022-09-15 at 09 40 11](https://user-images.githubusercontent.com/28265588/190345215-a5cb6dd7-631c-4be2-add5-910f5d23efcf.gif) | The region focus indicator (lightblue) influences the grid's dimensions |
| 2 | <img width="201" alt="image" src="https://user-images.githubusercontent.com/28265588/190345574-4e035205-5bcb-4a5f-bac6-93aaad139677.png"> | The cursor changed to `pointer` even for cells that don't support editing. Should be `default` |
[...]

PR #2228, opened 2022-08-25
feat(Datagrid): inline edit - useInlineEdit hook, text and number types

Ratheeshrajan:
Contributes to #1924

 Include Inline editing in the data table for data types text, number, date, and select. Some of the suggestions from Jan that have been included are:
1. Ability to navigate the active cell in the grid area with arrow keys or click
2. When a user leaves the grid (e.g. by tabbing away) and re-enters the grid later (with no page refresh or similar), the focus should be set on the last cell they focused before leaving
3. Added `title` attribute to inline edit cell component
4. Ability to pass props to Carbon component used for inline editing
5. If using `useInlineEdit`, cells should have role `gridcell`

Ps: Other data types will be covered later

#### What did you change?
- Hook to render different inline edit data types
- Text/String edit type 
- Number type

Storybook


matthewgallo:
Updates will be made to this PR to support some of the missing pieces @janhassel pointed out, specifically adding cell buttons when using `useInlineEdit` and arrow key support.

packages/cloud-cognitive/src/components/Datagrid/styles/_useInlineEdit.scss
```diff
@@ -0,0 +1,166 @@
...
+* restricted by GSA ADP Schedule Contract with IBM Corp.
+*/
+
+@import '../../../global/styles/project-settings';
+@import './variables';
+
+$row-heights: (
+  xs: 1.5rem,
+  sm: 2rem,
+  md: 2.5rem,
```
elycheea, line 15: I don’t mind us keeping this in here, but based on our [data table settings](https://pages.github.ibm.com/cdai-design/pal/components/data-table/settings/usage/#overview), `md` is just Carbon 11, although this way we’re less likely to miss this.

elycheea (approved):
LGTM!

As discussed over Slack, some keyboard interactions such as <kbd>Home</kbd>/<kbd>End</kbd> (also with <kbd>Ctrl/Cmd</kbd>) and then undo, redo, and clickOutside can be addressed in another PR.

Issue #2237, opened 2022-08-29
Datagrid: Inline edit add `date` type


matthewgallo:
## What will this achieve?

This will add support for the `date` type to the `useInlineEdit` hook for the Datagrid component.

![image](https://user-images.githubusercontent.com/10215203/187211691-0b2d5948-9e97-4748-bb3b-7419525ce41a.png)

## Additional information

- Designs
- Existing code
- etc

PR #2238, opened 2022-08-29
feat(Datagrid): #2237 datagrid inline edit `date` type

matthewgallo:
Contributes to #2237 

**The other two inline edit PRs should be reviewed first.

Adds the `date` type to Datagrid inline edit. This addition includes Carbon's `DatePicker` and `DatePickerInput` components. To add a column that uses `date` inline edit, it will look like the following:

```
{
      Header: 'Active since',
      accessor: 'activeSince',
      inlineEdit: {
        type: 'date',
        inputProps: {
          // optionally pass props here to be passed through to Carbon's DatePicker component
          onChange: (newDateObj, cell) => {
            console.log(newDateObj, cell);
          },
          labelText: 'Change active since date',
          // optionally pass props here to be passed through to Carbon's DatePickerInput component
          datePickerInputProps: {
            labelText: 'Change active since date',
          },
        },
      },
    },
```

The onChange gives back the new date object (which the consumer will need to format on their own) and the `cell` object from the datagrid data. The new data will need to be updated by the consumer because it's required that the date be formatted in the `dateFormat` passed to `DatePicker`. I've included an example in the storybook doing so.

See interaction below:

https://user-images.githubusercontent.com/10215203/189397232-1fad75b2-b300-4dd9-b1c4-5ca6166275d1.mov

#### What did you change?
`useInlineEdit.js`
`getInlineEditColumns.js`
`Datagrid.stories.js`
`_useInlineEdit.scss`
`makeData.js`
Storybook


Ratheeshrajan:
@matthewgallo Noticed a small alignment issue while using the column resize in `select` type columns

https://user-images.githubusercontent.com/305492/189659155-57c8b1a1-a1d8-4885-a719-078a087bbe4c.mov

Ratheeshrajan (approved):
The changes look good to me.

Issue #2215, opened 2022-08-23
<IdeNavigation> incorrectly requires both `aria-label` and `aria-labelledby`


s100:
## What package(s) are you using?

- [ ] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [x] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

`<IdeNavigation>` has the following prop:

https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/f3eb2a8ccda451aa4f7cb9a76925bdb4d0fc4e64/packages/cdai/src/components/IdeNavigation/Navigation/IdeNavigation.js#L96-L102

This requires both `aria-label` and `aria-labelledby` to be specified as strings. We have had an accessibility issue raised where in fact it is incorrect to have both of these. Either `aria-label` provides the actual label string, or `aria-labelledby` provides the selector for the element which provides the actual label string - having both makes the correct label ambiguous.

In practice, when both are present, `aria-labelledby` takes priority. In our case, we use a direct `aria-label`, so we have no choice but to specify `aria-labelledby=''`. Screen readers take `aria-labelledby=''` to be authoritative, and so the screen readers conclude that there is no element to provide the label, and therefore no label at all. It is not possible to satisfy both screen readers and the prop type requirements at once.

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

`@carbon/ibm-cloud-cognitive-cdai@1.22.12`

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM App Connect Enterprise. If this could be fixed well before mid-September 2022 that would be nice.

s100:
https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2243 was merged on 1 September 2022. Now awaiting a new release of `@carbon/ibm-cloud-cognitive-cdai` with this fix in.

elycheea:
@s100 Some changes to our auth in the Carbon Design org caused our recent publishes to fail, but we should have fixed it and your fixes should be out on the 13th! Thanks for contributing.

s100:
Fixed in `@carbon/ibm-cloud-cognitive-cdai@1.22.19`.

PR #2243, opened 2022-08-31
fix: Require only one of `arial-label` or `aria-labelledby` in `<IdeNavigation>`

s100:
Contributes to #2215

#### What did you change?

Previously, `<IdeNavigation>`'s prop `ariaAttributes` incorrectly required *both* `aria-labelledby` *and* `aria-label` to be specified. Now, only one or the other of `aria-labelledby` and `aria-label` is required. If both are supplied, there is no prop type error, so we will have backwards compatibility. (However, this will still cause the same accessibility issue mentioned in #2215.)

The default `ariaAttributes` prop now only specifies `aria-label`, not `aria-labelledby`. The example value of `aria-labelledby` was `'Example Aria Labeled by'` which was never [a meaningful value for this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) in any case.

Manually... also CI will go a long way here I hope.


davidmenendez (approved):
LGTM

Issue #2213, opened 2022-08-23
Accessibility issue with <IdeHome> cards with external links


s100:
## What package(s) are you using?

- [ ] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [x] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

We make use of `<IdeHome>`, which supports a prop `sets`, which is an array of set objects, where each set has a property `tasks`. In our case, one of those tasks is:

```js
const card = {
  cardLinkType: 'external',
  cardHref: 'https://ibm.biz/ACEv11ContainerDocs',
  cardSize: 'standard',
  taskHeading: translate('homepage.cards.learnMore.title'),
  taskHeadingDescription: translate('homepage.cards.learnMore.description'),
  taskGraphic: <Education />,
  taskId: 'learn_more'
}
```

Note the `cardLinkType` of `'external'`, and the supplied `cardHref`. The card renders like this:

![image](https://user-images.githubusercontent.com/9932290/186166465-2d7434fd-646a-4fc4-9545-e12fefc2c2ea.png)

When the user clicks on the card, our hyperlink is opened in a new browser tab. We have had an accessibility issue raised explaining that the user should be warned when a link opens in a new window, but the card does not give any such warning. It does provide a little icon indicating this, but the icon is not useful for users using screen readers. My suggestion would be to add alt text to this icon saying "Opens in new window", or similar, but I will leave it with you.

> Is this issue related to a specific component?

`<IdeHome>`

> What did you expect to happen? What happened instead? What would you like to
> see changed?

See above.

> What browser are you working in?

Chrome, although this isn't a browser-specific issue

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

`@carbon/ibm-cloud-cognitive-cdai@1.22.12`

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM App Connect Enterprise. We would like to be able to fix this issue by mid-September 2022.

## Steps to reproduce the issue

See above.

## Additional information

See above for screenshots.

elycheea:
@s100 The CDAI does not have active maintainers at this time. We’re happy to review any PRs if you’d like to contribute these changes though.

s100:
[...]

PR #2244, opened 2022-08-31
fix: Update IdeCard.js

s100:
Contributes to #2213

#### What did you change?

Added alt text to the "opens in new tab" icon on `<IdeCard>`s of type `'externalLink'`.

Manual verification, plus CI.


packages/cdai/src/components/IdeCard/IdeCard.js
```diff
@@ -294,6 +294,7 @@ export class IdeCard extends React.Component {
             <div
               className={`${idePrefix}-card--tile-img`}
               {...idAttribute(`ide-card-${cardId}-launch-Image`)}
+              alt="Opens in new tab"
```
davidmenendez, line 297: we shouldn't have any hard coded strings for translation purposes. this really should be sent as a prop.

davidmenendez (approved):

davidmenendez (changes requested):
one small comment

packages/cdai/src/components/IdeCard/IdeCard.js
s100: @davidmenendez Makes sense. I have updated the component to have a new optional string prop `externalLinkText`, which is passed down through `renderCardContent` and `renderCardHeader` to this point. Currently I'm retaining the English string "Opens in new tab" as the *default* value of this prop, as I see is done in a few other places in this codebase.

davidmenendez (approved):
thanks!

PR #2248, opened 2022-08-31
chore: update pre-commit hook

davidmenendez:
i noticed the formatter in the pre-commit hook wasn't working, so i wanted to do some investigating into it.

it seems to me like we probably have way too much bloat in our main `package` file. the `lint-staged` commands seemed a little redundant or unnecessary. like, i don't know why we need the `--check` flag (https://prettier.io/docs/en/cli.html#--check) to tell the user to run format when it should just be formatting for you. admittedly, i'm not an expert, but i did some research and this is what i came up with. i think making more adjustments to `package` by cleaning it up and streamlining it would be good for us in the long run.

would love to talk more about this topic.


davidmenendez:
tagging @lee-chase if you can give any insight on this topic it would be greatly appreciated

lee-chase (commented):
Hey, @davidmenendez not sure who set this up. These commands probably should use the default config for prettier, eslint and stylelint. I don't mind the addition of write/fix for all three (stylelint supports this too).  

With regards to the update:
- I am. not sure how the removal of `--loglevel warn` affects us.
- I think --report-needless-disables and --report-invalid-scope-disables should be moved to the global config files.

matthewgallo (approved):

PR #2255, opened 2022-09-06
fix: prevent auth checks during lerna publish

matthewgallo:
Last weeks release failed because lerna now tries to check npm auth via 2fa even during github actions. The `--no-verify-access` flag will disable these auth checks.

#### What did you change?
`release.yml`
`release-v11.yml`
Carbon has made this same change last week after experiencing the same issue.


matthewgallo:
Merging in order to fix issue with our release action tomorrow morning

Issue #2270, opened 2022-09-12
Datagrid stories: improve controls


elycheea:
## What will this achieve?

Allow users to have more control over datagrid settings. Currently, the controls only use the `datagridState` but it provides no way view use cases and examples that the datagrid hooks provide.

## How will success be measured?

For each story, users should be able to view scenarios matching those described in the in [design specs](https://pages.github.ibm.com/cdai-design/pal/components/data-table/overview/).

## Additional information

- Designs
- Existing code
- etc

elycheea:
This is a list of scenarios we should demonstrate in our stories but some of these may not require a separate story (e.g. alignment, truncation)

### Header
- [x] Default header w/ or w/o sublines
- [ ] Dense header w/ or w/o sublines
- Example with accordion headers??

For `PageHeader` or `Tabs` as table header, should double check accessibility.

### Settings
AKA Row height settings
- [ ] Rename to `Row height settings` or `Row size settings`
  - Overflow menu should align right
  - v11 should use Popover instead

### Row action buttons
- [ ] Overflow menu with more than 2 actions (including destructive action) -> `Sticky action column`
- [ ] Hover on row (with overflow or actions) -> `Is Hover on Row` uses text as example
- [x] Action buttons with 1-2 buttons
- [ ] Action buttons + batch actions (to show disabled in batch mode)

### Batch actions
- [x] Batch actions with more than 4 actions

Select all across all pages in the toolbar not implemented at this time (Carbon). More does not use caret yet due to limitations in ButtonMenu.

### Expandable rows
- [ ] Rename to `Expandable row`
  - Fix spacing shown with content
  - Border should be below expanded area — not above

### Nested rows
- [ ] Default rows
  - Expansion indicator is missing (should display on hover of parent row; hover on child should not show indicator)
  - Chevrons should be right/down not down/up
- [ ] Selectable rows
- [x] Multilevel nesting
- [ ] Selectable multilevel nesting
--
- Simple pagination — no option to select items on page
  - May not need an actual story, but should verify this works as expected

- [ ] Remove `Nested table` story

### Resizable columns
- Resizable columns
  - Individual column ✅ 
  - Double click to resize to fit — what about keyboard??
  - Multiple columns (visible only)
  - _Column header should wrap to two lines before truncating when header is long_
    - _compact data tables should keep to one line for truncation_
  - Should include label and icons (min-width shows icon only)
  - Column min-width of `1ch`?

[...]

PR #2279, opened 2022-09-15
chore(Datagrid): datagrid extensions continued, add expanded/nested rows directories

matthewgallo:
**_#2274 should be reviewed first_**

Contributes to #2270 

Adds the `Expanded rows` and `Nested rows` extensions to the new Datagrid storybook structure. While working on these, I fixed some styling issues as well.

#### What did you change?
Added Expanded rows and Nested rows directories to Datagrid stories.
Storybook

### DIFF
--- a/packages/cloud-cognitive/src/components/Datagrid/Datagrid.mdx
+++ b/packages/cloud-cognitive/src/components/Datagrid/Datagrid.mdx
@@ -17,12 +17,57 @@ import { Datagrid } from '.';
 
 <!-- TODO: Overview. -->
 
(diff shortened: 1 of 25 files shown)


elycheea:
@matthewgallo Just noticed that expanded row uses the same carets as the nested rows do now. I think expanded rows should be down/up chevrons while nested rows should use right/down chevrons. 😵

matthewgallo:
Whoops! 😱 Expandable rows are now using up/down chevrons again. The last border of the nested row seems to be more complex. Can we open an issue for that?

elycheea (approved):

PR #2282, opened 2022-09-16
fix(Data Grid): review feedback fix

Ratheeshrajan:
Contributes to #1924

- The row should also receive its hover effect (light gray background) when using keyboard navigation.
image
- The row should still have its hover effect (light gray background) when a cell is in editing mode
image
- The dropdown icons are baseline-aligned, should be center-aligned
image

#### What did you change?
`DatagridRow.js`
`_useInlineEdit.scss`
Storybook


packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridRow.js
```diff
@@ -43,6 +43,30 @@ const DatagridRow = (datagridState) => {
           `${blockClass}__carbon-row-expanded-hover-active`
         );
       }}
+      onKeyUp={(event) => {
```
matthewgallo, line 46: As discussed over slack, try just adding a conditional class to the row if the row index is equal to the `activeCellId` row

Ratheeshrajan: @matthewgallo The changes you suggested have been implemented. Thanks for the feedback.

matthewgallo (approved):
Before merging, can you create another PR with these same changes but  using `carbon-v11` as the base branch? The v11 token for `$hover-ui` is now `$layer-hover`.

PR #2284, opened 2022-09-16
chore(Datagrid): review feedback, region focus outline fix

matthewgallo:
Contributes to #1924 review

Addresses items 1 and 2 from Jan's design review. Because of the positioning and the scrolling nature of the grid elements, I used pseudo elements to form the outline around the focus region.

#### What did you change?
`DatagridContent.js`
`_useInlineEdit.scss`
Storybook


davidmenendez (approved):

elycheea (commented):
Noticed that the active cell (and region) remain highlighted after navigating back up to actions. Matt’s to look into it but looks like the active cell portion is already an issue so if it’s a bit more involved, we’ll address it a new PR. 🙂 

image

elycheea (commented):
Looks good minus some `console.log`s left in. Also looks like we have a conflict from the validation styles that were merged earlier today.

elycheea (approved):
🙌

Issue #2030, opened 2022-06-09
Inline edit: Save and change includes new line in user input in Firefox


elycheea:
## What package(s) are you using?

- Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)

## Detailed description

> **Describe in detail the issue you're having.**
> Hi, I was using the inlineEdit component with `v1.19.1`  of `@carbon/ibm-products`, and found there is different behaviors in Firefox and Chrome browser, that is, when user input the some char or integer in inlineEdit, and in Chrome, it can work well, the change function and save function can return the correct user input likes the following image1, but in Firefox, the change function and save function can return the incorrect user input with `[input]/n`  likes the following image2, so I am not sure if it is the issue for inlineEdit component, Could anyone help to give some comments for that? Thanks in advance.

> **Is this issue related to a specific component?**
InlineEdit

> **What did you expect to happen? What happened instead?What would you like to
see changed?**
Behavior should be the same between browsers — no new line should be added in Firefox

> **What browser are you working in?**
Chrome, Firefox

> **What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?**
1.19.1

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

In our Storybook on Firefox,
 
1. Hit backspace to clear example text “hello, world”
2. Check actions tab — should show
```
    change: "
    "
    save: "hello, world"
```

This appears to occur when the input value is deleted.

## Additional information

- Screenshots or code
- Notes

![image.png](https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/6779e98d-813c-44e9-852e-9efc541a8069)
![image.png](https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/874ee545-9cfd-41f9-a01d-0512b05ee170)

PR #2298, opened 2022-09-20
fix: inline edit newline fix

davidmenendez:
Contributes to #2030 

I was able to fix this by using `textContent` instead of `innerText`. i have been unable to find a reason as to why this interaction occurs in firefox, which is unfortunate because i think it's beneficial to understand these limitations. either way this seems to work better.


matthewgallo (approved):
🎉 Maybe because `textContent` filters out escape characters, such as the `\n` which seems to be causing the issues in #2030. Looks to be working great now!

PR #2303, opened 2022-09-21
chore(Datagrid): add column alignment story extension (v11)

matthewgallo:
Contributes to #2270 

Adds a new Datagrid extension directory specifically for column alignment. I also added new data to match the usage example for right and center aligned columns from the PAL site, see [here](https://pages.github.ibm.com/cdai-design/pal/components/data-table/column-alignment/usage).
Screen Shot 2022-09-21 at 11 12 17 AM

#### What did you change?
Added new Datagrid extension directory, only storybook presentational change
Storybook


davidmenendez (dismissed):

packages/cloud-cognitive/src/components/Datagrid/utils/makeData.js
```diff
@@ -87,6 +95,30 @@ const newPerson = () => {
         : statusChance > 0.33
         ? yesterdayDate
         : twoDaysAgoDate,
+    bonus: `$\r${getRandomInteger(100, 500, 2)}`,
+    status_icon:
```
davidmenendez, line 99: minor thing, but this feels a little redundant. i think it would be cleaner to have a helper function to get the props for the status icon. something like
```
const getStatus = () => {
const props = { size: 'sm', theme: 'light' }
if (statusChance > .66 ) {
 return { ...props, kind: 'critical', iconDescription: 'critical'}
}
...
}
...
const iconProps = getProps();
...
<StatusIcon {...iconProps} />
```
and this way you avoid all the conditional renders

davidmenendez (changes requested):
Looks good. just one comment

davidmenendez (approved):

PR #2309, opened 2022-09-22
docs: add details around carbon version support

matthewgallo:
Adds some of the new details around how we plan to support Carbon v10 and v11 for Carbon for IBM Products components.

#### What did you change?
Added new doc file


elycheea (approved):
I don’t think we need this for the v11 branch, so I think this one can be merged whenever you’re comfortable. Can always add more detail if we need to later.

PR #2316, opened 2022-09-28
feat(Datagrid): add validation to inline editing

matthewgallo:
Contributes to #1924 

@Ratheeshrajan I took over from where you had left off since you're looking at https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/1923 right now.

This change adds validation to the text and number types for Datagrid inline edit (`useInlineEdit`). In order for the component to know whether a field is invalid, there is a `validator` function that is included in the column inline editing configuration. Also included according to [PAL guidance](https://pages.github.ibm.com/cdai-design/pal/components/data-table/inline-editing/usage#validation), when in edit mode and input is invalid, `enter` and `tab` do not allow the cell to make new changes until the input is valid once again.

See example below:
```js
{
      Header: 'Age',
      accessor: 'age',
      width: 120,
      inlineEdit: {
        // required for including validation, this is used to set the invalid prop internally
        validator: (n) => n < 18,
        type: 'number',
        // These props are passed to the Carbon component used for inline editing
        inputProps: {
          invalidText: 'Invalid number, must be 18 or greater',
        },
      },
    },
```

#### What did you change?
```
packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/InlineEdit/InlineEditCell/InlineEditCell.js
packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/InlineEdit/handleGridKeyPress.js
packages/cloud-cognitive/src/components/Datagrid/styles/_useInlineEdit.scss
packages/cloud-cognitive/src/components/Datagrid/utils/getInlineEditColumns.js
```
Storybook

### DIFF
--- a/packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/InlineEdit/InlineEditCell/InlineEditCell.js
+++ b/packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/InlineEdit/InlineEditCell/InlineEditCell.js
@@ -77,6 +77,32 @@ export const InlineEditCell = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  // Reverts cellValue back to initialValue when exiting edit mode via clicking outside
+  // of the cell (either on a regular cell or clicking into another inline edit cell) and the
+  // edit input is in an invalid state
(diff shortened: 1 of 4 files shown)


elycheea (approved):
LGTM

Issue #2283, opened 2022-09-16
InlineEdit: tokens used inside `calc` should be interpolated


matthewgallo:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

Build issues because of token values inside of `calc` that are not being interpolated.
> Is this issue related to a specific component?

`_inline-edit.scss`
> What did you expect to happen? What happened instead? What would you like to
> see changed?

Expected build to pass
> What browser are you working in?

Chrome
> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

Latest
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

N/A
## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-cloud-cognitive/tree/main/examples/cloud-cognitive/codesandbox

## Additional information

- Screenshots or code
- Notes
See lines 108 and 109 of `_inline-edit.scss`
![image](https://user-images.githubusercontent.com/10215203/190676455-b2903cc8-3071-48f0-87de-bfbc3fa97a00.png)

matthewgallo:
Completed, will be available in tomorrow's release.

PR #2321, opened 2022-09-28
fix(InlineEdit): fix calc interpolation issue (v11)

glapadre:
Contributes to #2283 (v11)

Carbon tokens inside of calc needed to be interpolated in order to build correctly

#### What did you change?
`_inline-edit.scss`
`styles.test.js.snap`

Styling still works correctly and after carbon token is interpolated correctly


elycheea (approved):
Looks good 🙌

Issue #2278, opened 2022-09-15
[CreateNarrowTearsheet]: Mixed title required propTypes


tomleah:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

Today the `formTitle` prop is required on the CreateTearsheetNarrow and the `title` prop is optional.

This should be the other way around. `title` is always required when using this Tearsheet and `formTitle` should be optional.

With the `formTitle` being required it means we have propType errors 😢 

**Version:**

@carbon/ibm-products@1.25.1 but I think its present in @carbon/ibm-products@1.30.0 also

## Expected

- `formTitle` is optional and hides the `<h3>`
- `title` is required. Although this is not the main issue as this doesn't cause errors. Its also probably a breaking change?

## Additional information

- [Design discussion](https://ibm-cloud.slack.com/archives/CQGR0HC05/p1661360436223479)
- [Dev discussion](https://ibm-cloud.slack.com/archives/C013ZTX0N6B/p1661357371516469)

PR #2322, opened 2022-09-29
fix(CreateTearsheetNarrow): Update formTitle prop as optional - v11

punnoosewilson:
Contributes to #2278 

Need to make the `formTitle` prop as optional.

#### What did you change?
- Updated the `formTitle` prop as optional.

image

Storybook


packages/cloud-cognitive/src/components/CreateTearsheetNarrow/CreateTearsheetNarrow.js
```diff
@@ -145,7 +145,7 @@ CreateTearsheetNarrow.propTypes = {
   /**
    * Specifies a required field that provides a title for a form
```
matthewgallo, line 146:
```suggestion
   * Specifies a field that provides a title for a form
```

matthewgallo (changes requested):

Issue #1781, opened 2022-03-15
Datagrid component - unit tests coverage


holmansze:
Fix existing unit tests and improve test coverage.

markkulube:
testing

elycheea:
@Ratheeshrajan can you work with Matt on this?

elycheea:
Still relevant. We will be continuing this work as we prep Datagrid for release.

PR #2328, opened 2022-09-29
chore(Datagrid): add test for  useInlineEdit hook (v11)

matthewgallo:
Contributes to #1781 

Adds the first test for Datagrid `useInlineEdit` hook functionality. In the test, we check that a cell value can be edited via click and keyboard interactions.

#### What did you change?
Added new test for `useInlineEdit` hook
Ran datagrid tests


elycheea:
Weill reopen these as needed since these are pretty outdated now.

Issue #2310, opened 2022-09-22
InlineEdit refactor


davidmenendez:
after investigating issue #2112 i've come to the conclusion that the only way to really fix the bug is to refactor the component. it's currently implemented using `<div contentEditable />` which doesn't behave well with react. the main problem being that any time the state updates the cursor moves to the start of the input. it's possible to keep this kind of implementation, but i don't think this is the best way to do it. using refs and ranges adds unnecessary complexity. switching to just using a text input will solve most of these problems and make maintaining this component a lot easier.

after doing some research it seems like `contentEditable` is really only suited for things like building text editors. and while the same problems that i mentioned above still exist, the main benefit is that using `contentEditable` is better suited for doing things like formatting text and styling it similarly to the way a text editor behaves. see [draft.js](https://draftjs.org/) for example

lee-chase:
One reason to use `contenteditable` was to allow other tags e.g. h2 (as per comments in code). Not sure if this feature is used.

elycheea:
Caught this on the v11 version of InlineEdit. You’ll probably see it when you’re working on the v2 PR anyway, but wanted to drop this here anyway.
<img width="363" alt="image" src="https://user-images.githubusercontent.com/1717398/193141653-5021f25b-79b0-4107-a0f5-aa8d453f6eef.png">

PR #2330, opened 2022-09-30
feat: inline edit refactor

davidmenendez:
Contributes to #2310 

the main changes to this refactor revolve around moving from a div with the `contentEditable` attribute to using a text input. the purpose for this is to more easily allow the component to behave more like a carbon text input. now the user is able to access the value of the input as it changes, which wasn't how the original inline edit worked.

no tests added yet. i wanted to get the foundation set before adding those.


packages/cloud-cognitive/src/components/InlineEditV2/InlineEditV2.js
```diff
@@ -0,0 +1,214 @@
...
+        onSaveHandler(false);
+      }
+    };
+
+    const escapeHandler = () => {
+      onCancelHandler(false);
+    };
+
+    const onKeyHandler = (e) => {
+      switch (e.keyCode) {
```
matthewgallo, line 113: We should probably use `event.key` since `keyCode` is deprecated. See example [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/main/packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/InlineEdit/handleGridKeyPress.js#L125)

matthewgallo (changes requested):
The devtools attribute should be added to the component as well. And `packages/core/story-structure.js` should be updated with this new component too, otherwise it gets put inside of a "Lost and found" storybook directory.

image

packages/cloud-cognitive/src/components/InlineEditV2/InlineEditV2.js
davidmenendez: was unaware. thanks for the head's up!

matthewgallo (approved):

Issue #2333, opened 2022-10-01
NavList component throws error when using production build


jowen:
## What package(s) are you using?

- [ ] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [x] Legacy/Security

## Detailed description
When using a production build only, NavList component throws an error preventing the menu from rendering. With a  developer build, component functions with no issue.

This problem is caused by the use of child.type.name, in the Nav component, which is not safe to use with production builds that minify class/function names. Instead child.type.displayName should be used.

> Describe in detail the issue you're having.

> Is this issue related to a specific component?

Shell -> Nav -> NavList

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

@carbon/ibm-security 2.13.4.

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Perform a production build with compression, symbol/minification enabled
2. Create a menu system using the Shell component
3. Open menu

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-cloud-cognitive/tree/main/examples/cloud-cognitive/codesandbox

## Additional information

- Screenshots or code
- Notes

jowen:
fixed by https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2334

PR #2334, opened 2022-10-01
fix(NavList): use type.displayName

jowen:
Contributes to #2333

Fix for NavList throwing error when using production build.

#### What did you change?
Swap from using type.name to type.displayName.   

type.name is not safe to use with production builds which may minify symbol names.

- verified unit tests pass
- verified component functions correctly in storybook


matthewgallo (approved):
Thanks for this contribution @jowen 🎉

Issue #2341, opened 2022-10-06
inline edit refactor testing


davidmenendez:
add testing for inline edit refactor

matthewgallo:
Complete

PR #2344, opened 2022-10-06
chore: add inline edit v2 tests

davidmenendez:
Contributes to #2341 

adds tests for inline edit v2. code coverage is at 100%

Screen Shot 2022-10-06 at 2 56 00 PM


AlexanderMelox (approved):
Lgtm 👍🏻

PR #2347, opened 2022-10-11
feat: upgrades framer-motion version

AlexanderMelox:
Updates framer motion to latest version to fix build bug.

Screen Shot 2022-10-11 at 10 16 30 AM


AlexanderMelox:
This approach wouldn't have worked

Issue #2287, opened 2022-09-19
CreateTearsheet initialStep not honored when mounted with open=true


justindm234:
CreateTearsheet does not open to the initialStep when the component is mounted with open=true

Our application rehydrates state and needs to mount an open tearsheet that returns the user to a specific step.

## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Is this issue related to a specific component?

CreateTearsheet

> What did you expect to happen? What happened instead? What would you like to
> see changed?

CreateTearsheet should open to the step identified by initialStep. I believe it need to initialize its currentStep state value to `initialStep || 0`.

> What browser are you working in?

Chrome 105

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

1.7.0

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM System z Hardware Management Console

## Steps to reproduce the issue

1. Mount a CreateTearsheet with open=true and initialStep > 1

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-cloud-cognitive/tree/main/examples/cloud-cognitive/codesandbox

https://codesandbox.io/s/createtearsheet-initialstep-when-mounted-open-n9jjhm

Tearsheet is mounted with open=true and initialStep=4. On mount the tearsheet is open but the step is 1. If you close the tearsheet and click the "Open Tearsheet" button to re-open it, the tearsheet opens to step 4.

## Additional information

- Screenshots or code
- Notes

elycheea:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Development-602c6543cd8ba2001cb5ed88/issues/carbon-design-system/ibm-cloud-cognitive/2287?planning-poker) with Zenhub @davidmenendez @matthewgallo @AlexanderMelox

PR #2353, opened 2022-10-13
fix(CreateTearsheet): fix initialStep issue when open=true

punnoosewilson:
Contributes to #2287 

CreateTearsheet initialStep not honored when mounted with open=true.

#### What did you change?
- Updated the setCurrentStep = initialStep, if the TearSheet is opened and initialStep has some values

- Storybook


matthewgallo (changes requested):
This works great when there are no conditional `CreateTearsheetStep`s included. But in the case of conditional steps, this is causing the `CreateTearsheet` to render an incorrect step.

To account for hidden steps, try something like this:
```js
if (open && initialStep) {
  let numberOfHiddenSteps = 0;
  stepData.forEach((step, stepIndex) => {
    if (stepIndex + 1 > initialStep) {
      return
    }
    if (step.shouldIncludeStep === false) {
      numberOfHiddenSteps += 1;
    }
  });
  setCurrentStep(initialStep + numberOfHiddenSteps);
}
```

punnoosewilson:
> This works great when there are no conditional `CreateTearsheetStep`s included. But in the case of conditional steps, for example [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/main/packages/cloud-cognitive/src/components/CreateTearsheet/preview-components/MultiStepTearsheet.js#L197), this is causing the `CreateTearsheet` to render an incorrect step.
> 
> To account for hidden steps, try something like this:
> 
> ```js
> if (open && initialStep) {
>   let numberOfHiddenSteps = 0;
>   stepData.forEach((step, stepIndex) => {
>     if (stepIndex + 1 > initialStep) {
>       return
>     }
>     if (step.shouldIncludeStep === false) {
>       numberOfHiddenSteps += 1;
>     }
>   });
>   setCurrentStep(initialStep + numberOfHiddenSteps);
> }
> ```

@matthewgallo 
It works for the first time when `open = true`. However, if we close the TearSheet and then reopen it by clicking the button, the issue persists.

So I did the identical modification within the `useResetCreateComponent` hook and it worked.

But I've some concerns like: (Because I made modifications within the Hook)
- Will it have an impact on any other components?
- Will it affect any product that uses this component?

If it is a good to go, I'll bring those adjustments in the `v11 version` as well.

matthewgallo:
Ok, can you just move that code into a utility function for `CreateTearsheet` to reduce duplication? In terms of the concerns you brought up, that hook is only used in the `CreateTearsheet` so we don't need to worry about other components being effected by this change and this shouldn't introduce any changes for anyone using this component currently, except for fixing the `initialStep` issue. So I think we're ok!

punnoosewilson:
@matthewgallo 
The recommended changes are committed.

packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.js
```diff
@@ -98,15 +99,7 @@ export let CreateTearsheet = forwardRef(
...
-        let numberOfHiddenSteps = 0;
-        stepData.forEach((step, stepIndex) => {
-          if (stepIndex + 1 > initialStep) {
-            return;
-          }
-          if (step.shouldIncludeStep === false) {
-            numberOfHiddenSteps += 1;
-          }
-        });
+        let numberOfHiddenSteps = getNumberOfHiddenSteps(stepData, initialStep);
```
matthewgallo, line 102: Change this `let` to `const` since it doesn't need to be reassigned

packages/cloud-cognitive/src/global/js/hooks/useResetCreateComponent.js
```diff
@@ -37,15 +38,7 @@ export const useResetCreateComponent = ({
...
-        let numberOfHiddenSteps = 0;
-        stepData.forEach((step, stepIndex) => {
-          if (stepIndex + 1 > initialStep) {
-            return;
-          }
-          if (step.shouldIncludeStep === false) {
-            numberOfHiddenSteps += 1;
-          }
-        });
+        let numberOfHiddenSteps = getNumberOfHiddenSteps(stepData, initialStep);
```
matthewgallo, line 41: Change this `let` to `const` since it doesn't need to be reassigned

packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.js
```diff
@@ -97,7 +98,11 @@ export let CreateTearsheet = forwardRef(
       if (lastItem !== lastIncludedStep) {
         setLastIncludedStep(lastItem);
       }
-    }, [stepData, firstIncludedStep, lastIncludedStep]);
+      if (open && initialStep) {
+        let numberOfHiddenSteps = getNumberOfHiddenSteps(stepData, initialStep);
```
matthewgallo, line 102: Change this `let` to `const` since it doesn't need to be reassigned
```suggestion
        const numberOfHiddenSteps = getNumberOfHiddenSteps(stepData, initialStep);
```

packages/cloud-cognitive/src/global/js/hooks/useResetCreateComponent.js
```diff
@@ -35,7 +38,8 @@ export const useResetCreateComponent = ({
         Number(initialStep) <= Number(totalSteps) &&
         Number(initialStep) > 0
       ) {
-        setCurrentStep(Number(initialStep));
+        let numberOfHiddenSteps = getNumberOfHiddenSteps(stepData, initialStep);
```
matthewgallo, line 41: Change this `let` to `const` since it doesn't need to be reassigned

```suggestion
        const numberOfHiddenSteps = getNumberOfHiddenSteps(stepData, initialStep);
```

Issue #1864, opened 2022-04-01
SCSS Source maps invalid for index-full-carbon maybe others


lee-chase:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

When loading SCSS from index-full-carbon.css the source maps the source maps have an invalid path to the source files.

Numerous error reports like the following are shown. It appears to be targeting `node_modules/node_modules/`.

<img width="1133" alt="image" src="https://user-images.githubusercontent.com/15086604/161227401-b23963ea-2b29-4114-827c-27f0e6c1a732.png">

## Steps to reproduce the issue

1. Copy an example from the examples/carbon-for-ibm-products folder.
2. npm run install
3. npm run start

lee-chase:
This appears to be due to our build picking up SCSS files from outside the package being built.

In this case `@carbon/carbon/layout/scss/_breakpoint.scss` we reacy out of the package up to the top level.

lee-chase:
Raised https://github.com/sass/sass/issues/3281

elycheea:
@matthewgallo Let’s look into whether this is still an issue for v11 and figure out how to prioritize from here.

matthewgallo:
This is now fixed in `1.34.1`, confirmed with an app using react-scripts by running `react-scripts build` and no longer see source map errors

PR #2359, opened 2022-10-14
fix(builds): add script to fix `node_modules` path in sass source maps (v11)

matthewgallo:
Contributes to #1864 

This PR includes a script that changes the relative path to the `node_modules` directory in each of the sass source maps (ie `.css.map` files) which fixes the path used within the map files themselves.

#### What did you change?
```
packages/cloud-cognitive/package.json
scripts/updateSourceMaps.js
```
To test this change, I ran the build script from `packages/cloud-cognitive` which now includes a new script `"build-css-update-maps": node ../../scripts/updateSourceMaps.js` which is run after the source maps are created. This script changes the path to `node_modules` inside of each map file. After that was run, I copied over the generated `css` directory into a sample app that uses our package. Then I ran `react-scripts build` in the sample app and no longer see source map errors anymore.

### DIFF
--- a/packages/cloud-cognitive/package.json
+++ b/packages/cloud-cognitive/package.json
@@ -36,11 +36,12 @@
     "access": "public"
   },
   "scripts": {
-    "build": "run-s clean build-first build-all",
+    "build": "run-s clean build-first build-all build-css-update-maps",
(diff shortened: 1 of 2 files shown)


elycheea (approved):
Looks good to me!

PR #2377, opened 2022-10-18
chore(Combo button): v11 updates

qbi11y:
Contributes to #

{{short description}}


elycheea:
Closing this since we’re expecting a new version to be added in Carbon 11 that will replace the v2 version.

PR #2391, opened 2022-10-20
fix(PageHeader): button alignment logic based on variable value

### LINKED ISSUE #2262: Page header button alignment bug
Storybook shows buttons always aligning right, even at 672px and 320px breakpoints: 

image

However, guidance shows the buttons aligning left below 672px.

image

lizisawizard:
Closes #2262 

The button alignment for the PageHeader component was off because the widthIsNarrow variable was being calculated wrong. The value was calculated based on if the current window width was smaller than the breakpoint. However, breakpoint values were in rem and window values returned in px so the value was almost always false unless the window was realllyyy tiny. 

#### What did you change?
So I just divided the current width value by 16 (convert to rem) and added a parseInt to the breakpoint since it was returning a string so we can evaluate the int value (44rem -> 44)

On my local, I checked to see if the rightAlign class was being applied or not based on the window size. It worked uwu


matthewgallo (approved):
👏 👏

Issue #2342, opened 2022-10-06
inline edit refactor design review


davidmenendez:
## Design review

[for inline edit refactor (V2)](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-patterns-edit-and-update-inlineeditv2-canary--default)

### Standards

- [x] All pattern updates/changes/iterations have been discussed with the
      component developer
- [x] Patterns have been approved by either DSAG or another approving entity

### Pattern and behavior

- [x] The component behaves according to the guidelines set by the pattern
      maintainer
- [x] The component’s UI matches the pattern specifications set by the pattern
      maintainer
- [x] The component’s motion matches the specifications set by the pattern
      maintainer(s)
- [x] The UI produced is responsive, cross-browser, and responds to the
      currently set Carbon theme.
- [x] Colors, themes, sizes and additional props are true and correct, aligning
      with Carbon set tokens (unless otherwise specified by the pattern)
- [x] Paddings/margins/spacings are true and correct, in both desktop and mobile
      views

### Storybook

- [x] A functioning component renders in Storybook
- [x] The Storybook displays multiple scenarios that show how the component can
      be used
- [x] Changing props in the Storybook updates the component

sdignum:
@davidmenendez When clicking into the text field, once I make a change, the text field width should not change. The Save icon button should appear, but it shouldn't need to adjust the field width.

sdignum:
Looks good to me @davidmenendez. Let me know if you need for me to do anything else

PR #2397, opened 2022-10-21
Inline edit v2 design review

davidmenendez:
Contributes to #2342

address a few design issues pointed out during the review

* save button is now always visible but when a change hasn't been made it remain disabled
* a tooltip has been added to the edit button in read only mode
* some color issues have been addressed
* a width has been added to the storybook story to prevent width changing

### DIFF
--- a/packages/cloud-cognitive/src/components/InlineEditV2/_inline-edit-v2.scss
+++ b/packages/cloud-cognitive/src/components/InlineEditV2/_inline-edit-v2.scss
@@ -46,6 +46,8 @@
     }
 
     &__warning-icon {
+      /* stylelint-disable-next-line carbon/layout-token-use */
+      margin-right: 0.4375rem;
       color: $support-01;
     }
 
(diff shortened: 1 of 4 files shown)


packages/cloud-cognitive/src/components/InlineEditV2/_inline-edit-v2.scss
```diff
@@ -65,7 +67,8 @@
     outline: none;
   }
 
-  .#{$block-class}-readonly .#{$block-class}__text-input.#{$carbon-text-input} {
+  .#{$block-class}-readonly .#{$block-class}__text-input.#{$carbon-text-input},
+  .#{$block-class}-readonly .bx--btn.bx--btn--icon-only.bx--tooltip__trigger {
```
matthewgallo, line 71:
```suggestion
  .#{$block-class}-readonly .#{$carbon-prefix}--btn.#{$carbon-prefix}--btn--icon-only.#{$carbon-prefix}--tooltip__trigger {
```

matthewgallo (changes requested):
Just have a small suggestion to use `$carbon-prefix` in the component styles.

packages/cloud-cognitive/src/components/InlineEditV2/_inline-edit-v2.scss
davidmenendez: blah good catch. i always forget that haha

matthewgallo (approved):

PR #2405, opened 2022-10-25
docs: add initial content for v2 migration guide

matthewgallo:
Adds an initial markdown document containing migrations details from Carbon for IBM Products v1 => v2.

#### What did you change?
```
docs/guides/v2.md
```
vscode markdown preview.


elycheea (commented):
We should add a link back to the main Carbon 11 migration docs.

Also super minor but a few are missing periods — `EditSidePanel`, `ProductiveCard`, `SidePanel`, and `StatusIcon`.

Issue #2400, opened 2022-10-24
[useWindowScroll] Using PageHeader component causes the page to scroll at random times.


dkaushik95:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive) `"@carbon/ibm-products": "^1.34.1",`
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.
When I move away from a page with a pageheader and come back to the tab, I see something scrolls to top. After that, every time I interact with the UI components in the page, it always scrolls to the top. I investigated all our behaviors where we use the window.scrollTo function but it doesn’t seem to affect it. I looked at global listeners and found that useWindowScroll is causing the scroll to happen. When I remove the listener, my components starts to work fine.

> Is this issue related to a specific component?
`PageHeader` 

> What did you expect to happen? What happened instead? What would you like to
> see changed?
The page shouldn't scroll unless specified by a prop.

> What browser are you working in?
Chrome

### Important
On windows, the scroll behavior is even worse where it's scrolling on any user interaction.

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?
"^1.34.1"

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
CP4Waiops - DCUT November 2

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-cloud-cognitive/tree/main/examples/cloud-cognitive/codesandbox

## Additional information

- Screenshots or code

![Screen Shot 2022-10-20 at 6 00 25 PM](https://user-images.githubusercontent.com/8481567/197570964-aa040b0b-862f-4849-9c8b-20edbbfc8912.png)

- Notes

matthewgallo:
The fix for this was delivered in `1.35.1`. @dkaushik95 Please feel free to reopen if you're still noticing this behavior after upgrading.

PR #2410, opened 2022-10-26
fix(PageHeader): restrict useEffect logic to prevent scrolling

matthewgallo:
Contributes to #2400 

This addresses the scrolling issue in the PageHeader. The only portion of the PageHeader that scrolls the page is from a utility called `utilSetCollapsed` that is called from inside of a `useEffect`. I think the issue here is that the previous check to run this function only look to see if `collapseHeader` was a `boolean` type. But we only want to ever run this utility if `collapseHeader` is true.

#### What did you change?
```
packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
```
Haven't been able to fully reproduce this issue, but `utilSetCollapsed` should only ever be called when `collapsed` is true. So I think the useEffect containing `utilSetCollapsed` is being run when it shouldn't be which is causing the scrolling.

### DIFF
--- a/packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
+++ b/packages/cloud-cognitive/src/components/PageHeader/PageHeader.js
@@ -438,7 +438,7 @@ export let PageHeader = React.forwardRef(
     };
 
     useEffect(() => {
-      if (typeof collapseHeader === 'boolean') {
+      if (collapseHeader === true) {
         utilSetCollapsed(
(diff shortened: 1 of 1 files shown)


rvergeld:
@matthewgallo Hey, I'm currently experiencing the scrolling issue for one of our webpages. Do you know when this fix will be released?

matthewgallo:
Hey @rvergeld, it will be released this Tuesday morning.

PR #2418, opened 2022-10-31
docs: add initial content for v2 migration guide (v11)

matthewgallo:
Adds an initial markdown document containing migrations details from Carbon for IBM Products v1 => v2.

#### What did you change?
```
docs/guides/v2.md
```
vscode markdown preview.


davidmenendez (approved):
this looks great. i noticed `InlineEdit` has the no changes label, but if we're defaulting v2 in v11 there will be some changes. i guess i can actually make those changes once it's moved from canary though.

PR #2466, opened 2022-11-16
feat: adds FilterSummary

AlexanderMelox:
This adds FilterSummary component

Screen Shot 2022-11-15 at 7 30 04 PM

### DIFF
--- a/packages/cloud-cognitive/src/components/FilterSummary/FilterSummary.js
+++ b/packages/cloud-cognitive/src/components/FilterSummary/FilterSummary.js
@@ -1,3 +1,9 @@
+/**
+ * Copyright IBM Corp. 2022, 2022
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
 import { Button } from 'carbon-components-react';
 import React from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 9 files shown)


packages/cloud-cognitive/src/components/FilterSummary/FilterSummary.js
```diff
@@ -0,0 +1,51 @@
...
+        />
+        <Button kind="ghost" size="sm" onClick={clearFilters}>
+          {clearFiltersText}
+        </Button>
+      </div>
+    );
+  }
+);
+
+FilterSummary.displayName = 'FilterSummary';
```
elycheea, line 42: I think more often, we see this set as 

```js
const componentName = 'FilterSummary';
```

then later

```js
FilterSummary.displayName = componentName;
```

Not critical, but probably good for more consistent standards.

packages/cloud-cognitive/src/components/FilterSummary/FilterSummary.stories.js
```diff
@@ -0,0 +1,79 @@
+//cspell: disable
```
elycheea, line 1: Probably better to disable cspell by words or lines or just around the particular lines in question rather than disabling for the whole file.

```
// cspell:disable
....
// cspell:enbable
```

elycheea (commented):
Mostly looks good, and I think we can merge as is, but did want to call out a few small things that should be addressed sooner than later.

- Copyright headers! Missing in a few files with inconsistent dates. These should all be 2022.
- Minor suggestion on setting the displayName more consistently with how we do in the majority of patterns
- More selective use of cspell than disabling the entire file.

These are probably all more akin to release review changes, but think we’re less likely to miss these changes if we do them now than later.

elycheea (approved):

Issue #2409, opened 2022-10-26
PageHeader: `title` truncation no longer being applied


matthewgallo:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

I noticed that the title truncation in the PageHeader component is no longer being applied. I'm not sure when this stopped working, but when the page title is long enough it should truncate.
> Is this issue related to a specific component?

PageHeader
> What did you expect to happen? What happened instead? What would you like to
> see changed?

Expected the title to truncate if long enough.
> What browser are you working in?

Chrome
> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

Latest
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

N/A
## Steps to reproduce the issue

1. View page header story with [long values](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-page-header-pageheader--fully-loaded-and-some).

elycheea:
@glapadre Do you think you can look into this?

glapadre:
10-4!

PR #2473, opened 2022-11-16
fix(PageHeader): Page header title truncation v11

glapadre:
Contributes to #2409
Update h1 tag to truncate page title

What did you change?
`_page-header.scss`
`PageHeaderTitle.js`
`styles.test.js.snap`

Updated snapshot tests and checked styling shows as expected


davidmenendez:
this PR is incorrectly targeting the main branch when it should be targeting the v11 branch

Issue #2367, opened 2022-10-17
Multi add select release review


elycheea:
### Readiness

- [x] One or more scenarios for a design pattern have been identified as a
      useful unit of functionality to publish.
- [x] A functioning component or components delivering those scenarios have been
      delivered and merged to main.
- [x] Design maintainer has approved the implementation for those scenarios (via
      a comment on the relevant issue/epic).

### Engineering review

- [x] Breaking changes have only been introduced with prior approval, discussion
      and documented in release notes. Ideally deprecation notices in the prior
      major version must have been added in a timely fashion.
- [x] The implementation takes into account, and does not impair, remaining and
      anticipated design scenarios.
- [x] Public component features (names, props, etc) are consistent with
      Carbon-defined conventions and are consistent internally. Where there
      isn't an existing convention to apply, ensure robust precedents are being
      established.
- [x] The UI produced is accessible, responsive, translatable, cross-browser,
      and responds to the currently set Carbon theme.
- [x] Components are functional components using hooks.
- [x] Public components which produce DOM structures support className.
- [x] Public components support a ref (via React.forwardRef).
- [x] Public component supports a Devtools attribute
- [x] All significant DOM elements have meaningful classes.
- [x] Additional attributes that are not identified as props (such as title,
      aria-\*, etc) are passed through to an appropriate DOM node of the
      component as HTML attributes.
- [x] No warnings, errors or log messages in the console.
- [x] Each public component JS is exported in /src/components/index.js, each
      public component SCSS is included in /src/components/\_index.scss, and
      each public component has a flag in package-settings.js.
- [x] Each public component SCSS lists all of the Carbon and C&CS components
      imported and used by the JavaScript code and explicitly imports the SCSS
      for each of these components.
     - [x] Imports updated accordingly for v11 branch

### Standards

- [x] No linter warnings or errors are produced.
- [x] Carbon tokens (theme, layout, motion) are used unless the design specifies
      otherwise.
- [x] All components utilizing motion must include reduced-motion queries for
      accessibility purposes - read more here.
- [x] Code is formatted according to prettier rules (no use of
      //prettier-ignore).
- [x] Code is clear, maintainable and follows standard React practices and the
      code guidelines.
- [x] Copyright header in every source file (js, css, scss etc.) with the
      appropriate years.

### Testing

- [x] There is a set of test cases for the components.
- [x] The tests exercise all inputs (props, slots, etc) and verify appropriate
      outputs.
[...]

PR #2487, opened 2022-11-22
fix: Multi add select review fixes

davidmenendez:
Contributes to #2367 

addresses all the points in the above issue with the exception of the accessibility tests are failing and i think they might be related to carbon, so more investigation is necessary.


packages/cloud-cognitive/src/components/AddSelect/AddSelect.test.js
```diff
@@ -161,4 +162,28 @@ describe(componentName, () => {
     render(<AddSelect {...newProps} />);
     expect(screen.getByTitle('editor')).toBeInTheDocument();
   });
+
+  // it('has no accessibility violations', async () => {
+  //   const { container } = render(<AddSelect {...defaultProps} />);
+  //   await expect(container).toBeAccessible(componentName);
+  //   await expect(container).toHaveNoAxeViolations();
+  // });
```
elycheea, line 170: You can use `xit` to skip instead of commenting.

packages/cloud-cognitive/src/components/AddSelect/_add-select.scss
```diff
@@ -440,7 +448,7 @@
 
   .#{$block-class}__multi .#{$pkg-prefix}--empty-state {
     max-width: 16rem;
-    margin-top: $layout-04;
+    margin-top: $spacing-04;
```
elycheea, line 451: `$layout-04` was 3rem so the equivalent spacing should actually be `$spacing-09`.

elycheea (commented):
Looks good so far! Glad `:focus-within` worked out. 🙌

I just caught one tiny issue with flipping from the `$layout` token to `$spacing`. Is the plan to address the accessibility tests in another PR?

davidmenendez:
@elycheea thanks for the feedback! yeah i'm going to work on fixing that test in a follow up PR.

elycheea (approved):

Issue #2291, opened 2022-09-19
Data table extension: Nested rows fixes


elycheea:
## Data table extension: Nested rows

- [x] Fix last border of in a set of nested children
- [x] Add expansion indicator for child rows that are also parent rows (only on hover — should verify with design maintainers)

From #2279, last border of nested row should reach all the way left on the data table.

> Also think the last border of the nested rows (or the top border of the next row) should reach all the way left on the data table based on the guidance. We can also check with the designers on the `$active-ui` color since that’s in the guidance, but the screenshot uses more of `$ui-03` (or `$border-subtle-01`).
>
> Current | Guidance
> --------|-----------
> ![image](https://user-images.githubusercontent.com/1717398/190686168-e7eff5c2-1ac0-4f4b-bbf6-5afa32e0f564.png) | ![image](https://user-images.githubusercontent.com/1717398/190686433-a09af1c6-72be-4247-a2ea-522a83dba7e6.png)

We may also need to check with design maintainers if expansion indicator should show for all parents (even if they are also children) for multiple levels of nesting. (e.g. should hovering the chevron of a child row that is also a parent show an indicator?)

## Design maintainers
- Ashley Bock
- Johanna Koval

matthewgallo:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Development-602c6543cd8ba2001cb5ed88/issues/carbon-design-system/ibm-cloud-cognitive/2291?planning-poker) with Zenhub @AlexanderMelox @davidmenendez @elycheea

PR #2496, opened 2022-11-29
chore(Datagrid): add nested row indicator for child rows

matthewgallo:
Contributes to #2291 

This PR adds the indicator line for nested rows that are also parent rows (ie they have sub rows within them). See attached screen recording for example:

https://user-images.githubusercontent.com/10215203/204433944-a31c5e75-f4d4-47d6-acc4-04a4916bf2ef.mp4

#### What did you change?
```
packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridRow.js
packages/cloud-cognitive/src/components/Datagrid/useNestedRows.js
packages/cloud-cognitive/src/components/Datagrid/Datagrid.test.js
```
Storybook and ran tests for Datagrid


packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridRow.js
```diff
@@ -18,12 +19,39 @@ const blockClass = `${pkg.prefix}--datagrid`;
...
+
 // eslint-disable-next-line react/prop-types
 const DatagridRow = (datagridState) => {
-  const { row } = datagridState;
+  const { row, rowSize, withNestedRows } = datagridState;
   const { state } = useContext(InlineEditContext);
   const { activeCellId } = state;
   const activeCellObject = activeCellId && getCellIdAsObject(activeCellId);
+
+  const getVisibleNestedRowCount = ({ isExpanded, subRows }) => {
```
davidmenendez, line 37: i think i get the gist of what this is trying to do. i think you can clean this up by using recursion instead of nesting if and foreach statements. you could do something like this
```
const getRowHeight = ({ isExpanded, subRows }) => {
  let size = 0;
  if (isExpanded && subRows) {
    size += subRows.length;
    subRows.forEach((child) => {
      size += getRowHeight(child);
    })
  };
  return size;
};
```

davidmenendez (changes requested):
looks good. just one comment

packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridRow.js
matthewgallo: Ah yes, I like this much better too! I'll update it

davidmenendez (approved):

Issue #2510, opened 2022-11-30
ActionBar: actions which appear on the overflow menu are not clickable


jowen:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description
ActionBar actions which appear on the overflow menu are not clickable.  This is a blocking issue as it impacts PageHeader preventing actions from being triggered.

![image](https://user-images.githubusercontent.com/802501/204822546-fa69dd96-5195-47b0-87af-40ed5a1a5a42.png)

> Describe in detail the issue you're having.

> Is this issue related to a specific component?
ActionBar and components that depend on ActionBar like PageHeader

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?
1.36.0

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

You can reproduce this using the ActionBar page in the storybook, https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-internal-actionbar--default&args=actions:2

1. Modify the story so an alert is shown when clicking an action
2. Click on an action in the bar, the alert is shown
3. Now click on an action in the bar's overflow menu, the alert is not shown
4. The action's onclick handler is not being forwarded to the overflow menu item

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-cloud-cognitive/tree/main/examples/cloud-cognitive/codesandbox

## Additional information

- Screenshots or code
- Notes

matthewgallo:
This will be included in next week's release (Dec. 6). Thanks again for contributing @jowen!

PR #2511, opened 2022-11-30
fix(ActionBar): overflow menu item onClick

jowen:
Contributes to #2510

#### What did you change?
forward the action's onClick handler to the OverflowMenuItem

verified in storybook, with and without a click handler


matthewgallo (approved):
Thanks for the contribution @jowen! 🎉

PR #2514, opened 2022-11-30
fix: use v1 inline edit in page header

davidmenendez:


packages/cloud-cognitive/src/components/PageHeader/PageHeaderTitle.js
```diff
@@ -48,7 +48,8 @@ export const PageHeaderTitle = ({ blockClass, hasBreadcrumbRow, title }) => {
           <SkeletonText className={`${blockClass}__title-skeleton`} />
         ) : isEditable ? (
           <InlineEdit
-            hideLabel
+            v1
+            hideLabe
```
matthewgallo, line 52: Typo
```suggestion
            hideLabel
```

matthewgallo (dismissed):
I noticed that the icons from the InlineEdit component are very small, see attached screenshot
image

Issue #2438, opened 2022-11-04
CreateFullPage component: impossible to Cancel after closing the confirmation modal by clicking outside of it


ashpc:
Steps to reproduce
How do we reproduce the error you described above: (ex. below)

In the CreateFullPage component, press Cancel - the confirmation modal will appear.
image

Close the modal by clicking outside of it.

Try to press Cancel again - the confirmation modal won't be displayed.
<img width="1671" alt="Screen Shot 2022-11-03 at 7 47 22 PM" src="https://user-images.githubusercontent.com/1907155/199861482-4b8feeb5-cbf9-4478-9e35-1a639c30ebee.png">

Reproducible here: https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-patterns-create-flows-createfullpage--create-full-page

elycheea:
@jagathgj if you have bandwidth, do you want to take a look at this? If not, we can reassign to someone else.

jagathgj:
@elycheea, I did a quick look at the issue. Create flows of FullPage(all) and Modals(Create Modal, Create Modal using dark theme) are seen as opened before clicking the open button. It shouldn't be open until the user clicks the open button.

PR #2522, opened 2022-12-05
fix(create full page): modal close issue fixed

jagathgj:
Contributes to #2438 

Modal not opening after closing it with outside click

#### What did you change?

Added onClose function with a state change(onClose={()=>setModalIsOpen(false)})

Storybook


elycheea (approved):
🎉

Issue #2524, opened 2022-12-05
FilterFlyout and FilterSummary to v11


AlexanderMelox:
## What will this achieve?
Add FilterFlyout and FilterSummary to v11

PR #2525, opened 2022-12-05
feat(Datagrid): adds FilterFlyout and FilterSummary v11

AlexanderMelox:
Contributes to #2524 

Adds FilterFlyout and FilterSummary to v11


davidmenendez (approved):

matthewgallo (changes requested):
One small request here, since the data model that is being used across stories has changed, can you update ClickableRow stories to include the fix I added [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2544/files). There is a date object that needs `.toString()`.

AlexanderMelox:
> One small request here, since the data model that is being used across stories has changed, can you update the `Clickable row with side panel` story to include the fix I added [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2544/files). There is a date object that needs `.toString()` called in order to render.

Added it to latest commit

matthewgallo (approved):

Issue #2543, opened 2022-12-07
Datagrid: clickable row with side panel story has error


matthewgallo:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

The story for clickable row with side panel in the Datagrid component is throwing an error and does not render the side panel. This is happening because there is a date object in the data that is trying to render without the use of `.toString()`.
> Is this issue related to a specific component?

Datagrid
> What did you expect to happen? What happened instead? What would you like to
> see changed?

Expected to be able to click the row without any errors occuring.
> What browser are you working in?

Chrome
> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

Latest
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

N/A
## Steps to reproduce the issue

1. Go to storybook and find the Datagrid extension "Clickable Row items with side panel"
2. Click one of the rows and you'll see the issue
![image](https://user-images.githubusercontent.com/10215203/206208689-732abd44-6ea3-4f1b-93d1-e13d2818aef5.png)

PR #2544, opened 2022-12-07
fix(Datagrid): address date issue with clickable row side panel

matthewgallo:
Contributes to #2543 

Fixes the Datagrid clickable row with side panel story. There was a date that wasn't calling `.toString()` which was causing that story to throw an error and not render properly.

#### What did you change?
```
packages/cloud-cognitive/src/components/Datagrid/Extensions/ClickableRow/ClickableRow.stories.js
```
Storybook


AlexanderMelox (approved):
LGTM 🚀

Issue #2545, opened 2022-12-07
Datagrid: inline edit extension with `isFetching` property causes error


matthewgallo:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

When using inline edit Datagrid with the `isFetching` property, an error is thrown because `value.text` is undefined at that point. Adding an optional chain (ie `value?.text`) should easily fix this issue.
> Is this issue related to a specific component?

Datagrid (inline edit)
> What did you expect to happen? What happened instead? What would you like to
> see changed?

Expected no errors using inline edit with the `isFetching` property.
> What browser are you working in?

Chrome
> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

Latest
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

N/A

georgespano:
As part of this work, one other small suggestion that would help with our modelling structure is to have the inline edit drop down able to support items defined simply by an array of strings: items: ["item1", "item2", "item3"]. The drop down component supports this already, but a small change would still be required to remove the assumption that each item has a text field in it. In InlineEditCell file the method renderDropdownItem can be changed in these areas:

```
const renderDropdownItem = (item) => {
    const includesIcon = !!item?.icon;
    return includesIcon ? (
      <>
        {React.createElement(item.icon)}
        <span className={cx(`${blockClass}__inline-edit--select-item`)}>
          **{item?.text}**
        </span>
      </>
    ) : (
      **item?.text**
    );
  };
```

Also the label parameter for the InlineEditButton would need a small change:

```
label={
            type === 'selection'
              ? **value.text**
              : type === 'date'
              ? buildDate(value)
              : value
          }
```

matthewgallo:
Complete, will be included in next release

PR #2546, opened 2022-12-07
fix(Datagrid): update inline edit selection type to support string items

matthewgallo:
Contributes to #2545 

Allows the selection inline edit type in Datagrid to now use either an array of strings or array of objects. I tested this by passing an array of strings in `getInlineEditColumns.js` to the column that supports the `selection` type after I made the changes and confirmed that the interaction still worked as expected.

#### What did you change?
```
packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/InlineEdit/InlineEditCell/InlineEditCell.js
```
Storybook


georgespano:
Thanks @matthewgallo for the quick turnaround!

davidmenendez (approved):

PR #2551, opened 2022-12-08
fix(CreateTearsheet): Add firstFocusElement

### LINKED ISSUE #2361: [CreateTearsheet] Step with tooltip as the first focusable element triggers the tooltip

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

Our designs sometimes include tooltips in the `labelText` prop of inputs. When combining it with a CreateTearsheet, this tooltip gets focused and it triggers the tooltip.

CreateTearsheet

> see changed?

 We expected the tooltip to be ignored in this case.

Google Chrome

> package are you using?

@carbon/ibm-products v1.25.1

> should be aware of?

IBM RPA. No important dates.

1. Open the Tearsheet
2. Navigate to second step and back to the first
3. Tooltip will trigger instead of focusing the input

>
> https://codesandbox.io/s/beautiful-platform-tumcnw?file=/src/index.js

lizisawizard:
Closes #2361 

The issue was that if given a tooltip as the first element in the CreateTearsheet Pattern, it'll focus/open it first

#### What did you change?
I added skipFocusElements as a prop to CreateTearsheet and CreateFullPage Patterns so uses can specify if they want to skip over elements with the useCreateComponentFocus custom hook. The skipFocusElements prop takes in an array of classNames to skip over until it finds an element not in skipFocusElements to focus on.

I created a test case in storybook and verified it worked properly for both CreateTearsheet and CreateFullPage


davidmenendez:
i think i get the gist of the problem and what you're trying to accomplish. i'm not as familiar with this component as @matthewgallo but it seems to me like maybe it would be easier to specify a specific id or class name to focus instead of using an array of class names to not focus, which i believe is my understanding of the proposed solution.

lizisawizard:
> i think i get the gist of the problem and what you're trying to accomplish. i'm not as familiar with this component as @matthewgallo but it seems to me like maybe it would be easier to specify a specific id or class name to focus instead of using an array of class names to not focus, which i believe is my understanding of the proposed solution.

That makes sense, I can't go ahead and change it

lizisawizard:
@davidmenendez I changed it to first focusable element just now, please lmk if that works!

elycheea (commented):
Just taking an initial look, I think this should not only work with classNames but IDs since IDs are unique and probably better suited for specifying which element should get initial focus.

davidmenendez:
yeah i agree with @elycheea . i think you can simplify this to just have the user provide the css selector they to use. this way you don't have to try to do any guess work. i'd just update the code to use something like `document.querySelector(focusSelector).focus()` where `focusSelector` is whatever selector string the user passes. and then i'd specify in the documentation that you need a proper selector like `.class` or `#id` and not just `class` or `id`. and this way they can be as specific as they need to be without the code trying to jump through hoops to locate it.

davidmenendez:
my fix basically addresses what i believe the problem to be- that code was running by default to try to focus something, which was causing the issue with tooltips. this also addresses a problem, which i commented about in the code itself, that's caused by code in `TearsheetShell` that uses a timeout to focus on stacked tearsheets. IMO we should probably look to clean this up in the future.

elycheea (approved):
Looks great! Thanks for investigating and for adding comments. 🙌 

I think we should update the PR name for the commit though.

PR #2552, opened 2022-12-08
fix(CreateTearsheet): add firstFocusElement v11

lizisawizard:
…s for skipping first focusable component

Contributes to #2361 

The issue was that if given a tooltip as the first element in the CreateTearsheet Pattern, it'll focus/open it first

What did you change?
I added skipFocusElements as a prop to CreateTearsheet and CreateFullPage Patterns so uses can specify if they want to skip over elements with the useCreateComponentFocus custom hook. The skipFocusElements prop takes in an array of classNames to skip over until it finds an element not in skipFocusElements to focus on.

I created a test case in storybook and verified it worked properly for both CreateTearsheet and CreateFullPage


elycheea (approved):
Thanks for making the fix and update on this one @davidmenendez!

PR #2554, opened 2022-12-09
chore: add tests for code sandbox examples

lee-chase:
- Adds to the example-gallery-config to create a test suite for the code sandbox examples.
- Runs the tests as part of the standard tests
- new script `test:examples` tests just the examples.

NOTES: 
1. The construction of the tests expects an `Example` component to be found in `src/Example`.
2. @davidmenendez I had to remove v2 props from the example to get it to pass the tests. Needs a bit of thought to get it to work with V1 and V2.
3. Three of the tests `carbon-v11-template-example`, `TagSet` and `WebTerminal` do not work and are skipped. Perhaps raise a separate PR to resolve these.


elycheea (approved):
Thanks for running through changes with me this morning.

Looks good to me. Node 16 makes sense. 👍 
Now to CodeSandbox. 🙃

Issue #2565, opened 2022-12-13
inline edit v2 motion


davidmenendez:
inline edit has some motion baked in and i'd like to add some into the v2 version

davidmenendez:
received some initial feedback. moving this to backlog for now while i work on other more pressing issues.

PR #2566, opened 2022-12-13
chore: [WIP] adds motion for inline edit v2

davidmenendez:
Contributes to #2565 

inline-edit-motion

@AlexanderMelox @kristastarr i'd love to get your feedback on this since both of you have a lot more motion knowledge than i do


lee-chase:
@davidmenendez the original motion was a little different with the checkmark sliding in/out depending on whether or not the value has been edited. The reason for this was to ensure the cancel button appears underneath the mouse when the user clicks the edit button. This did require a rather odd-shaped clip to hide the slide-in and tooltip, which you should probably copy.

Personally, I think the animation from edit mode to non-edit mode is should feel instant, so a simple fade-out, fade-in. Well worth looping in Mike Olasov for an opinion.

P.S. I notice there are some differences between V1 and V2 hover effects. Not sure if this is a mistake in V1 or a deliberate change for V2.

Kapture 2022-12-14 at 10 50 04

kristastarr:
My first impression is that this is really fun to interact with!  I like the "zoom in" animation, but bearing in mind that functionality is more important than "delight",  @lee-chase brings up some good points.  Maybe making the animation speed faster would make it feel more instant? After interacting with the component more, I see how it makes since for the cancel button to appear underneath/in same location as the edit button when the user clicks the edit button (as in v1), so it's easy to switch between edit/cancel without moving your mouse.  

I kinda think it would be interesting to see the Edit icon zoom in from L to R, and Cancel zoom in from R to L, so moving from edit to cancel was like a toggle type action, as seen in Carbon's toggle component.  I don't mean have Edit move in from the L side of the input field,  I mean like as if it appeared just to the L of the cancel icon and moved into position from there.  But also I think that would be a little weird since the pencil icon itself looks like the pencil is drawing a line from L to R.  I agree that Motion Mike might have some good thoughts! (I'm not suggesting this as a solution for the edit/cancel overlap mentioned above, although it might give some different options to explore to make that happen.)

lee-chase (approved):
@davidmenendez I prefer the original animation, but not sure how to tweak Framer motion to achieve this.

So instead I've adjusted your animation to hide the overflow.

In some ways this is better than the original animation as it does not require the overflow to be clipped in a comples way.

davidmenendez:
@lee-chase would you like to approve?

Issue #2523, opened 2022-12-05
Organize filter flyout and filter left panel storybook 


AlexanderMelox:
## What will this achieve?
Organizes the filter flyout and filter left panel into a folder in storybook.

PR #2570, opened 2022-12-14
feat(Datagrid): makes seperate story file and adds docs for FilterFlyout (v11)

AlexanderMelox:
Contributes to #2523 

This adds documentation for filter flyout and adds a story file for data table filtering.

#### What did you change?
- Changed filtering stories from `Datagrid.stories.js` and moved it to `Filtering.stories.js`
- Fixed typo with `SelectAllWithToggle` import and updated path to fix error
- Adds Filtering docs to `Datagrid.mdx`

### DIFF
--- a/packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridToolbar.js
+++ b/packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridToolbar.js
@@ -161,8 +161,6 @@ const DatagridToolbar = (datagridState) => {
     datagridState;
   const { filterTags, EventEmitter } = useContext(FilterContext);
 
-  console.log({ filters: state.filters });
-
   const renderFilterSummary = () =>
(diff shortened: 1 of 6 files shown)


packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridToolbar.js
```diff
@@ -161,6 +161,8 @@ const DatagridToolbar = (datagridState) => {
     datagridState;
   const { filterTags, EventEmitter } = useContext(FilterContext);
 
+  console.log({ filters: state.filters });
```
matthewgallo, line 164: Console log should be removed

matthewgallo (changes requested):
There also appears to be some small lint issues to fix as well

matthewgallo (approved):

Issue #2290, opened 2022-09-19
Data table extension: Column customization should use Tearsheet


elycheea:
## Data table extension: Column customization

Currently, column customization uses a modal (see [Storybook](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary--customizing-columns)). In the updated guidelines, column customization should occur through a narrow tearsheet[^1].

Some styling changes may also be addressed during this change:
- [x] Instructions need padding on right and should use _most likely_ `$text-05` / `$text-helper` (v11) — verify with designers
- [x] Fix search icon alignment in search bar; should be 16px from left edge
- [x] Fix clear icon in search bar — sizing and styles[^2]
- [x] Spacing after drag and drop icon looks like it should be larger
- [x] During search, drag and drop is disabled but visibly still looks active; `$disabled-02` / `$icon-disabled` (v11)?
- [x] Header row (Column name) should be bold and fix on scroll

Reach out to design maintainers with any questions.

### Design maintainers
- Ashley Bock
- Johanna Koval

[^1]: https://pages.github.ibm.com/cdai-design/pal/components/data-table/column-customization/usage#anatomy
[^2]: https://pages.github.ibm.com/cdai-design/pal/components/data-table/column-customization/usage#search

matthewgallo:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Development-602c6543cd8ba2001cb5ed88/issues/carbon-design-system/ibm-cloud-cognitive/2290?planning-poker) with Zenhub @AlexanderMelox @davidmenendez @elycheea

PR #2582, opened 2022-12-16
fix(Datagrid): ColumnCustomizationModal to Tearsheet Narrow

lizisawizard:
Contributes to #2290 

The guidelines for the column customization for the datagrid changed to be a tearsheet instead of a modal

#### What did you change?
I changed the modal to a tearsheet for Column Customization - I added styling adjustments and tried to start changing 'Modal' to 'Tearsheet'

Checked the storybook on my local


packages/cloud-cognitive/src/components/Datagrid/styles/addons/_CustomizeColumnsTearsheet.scss
```diff
@@ -0,0 +1,100 @@
...
+
+.#{$block-class}__customize-columns-tearsheet--actions > button {
+  margin-left: $spacing-05;
+}
+
+.#{$block-class}__customize-columns-instructions {
+  padding-left: $spacing-05;
+  margin-bottom: $spacing-06;
+
+  @include carbon--type-style('body-long-01');
```
elycheea, line 59: We usually place `@include`s at the top of the block.

packages/cloud-cognitive/src/components/Datagrid/styles/addons/_CustomizeColumnsTearsheet.scss
```diff
@@ -0,0 +1,100 @@
...
+}
+
+.#{$block-class}__customize-columns-select-all {
+  display: flex;
+  height: $spacing-09;
+  padding-left: $spacing-08;
+  border-bottom: 1px solid $layer-active;
+  background-color: $layer;
+
+  label {
```
elycheea, line 75: Generally good practice to target the actual class name here — you can use the carbon-prefix to do so.
Also should use a type-style token here.

packages/cloud-cognitive/src/components/Datagrid/styles/addons/_CustomizeColumnsTearsheet.scss
```diff
@@ -0,0 +1,100 @@
...
+    font-weight: 600;
+  }
+}
+
+.#{$block-class}__customize-columns-select-all:hover {
+  background-color: $layer-hover;
+}
+.#{$block-class}__customize-columns-select-all-selected {
```
elycheea, line 83: Missing space above this block.

Notice a lot of repeated styles between `.-__customize-columns-select-all` and `.-__customize-columns-select-all-selected` — if there are styles we always want to apply, we should use a common class (probably just the former) and then add the state specific changes on the `--selected` version. (Also two hyphens for state to follow BEM more closely).

```scss
  position: sticky;
  top: 0;
  z-index: 1;
```

these should get moved to the common class since I don‘t think the stickiness is intended to be based on the state.

elycheea (changes requested):
Nice work so far, but lots of small fixes — mostly related to token usage and CSS practices. I’ll try to summarize below.

- Update size prop on search in `Actions.js`; this should allow you to remove some of the SCSS overrides
- Re-add `isDirty` to disable primary button until columns have been edited
- Remove `CustomizeColumnsModal.scss` — looks like we still have both files
- Restructure modal to tearsheet styles so sticky header applies to both search and column header
- Update formatting and token usage

Then this isn’t really on you but since we have some new files, I think we should go ahead and change this on the files you’ve modified — use the correct copyright header. We should use the version you see in Tearsheet or SidePanel.

packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/CustomizeColumns/TearsheetWrapper.js
```diff
@@ -9,18 +9,18 @@
...
-import CustomizeColumnsModal from './CustomizeColumnsModal';
+import CustomizeColumnsTearsheet from './CustomizeColumnsTearsheet';
 
-const ModalWrapper = ({ instance }) => {
-  const { onSaveColumnPrefs, isModalOpen, setIsModalOpen, labels, ...rest } =
+const TearsheetWrapper = ({ instance }) => {
+  const { onSaveColumnPrefs, isTearsheetOpen, setIsTearsheetOpen, labels, ...rest } =
     instance.customizeColumnsProps;
-  if (isModalOpen) {
+  if (isTearsheetOpen) {
```
elycheea, line 17: Do we actually need to check this state here? I think this is what’s causing the broken animation.
CC @AlexanderMelox @davidmenendez

packages/cloud-cognitive/src/components/Datagrid/styles/addons/_CustomizeColumnsTearsheet.scss
```diff
@@ -0,0 +1,100 @@
...
+
+.#{$block-class}__customize-columns-tearsheet--actions {
+  display: flex;
+  height: $spacing-09;
+  flex-flow: row;
+  background-color: $field-02;
+
+  .bx--search-magnifier svg {
+    left: 1rem;
+  }
```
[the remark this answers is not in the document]
lizisawizard: I changed it to XL - should i confirm with a designer tho?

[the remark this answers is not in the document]
elycheea: Hmm, probably not necessary in this case. I don’t think the size was explicitly defined in the design, but we’re able to infer the correct Carbon size/prop based on the provided designs.

packages/cloud-cognitive/src/components/Datagrid/styles/addons/_CustomizeColumnsTearsheet.scss
```diff
@@ -0,0 +1,100 @@
...
+
+@import '../variables';
+
+.#{$block-class}__customize-columns-tearsheet {
+  .#{$carbon-prefix}--tearsheet-content {
+    position: inherit;
+    display: flex;
+    flex-flow: column;
+    padding: 0;
+    margin: 0;
```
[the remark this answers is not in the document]
lizisawizard: The overflow: auto is causing issues with the sticky positioning tho - is it ok that I leave it off?

elycheea (dismissed):
Made a few small fixes, but overall looks great! Please open the v11 version once you’re ready. 🙌

elycheea (approved):

Issue #2366, opened 2022-10-17
Single add select release review


elycheea:
### Readiness

- [x] One or more scenarios for a design pattern have been identified as a
      useful unit of functionality to publish.
- [x] A functioning component or components delivering those scenarios have been
      delivered and merged to main.
- [ ] Design maintainer has approved the implementation for those scenarios (via
      a comment on the relevant issue/epic).

### Engineering review

- [x] Breaking changes have only been introduced with prior approval, discussion
      and documented in release notes. Ideally deprecation notices in the prior
      major version must have been added in a timely fashion.
- [x] The implementation takes into account, and does not impair, remaining and
      anticipated design scenarios.
- [x] Public component features (names, props, etc) are consistent with
      Carbon-defined conventions and are consistent internally. Where there
      isn't an existing convention to apply, ensure robust precedents are being
      established.
- [x] The UI produced is accessible, responsive, translatable, cross-browser,
      and responds to the currently set Carbon theme.
- [x] Components are functional components using hooks.
- [x] Public components which produce DOM structures support className.
- [x] Public components support a ref (via React.forwardRef).
- [x] Public component supports a Devtools attribute
- [x] All significant DOM elements have meaningful classes.
- [x] Additional attributes that are not identified as props (such as title,
      aria-\*, etc) are passed through to an appropriate DOM node of the
      component as HTML attributes.
- [x] No warnings, errors or log messages in the console.
- [x] Each public component JS is exported in /src/components/index.js, each
      public component SCSS is included in /src/components/\_index.scss, and
      each public component has a flag in package-settings.js.
- [x] Each public component SCSS lists all of the Carbon and C&CS components
      imported and used by the JavaScript code and explicitly imports the SCSS
      for each of these components.

### Standards

- [x] No linter warnings or errors are produced.
- [x] Carbon tokens (theme, layout, motion) are used unless the design specifies
      otherwise.
- [x] All components utilizing motion must include reduced-motion queries for
      accessibility purposes - read more here.
- [x] Code is formatted according to prettier rules (no use of
      //prettier-ignore).
- [x] Code is clear, maintainable and follows standard React practices and the
      code guidelines.
- [x] Copyright header in every source file (js, css, scss etc.) with the
      appropriate years.

### Testing

- [x] There is a set of test cases for the components.
- [x] The tests exercise all inputs (props, slots, etc) and verify appropriate
      outputs.
- [x] The tests validate the behaviors and interactions defined in the design
[...]

PR #2583, opened 2022-12-20
fix: single add select review fixes

davidmenendez:
Contributes to #2366 

addresses these issues

*  When the description is long, should this be truncated and have a title prop so the user can hover and read the text
* itemsLabel on storybook doesn't update the UI

looks like there might be an issue with our codesandbox examples, so i'll revisit that in another issue / PR


davidmenendez:
@AlexanderMelox fixes from the release review. i mentioned in the original description that i need to look to possibly fixing an issue with our codesandbox examples. i also didn't fix this issue-

> globalSearchLabel on storybook doesn't update the UI

you don't see anything because the carbon `Search` component hides the label, but if you inspect the html it's updating.

elycheea (approved):
🎉

PR #2587, opened 2022-12-27
fix(Datagrid): ColumnCustomizationModal to Tearsheet Narrow - v11

lizisawizard:
Contributes to #2290 

The guidelines for the column customization for the datagrid changed to be a tearsheet instead of a modal

#### What did you change?
I changed the modal to a tearsheet for Column Customization - I added styling adjustments and tried to start changing 'Modal' to 'Tearsheet'

Checked the storybook on my local


elycheea:
I’m seeing the latest changes from #2552 in here. Since @lizisawizard is no longer on this project, @davidmenendez I think the best approach is to merge this after we merge the skip focus changes or create a new PR with just the Tearsheet changes for ColumnCustomization.

elycheea (approved):
Resolved conflicts from firstFocusElement PR that David worked on and Ratheesh’s customize columns highlighting.

Made one more change in the test so that reusing the text `Customize columns` works and we don’t need to change to titlecase as a workaround.

Issue #2577, opened 2022-12-15
Datagrid Inline edit: date option only accepts Date object


matthewgallo:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

The Carbon `DatePicker` component also accepts strings and is able to render the date. The Datagrid inline edit component only works with Date objects and should be updated to accept strings as well.
> Is this issue related to a specific component?

Datagrid, specifically the `InlineEditCell.js` component, change would need to be made [here](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/main/packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/InlineEdit/InlineEditCell/InlineEditCell.js#L400-L414).
> What did you expect to happen? What happened instead? What would you like to
> see changed?

If my data set includes dates as strings, they should be able to work just the same as Date objects.
> What browser are you working in?

Chrome
> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

Latest
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

N/A
## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-cloud-cognitive/tree/main/examples/cloud-cognitive/codesandbox

## Additional information

- Screenshots or code
- Notes

elycheea:
Linking to this comment to make sure I don’t accidentally close this issue before we review if we can refactor some things to support entering dates as strings. 😅 

https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2592#pullrequestreview-1238293689

PR #2592, opened 2023-01-05
fix(DataGrid): date input to accept string date

Ratheeshrajan:
Contributes to #2577 

The Datagrid inline edit component only works with Date objects and should be updated to accept strings as well.

#### What did you change?
`InlineEdit/InlineEditCell/InlineEditCell.js`
`utils/makeData.js`
`utils/getInlineEditColumns.js`
Storybook


elycheea (changes requested):
Based on how I interpreted the ticket, I think we also need to support typing in a date (as a string) in the DatePicker, which Carbon does already support.

image

But right now it looks like we set the DatePicker value so that it can’t be edited outside of the dropdown.

See Carbon [DatePicker docs](https://v7-react.carbondesignsystem.com/?path=/docs/components-datepicker).

image

georgespano:
Hello @elycheea. The issue is opened based on discussions @matthewgallo and I had with how we are using the DataGrid. The discussion was around the buildDate function returning null if the cell value was not an instance of date. Our dates are represented as strings and as such the function would return null and the cell would render empty, even though there was data for it. Looking at the changes made by @Ratheeshrajan I see that the issue would be resolved. Typing in dates is not something we are necessarily interested in, so you can handle adding that as a feature if you prefer.

One thing to mention to @Ratheeshrajan is that in the InlineEditCell.js DatePicker onChange function, it would be good to have it save the cell in the same format as the value. ie. if the incoming cell value was provided a string, then save the cell value as a string. Or even better, have a setting in the inlineEdit configuration for the date picker props that lets a user specify the format to set ie. "date", "string".

You can reach out to me in slack if you would like to discuss more.

Thanks!

elycheea (approved):
Approving after discussion with @georgespano although before we close #2577, I think we should try to simplify and support more default from Carbon date picker (if possible). We can discuss the current implementation once you’re back!

elycheea (approved):

PR #2599, opened 2023-01-06
chore: lint and fix examples

lee-chase:
Examples and scripts were not being linted. Added these folders to the lint scripts and patched as appropriate.

This fix includes a duplicate `.babelrc` in each example folder. This was required in order to run esLint.

Otherwise removing unused code from examples and scripts.


elycheea (approved):
Looks good. Thanks, Lee!

PR #2605, opened 2023-01-09
fix: codesandboxes again

lee-chase:
After another run through the codesandboxes, I'm starting to wonder if this is just going in circles.

The trickiest ones to fix are those requiring a prefix or are currently canary. The other changes are all to try and tweak how or if import-once is pulled in. Codesandbox appears to have a real issue with SCSS node_modules files referring to node_modules.

Prefix example seems to work, on codesandbox with these changes. Not sure why as import-once is definitely being imported.

The two AddSelect components were adjusted slightly and had the Canary controlling switch set to true.


elycheea (approved):
🔨  continuing the whack-a-mole. Thanks Lee.

PR #2611, opened 2023-01-12
fix: tagset overflow first open

lee-chase:
Fixes to #1677 Focus Issue in TagSet component when tooltip is opened for first time

#### What did you change?

Updates the handleChange method to check the event type as indicated in the related Carbon issue.

1. In storybook change looking at the many tags story for TagSet. 
2. Change the width of the container until there are fewer than 10 items in the overflow.
3. Click the over flow.

Previously the tip only opened if the overflow already had focus (before the click). It should now correctly open and close as expected.


elycheea (approved):
🙌 

Doesn’t look like this was an issue on v11, which I think was using the newer ToggleTip so no v11 PR necessary.

PR #2617, opened 2023-01-18
docs(tearsheet): update "portalTarget" prop name in stories

janhassel:
While working with `Tearsheet` I've found myself digging through the source code to find out how to render it in a different node rather than `document.body` because `portalTargetCustomDomNode` didn't work despite being listed in the storybook controls.
I found out that the prop is actually called `portalTarget`.

This PR updates the storybook control to reflect this.

@lee-chase Please let me know if this was intentional / I'm overseeing something.

#### What did you change?
- Update `portalTargetCustomDomNode` to `portalTarget` in the storybook controls for `TearsheetSheel`, `Tearsheet` and `TearsheetNarrow` stories.

Storybook


davidmenendez (dismissed):

lee-chase (changes requested):
This was set to something other than portalTarget to avoid implying that the property is a boolean. 

A better solution would be to change the storybook property to portalTarget and the control type to "control: { type: 'radio' }" or "control: { type: 'select' }" and map the desired value to the property. As it's only two states the radio is probably the best option.

janhassel:
@lee-chase So if I understand correctly, you'd like to see the control as a radio group with two options (e.g. `document.body (default)` and `ref.current`)?

lee-chase:
> 

That's right @janhassel, perhaps check the description is clear to you too.

janhassel:
@lee-chase Upon thinking more about this: should there even be a control for this prop or should it just be disabled in the storybook as there's nothing that really changes visibly to the user playing around with the knobs. This could avoid confusion over what you can pass into `props.portalTarget`

lee-chase:
@janhassel Storybook is both a demonstration and documentation. As a result, we should list all properties.

janhassel:
@lee-chase I'm not talking about removing the prop from the list but just disabling the control as it doesn't produce any visually different output and doesn't really have an appropriate control. So doing the same as you currently do with `props.influencer`, `props.onClose`, `props.navigation` and so on...

janhassel:
@lee-chase I just pushed an update outlining what I mean. Let me know what you think!

lee-chase:
@janhassel a fair point and an acceptable solution. In order for the `portalTarget` to appear on the docs page with a suitable description it appears to be necessary to duplicate the `TearsheetShell.propTypes` entry in Tearsheet and TearsheetNarrow, as per influencer.

janhassel:
@lee-chase Pushed an update. The `portalTarget` prop now appears in the storybook docs argstable.

lee-chase (approved):
Happy with the changes

janhassel:
@davidmenendez Looks like GitHub marked your review as stale over the time. Would you mind checking over it again?

Issue #2505, opened 2022-11-30
Datagrid design review: Nested rows


elycheea:
## Design review

- Component epic #1607 #2499
- Designers/reviewers: Johanna Koval (~@johanna-koval~ @johannasuarez-koval)
- Links: [PAL guidance](https://pages.github.ibm.com/cdai-design/pal/components/data-table/nested-rows/usage) / [Storybook](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-nestedrows)

--- 

### Standards

- [ ] All pattern updates/changes/iterations have been discussed with the
      component developer
- [x] Patterns have been approved by either DSAG or another approving entity

### Pattern and behavior

- [ ] The component behaves according to the guidelines set by the pattern
      maintainer
- [ ] The component’s UI matches the pattern specifications set by the pattern
      maintainer
- [ ] The component’s motion matches the specifications set by the pattern
      maintainer(s)
- [ ] The UI produced is responsive, cross-browser, and responds to the
      currently set Carbon theme.
- [ ] Colors, themes, sizes and additional props are true and correct, aligning
      with Carbon set tokens (unless otherwise specified by the pattern)
- [ ] Paddings/margins/spacings are true and correct, in both desktop and mobile
      views

### Storybook

- [x] A functioning component renders in Storybook
- [ ] The Storybook displays multiple scenarios that show how the component can
      be used
- [ ] Changing props in the Storybook updates the component

johanna-koval:
Hi @davidmenendez, I have done an initial review of the coded expansion compared to the design guidance, and here there a few considerations:

**Expansion indicator**
The expansion line that displays on hover looks great! and display for all levels of nesting. I believe the only adjustment needed is the color style, the color token should be $active-ui (gray 30 in Gray10 theme).
We decided to avoid using blue for these elements since, in general, blue is used to indicate a CTA and in the case of the data grid expansions, we are using blue to indicate a row is selected. 
<img width="500" alt="CleanShot 2022-12-20 at 17 15 03@2x" src="https://user-images.githubusercontent.com/5809654/208777217-ba907519-4081-49c2-9c90-3bd94309898a.png">

**Row divider**
Row dividers are not aligned in the same way as the design. Dividers of a nested row align to the left with the text of the parent row above, similar to how the simple row expansion works 
[...]

PR #2622, opened 2023-01-24
fix(nested rows): fixed border color and align

jagathgj:
Contributes to #2505 

Design review fixes done for the nested rows. Alignment of columns fixed as well as border color changed

#### What did you change?
scss changed and also condition for getCellProps and geRowStyles in useNestedRows changed

storybook


davidmenendez:
tests are failing

davidmenendez (dismissed):

packages/cloud-cognitive/src/components/Datagrid/useNestedRows.js
```diff
@@ -32,12 +32,10 @@ const useNestedRows = (hooks) => {
       style: {
         marginLeft: `${row.depth > 0 ? marginLeft : 0}px`,
```
elycheea, line 33: It looks like we need to remove this margin too in v11?

Seeing extra spacing and text misalignment here.
image

packages/cloud-cognitive/src/components/Datagrid/styles/_useNestedRows.scss
```diff
@@ -50,3 +50,21 @@
...
+.#{$block-class} tr.#{$block-class}__carbon-nested-row {
+  td:first-child:empty,
+  .#{$block-class}__expandable-row-cell {
+    border-bottom: none;
+  }
+}
+
+.#{c4p-settings.$carbon-prefix}--data-table
+  td.#{$block-class}__expandable-row-cell {
+  padding-left: 0;
```
elycheea, line 63: I think this should actually use `$spacing-03`.

Using `padding-left: 0`, the button gets shifted a little to the left.

Collapsed | Expanded
-----------|----------
image | image

elycheea (changes requested):
This one we also need to remove the extra left margin plus adjust the spacing.
I do have this in a PR already but wasn’t able to push to your branch!

packages/cloud-cognitive/src/components/Datagrid/useNestedRows.js
jagathgj: Margin removed.

elycheea (commented):
Looks good but just realized these should probably work onFocus as well.

image

Probably only when the arrow is focused so we may need to add some fixes to  `useNestedRowExpander.js` as well. (Applies to both v10 and v11 PRs.)

Issue #2503, opened 2022-11-30
Datagrid design review: Row height settings


elycheea:
## Design review

- Component epic #1607 #2499
- Designers/reviewers: Virginia Nicholson, Sammy Schuckert
- Links: [PAL guidance](https://pages.github.ibm.com/cdai-design/pal/components/data-table/settings/usage/) / [Storybook](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-rowheightsettings)

---

### Standards

- [x] All pattern updates/changes/iterations have been discussed with the
      component developer
- [x] Patterns have been approved by either DSAG or another approving entity

### Pattern and behavior

- [x] The component behaves according to the guidelines set by the pattern
      maintainer
- [x] The component’s UI matches the pattern specifications set by the pattern
      maintainer
- [x] The component’s motion matches the specifications set by the pattern
      maintainer(s)
- [x] The UI produced is responsive, cross-browser, and responds to the
      currently set Carbon theme.
- [x] Colors, themes, sizes and additional props are true and correct, aligning
      with Carbon set tokens (unless otherwise specified by the pattern)
- [x] Paddings/margins/spacings are true and correct, in both desktop and mobile
      views

### Storybook

- [x] A functioning component renders in Storybook
- [x] The Storybook displays multiple scenarios that show how the component can
      be used
- [x] Changing props in the Storybook updates the component

sammyschuckert:
In general the implementation works as expected and follows most of the PAL guidance. I needed to very nit-picky to find inconsistencies and issues.

### 1. Order of size settings in the menu 
The order of the size settings is descending in the storybook while the PAL guidance uses a ascending order for the size settings.

![image](https://user-images.githubusercontent.com/6014057/211615792-9ab3872e-0652-4a12-b2e8-56fe7053058c.png)

### 2. One size setting labeled wrong and "Small" setting missing
In the Storybook the default rowSize array is listing size settings in the wrong order and labels "Extra small (xs)" as "Small", hence "Small (sm)" is missing.

![image](https://user-images.githubusercontent.com/6014057/211616359-7df000b5-4067-4123-b718-0972fcb77eaf.png)

### 3. Placement of the settings button
[...]

PR #2625, opened 2023-01-24
fix: datagrid design review row height settings

davidmenendez:
Contributes to #2503


packages/cloud-cognitive/src/components/Datagrid/Extensions/RowHeightSettings/RowHeightSettings.stories.js
```diff
@@ -151,7 +155,7 @@ const BasicUsage = ({ ...args }) => {
       {
         Header: 'Someone 11',
         accessor: 'someone11',
-        multiLineWrap: true,
+        multiLineWrap: false,
```
elycheea, line 158: We can probably remove this whole section if we aren’t trying to show multiline in this story anymore.

elycheea (changes requested):
Saw one missing change from the design review that Virginia added afterwards.

The row settings button should match the “flyout” as it does with filter flyout.

Actual | Expected
--|--
image | image

packages/cloud-cognitive/src/components/Datagrid/styles/addons/_RowSizeDropdown.scss
```diff
@@ -16,7 +16,7 @@
   box-shadow: 1px 4px 8px -3px $overlay-01, -1px 6px 8px -5px $overlay-01;
 }
 
-.#{$block-class}__row-size-button--open {
+.#{$carbon-class}--btn--ghost.#{$block-class}__row-size-button--open {
```
elycheea, line 19: This should still work if you just use 

```suggestion
.#{$carbon-prefix}--btn--ghost.#{$block-class}__row-size-button--open {
```

which is how we usually have it in other parts of the Datagrid component.

Then you can remove `$carbon-class` from variables since I don’t think we need to add this one separately?

elycheea (approved):

PR #2639, opened 2023-01-31
feat(Datagrid): exports FilterFlyout from datagridState (v11)

AlexanderMelox:
Exports FilterFlyout from datagridState


elycheea (approved):
Had to fix the base branch to point to v11 instead of `main` but looks good!

Issue #2506, opened 2022-11-30
Datagrid design review: Inline editing


elycheea:
## Design review

- Component epic #1607 #2499
- Designers/reviewers: Jennifer Jennings (@jjennings7)
- Links: [PAL guidance](https://pages.github.ibm.com/cdai-design/pal/components/data-table/inline-editing/usage) / [Storybook](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-inlineedit)

---

### Standards

- [x] All pattern updates/changes/iterations have been discussed with the
      component developer
- [x] Patterns have been approved by either DSAG or another approving entity

### Pattern and behavior

- [ ] The component behaves according to the guidelines set by the pattern
      maintainer
- [ ] The component’s UI matches the pattern specifications set by the pattern
      maintainer
- [x] The component’s motion matches the specifications set by the pattern
      maintainer(s)
- [x] The UI produced is responsive, cross-browser, and responds to the
      currently set Carbon theme.
- [ ] Colors, themes, sizes and additional props are true and correct, aligning
      with Carbon set tokens (unless otherwise specified by the pattern)
- [x] Paddings/margins/spacings are true and correct, in both desktop and mobile
      views

### Storybook

- [x] A functioning component renders in Storybook
- [x] The Storybook displays multiple scenarios that show how the component can
      be used
- [x] Changing props in the Storybook updates the component

matthewgallo:
• Would be helpful to include variant with selectable rows (include `$ui-05` border around selected row)
• Add note about what columns use validation
• Complex cells are missing

elycheea:
Re-assigning from Matt to @Ratheeshrajan 🚀

jjennings7:
@matthewgallo @elycheea @Ratheeshrajan I checked the color tokens the best I could using Inspect in my browser and Carbon Dev Tools, and I found a few discrepancies:
| Interactive states      | Background Tokens | Implementation in Storybook |
| ----------- | ----------- | ----------- |
| Editable cell      | Background: $field-01 | $ui-01 |
| Cell hover   | Background: $hover-selected-ui       | $active-ui |
| Active filled cell      | Background: $active-ui <br> :rotating_light: think the documentation is wrong | $field-02 |
| Focus cell      | Background: $field-01 | $active-ui |
| Focus cell      | Text color: $text-02 | $text-01 |
[...]

PR #2640, opened 2023-01-31
fix(Datagrid): inline edit cell tokens

davidmenendez:
Contributes to #2506


davidmenendez:
hello, @jjennings7, for the time being i'm taking over some datagrid work from Matt. can you please review the changes in this [preview](https://deploy-preview-2640--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-inlineedit--inline-edit-usage-story) for issue #2506

elycheea (approved):
Looks right when comparing against the guidance! Only active filled cell is different than documented, but agree with @jjennings7 that the guidance may need to be updated in this case.

Issue #2648, opened 2023-02-02
Reduce frequency of update workflow


elycheea:
## What will this achieve?

With contributors’ time split between product and library, we should reduce dependency updates to be once a month for the time being.

## Additional information

- Designs
- Existing code
- etc

PR #2652, opened 2023-02-03
chore: reduce dependency update frequency

elycheea:
Contributes to #2648 

Reduce how often we update our dependencies to once a month.

#### What did you change?
GitHub Action for updates.

```
.github/workflows/update.yml
```


davidmenendez (approved):
🤝

Issue #2509, opened 2022-11-30
Datagrid design review: Horizontal scrolling and responsiveness


elycheea:
## Design review

- Component epic #1607 #2499
- Designers/reviewers: Virginia Nicholson, Jennifer Jennings (@jjennings7)
- Links: [PAL guidance](https://pages.github.ibm.com/cdai-design/pal/components/data-table/scrolling/usage) / [Storybook v2](https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-datagrid-datagrid-canary--frozen-columns) / [Storybook v1](https://v1-ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-datagrid-datagrid-canary--frozen-columns) — for unfrozen rows, see batch actions or actions dropdown

---

### Standards

- [x] All pattern updates/changes/iterations have been discussed with the
      component developer
- [x] Patterns have been approved by either DSAG or another approving entity

### Pattern and behavior

- [ ] The component behaves according to the guidelines set by the pattern
      maintainer
- [x] The component’s UI matches the pattern specifications set by the pattern
      maintainer
- [x] The component’s motion matches the specifications set by the pattern
      maintainer(s)
- [x] The UI produced is responsive, cross-browser, and responds to the
      currently set Carbon theme.
- [x] Colors, themes, sizes and additional props are true and correct, aligning
      with Carbon set tokens (unless otherwise specified by the pattern)
- [x] Paddings/margins/spacings are true and correct, in both desktop and mobile
      views

### Storybook

- [x] A functioning component renders in Storybook
- [ ] The Storybook displays multiple scenarios that show how the component can
      be used
- [x] Changing props in the Storybook updates the component

matthewgallo:
• Rename `Sticky column` story to `Frozen columns`

elycheea:
^ FYI @Ratheeshrajan

jjennings7:
@matthewgallo @elycheea @Ratheeshrajan I'm unable to check the boxes in the design review section for some reason. Can one of you please check them off for me? With the exception of the following, all the rest can be marked:

> - [ ] The component behaves according to the guidelines set by the pattern maintainer

See Matt's comment above with the few small changes needed 


> - [ ] The Storybook displays multiple scenarios that show how the component can be used

[...]

PR #2657, opened 2023-02-08
fix(DataGrid): design review horizontal scrolling

Ratheeshrajan:
Contributes to #2509 

- Rename Sticky column story to Frozen columns
- The width of the sticky actions column on right is too wide (width should be 48px, in addition, the column header should also be 48px): Scope - frozen columns, row action buttons
- There should not be any white space/margins at end of the scrollable column header row: Scope - frozen columns, row action buttons
- Frozen column border should use $active-light-ui: Scope - frozen columns, row action buttons, select row, radio select
- Control for multi-wrap column property, multiLineWrap, also potential for adding it once for all columns and not on a per-column basis
- Unfreeze the first column at <672px, and the frozen column header not sticking anymore: Scope - frozen columns, row action buttons, select row, radio select
- Remove "click action menu" from frozen columns story
- Test failure occurred while shifting the select column position.

#### What did you change?
`Datagrid/Datagrid.mdx`
`Datagrid/Datagrid.stories.js`
`Datagrid/useActionsColumn.js`
`Datagrid/useDefaultStringRenderer.js`
`Datagrid/useSelectAllToggle.js`
`Datagrid/useSelectRows.js`
`Datagrid/useStickyColumn.js`
`Datagrid/Datagrid/DatagridSelectAll.js`
`Datagrid/DatagridSelectAllWithToggle.js`
`Datagrid/Extensions/RowActionButtons/RowActionButtons.stories.js`
`Datagrid/styles/_datagrid.scss`
`Datagrid/styles/_useSelectAllToggle.scss`
`Datagrid/styles/_useStickyColumn.scss`
`Datagrid/Datagrid.test.js'

Storybook: stories - frozen columns, row action buttons, select row, radio select


AlexanderMelox:
@Ratheeshrajan the checkboxes seems to be off centered
Screenshot 2023-02-14 at 11 34 32 AM

Ratheeshrajan:
@AlexanderMelox fixed the issue mentioned above.

packages/cloud-cognitive/src/components/Datagrid/styles/_datagrid.scss
```diff
@@ -318,11 +321,13 @@
...
 .#{$block-class}__head-hidden-select-all {
-  padding-right: $spacing-08;
+  padding-right: $spacing-09;
+
   &.#{$block-class}__select-all-sticky-left {
     position: sticky;
     z-index: 1;
-    left: 0;
+    left: $spacing-09;
+    width: 3.4rem;
```
elycheea, line 330: Where is the extra .4rem come from?
It looks like the radios should use a width of 3rem from what I can see.

image

image

packages/cloud-cognitive/src/components/Datagrid/styles/_useStickyColumn.scss
```diff
@@ -23,25 +23,41 @@
...
-.#{$block-class}__left-sticky-column-cell {
+.#{$block-class}__left-sticky-column-header {
   /* stylelint-disable-next-line declaration-no-important */
   position: sticky !important;
+  z-index: 1;
   left: 0;
-  display: flex;
-  align-items: center;
-  border-right: 1px solid $ui-03;
 }
```
elycheea, line 31: Looks like this change actually affects the left sticky columns when there is no index column to the left too — right border is now missing but should also use `$active-light-ui`.

image

elycheea (changes requested):
I’m not sure if we should add the row index scope within this body of work since that’s not actually covered in the PAL guidelines yet and the use case may be a little more specific.
I’m happy to see that you got it working though, but also think there are other design considerations necessary before we include this enhancement — e.g. the border colors when there are multiple fixed columns are a bit different.

Also, because we added this in the default headers, it looks like we needed to make quite a few changes to the test — to me, this just shows _how rigid our current tests are_. We really should be [targeting the queries](https://testing-library.com/docs/queries/about/#priority) more specifically — I do understand that a lot of this rigidity was from previous work though so I’d also be fine with us addressing this later, depending how we move this PR forward.
If we do keep the row index to the left of the sticky columns, we may need to update the tests so they work in multiple scenarios both with and without the index rows since you can see in [customize columns + frozen columns](https://deploy-preview-2657--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-columncustomization--column-customization-with-fixed-column-story), the frozen row border disappears.

packages/cloud-cognitive/src/components/Datagrid/Datagrid.stories.js
```diff
@@ -48,14 +48,19 @@ export default {
...
   {
-    Header: 'Row Index',
+    Header: '',
     accessor: (row, i) => i,
     sticky: 'left',
     id: 'rowIndex', // id is required when accessor is a function.
+    width: 48,
+    centerAlignedColumn: true,
+    disableResizing: true,
+    disableSortBy: true,
```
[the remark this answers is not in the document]
Ratheeshrajan: @elycheea As the row index scenario is not covered in PAL design guidelines, shall i revert the changes related to row index column shift and keep the other changes in this PR?

packages/cloud-cognitive/src/components/Datagrid/styles/_datagrid.scss
Ratheeshrajan: This is a mistake on my side. I'll fix it in the next commit.

packages/cloud-cognitive/src/components/Datagrid/styles/_useStickyColumn.scss
Ratheeshrajan: If we are reverting the row index changes, I hope the border will restore. Otherwise, I'll fix it.

packages/cloud-cognitive/src/components/Datagrid/Datagrid.test.js
```diff
@@ -904,7 +904,7 @@ describe(componentName, () => {
         .getByRole('table')
         .getElementsByTagName('thead')[0]
         .getElementsByTagName('tr')[0]
-        .getElementsByTagName('div')[0]
+        .getElementsByClassName('c4p--datagrid__head-select-all')[0]
         .getElementsByTagName('th')[0]
         .getElementsByTagName('div')[0]
         .getElementsByTagName('input')[0]
```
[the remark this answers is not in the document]
elycheea: That sounds good. :) I just happened to catch that since you had to update these tests for this change.

packages/cloud-cognitive/src/components/Datagrid/Datagrid.stories.js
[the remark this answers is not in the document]
elycheea: Yes, I think we can revery just the row index scenario and open that as an possible enhancement later.

elycheea (approved):
Made one tiny fix to remove a border that I think was added for the row index scenario — only caught it because it appeared in the Select all items story.

image

Rest of the changes looked good. 👍

Issue #2610, opened 2023-01-12
DatagridContent element type is invalid


Billzg13:
## What package(s) are you using?

```
"@carbon/charts": "1.6.1",
"@carbon/charts-react": "1.6.1",
"@carbon/ibm-products": "2.0.0-rc.10",
"@carbon/react": "1.19.0",
```

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

Trying to use a Datagrid component in my app and it breaks on runtime.

I created a fresh react app with these dependencies:
```
"dependencies": {
    "@carbon/charts": "1.6.1",
    "@carbon/ibm-products": "2.0.0-rc.10",
    "@carbon/react": "1.19.0",
    "react": "17.0.2",
    "react-dom": "17.0.2",
    "react-scripts": "5.0.1",
}
```
that builds with `react-scripts` and Datagrid also fails there with the same error.

I would expect `Datagrid` to work with the default webpack configuration, like the rest of  `ibm-product` components. Is there a special configuration required in order to make this work?

> Is this issue related to a specific component?

Yes **Datagrid**

> What did you expect to happen? What happened instead? What would you like to
> see changed?

We were able to find what causes the issue, it is the way DataTable is imported.
For example, at DatagridContent.js file this is how DataTable is de-structured:

```
import { DataTable} from '@carbon/react';

const { TableContainer, Table } = DataTable; <--- TableContainer and Table are undefined
```

Making this change at all the places where DataTable is imported solves the issue.

```
import { TableContainer, Table} from '@carbon/react';
```

> What browser are you working in?

Google chrome

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

I am working on `Turbonomic ux-app` project

## Steps to reproduce the issue

1. Try to use a Datagrid component with any configuration with  `react-scripts`

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-cloud-cognitive/tree/main/examples/cloud-cognitive/codesandbox

### Console errors

![Screenshot from 2023-01-05 16-30-56](https://user-images.githubusercontent.com/36958390/212064069-21c53716-5a04-4020-a063-453ed0ca416d.png)
[...]

PR #2662, opened 2023-02-09
feat: imports all DataTable pieces from `@carbon/react` (v11)

AlexanderMelox:
Contributes to #2610 

Should fix the DatagridContent invalid import bug? Not really sure but someone hinted that this might be a fix.

#### What did you change?
All files that destructure from `Datatable`


elycheea (approved):
Stories and builds look good. Also find this easier to scan what’s being imported so hoping this does fix that issue. 🤞

PR #2664, opened 2023-02-10
chore: add select release

davidmenendez:
moves add select from canary to release. there appears to be a bug in the v11 version, but i'll handle that separately prior to the v11 release.


elycheea (dismissed):

elycheea (approved):
Reapproving after removing `-canary` in `carbon.yml` Storybook links. 🙃

Issue #2579, opened 2022-12-15
Datagrid: Update sortable column headers to use Carbon table


elycheea:
## What will this achieve?

Carbon `DataTable` now includes tabbable column headers for sorting. ([v10](https://v7-react.carbondesignsystem.com/?path=/docs/components-datatable-sorting--usage), [v11](https://react.carbondesignsystem.com/?path=/docs/components-datatable-sorting--default#sorting))

Not sure when this was added, but we should be able to remove the current extension and use the common behavior for sorting instead.

The keyboard interactions came up as something we want to look into alongside with resize during #2507.

## How will success be measured?

Users should be able to sort via keyboard. Expecting that we can replace `useSortableColumns` and clean up some code.

This should be addressed for both v10 and v11.

## Additional information

- Designs
- Existing code
- etc

Ratheeshrajan:
@elycheea Carbon default isSortable attribute not mutating datagridState rows rendering for some reasons. So I have tried to implement keyboard focus and tabbable sorting heders with the existing `useSortableColumns` hook. Could you please have a look at PR #2673 and let me know if this works?

I have tried the following method from carbon [sorting documentation ](https://v7-react.carbondesignsystem.com/?path=/docs/components-datatable-sorting--usage)
`<TableHeader {...getHeaderProps({ header, isSortable: true })}>
  {header.header}
</TableHeader>`

it's rendering the sort icons in the header, but not invoking the sorting in Datagrid.

PR #2674, opened 2023-02-16
fix(DataGrid): sort column focus v11

Ratheeshrajan:
Contributes to #2579
Carbon default `isSortable` attribute not mutating `datagridState` rows rendering for some reasons. So implemented focus and tabbable column headers for sorting in the existing `useSortableColumns` hook.

Tabbable column headers for sorting, users should be able to sort via keyboard. 

#### What did you change?
`styles/_useSortableColumns.scss`
`useSortableColumns.js'
Storybook


elycheea (approved):
Made a few small tweaks so we don’t introduce any new `!important`s and fixed padding on tables that weren’t using the `useSortableColumns` hook. 

Note: Animation was not included in this PR since we weren’t able to get it working consistently.

PR #2676, opened 2023-02-16
feat(Datagrid): Custom hook to handle `ActionSet` disabled state for `FilterPanel` & `FilterFlyout`

AlexanderMelox:
Simple PR, all it does is it encapsulates the logic for disabling the ActionSet buttons into a custom hook and uses it for FilterFlyout too.


elycheea (approved):
Updated the copyright headers but looks good.

PR #2680, opened 2023-02-17
fix: datagrid inline edit focus background color

davidmenendez:
Contributes to #2506


packages/cloud-cognitive/src/components/Datagrid/styles/_useInlineEdit.scss
```diff
@@ -392,6 +392,10 @@ $row-heights: (
   outline-color: $support-01;
 }
 
+.#{$block-class} .#{$carbon-prefix}--text-input:focus {
+  background: $active-ui;
```
elycheea, line 396: I believe @jjennings7 suggested 

```suggestion
  background: $field-01;
```

lee-chase (approved):

Issue #2647, opened 2023-02-02
Increase TearsheetNarrow description width


elycheea:
## What will this achieve?

Enhancement to increase TearsheetNarrow description width from 60% to 80%.

## How will success be measured?

Description width should match that of Carbon modal. (`sm`?)

## Additional information
https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/659

Discussed with @cameroncalder based on previous issue — 60% for narrow tearsheets is a bit too small and we will follow the spacing from the Carbon modal.

- Designs
- Existing code
- etc

PR #2685, opened 2023-02-20
fix(TearsheetNarrow): changed width from 60% to 80%

qbi11y:
Contributes to #2647 

Change the description width from 60% to 80%


elycheea (changes requested):
The increased description width should only apply to TearsheetNarrow. This currently affects the wide Tearsheet as well, which was intended to have 60% max to the description.

PR #2690, opened 2023-02-21
feat(Datagrid): Extracts Filtering logic into `useFilters` hook

AlexanderMelox:
This PR takes the combined logic from filter flyout and panel and merges them into a hook so that the code stays DRY.


Ratheeshrajan:
@AlexanderMelox Noticed a small issue when we are applying filters, the horizontal scrollbar is not visible

AlexanderMelox:
> @AlexanderMelox Noticed a small issue when we are applying filters, the horizontal scrollbar is not visible image

We noticed this and are making a separate issue for it

Ratheeshrajan (approved):
Looks good to me.

Issue #2329, opened 2022-09-29
Fields in OptionsTile v11 using incorrect color


elycheea:
## Detailed description

> Describe in detail the issue you're having.

Fields (dropdowns in this case) match the layer color instead of contrasting against them.

![image.png](https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/befcb29d-9bff-4536-8774-053dcd2f4eb8)

> Is this issue related to a specific component?

OptionsTile

> What did you expect to happen? What happened instead? What would you like to
> see changed?

![image.png](https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/3c0ba12b-12fb-4aad-a2bd-aaf6d7727c3d)

PR #2693, opened 2023-02-22
fix(options-tile): use layer

janhassel:
Closes to #2329 

(v11 only change)

#### What did you change?
- Wrap OptionsTile content in `<Layer>`
- Remove `props.light` from demo components in storybook

Storybook


AlexanderMelox (approved):
LGTM thanks @janhassel 🚀

PR #2694, opened 2023-02-22
docs(tearsheet): update "portalTarget" prop name in stories (v11)

janhassel:
v11 port of #2617 

See original PR for details.


elycheea (approved):
Thanks for this change.

Discussion on approach documented in #2617

PR #2697, opened 2023-02-22
feat(Datagrid): Extracts Filtering logic into useFilters hook (v11)

AlexanderMelox:
This PR takes the combined logic from filter flyout and panel and merges them into a hook so that the code stays DRY.

### DIFF
--- a/packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridContent.js
+++ b/packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridContent.js
@@ -27,7 +27,7 @@ const blockClass = `${pkg.prefix}--datagrid`;
 
 export const DatagridContent = ({ datagridState }) => {
   const { state: inlineEditState, dispatch } = useContext(InlineEditContext);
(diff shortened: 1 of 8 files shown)


AlexanderMelox:
Found an issue with checkbox, need to look into it

davidmenendez (approved):

Ratheeshrajan:
I am not sure this is an issue, some times I need to click twice on the button to open the filter panel/flyout.

https://user-images.githubusercontent.com/305492/223712236-86ce75b3-1f97-48e8-9383-11d786b06b86.mov

AlexanderMelox:
@Ratheeshrajan This is because our `DatagridToolbar` re renders multiple times in run time and causes the state to not go through sometimes. @elycheea This was one area I felt like we needed to refactor.

elycheea:
@AlexanderMelox @Ratheeshrajan Let’s open an issue for this — I think the re-rendering issues are going to be noticeable to consumers/users so worth addressing.

Ratheeshrajan:
@elycheea @AlexanderMelox opened a new issue for the above #2725. Otherwise, all the changes look good to me.

Ratheeshrajan (approved):

Issue #2695, opened 2023-02-22
React 17 support for `@carbon/ibm-cloud-cognitive-cdai`


s100:
## What will this achieve?

Currently `@carbon/ibm-cloud-cognitive-cdai` has a peer dependency on either React 16 or React 17. In practice, however, it also has a production dependency on [`react-highlighter`](https://github.com/helior/react-highlighter), which [only supports React 16](https://github.com/helior/react-highlighter/issues/66) and has been unmaintained for 4.5 years. This means that downstream consumers are locked into React 16, which in turn is blocking our React 17 migration.

I suggest that you stop using `react-highlighter` and replace it with something else, either built by yourself or an equivalent replacement which does support React 17+.

## How will success be measured?

If I can specify a dependency on React 17, then `npm install @carbon/ibm-cloud-cognitive-cdai` without hitting peer dependency clashes, this will be a success.

## Additional information

I can provide an implementation of a very simple replacement `<Highlight>` component, with much more modern code, if that would be useful.

elycheea:
@s100 Can we close this issue or are there more related updates required?

s100:
As of `@carbon/ibm-cloud-cognitive-cdai@1.24.2` it looks like React 17 is fully supported. Thank you.

PR #2698, opened 2023-02-23
fix: Replace `react-highlighter`

s100:
Contributes to #2695

#### What did you change?

Removed `react-highlighter` as a dependency, replacing it with a very small inline component `<Highlight>` which fulfills all of the functionality we use.

I'm relying on CI to do this... this code has also worked locally for us.


asfordmatt (approved):
LGTM

Issue #2649, opened 2023-02-02
[Datagrid]: FilterLeftPanel add "Any" options to Radio buttons and Dropdowns


AlexanderMelox:
## What will this achieve?
This will allow the user to reset their selection after applying filters. 

For example, a user selects an item in the dropdown. Now there is no way to reset that value. This should show an i

## How will success be measured?
User should be able to select any on dropdown and radio buttons to reset their selections

## Additional information

- Designs
- Existing code
- etc

Ratheeshrajan:
Closing as this is addressed in #2701 #2687

PR #2701, opened 2023-02-24
fix(DataGrid): any option in filter panel v11

Ratheeshrajan:
Contributes to #2649 

Filter left panel & filter flyout: add "Any" option to Radio buttons and Dropdowns and make the default selected

#### What did you change?
`addons/Filtering/FilterFlyout.js`
`addons/Filtering/FilterPanel.js`

Storybook


AlexanderMelox (approved):
LGTM 🚀

Issue #2557, opened 2022-12-09
Text is cutoff in TagSet overflow list


AlexanderMelox:
<img width="194" alt="Screenshot 2022-12-09 at 6 40 54 PM" src="https://user-images.githubusercontent.com/12755042/206812614-d955d85d-fbed-4982-9035-fdb44f00b0c9.png">
<img width="169" alt="Screenshot 2022-12-09 at 6 43 20 PM" src="https://user-images.githubusercontent.com/12755042/206812793-2fdb1015-e8cb-4538-a872-dc8cf702911d.png">

## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

When using letters like 'p', the line heights cuts off

> Is this issue related to a specific component?

`<TagSet />`

> What browser are you working in?

Google chrome

elycheea:
During team scrum today, we investigated and found that there’s border radius from the Carbon tag that is the cause and not line height.

PR #2704, opened 2023-02-24
fix(TagSet): prevent text from getting cut off

qbi11y:
Contributes to #2557 
Descenders were being cut off due to border-radius

#### What did you change?
I created an override to remove the border-radius


AlexanderMelox:
@qbi11y Looks like you have some Tearsheet files on here

qbi11y:
Creating new PR: https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2707

PR #2705, opened 2023-02-27
fix(options-tile): add safe area around toggle

janhassel:
Based on a discussion with @aagonzales, to separate the two click areas (header + toggle), this PR adds a safe area around the toggle of an `OptionsTile` which will not trigger the header hover or expand action.
The header's hover however should still be visible behind the toggle when hovering the left area of the header. See mockup below:

image

#### What did you change?
- Added safe area around toggle in `OptionsTile`

Storybook


davidmenendez:
could you please add an additional story in storybook to demonstrate this change?

janhassel:
@davidmenendez I'm not sure if I understand what you mean. You can check out the behaviour in the existing story: https://deploy-preview-2705--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-options-tile-optionstile--options-tile&args=enabled:true

elycheea (approved):
Changes look good and behave as described. 🎉

elycheea (approved):
Re-approving — updated yarn to fix the build errors.

PR #2709, opened 2023-02-28
fix: add select keyboard accessibility adjustments

davidmenendez:
takes a look at some basic keyboard navigation enhancements for add select as well as a few other minor v11 styling fixes. 
this does remove some key "features", namely row focusing styling and displaying hidden functionality until row hover, but i believe these specific interactions are directly responsible for some of the accessibility problems with add select. so if design has a problem i suggest they work with accessibility to find a solution. breadcrumbs for hierarchy mode should now also be accessible via the keyboard.


lee-chase:
@davidmenendez need to run prettier?

davidmenendez:
@lee-chase fixt

davidmenendez:
@lee-chase checks are fixed now

packages/cloud-cognitive/src/components/AddSelect/_add-select.scss
elycheea, line 1: I think you can regain the visibility / hover with something like this ....

```scss
.#{$block-classs}__selections-dropdown {
  visibility: hidden;
}


.#{$block-classs}__selections-row:focus-within .#{$block-classs}__selections-dropdown
.#{$block-classs}--add-select__selections-row:hover .#{$block-classs}__selections-dropdown {
  visibility: visible;
}

```

elycheea (commented):
Overall, the interactions actually feel much smoother via keyboard.

I shared a suggestion for a CSS based solution for getting the dropdowns back on hover/focus-within.

There are some definite visual regressions between v1 and v2 but they’re already in v2 so I will open a migration review once we wrap up the accessibility improveents.

davidmenendez:
> There are some definite visual regressions between v1 and v2 but they’re already in v2 so I will open a migration review once we wrap up the accessibility improveents.

yeah absolutely. this PR was opened prior to the greater accessibility inspection. i think we can just get this in for now to alleviate some temporary headaches, but most of this will be revisited in the follow up PR for the treegrid refactor.

elycheea (approved):
Marking as approved since other accessibility fixes are expected in another PR.

Issue #2710, opened 2023-03-01
DataGrid: dense header not aligned with out description 


Ratheeshrajan:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)

## Detailed description
Datagrid title alignment is off if there is only grid title supplied.

![image](https://user-images.githubusercontent.com/305492/222065425-e9289c6e-57a9-4ced-89ac-53e61449a842.png)

The title and toolbar are aligned on the same line.

## Steps to reproduce the issue

https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-header--header-basic-usage-story&args=gridDescription:;useDenseHeader:true

## Additional information
![image](https://user-images.githubusercontent.com/305492/222066190-674e71b9-05ef-4bb5-8588-b17ecb6b3a8a.png)

Ratheeshrajan:
Closing as #2711 & #2712 merged

PR #2712, opened 2023-03-01
fix(DataGrid): dense header alignment fix v11

Ratheeshrajan:
Contributes to #2710 

Datagrid title alignment is off if there is only grid title supplied.

#### What did you change?
`styles/_datagrid.scss`
Storybook

Check `useDenseHeader` true
Remove value from `gridDescription`


elycheea (approved):
Caught a separate issue #2724 during review, but this addresses the current issue with the dense header and description. 👍

Issue #2716, opened 2023-03-03
Datagrid - Not able to Hide a column from Column customization if there should not be an option to unselect it from manage column modal


rakesh-musturi:
## Title line template: [Title]: Brief description
Datagrid Component has ColumnCustomization option where we can check uncheck columns which can be hidden from the datagrid, how can we restrict user from managing a particular column which should not be able to select deselect even from the manage columns options.

## What package(s) are you using?

- [ ] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)

## Detailed description

> Describe in detail the issue you're having.

> Is this issue related to a specific component?

> What did you expect to happen? What happened instead? What would you like to
> see changed?

There should be a prop, which helps disabling  column being customized

> What browser are you working in?
Chrome

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Go to Column Customization story in extension and make any column header title blank and make that column as sticky with any icon as column value
2. Go to Customize column action and click the Customize column icon
4. You will see column names with checkboxes, but for the column which has header title as blank only shows checkbox without title, this column should be hidden where user should not be able to customize this column as this has an icon with a custom action which should not be hidden at any cost

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-cloud-cognitive/tree/main/examples/cloud-cognitive/codesandbox

## Additional information

- Screenshots or code
-
<img width="534" alt="Screenshot 2023-03-03 at 12 43 16 PM" src="https://user-images.githubusercontent.com/123563467/222655565-607f9637-cb74-4e02-ade6-81bc82eedcb7.png">
<img width="906" alt="Screenshot 2023-03-03 at 12 42 55 PM" src="https://user-images.githubusercontent.com/123563467/222655629-28789b78-38a0-4fd9-97f3-934d44b8cf82.png">

PR #2717, opened 2023-03-03
Fix for Issue #2716

rakesh-musturi:
Contributes to #

The fix is related to the below issue 
[https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/2716](url)
This Fix Gives an option in Datagrid to restrict a column from being customized based on a prop

#### What did you change?

A new boolean prop `hideFromCustomizeColumn` has been created which will be used as a check in the `Columns.js` files

Have the made the changes in `ColumnCustomization.stories.js` so that people can view the same and also added the props details in `Datagrid.mdx`

### DIFF
--- a/packages/cloud-cognitive/src/components/Datagrid/Datagrid.mdx
+++ b/packages/cloud-cognitive/src/components/Datagrid/Datagrid.mdx
@@ -137,6 +137,24 @@ const columns = [
   },
 ];
 ```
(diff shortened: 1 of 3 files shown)


lee-chase:
@elycheea @rakesh-musturi  Suggest close in favour of #2745 

Please re-open if in error

rakesh-musturi:
Issue is not fixed

PR #2723, opened 2023-03-07
fix: datagrid inline edit focus background color v11

davidmenendez:
Contributes to https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/2506


lee-chase (commented):
@davidmenendez should this not be $field-01?

PR #2730, opened 2023-03-09
fix: cardfooter button icon props update v11

davidmenendez:
Contributes to #2729


elycheea (commented):
The `secondaryButtonIcon` still uses `PropTypes.string` — probably worth going ahead and updating that as well.

elycheea (approved):

PR #2731, opened 2023-03-10
feat: upgrade Carbon dependencies to latest compatible versions

SimonFinney:
Contributes to #1869

Upgrade Carbon 11 dependencies to the latest compatible versions — later `@carbon/react` versions fail tests


elycheea (approved):
Thanks for making this update!

FYI, I think some of the semver syntax was still getting overwritten during `yarn upgrade:automatic` in our v10 branch. Might be related to `npm-check-updates`, but hopefully not an issue for long once we address the deprecated props in AddSelect 🤞 .

PR #2738, opened 2023-03-14
fix(Tearsheet): Correct Tearsheet description width

qbi11y:
Contributes to #
Description width for wide tearsheet is now 60% while narrow is 80%

#### What did you change?
Changed the description width from 80% to 60%


elycheea (approved):
Changes look good now. Thanks @qbi11y 👍

Issue #2747, opened 2023-03-17
Investigate and define mechanisms to allow partial release of work through feature flags


lee-chase:
Some components are very large with lots of features that could perhaps be delivered separately. 

While we currently have a feature flag mechanism there is no way to indicate a feature is flagged in storybook. It is also unclear how we could deliver some features such as hooks and extensions in like those in DataTable without breaking some react rules e.g. hooks cannot be called conditionally.

Investigate and create a proposal for discussion.

PR #2748, opened 2023-03-17
feat: possible feature flag usage

lee-chase:
Contributes to #2747 

#### What did you change?

- Updated the cloud-cognitive README on feature flags
- Added an example around `useInfiniteScroll` which should not be merged. 
- Added a sample test using the feature flag
- Augmented the useInfiniteScroll story
- Added a notification for stories that are flagged via `preview.js`

image

#### Other options

- Add a wrapper around functions such as useInfiniteScroll as we do with components. However given the rules of hooks do not allow conditional calling we would need to test such a mechanism very careful.
- Move flagged stories to an experimental section. As the location of all stories is determined by component displayName we would need to update this mechanism which `title` defined in a story file default export.


packages/cloud-cognitive/src/components/Datagrid/Datagrid.stories.js
```diff
@@ -236,6 +241,9 @@ export const InfiniteScroll = () => {
     </Wrapper>
   );
 };
+InfiniteScroll.args = {
+  usesFeatureFlags: ['DataGrid.useInfiniteScroll'],
```
elycheea, line 245:
```suggestion
  usesFeatureFlags: ['Datagrid.useInfiniteScroll'],
```

elycheea (commented):
As mentioned on scrum, probably normal warning (yellow) is more appropriate than error. I think you’re changing the example anyway but wanted to post the comment so it doesn’t get lost in limbo. :)

packages/cloud-cognitive/src/components/Datagrid/Datagrid.stories.js
lee-chase: Fixed.

elycheea (approved):
Made some pretty minor fixes:
- Changed to low contrast notification (when in doubt)
- Renamed “More information” to “Learn more” (following Carbon content guidance on actions)
- Removed redundant “view README” since this is linked in the action

Issue #2485, opened 2022-11-22
Tooltip truncates for the Sidepanel toolbar buttons


murtdagi:
If the tooltip is longer which is the case for some of the non-english languages, the tooltip for the Sidepanel toolbar action buttons are truncated as seen in the attached screenshot.

can we please have this fixed?

![image](https://user-images.githubusercontent.com/7819852/203297078-d86b8dc9-03d3-49ce-8043-c09447462593.png)

elycheea:
@murtdagi We are closing this issue since it is working in our examples, but if you are still seeing this issue, feel free to reopen with a CodeSandbox so we can investigate further.

murtdagi:
@elycheea Just realized that the tooltip truncates when the `animateTitle` is set to `true`.

murtdagi:
@elycheea can you please reopen this issue for further investigation?

elycheea:
@murtdagi Thanks for providing that additional information. We can look into this further.

elycheea:
@murtdagi Wanted to follow up that `tooltipPosition` and `tooltipAlignment` have been exposed and the default alignment has been changed to `start` in the PRs that are linked above. Closing this as completed, but let us know if you still run into issues. 👋

murtdagi:
Thanks @elycheea , will test this out soon and will let you know if i still see any issues.

PR #2757, opened 2023-03-21
feat(SidePanel): expose tooltip props

lily-peng:
Contributes to #2485

Exposes the `tooltipPosition` and `tooltipAlignment` props for the SidePanel action toolbar buttons. Users will no longer be locked into bottom, centered tooltips for action toolbar buttons and can change them for their uses.

#### What did you change?
`tooltipPosition` and `tooltipAlignment` are now exposed props under `actionToolbarButtons`

So far I have tested by setting the props' values in `actionToolbarButtons` and see the correct change in Storybook.

I also tested with `yarn test:c4p` and it passed all tests.


elycheea (approved):
Thank you @lily-peng 🚀

Issue #2628, opened 2023-01-26
Carbon 11 Tearsheet should use Layer for content


lee-chase:
The Tearsheet sets a background color around the tearsheet. As a result, if the user wishes to use Layers they have to nest two layers.

## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

- The background of the content area is set to --cds-background
<img width="1070" alt="image" src="https://user-images.githubusercontent.com/15086604/214828290-26b2bc5d-a021-4813-a758-21d3fa0e359e.png">
- The first Layer also does this
<img width="1153" alt="image" src="https://user-images.githubusercontent.com/15086604/214828455-8d178012-9200-46c0-8538-63cb322532b8.png">
- A second layer is required when it should not be
<img width="1150" alt="image" src="https://user-images.githubusercontent.com/15086604/214828551-3f1e9652-a5d5-4761-aa1d-ade66951a4fd.png">

> Describe in detail the issue you're having.

> Is this issue related to a specific component?

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> What browser are you working in?

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

Create a tearshet with Carbon11 layers and note the first layer has no effect.

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-cloud-cognitive/tree/main/examples/cloud-cognitive/codesandbox

## Additional information

- Screenshots or code
- Notes

PR #2762, opened 2023-03-22
fix: add `Layer` to `Tearsheet`

SimonFinney:
Contributes to #2628


elycheea (approved):
Clean! Didn‘t actually look into `Wrap` in utils before now but looks good.

FYI — I did see some minor differences in typography between v1 and v2, but it looks like it’s already on our v2 Storybook so we’ll open a separate issue if needed.

Issue #2755, opened 2023-03-21
Novice to pro: Non-linear reading


elycheea:
# Novice to pro: non-linear reading component

## Brief description

The non-linear reading component provides a way to explain concepts more broadly and succinctly while allowing users to read more detail as needed without leaving their context.

## Details

[Novice to pro](https://pages.github.ibm.com/cdai-design/pal/novice-to-pro/overview) is a continuous learning framework aimed to orient, onboard, explain, educate, and cultivate novice users into high-functioning power users.

[Link to design](https://pages.github.ibm.com/cdai-design/pal/novice-to-pro/components/nonlinear-reading/usage)

### Links to other materials

e.g. sketch files, icons, images, prototypes

## Design Maintainer(s)

- Shelley Dignum
- Rebecca Love

## Tasks

Before starting work on this epic, please review and complete the following.

- [x] Initial review of design/existing code

### Working in Carbon for IBM Products package

- [x] Have you reviewed our
      [contribution guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/CONTRIBUTING.md)?
- [x] Have you noted our
      [code guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/docs/CODE_GUIDELINES.md)?
- [x] Did you use the CLI to scaffold your new component?

PR #2768, opened 2023-03-23
feat: add new NonLinearReading component

paul-balchin-ibm:
Contributes to #2755 

The non-linear reading component (see [design](https://pages.github.ibm.com/cdai-design/pal/novice-to-pro/components/nonlinear-reading/usage/)) provides a way to explain concepts more broadly and succinctly while allowing users to read more detail as needed without leaving their context.

_[Novice to pro](https://pages.github.ibm.com/cdai-design/pal/novice-to-pro/overview) is a continuous learning framework aimed to orient, onboard, explain, educate, and cultivate novice users into high-functioning power users._

#### What did you change?

This is a new component.

- Passed design review
- Used the CLI to scaffold a new component 
- Added Storybook stories
- Verified default tests (see one comment)
- Added tests

### DIFF
--- a/packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.js
+++ b/packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.js
@@ -1,0 +1,114 @@
+/**
+ * Copyright IBM Corp. 2023, 2023
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useState } from 'react';
+
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import { ChevronUp16 } from '@carbon/icons-react';
+
+import { getDevtoolsProps } from '../../global/js/utils/devtools';
+import { pkg } from '../../settings';
+
+// The block part of our conventional BEM class names (blockClass__E--M).
+const blockClass = `${pkg.prefix}--non-linear-reading`;
+const componentName = 'NonLinearReading';
+
+// Default values for props
+const defaults = {
+  theme: 'light',
+};
+
+/**
+ * Use non-linear reading when space is limited to share a
+ * brief, at-a-glance, summary of a concept that may require
+ * more explanation for some users.
+ */
+export let NonLinearReading = React.forwardRef(
+  (
+    {
+      // The component props, in alphabetical order (for consistency).
+      children,
+      className,
+      /* other props for NonLinearReading, with default values if needed */
+      definition,
+      theme = defaults.theme,
+      ...rest
+    },
+    ref
+  ) => {
+    const [isOpen, setOpen] = useState(false);
+
+    const handleToggle = () => {
+      setOpen((prevState) => !prevState);
+    };
+
+    const handleKeyDown = (event) => {
+      const { key } = event;
+      if (key === ' ' || key === 'Enter') {
+        handleToggle();
+      }
+    };
+
+    return (
+      <span
+        {...rest}
+        className={cx(blockClass, className)}
+        ref={ref}
+        role="main"
+        {...getDevtoolsProps(componentName)}
+      >
+        {' '}
+        <button
+          type="button"
+          className={cx(`${blockClass}--term-${theme}`, [
+            isOpen
+              ? [`${blockClass}--term-${theme}-open`]
+              : [`${blockClass}--term-${theme}-closed`],
+          ])}
+          onClick={handleToggle}
+          onKeyDown={handleKeyDown}
+        >
+          {children}
+          {isOpen && <ChevronUp16 />}
+        </button>{' '}
+        {isOpen && (
+          <span className={`${blockClass}--body-${theme}`}>{definition}</span>
+        )}{' '}
+      </span>
+    );
+  }
+);
+
+// Return a placeholder if not released and not enabled by feature flag
+NonLinearReading = pkg.checkComponentEnabled(NonLinearReading, componentName);
+NonLinearReading.displayName = componentName;
+
+// The types and DocGen commentary for the component props,
+// in alphabetical order (for consistency).
+// See https://www.npmjs.com/package/prop-types#usage.
+NonLinearReading.propTypes = {
+  /**
+   * The term of the component appears as a pill.
+   */
+  children: PropTypes.node.isRequired,
+
+  /**
+   * Provide an optional class to be applied to the containing node.
+   */
+  className: PropTypes.string,
+
+  /**
+   * The content that appears when the term is toggled open.
+   */
+  definition: PropTypes.node.isRequired,
+
+  /**
+   * Determines the theme of the component.
+   */
+  theme: PropTypes.oneOf(['light', 'dark']),
+};
--- a/packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.stories.js
+++ b/packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.stories.js
@@ -1,0 +1,120 @@
+/**
+ * Copyright IBM Corp. 2023, 2023
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { Link } from 'carbon-components-react';
+
+import {
+  getStoryTitle,
+  prepareStory,
+} from '../../global/js/utils/story-helper';
+
+import { NonLinearReading } from '.';
+import mdx from './NonLinearReading.mdx';
+
+import styles from './_storybook-styles.scss';
+
+export default {
+  title: getStoryTitle(NonLinearReading.displayName),
+  component: NonLinearReading,
+  parameters: {
+    styles,
+    docs: {
+      page: mdx,
+    },
+  },
+};
+
+const defaultProps = {
+  theme: 'light',
+};
+
+const Template = (args) => {
+  return <NonLinearReading {...args} />;
+};
+
+export const nonLinearReading = prepareStory(Template, {
+  args: {
+    children: 'Click to expand this "term"...',
+    definition:
+      '...to reveal its definition. See the Docs tab for more examples.',
+    ...defaultProps,
+  },
+});
+
+export const inSeries = prepareStory(() => (
+  <>
+    After we create your
+    <NonLinearReading
+      definition={
+        <>
+          <Link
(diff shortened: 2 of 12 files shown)


packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+describe(componentName, () => {
+  it('renders a component NonLinearReading', () => {
+    renderComponent();
+    expect(screen.getByRole('main')).toHaveClass(blockClass);
+  });
+
+  it('has no accessibility violations', async () => {
+    const { container } = renderComponent();
+    // await expect(container).toBeAccessible(componentName);
```
paul-balchin-ibm, line 43: I could not pass the accessibility test. 

Adding these extra tests passed, but still threw the error below:
- Focus on body > tab > button has focus? True. 
- Button is visible? True. 

Debug
```jsx
<body>
  <div>
    <span> // <-- Component starts here.
      <button>
        term
      </button>
```
Report ([link](https://able.ibm.com/rules/archives/latest/doc/en-US/element_tabbable_visible.html#%7B%22message%22%3A%22Confirm%20the%20element%20should%20be%20tabbable%2C%20and%20is%20visible%20on%20the%20screen%20when%20it%20has%20keyboard%20focus%22%2C%22snippet%22%3A%22%3Cbutton%20class%3D%5C%22c4p--non-linear-reading--term-light%20c4p--non-linear-reading--term-light-closed%5C%22%20type%3D%5C%22button%5C%22%3E%22%2C%22value%22%3A%5B%22VIOLATION%22%2C%22POTENTIAL%22%5D%2C%22reasonId%22%3A%22potential_visible%22%2C%22ruleId%22%3A%22element_tabbable_visible%22%7D))
```
NonLinearReading › has no accessibility violations
Message: Confirm the element should be tabbable, and is visible on the screen when it has keyboard focus
Level: potentialviolation
Snippet: <button class="..." type="button">
```

Following the link, the documentation for this issue includes a "needs review" notification, but no details.

image

packages/core/story-structure.js
```diff
@@ -140,6 +140,10 @@ const s = [
           'c/TearsheetShell',
         ],
       },
+      {
+        n: 'Novice to pro',
```
paul-balchin-ibm, line 144: **Novice to pro** has been designated as a folder directly beneath **IBM Products**

packages/cloud-cognitive/src/components/NonLinearReading/_non-linear-reading.scss
```diff
@@ -0,0 +1,156 @@
...
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+// Standard imports.
+@import '../../global/styles/project-settings';
+
+@mixin term-closed-defaults {
+  // stylelint-disable-next-line carbon/layout-token-use
+  padding: 1px $spacing-03 1px $spacing-03;
```
elycheea, line 13: We try to stick with spacing tokens or at least rems where we can. Is the 1px just used to achieve a 20px height?

packages/cloud-cognitive/src/components/NonLinearReading/_non-linear-reading.scss
```diff
@@ -0,0 +1,156 @@
...
+    cursor: pointer;
+  }
+}
+@mixin term-open-defaults {
+  @include term-closed-defaults();
+
+  padding-right: 0;
+
+  svg {
+    height: 18px;
```
elycheea, line 30: Was going to recommend

```suggestion
    height: rem(18px);
```
But upon closer inspection, doesn’t seem like this line takes effect right now anyway?

image

packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.stories.js
```diff
@@ -0,0 +1,120 @@
...
+  theme: 'light',
+};
+
+const Template = (args) => {
+  return <NonLinearReading {...args} />;
+};
+
+export const nonLinearReading = prepareStory(Template, {
+  args: {
+    children: 'Click to expand this "term"...',
```
elycheea, line 42: I think curly quotes are more appropriate, but pretty small nit. 🙈 

```suggestion
    children: 'Click to expand this “term”...',
```

elycheea (changes requested):
Overall looking good so far. Fixes for BEM naming conventions are probably more of a recommendation than required since we haven’t been super consistent in the past. (But doesn’t mean we _shouldn’t_ start being more consistent, of course.)

Would recommend enabling the theme prop on all stories, and not just the first one.

Lastly, I do wonder if we should create variables for the NLR colors. It might be fine for now, but I can foresee this being an area where teams may want a little more control in the future. Not sure other products use purple as much so might be worth taking the extra step to give consumers control even if it’s not the preferred/recommended colors. Curious if there was much discussion on color/theming or if the color was selected intentionally for all products/cases?

packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.js
```diff
@@ -0,0 +1,114 @@
...
+        role="main"
+        {...getDevtoolsProps(componentName)}
+      >
+        {' '}
+        <button
+          type="button"
+          className={cx(`${blockClass}--term-${theme}`, [
+            isOpen
+              ? [`${blockClass}--term-${theme}-open`]
+              : [`${blockClass}--term-${theme}-closed`],
```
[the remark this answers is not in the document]
paul-balchin-ibm: I've never worked on a project that rigidly enforced BEM because it was never _that_ important, and I've never worked on a component library where it was very important. Ergo, I'm not an expert. I shall be guided by you :).

I expect there will be more BEM comments with future Novice to Pro comments.

packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.stories.js
paul-balchin-ibm: Had a conversation with Cameron Calder a while back: these are required. Thanks for pointing it out. I'll make the change.

packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.stories.js
```diff
@@ -0,0 +1,120 @@
...
+    definition:
+      '...to reveal its definition. See the Docs tab for more examples.',
+    ...defaultProps,
+  },
+});
+
+export const inSeries = prepareStory(() => (
```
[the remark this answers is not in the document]
paul-balchin-ibm: The `theme` is included in the `{...defaultProps}`.

packages/cloud-cognitive/src/components/NonLinearReading/_non-linear-reading.scss
paul-balchin-ibm: That's weird. Technically that tooltip information is correct. And yet turning that prop on and off, or increasing the height, actually effects a change.

I did some testing, and it turned to be even weirder:
stylelint-errors

packages/cloud-cognitive/src/components/NonLinearReading/_non-linear-reading.scss
paul-balchin-ibm: Yes, the `1px` is to set the height.

Same problem as detailed below: using `rem()` and `carbon--rem()` throw linting errors.

packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.stories.js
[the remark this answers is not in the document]
elycheea: I think you still need to pass `defaultProps` in though — if you check the story, the theme prop isn’t available in the controls.

packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.js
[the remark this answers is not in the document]
elycheea: This is true, which is probably why we (including me) haven’t been consistent about it in the past. I might be the only one who will call out BEM though. 😅  Most of our standards are caught through the linting/CI checks.

That said, if only classnames are an issue, I can hold off BEM comments since it wouldn’t be a very productive to just replace these and probably more prone to error if we were to miss one.

packages/cloud-cognitive/src/components/NonLinearReading/_non-linear-reading.scss
elycheea: I would guess that border is excluded from the the lint rules since a `1px` border is pretty common to add. I think our linting flags these because we prefer to follow the spacing patterns. A lot of our components are based on composing with others. NLR is more of a newly constructed one so I’m not too surprised by this? There may be less need to flag/disable in the other n2p components?

packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.stories.js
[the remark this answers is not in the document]
paul-balchin-ibm: After doing some testing, I found the following: it all requires running through the `Template()`.

```
const Template = (args) => {
  return <NonLinearReading {...args} />;
};
```

This typical story will provide working light/dark radio buttons
```
export const Default = prepareStory(Template, {
  args: {
    ...defaultProps,
    children: 'Term',
    definition: 'Definition',
  },
});
```

This story will not (`prepareStory` is not referring to `Template`)

```
export const Contextual = prepareStory(() => (
  <>
    XDR Connect’s correlation
    <NonLinearReading
      definition="This is a technical component that uses a set of rules to process alerts from your sources, and streamline threat analysis."
      {...defaultProps}
    >
      engine,
    </NonLinearReading>
    creates a case by processing data from alerts across multiple security
    tools.
  </>
));
```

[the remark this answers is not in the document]
paul-balchin-ibm: I'm going to try to create a second template that will accept contextual, surrounding text.

paul-balchin-ibm:
Closing. This PR was made directly against this repo, instead of using a fork.

The updated PR with the same name can be found at #2793

paul-balchin-ibm:
Closing. This PR was made directly against this repo, instead of using a fork.

The updated PR with the same name can be found at #2793

Issue #2324, opened 2022-09-29
DataSpreadsheet release review


elycheea:
### Readiness

- [x] One or more scenarios for a design pattern have been identified as a
      useful unit of functionality to publish.
- [x] A functioning component or components delivering those scenarios have been
      delivered and merged to main.
- [ ] Design maintainer has approved the implementation for those scenarios (via
      a comment on the relevant issue/epic).

### Engineering review

- [x] Breaking changes have only been introduced with prior approval, discussion
      and documented in release notes. Ideally deprecation notices in the prior
      major version must have been added in a timely fashion.
- [x] The implementation takes into account, and does not impair, remaining and
      anticipated design scenarios.
- [ ] Public component features (names, props, etc) are consistent with
      Carbon-defined conventions and are consistent internally. Where there
      isn't an existing convention to apply, ensure robust precedents are being
      established.
- [ ] The UI produced is accessible, responsive, translatable, cross-browser,
      and responds to the currently set Carbon theme.
- [ ] Components are functional components using hooks.
- [ ] Public components which produce DOM structures support className.
- [ ] Public components support a ref (via React.forwardRef).
- [ ] Public component supports a Devtools attribute
- [ ] All significant DOM elements have meaningful classes.
- [ ] Additional attributes that are not identified as props (such as title,
      aria-\*, etc) are passed through to an appropriate DOM node of the
      component as HTML attributes.
- [ ] No warnings, errors or log messages in the console.
- [ ] Each public component JS is exported in /src/components/index.js, each
      public component SCSS is included in /src/components/\_index.scss, and
      each public component has a flag in package-settings.js.
- [ ] Each public component SCSS lists all of the Carbon and C&CS components
      imported and used by the JavaScript code and explicitly imports the SCSS
      for each of these components.

### Standards

- [ ] No linter warnings or errors are produced.
- [ ] Carbon tokens (theme, layout, motion) are used unless the design specifies
      otherwise.
- [ ] All components utilizing motion must include reduced-motion queries for
      accessibility purposes - read more here.
- [ ] Code is formatted according to prettier rules (no use of
      //prettier-ignore).
- [ ] Code is clear, maintainable and follows standard React practices and the
      code guidelines.
- [ ] Copyright header in every source file (js, css, scss etc.) with the
      appropriate years.

### Testing

- [ ] There is a set of test cases for the components.
- [ ] The tests exercise all inputs (props, slots, etc) and verify appropriate
      outputs.
[...]

PR #2769, opened 2023-03-23
fix: additional coverage testing for data spreadsheet

davidmenendez:
Contributes to #2324 

it doesn't look like much, but believe me this required extensive testing and code running to get to this point. there's still a lot of coverage left untested that i'm still not 100% sure how to tackle at this point. some of which just requires more deep diving into RTL and data spreadsheet and some of which i don't think is really easily achievable because it involves mocking OS level functionality.

we should definitely look to continue to improve this incrementally. this is probably the best i can do at this point though.


packages/cloud-cognitive/src/components/DataSpreadsheet/DataSpreadsheet.js
```diff
@@ -77,7 +76,7 @@ export let DataSpreadsheet = React.forwardRef(
   (
     {
       // The component props, in alphabetical order (for consistency).
-      cellSize = defaults.cellSize,
+      cellSize,
```
lee-chase, line 79: Why was the default cellSize removed?

lee-chase: @davidmenendez defaults.cellSize setting was also removed.

davidmenendez:
@lee-chase changes made

davidmenendez:
@lee-chase ^^

lee-chase (approved):

Issue #2750, opened 2023-03-20
Decorator with enhanced accessibility


liunate:
# Name of component or pattern

Decorator

## Brief description

Current Decorator can only be a single `<button>` that is the main interaction interface. A secondary interaction interface is via `onContextMenu` property which poses accessibility issue in both discoverability and usability. We need an accessible way to bring out additional context menu as a secondary interaction interface without relying on right-pointer-click.

## Details

[link to figma design](https://www.figma.com/file/U41FcMocAzV2k6ILyIm1Ap/branch/xpExgSyopy9gYkU9eWWCc4/Narrative-View%2F-Columns-Explorations?node-id=602-461226&t=DTG4CfeFYk3i8ewU-0) (IBM internal)

### Links to other materials

e.g. sketch files, icons, images, prototypes

**new decorator storybook story**
- single button
<img width="212" alt="image" src="https://user-images.githubusercontent.com/38996291/226350292-f4354c20-e8c1-47d8-95da-ef2190b52089.png">

- two buttons(left/right)
<img width="440" alt="image" src="https://user-images.githubusercontent.com/38996291/226350466-3d9fc055-322f-4a90-ac8a-1b3cf1e73a21.png">

- screenshot from figma showing new decorator variants:
1. as a table cell in square
2. pill-shaped with single button click interaction
3. pill-shaped with two buttons click interaction(left/right)
<img width="1042" alt="image" src="https://user-images.githubusercontent.com/38996291/226311818-4629f018-e6e1-4c7d-ba64-d293f2ce2fdc.png">

## Design Maintainer(s)

- Owner/main maintainer(s)
designer: jamie.godin@ibm.com
developer: me

- Second/backup maintainer(s)
none

## Tasks

Before starting work on this epic, please review and complete the following.

- [x] Initial review of design/existing code

We did some study on existing [decorator](https://carbon-for-ibm-products.netlify.app/?path=/story/security-components-decorator--default) and in figma link attached above jamie has purposed new design.

We have already coded by the purposed design in another repository base one copied `decorator` source code, and want to contribute it back to package `security` to a wider audience could consume. I am in middle of following contribution guidelines to migrate the new decorator codebase to this repository into package `security`.

### Working in Carbon for IBM Products package

- [x] Have you reviewed our
      [contribution guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/CONTRIBUTING.md)?

yup

- [x] Have you noted our
[...]

PR #2772, opened 2023-03-24
feat(decorator): support second button in round-shaped decorator

liunate:
Contributes to #2750 - decorator with enhanced accessibility

#### What did you change?

1. Add one more decorator variants with a second user interaction interface `<button>`, which could play the same role as pointer-right-click(`onContextMenu`).
2. Round shaped decorator. It can replace deprecated `<Pill>`.

1. Keep original jest test cases with additional tests on the 2nd `<button>`. 
2. Keep original storybook stories while adding some new stories showcase 2nd `<button>`.


liunate:
Hello 👋 while I am migrating the new decorator code base, I am wondering whether we need some feature flag for people don't want the new round-shaped decorator? Like in [this style file line](https://github.com/liunate/ibm-cloud-cognitive/blob/a04bc55da2767e10605fcd7e79bd617dc9a6c4af/packages/security/src/components/DataDecorator/_index.scss#L16
https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/a04bc55da2767e10605fcd7e79bd617dc9a6c4af/packages/security/src/components/DataDecorator/_index.scss#L16) some overriding rules making decorator border looks round followed. Will it be a good idea to leave the option to consumers whether they want to move over to new decorator look, or stay at original square-shaped decorator?

FYI you can see new decorator in action by spinning up storybook and go to:
```
iframe.html?knob-Active (`active`)=true&knob-Link (`href`)=&args=&id=decorator--default
```

packages/security/src/components/__tests__/__snapshots__/public-api-test.js.snap
```diff
@@ -3555,6 +3571,7 @@ Map {
             "danger--ghost",
             "danger--tertiary",
             "tertiary",
+            "ghost-danger",
```
liunate, line 3574: Not quite sure if my change is relevant to the prop type of `FileUploader`(and several snapshot mismatches below). It seems `ghost-danger` is not part of carbon button kind but manually inserted at https://github.com/liunate/ibm-cloud-cognitive/blob/a04bc55da2767e10605fcd7e79bd617dc9a6c4af/packages/security/src/components/Button/Button.js#L20
https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/a04bc55da2767e10605fcd7e79bd617dc9a6c4af/packages/security/src/components/Button/Button.js#L20

In that case, could these `buttonKind` snapshot mismatches be something not updated back at when the line I just mentioned 👆was added?

liunate: > Hello 👋 while I am migrating the new decorator code base, I am wondering whether we need some feature flag for people don't want the new round-shaped decorator?

That said, there is a property `<Decorator noBorderRadius>` which renders the original square-shaped decorator, it's just not `true` by default. 🤷

liunate: Or just the opposite, should I just make it square-shaped by default, and expose some property like `<Decorator pillShaped>` so ppl want to keep original decorator look don't need to make any change at `@carbon/ibm-security` upgrade?

liunate:
@amtoussam @jlongshore I just checked up on the PR status and I'm wondering if there's anything else I can help to progress this PR? Tks~

amtoussam:
@liunate There are conversations designers are having with each other but this implementation. I hope to have an answer for you by next week.

CC: @jlongshore

bobbinrobyn:
@amtoussam - have you synced with Jamie, Cameron, or whoever else needed to be involved? I know Jamie is planning to do a design review ASAP.

jgodin2023:
1. The decorator should not have an underline on hover, as per HTML standards this is not a link as it opens the right side panel after clicking it.

https://user-images.githubusercontent.com/130777697/232563127-cdc12402-0667-41c0-9bc3-6f0b1cea9c69.mov

2. I didn't see if the left-click context menu was available in the repo. 

Screenshot 2023-04-17 at 1 24 35 PM

liunate:
> 1. decorator should not have an underline on hover

The underline is only drawn when the right part of decorator alone is clickable to match existing decorator behavior and the new purposed design in [figma](https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/2750)

Do we want to change that to better match HTML standard?

liunate:
> 2. I didn't see if the left-click context menu was available in the repo.

There are two reasons it's not part of this PR:
1. Carbon haven't had their [Menu component](https://react.carbondesignsystem.com/?path=/story/experimental-unstable-menu--playground) ready for production(still in experiment).
2. I saw various decorator usages: with or without context menu(by various libraries). It seems to me decorator could be one general component that we don't force the context menu, or that context menu could be attached by decorator consumer by their own needs.

jgodin2023:
> > 1. decorator should not have an underline on hover
> 
> The underline is only drawn when the right part of decorator alone is clickable to match existing decorator behavior and the new purposed design in [figma](https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/2750)
> 
> Do we want to change that to better match HTML standard?

Hey Nate, yes! Design has given this some thought as we want to move toward links being navigational, so removing the underline on hover and click will work well in this case.

liunate:
>  if we don't have a context menu showing when the user clicks the left side of the decorator button, what happens?

Current implementation leaves the decision to new decorator consumer. The reason is [mentioned here](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2772#issuecomment-1512291110). 

The biggest obstacle here is we don't have a context menu in shared code base. Current decorator users are using their own menu implementation(home-brewed, external libraries). By leaving the decision of what to do at click to them, they can keep their current implementation. It's also when Carbon menu is ready for production, we could enhance new decorator with that to provide a built-in menu feature on new decorator.

amtoussam:
@jgodin2023

elycheea:
@liunate Any updates on this progress?

liunate:
hi @elycheea sorry for the delay, I was busy with another assignment but will definitely progress this 1st thing next week.

liunate:
What I plan to do is:
1. Rebase to keep good base of this PR head branch
1. Remove the underline per [previous design comment](https://github.com/carbon-design-system/ibm-products/pull/2772#issuecomment-1516308441)
2. Check back here again see if we could move on next stage

As a side note, I just reviewed latest [carbon menu](https://react.carbondesignsystem.com/?path=/docs/components-menu--overview) and it is now production ready. That menu could be the next thing we could include in this component given that context menu is the common decorator use case.

liunate:
> What I plan to do is:
> 
> 1. Rebase to keep good base of this PR head branch
> 2. Remove the underline per [previous design comment](https://github.com/carbon-design-system/ibm-products/pull/2772#issuecomment-1516308441)
> 3. Check back here again see if we could move on next stage

@elycheea @jgodin2023 I just rebased and updated the component by previous comment. Let me know if there's anything else needs to be done to proceed with this PR. Many thanks!

elycheea:
@jgodin2023 I’m running the checks now so once they all pass, there should be a new preview if you could re-review. @jlongshore and I can help with the review from here. Thank you!

liunate:
@elycheea tks for triggering the CI run for me and it's obvious I missed some linting and snapshot things, so I just fixed them right away. What's still blocking is I saw there are couple a11y test failures from `c4p` that does not seem to be related in any way to the `security Decorator`  I am working on. These three `c4p` test cases all failed at the a11y:

```
@carbon/ibm-products:  FAIL  src/components/EditSidePanel/EditSidePanel.test.js (14.743 s)
@carbon/ibm-products:   ● EditSidePanel › has no accessibility violations

@carbon/ibm-products:  FAIL  src/components/CreateSidePanel/CreateSidePanel.test.js (15.734 s)
@carbon/ibm-products:   ● CreateSidePanel › has no accessibility violations

@carbon/ibm-products:  FAIL  src/components/APIKeyModal/APIKeyModal.test.js (18.969 s)
@carbon/ibm-products:   ● APIKeyModal › has no accessibility violations
```

I can reproduce the same locally by running `yarn run ci-check:test:c4p`. Then I logged the rendered DOM to another browser tab and ran DevTool IBM achecker, it shows me the same violation as above.
```jsx
    await wait(3000);
    screen.logTestingPlaygroundURL()screen.logTestingPlaygroundURL();
```

I know at this moment these packages `c4p` and `security` are configured independent versioning, and `c4p` could only actively maintained at another branch `main` instead of `main_v1`. Could that be the reason over time package `c4p` on `main_v1` generating more a11y violation as accessibility check rule updated? Or are those a11y or whatever test fails in that `c4p` package not something we care about for this PR head branch based on `main_v1`?

elycheea:
@liunate I can help look into the that — I think it’s this issue that we fixed in `main` (#3177) but might not have opened a PR for `main_v1` yet.

I think better to keep it as a separate PR that we merge before this one so you can keep the PR scope small here. Will let you know once we address it!

elycheea:
We merged https://github.com/carbon-design-system/ibm-products/pull/3233 this morning — this _should_ resolve the issue now. 🤞

liunate:
> We merged #3233 this morning — this _should_ resolve the issue now. 🤞

@elycheea Thanks for the prompt fix 🙏 Just ran all tests locally again and all passed, so could you please help trigger another round of CI check?

liunate:
Hello there~ just checking in and wondering if there's anything else needed to move forward with this PR? Tks~

jlongshore (approved):
This all LGTM

PR #2773, opened 2023-03-24
chore: add two side panel story

### LINKED ISSUE #2450: SidePanel: slideIn doesn't work when switching multiple side panels

- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

When we switch two side panels, `slideIn` works only for one of them.
For example, in the following animation, the two side panels (side panel 0 and 1) are setup with `slideIn`, but side panel 0 overlays instead when it's switched from side panel 1.
sidepanel

SidePanel

> see changed?

`slideIn` should work with multiple side panels

Google Chrome

> package are you using?

1.35.2

1. Open the CodeSandbox example below
2. Click "Open side panel 0" -> slide in
3. Click "Open side panel 1" -> slide in
4. Click "Open side panel 0" -> overlay

https://codesandbox.io/s/side-panel-example-forked-7hqy9o

lee-chase:
Closes #2450

Adds a demonstration story showing a work around for a issue #2450 

It essentially delays the opening of the next sidepanel.

It may be possible to do this internally in the SidePanel component but further investigation is needed.


lee-chase:
The underlying issue could be marked as resolved, alternatively someone needs to ensure open/close of multiple side panels happens in the desired order.

Issue #2764, opened 2023-03-23
Carbon v11 page header breadcrumbs have wrong styling for the back button.


omar-zahid:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

> Is this issue related to a specific component?
`packages/cloud-cognitive/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js`

> What did you expect to happen? What happened instead? What would you like to
> see changed?
Button should not be of class `cds--btn`

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?
@carbon/ibm-products": "^2.0.0-rc.22

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Additional information

- Screenshots or code
![image](https://user-images.githubusercontent.com/57479337/227104968-0b886ad6-afdb-4c4e-8ff6-ebb2c477c802.png)

- Notes
In the stable version, the button is not a CDS Button.

PR #2774, opened 2023-03-24
fix(page-header): fix 'back' breadcrumb regression

SimonFinney:
Contributes to #2764


packages/cloud-cognitive/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -310,17 +309,15 @@ export let BreadcrumbWithOverflow = ({
         >
           {backItem?.href && (backItem?.label || backItem?.title) && (
             <BreadcrumbItem className={cx(`${blockClass}__breadcrumb-back`)}>
-              <Link
+              <IconButton
```
lee-chase, line 312: This results in an additional 4px of spacing (+6 above and -2px below) over V10.

Applying the following would address this change

```scss
.c4p--breadcrumb-with-overflow__space {
  margin-top: -6px;
 }
```

elycheea (commented):
Not opposed to use including background-color on hover, but if we keep it, might be worth ensuring that back button uses our the good ole’ square style. But might be simpler to remove the hover in this case since v1 also didn’t include it.

elycheea:
@lee-chase pushed a fix for the background-color on the back arrow in breadcrumbs (small screens). Mind taking a look so we can get this one merged?

lee-chase (approved):

Issue #2770, opened 2023-03-24
Tearsheet buttons default to `xl` when they should default to `2xl` in Carbon v11


fbarroso24:
What is the proper way to pass `size` to the [Tearsheet ](https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-components-tearsheet-tearsheet--tearsheet#component-api) `actions` buttons?  The `actions` description says it will pass along additional props to the [Button ](https://react.carbondesignsystem.com/?path=/docs/components-button--default#component-api) component.  When i try `className` it correctly gets passed along to the [Button](https://react.carbondesignsystem.com/?path=/docs/components-button--default#component-api). However, passing along size does not appear to work.  

It appears to default to size=xl every time. 
Screenshot 1 below shows how I'm instantiating size 
![image](https://user-images.githubusercontent.com/8657768/227393148-280673f0-c32c-47e0-9e35-1369a8f9bcd5.png)

& screenshot 2 below shows how it keeps using size='xl' (e.g. cds--btn--xl) regardless.
![image](https://user-images.githubusercontent.com/8657768/227393176-40026193-1e99-4909-bfeb-ddbbdcf258ac.png)

## Title line template: [Title]: Brief description
Tearsheet buttons default to `xl` when they should default to `2xl` in Carbon v11.  Also, consider updating component to allow users to pass in button size as per what is described in the storybook (see below)
![image](https://user-images.githubusercontent.com/8657768/227393088-3fcfb4c6-0023-4177-990a-a40edca9edcc.png)
-->

## What package(s) are you using?
 I'm using @carbon/ibm-products": "2.0.0-rc.13" and @carbon/react": "1.23.0".

- [ ] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.
1.  The default button size for tearsheet should be size='2xl'.  Currently it is `xl`.
2. The size prop is not being forwarded via the [Tearsheet actions](https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-components-tearsheet-tearsheet--tearsheet#component-api) prop? Even though the Storybook currently states that it should be forwarded.
![image](https://user-images.githubusercontent.com/8657768/227393453-e5c0b220-0333-4eb5-996a-a70c061242c1.png)

> Is this issue related to a specific component?
Tearsheet buttons

[...]

PR #2776, opened 2023-03-27
fix(tearsheet): fix button sizes

SimonFinney:
Contributes to #2770


elycheea (commented):
Pretty minor change request, but I do think it’d be worth mentioning in the Tearsheet.propTypes for actions that we exclude `size` from the fields passed to Button.

> Additional fields in the object will be passed to the Button component, and these can include 'disabled', 'ref', 'className', and any other Button props. Any other fields in the object will be passed through to the button element as HTML attributes.

This line seemed to create a little confusion on how `size` should be treated, but verified with Cameron that these should be fixed.

elycheea (approved):

Issue #2752, opened 2023-03-20
Empty state & HTTP errors design review fixes


elycheea:
From the [empty state](https://github.ibm.com/CDAI-design/pal/issues/1139) and [HTTP error](https://github.ibm.com/CDAI-design/pal/issues/1140) migration design reviews —

#### Empty state
- [x] Update spacing under `.#{$block-class}__header` to `$spacing-03` (.5rem) — currently using a derivative .375rem
  - Should be updated in C11 but nice to have in C10
- [x] Update type tokens
  - [x] `lg` variant header/title to use `heading-03` instead of `productive-heading-03` (C10)
  - [x] `sm` variant header/title to use `body-compact-02` instead of `body-short-02` (C10)
  - [x] Description/subtitle to use `body-01` instead of `body-long-01` (C10)

#### HTTP errors
- [x] Update type tokens
  - [x] `productive-heading-02` ➡️ `heading-compact-02`
  - [x] `productive-heading-05` ➡️ `heading-05`
  - [x] `body-short-01` ➡️ `body-compact-01`

C11 changes should be prioritized. C10 changes are nice to have but since it only requires one small change, okay to address C11 only unless there are other related C10 changes we identify.

elycheea:
@allcontributors add @Laura-Marshall for design

PR #2781, opened 2023-03-28
fix(empty-state): update header spacing (#2766)

SimonFinney:
Contributes to #2752 — `v2`


davidmenendez (approved):

elycheea:
Accidentally retagged with https://github.com/carbon-design-system/ibm-cloud-cognitive/labels/status%3A%20ready%20to%20merge instead of the kodiak merge method label so we’ll need to keep testing the label on another branch. 🙃

PR #2782, opened 2023-03-28
fix(Tagset): descender cutoff fix

qbi11y:
Contributes to #

V11 fix for the descenders being cut off

#### What did you change?
Added a class that removes the border radius that is cutting of the descenders if they start a word.


davidmenendez:
looks like there's a problem here. you've got 3k+ file changes 😱

PR #2786, opened 2023-03-30
fix(sidepanel toolbar): tooltip truncate issue fix

jagathgj:
Contributes to #2485 

Sidepanel toolbar's icon tooltip truncate issue fixed 

#### What did you change?
position property changed from sticky to fixed 

storybook


matthewgallo:
Changing from `sticky` to `fixed` seems to cause some layout issues, notably after the panel has scrolled.

image

elycheea (changes requested):
Looks like this addresses the tooltip overflow issue. Also may need some adjustment to the `top` and width with the switch to `position: fixed`. 👀 

image

jagathgj:
I am pushing another change with tooltip alignment on 'start' because it would be better choice while in small screens. Here s the screenshot

image

elycheea (approved):

PR #2787, opened 2023-03-30
fix(sidepanel toolbar): tooltip truncate issue fix

jagathgj:
Contributes to #2485 

Sidepanel toolbar's icon tooltip truncate issue fixed

#### What did you change?
position property changed from sticky to fixed

storybook


davidmenendez (dismissed):

jagathgj:
I am pushing another change with tooltip alignment on 'start' because it would be better choice while in small screens. Here s the screenshot

image

matthewgallo (approved):

PR #2793, opened 2023-03-30
feat: add new NonLinearReading component

paul-balchin-ibm:
Contributes to #2755 

The non-linear reading component (see [design](https://pages.github.ibm.com/cdai-design/pal/novice-to-pro/components/nonlinear-reading/usage/)) provides a way to explain concepts more broadly and succinctly while allowing users to read more detail as needed without leaving their context.

_[Novice to pro](https://pages.github.ibm.com/cdai-design/pal/novice-to-pro/overview) is a continuous learning framework aimed to orient, onboard, explain, educate, and cultivate novice users into high-functioning power users._

#### What did you change?

This is a new component.

- Passed design review
- Used the CLI to scaffold a new component 
- Added Storybook stories
- Verified default tests (see [one comment](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2768/files#r1146718060))
- Added tests

### DIFF
--- a/packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.js
+++ b/packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.js
@@ -1,0 +1,114 @@
+/**
+ * Copyright IBM Corp. 2023, 2023
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useState } from 'react';
+
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import { ChevronUp16 } from '@carbon/icons-react';
+
+import { getDevtoolsProps } from '../../global/js/utils/devtools';
+import { pkg } from '../../settings';
+
+// The block part of our conventional BEM class names (blockClass__E--M).
+const blockClass = `${pkg.prefix}--non-linear-reading`;
+const componentName = 'NonLinearReading';
+
+// Default values for props
+const defaults = {
+  theme: 'light',
+};
+
+/**
+ * Use non-linear reading when space is limited to share a
+ * brief, at-a-glance, summary of a concept that may require
+ * more explanation for some users.
+ */
+export let NonLinearReading = React.forwardRef(
+  (
+    {
+      // The component props, in alphabetical order (for consistency).
+      children,
+      className,
+      /* other props for NonLinearReading, with default values if needed */
+      definition,
+      theme = defaults.theme,
+      ...rest
+    },
+    ref
+  ) => {
+    const [isOpen, setOpen] = useState(false);
+
+    const handleToggle = () => {
+      setOpen((prevState) => !prevState);
+    };
+
+    const handleKeyDown = (event) => {
+      const { key } = event;
+      if (key === ' ' || key === 'Enter') {
+        handleToggle();
+      }
+    };
+
+    return (
+      <span
+        {...rest}
+        className={cx(blockClass, className)}
+        ref={ref}
+        role="main"
+        {...getDevtoolsProps(componentName)}
+      >
+        {' '}
+        <button
+          type="button"
+          className={cx(`${blockClass}__term-${theme}`, [
+            isOpen
+              ? [`${blockClass}__term-${theme}--open`]
+              : [`${blockClass}__term-${theme}--closed`],
+          ])}
+          onClick={handleToggle}
+          onKeyDown={handleKeyDown}
+        >
+          {children}
+          {isOpen && <ChevronUp16 />}
+        </button>{' '}
+        {isOpen && (
+          <span className={`${blockClass}--body-${theme}`}>{definition}</span>
+        )}{' '}
+      </span>
+    );
+  }
+);
+
+// Return a placeholder if not released and not enabled by feature flag
+NonLinearReading = pkg.checkComponentEnabled(NonLinearReading, componentName);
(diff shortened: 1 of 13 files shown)


packages/core/story-structure.js
```diff
@@ -140,6 +140,10 @@ const s = [
           'c/TearsheetShell',
         ],
       },
+      {
+        n: 'Novice to pro',
+        s: ['c/NonLinearReading'],
```
paul-balchin-ibm, line 145: In the Storybook directory layout, **Novice to pro** has been designated as a folder directly beneath **IBM Products**

packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+describe(componentName, () => {
+  it('renders a component NonLinearReading', () => {
+    renderComponent();
+    expect(screen.getByRole('main')).toHaveClass(blockClass);
+  });
+
+  it('has no accessibility violations', async () => {
+    const { container } = renderComponent();
+    // await expect(container).toBeAccessible(componentName);
```
paul-balchin-ibm, line 43: I could not pass the accessibility test. 

Adding these extra tests passed, but still threw the error below:
- Focus on body > tab > button has focus? True. 
- Button is visible? True. 

Debug
```jsx
<body>
  <div>
    <span> // <-- Component starts here.
      <button>
        term
      </button>
```
Report ([link](https://able.ibm.com/rules/archives/latest/doc/en-US/element_tabbable_visible.html#%7B%22message%22%3A%22Confirm%20the%20element%20should%20be%20tabbable%2C%20and%20is%20visible%20on%20the%20screen%20when%20it%20has%20keyboard%20focus%22%2C%22snippet%22%3A%22%3Cbutton%20class%3D%5C%22c4p--non-linear-reading--term-light%20c4p--non-linear-reading--term-light-closed%5C%22%20type%3D%5C%22button%5C%22%3E%22%2C%22value%22%3A%5B%22VIOLATION%22%2C%22POTENTIAL%22%5D%2C%22reasonId%22%3A%22potential_visible%22%2C%22ruleId%22%3A%22element_tabbable_visible%22%7D))
```
NonLinearReading › has no accessibility violations
Message: Confirm the element should be tabbable, and is visible on the screen when it has keyboard focus
Level: potentialviolation
Snippet: <button class="..." type="button">
```

Following the link, the documentation for this issue includes a "needs review" notification, but no details.

image

---

I'll open an issue so we can remember to investigate further.

packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.js
```diff
@@ -0,0 +1,114 @@
...
+    },
+    ref
+  ) => {
+    const [isOpen, setOpen] = useState(false);
+
+    const handleToggle = () => {
+      setOpen((prevState) => !prevState);
+    };
+
+    const handleKeyDown = (event) => {
```
elycheea, line 50: It looks like this is getting called twice? Checking keyboard actions, it looks like the NLR pills opens and closes immediately. I think you can fix this with an `event.preventDefault()` but you probably don’t need [keydown](https://stackoverflow.com/questions/71915364/do-i-need-to-add-onkeydown-to-button-for-accessibility-compliance-wcag-2-0) at all here?

Seeing that issue though, it might be worth included [`aria-expanded`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded) since NLR includes expanded and collapsed state. (Also okay to investigate in a later PR to keep this initial PR small).

packages/cloud-cognitive/src/components/NonLinearReading/_non-linear-reading.scss
```diff
@@ -0,0 +1,157 @@
...
+          @include term-open-defaults();
+
+          // Novice to pro does not always use Carbon defaults/tokens
+          // stylelint-disable-next-line carbon/theme-token-use
+          background-color: $purple-20;
+          color: $text-01;
+
+          &:hover {
+            // stylelint-disable-next-line carbon/theme-token-use
+            background-color: #dcc7ff;
```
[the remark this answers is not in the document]
elycheea: Works for now, haha. I do worry the theming subject could come up later for teams where purple is less prevalent, but I’m always okay with YAGNI.

elycheea (changes requested):

packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.js
paul-balchin-ibm: Just removing the `onKeyDown` and `handleKeyDown` worked great. This is a holdover when the button was originally a span. Accessibility test still fails, though. :/

I've added the `aria-expanded` as suggested, but clearly there's more work to be done hooking up all the different "aria" parts. Thanks for letting us pushing that back to a later date. I'll make a note.

elycheea (approved):
Agree with the decision to move the pill in context only for the story. 🎉 
Rest looks good too!

packages/cloud-cognitive/src/components/NonLinearReading/NonLinearReading.test.js
paul-balchin-ibm: This has been opened as a separate issue.

See https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/2857

PR #2798, opened 2023-03-31
build: update lock file after bumping package version

SimonFinney:
Contributes to #2792 — update lock file after bumping package versions during publishing.


elycheea:
Need to add this to `main` package.json as well and update v1/v2 release actions in `main`.

Issue #2714, opened 2023-03-02
feat(Datagrid): Checkbox view all button in panel


AlexanderMelox:
## What will this achieve?

Adds view all button to checkboxes that surpasses 10 or more

<img src="https://user-images.githubusercontent.com/12755042/216653645-b0e10d69-acb7-4381-9322-087136949792.png" />

https://user-images.githubusercontent.com/12755042/216653645-b0e10d69-acb7-4381-9322-087136949792.png

elycheea:
Just kidding. @matthewgallo reminded that this is actually done.

PR #2802, opened 2023-04-03
feat(Datagrid): adds checkbox view all filters

AlexanderMelox:
Contributes to #2714 

Adds checkbox view all


AlexanderMelox:
This PR is dependent on MultiAddSelect being able to have pre-selected items to work. For now we can close the PR and re open or leave it here until we add that feature to multi add select

PR #2803, opened 2023-04-03
fix: add select keyboard accessibility fixes

davidmenendez:
the v10 version of https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2709


elycheea (commented):
In #2709 we changed the filter to `md` from `lg` but looks like this one is still using the larger search input. Did we miss that change here?

elycheea (approved):
Checked with David — seems like the size is intended.

Issue #2801, opened 2023-03-31
In Datagrid and all the datagrid extensions: Keyboard control no longer working


jjennings7:
## Detailed description

> Describe in detail the issue you're having.

My dev team is preparing to use the Data table extension for inline editing. Until recently, this item used to have really good keyboard control, which was super important for a11y and power users. In the most current version of the storybook, https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-inlineedit--inline-edit-usage-story the keyboard control is completely absent. 

> Is this issue related to a specific component?
This seems to be an issue for the entire Data grid component and not specific to the inline editing extension.
 
> What did you expect to happen? What happened instead? What would you like to
> see changed?
I expected to be able to navigate data tables using my keyboard

> What browser are you working in?
Chrome, firefox

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
DataStage, would like to include usage of this component in CPD 4.7 release

## Steps to reproduce the issue

1. Go to Storybook code
2. Try to navigate using the keyboard (tabbing, return, arrows keys)
3. It doesn't work (keyboard input no longer works at all)

matthewgallo:
Hey @jjennings7, thanks for bringing this up! It will be fixed with the above PRs and hopefully will be included in the release tomorrow.

PR #2804, opened 2023-04-03
fix(Datagrid): address bug with keyboard interaction for inline edit

matthewgallo:
Contributes to #2801

A change to one of the property names in the object passed to the `handleGridKeyPress` function caused the inline editing to stop working. I changed the property name back and things are now working again.

#### What did you change?
```
packages/cloud-cognitive/src/components/Datagrid/Datagrid/DatagridContent.js
```
Storybook


SimonFinney (approved):
LGTM — thanks 🙏🏻

PR #2817, opened 2023-04-05
fix: Add select accessibility refactor v11

davidmenendez:
for #2819 & #2821 

this initial PR addresses keyboard concerns with regards to add select. the main feature being added is arrow key navigation within add select. you should now be able to scroll items with up and down as well as traverse the hierarchy with the right arrow key.

this also breaks up `AddSelectList` into several new components including `AddSelectRow` and `AddSelectFormControl`. the main concern with add select is the usage of native checkboxes and radio buttons. with `AddSelectFormControl` the form controls are now svg's.


packages/cloud-cognitive/src/components/AddSelect/AddSelectFormControl.js
```diff
@@ -0,0 +1,90 @@
+import React from 'react';
+import {
+  RadioButton,
+  RadioButtonChecked,
+  Checkbox,
+  CheckboxChecked,
+} from '@carbon/react/icons';
+import PropTypes from 'prop-types';
+import { pkg } from '../../settings';
+import { UserProfileImage } from '../UserProfileImage';
```
elycheea, line 10: Wonder if we should consider having consumers pass in their own `UserProfileImage` later since this will eventually get deprecated and replaced with `UserAvatar` in v2 ... we might need to support both types? Or we could just update to `UserAvatar` when that time comes. 🤔

packages/cloud-cognitive/src/components/AddSelect/AddSelectFormControl.js
```diff
@@ -0,0 +1,90 @@
...
+
+const componentName = 'AddSelectFormControl';
+const blockClass = `${pkg.prefix}--add-select__selections`;
+
+export let AddSelectFormControl = ({ item, onClick, selected, type }) => {
+  const controlProps = {
+    onClick,
+    id: item.id,
+    className: `${blockClass}-form-control-wrapper`,
+    [`aria-label`]: item.title,
```
elycheea, line 20: Should probably add `size: 20` here — icons appearing a bit smaller than in previous versions.

Current | Previous
--------|---------
image | image

Or was this a conscious decision to use smaller checkboxes since the icons have a thicker stroke at this size also still a bit smaller. 👀 
image

packages/cloud-cognitive/src/components/AddSelect/AddSelectFormControl.js
```diff
@@ -0,0 +1,90 @@
...
+  };
+
+  const getCheckbox = () => {
+    if (selected) {
+      return <CheckboxChecked {...controlProps} />;
```
elycheea, line 25:
```suggestion
      return <CheckboxCheckedFilled {...controlProps} />;
```

davidmenendez:
there's still a few kinks to work out before this is ready. i'll be working on this today.

davidmenendez:
there's still a few kinks to work out before this is ready. i'll be working on this today.

davidmenendez:
@elycheea i addressed most of your concerns and stuff you pointed out. the only exception being multi add select with hierarchy. that's going to require some special consideration and i've already got an issue open specifically to address that.

packages/cloud-cognitive/src/components/AddSelect/AddSelectRow.js
```diff
@@ -0,0 +1,227 @@
...
+    },
+    [setFocus, index]
+  );
+
+  const handleSingleSelection = () => {
+    setSingleSelection(item.id);
+  };
+
+  const onSelectKeyDown = ({ key }) => {
+    if (key === 'Enter') {
```
elycheea, line 81: Hopefully a quick fix, but can we get this working with both <kbd>Enter</kbd> and <kbd>Space</kbd>? I noticed that the actual checkboxes in MultiAddSelect require `space` so this inconsistency feels unexpected.

elycheea (approved):
Approved for improvements to accessibility.

Things to be addressed later, continued improvements to `MultiAddSelect`, `UserProfileImage`/`UserAvatar` (future) handling, surprise updates to profile image background colors. As discussed, these will be logged and addressed separately.

PR #2842, opened 2023-04-10
chore: prototype col header actions

matthewgallo:
Prototype for DSAG showing possible interaction for column header actions within the datagrid, specifically with sortable columns.


janhassel:
@elycheea As mentioned in DSAG, I think it'd be best to bring this up and ideally contribute to Carbon directly by using their `useNoInteractiveChildren` hook – similar to how `ExpandableTile` handles the automatic detection of interactive content and changes its behaviour based on that.

Issue #2810, opened 2023-04-05
Side panel migration review fixes


elycheea:
From the [side panel](https://github.ibm.com/CDAI-design/pal/issues/1142) migration design reviews —

#### Standards
- [ ] Reduce padding of close icon to `8px` (`$spacing-03`)
  - Looks like this occurs specifically when there’s no scrollbar so may need to look into positioning
  - Should fix in C11 but nice to have in C10

    > ![image.png](https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/90e8b42b-ffdc-409b-8d4d-9a4f7c1aaa46)
- [ ] Fix icon button alignment (slightly off-centered)

     > <img src="https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/3c5a7855-3916-4100-a7d2-818bf3374b00" width="270">

### Enhancements
These account for things that may have been missed previously but would be nice to fix if straightforward.

- [ ] Add consistent padding to `labelText` for consistency with title
    - Title appears to use `$spacing-09`
    > ![image.png](https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/1aaf7bba-bc5e-4071-b900-1bd70fa58152)
- [ ] Increase scrolled / sticky title to use `$spacing-05` (1rem) bottom padding

    > ![image.png](https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/461e9a0e-72a6-41d6-b1f0-39ab128df932)
- [ ] Align icon with content when first item in action toolbar is icon
  - Icon should be aligned with the text, `.5rem` from the edge of the panel

  > ![image.png](https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/0c79cbfd-5bee-4a27-bf72-828ba1bee05b)
- [ ] Add `$spacing-03` below action icons when content is scrolled
- [ ] Add top and bottom padding to `.#{$block_class}__body-content`
  - Padding top should use `$spacing-03` / `8px`
  - Padding bottom should use `$spacing-10` / `64px`
- [ ] Remove extra padding from ghost button in action set (other buttons are aligned with text, but ghost button has an extra `16px`

    > ![image.png](https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/72757537-0aa8-475d-96e2-cd855b83cd91)
- [ ] Fix close icon hover state (background-color seems to be overridden)

SimonFinney:
[...]

PR #2848, opened 2023-04-11
fix(side-panel): reduce close button spacing

SimonFinney:
Contributes to #2810


matthewgallo (changes requested):
I'm wondering if we can have the designer/s for the side panel review the deploy preview for this change. The close button overlaps the scrollbar which looks incorrect.
image

lee-chase:
Designers often don't see the scroll bar because they don't have the right system settings.

On a Mac this should be set to `Always`

image

elycheea:
Closing this for now — believe we might need to take another look for how to address this both with/without the scrollbars.

PR #2858, opened 2023-04-12
chore: copy v1 animation for inline edit

lee-chase:
Contributes to #1649 

- Corrects the icon size used in V11 inline-edit (v1 and v2)
- Corrects the animation to match v10 version

NOTE: There may be other issues but this PR is only to match the icon size and animation behaviour of V1.

#### What did you change?

Updated inline edit v1 and 2 CSS and js

Storybook updated CSS snapshot


davidmenendez (changes requested):
i'm seeing a weird bug here when invalid is true and then the input is focused. it probably has to do with the fact that the warning svg is outside the motion

lee-chase:
@davidmenendez yeah, that bug happens in both V10 and V11. Just wanted to get V10 and V11 in the same state before making additional PRs.

Also not sure if recent changes trashed the tooltip.

lee-chase:
Looks like the styling of the icon only button changed somewhat in V11. I've removed some of the CSS which was managing focus, hover etc as Carbon now appears to be styled in a way that is compatible with our input.

davidmenendez (approved):

Issue #2856, opened 2023-04-12
[Datagrid] using useRowSize and useCustomizeColumns hooks together breaks row size customization.


xmannyxfreshx1:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

Using useRowSize and useCustomizeColumns hooks together breaks row size customization. Sending in the necessary props for row size customization works on it's own but fails to load the overflow menu when useCustomizeColumns is enabled.

> Is this issue related to a specific component?

Datagrid

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Both functions should be available when used together.

> What browser are you working in?

Chrome

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

`"@carbon/ibm-products": "1.48.0",`

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

CP4WAIOps

## Steps to reproduce the issue

1. Load a datagrid and enable either useCustomizeColumns or useRowSize hooks and notice they work as expected
2. Use both hooks at the same time, then notice the row size overflow menu does not render when clicking the gear button

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-cloud-cognitive/tree/main/examples/cloud-cognitive/codesandbox

https://codesandbox.io/s/angry-stallman-muehg2?file=/src/Example/Example.js
## Additional information

- Screenshots or code
### Just using useRowSize hook
![Screenshot 2023-04-12 at 12 30 53 PM](https://user-images.githubusercontent.com/13371221/231525514-10ec6a69-b787-44c6-bba3-452123f9003d.png)

Using both hooks
![Screenshot 2023-04-12 at 12 45 23 PM](https://user-images.githubusercontent.com/13371221/231526182-5ec04357-f557-486c-8c11-e0bb1c2588da.png)
(Options do not appear)

- Notes

matthewgallo:
This is now available is `1.49.0`, @xmannyxfreshx1 feel free to reopen if you still see any related issues here!

PR #2864, opened 2023-04-13
fix(Datagrid): fix how `RowSizeDropdown` open state value is set

matthewgallo:
Contributes to #2856 

When using row height settings with customize columns the open value for the `RowSizeDropdown` component does not update (see this [code sandbox](https://codesandbox.io/s/epic-fire-q8c41o?file=/src/Example/Example.js)), this prevents the overflow menu from opening at all. I think setting it with a function that gives us a reliable previous state value will help to fix this.

#### What did you change?
```
packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/RowSize/RowSizeDropdown.js
```
Storybook


AlexanderMelox:
I wonder if this has to do with the datagrid toolbar re rendering issues 🤔

AlexanderMelox (dismissed):
Quick and easy 🤖

AlexanderMelox:
> Ah yeah I wonder if that's a factor here as well. I haven't noticed this behavior except for in the code sandbox that Manny shared, we have that Datagrid combination on our storybook as well which leads me to believe maybe this is an edge case with setting the `isOpen` value.

I'm working on a prototype for a refactor for that so hopefully I can find an easy way to fix this 😓

AlexanderMelox (approved):

PR #2865, opened 2023-04-13
fix(Datagrid): fix how `RowSizeDropdown` open state value is set (v11)

matthewgallo:
Contributes to #2856 

When using row height settings with customize columns the open value for the `RowSizeDropdown` component does not update (see this [code sandbox](https://codesandbox.io/s/epic-fire-q8c41o?file=/src/Example/Example.js)), this prevents the overflow menu from opening at all. I think setting it with a function that gives us a reliable previous state value will help to fix this.

#### What did you change?
```
packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/RowSize/RowSizeDropdown.js
```
Storybook

### DIFF
--- a/packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/RowSize/RowSizeDropdown.js
+++ b/packages/cloud-cognitive/src/components/Datagrid/Datagrid/addons/RowSize/RowSizeDropdown.js
@@ -1,10 +1,8 @@
-// @flow
-/*
- * Licensed Materials - Property of IBM
- * 5724-Q36
- * (c) Copyright IBM Corp. 2021
- * US Government Users Restricted Rights - Use, duplication or disclosure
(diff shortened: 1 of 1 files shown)


AlexanderMelox:
Screenshot 2023-04-13 at 11 14 45 AM

Can we change the icon location of this and customize column they display on the left

matthewgallo:
This one is a bit more complex than it appears, because `RowSizeDropdown` uses `Popover`. But internally the icon button that is also used also uses `Popover` and because of the nested `Popover`s changing the `align` prop to `bottom` introduces some layout issues. Ok if we open a separate issue for this @AlexanderMelox?

AlexanderMelox:
Okay sounds good, will approve and made a new issue for this

https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/2868

AlexanderMelox (approved):

Issue #2700, opened 2023-02-23
Datagrid FilterPanel should support i18n v10


elycheea:
During [PR review for v11 FilterPanel](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2669), we caught a few things that were missed during the v10 PR.

- [ ] Label props for FilterPanel action set and placeholders

Would normally open a separate issue for this, but since it was also found during the FilterPanel PR, I’m okay with us including them together.

- [ ] Datagrid `.--table-container` should [use `overflow-y: auto` instead of `overflow-y: scroll`](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2669/files#diff-530c589e1e7f0592e6567c9125265c7ebc6ede38dfa073ff73eb1b0ecf59ed72L415)

PR #2869, opened 2023-04-13
feat(Datagrid): adds i18n support and change overflow-y

AlexanderMelox:
Contributes to #2700 

Adds support for i18n and changes overflow-y from scroll to auto in the table container


AlexanderMelox:
⚠️ **Attention**: This does not need a v2 since it's already in v2.

elycheea (approved):

Issue #2839, opened 2023-04-06
Improve focus state of user profile image


elycheea:
From the [user profile image](https://github.ibm.com/CDAI-design/pal/issues/1154) migration design reviews —

The focus ring around the user profile image is very thing and can be difficult to distinguish on some colors of the `UserProfileImage`.

![image](https://user-images.githubusercontent.com/1717398/230484373-57c22875-5ecc-4dc2-9ad0-30318e8dac94.png)

davidmenendez:
![profile-focus](https://user-images.githubusercontent.com/6370760/232112258-5cf6a064-2710-4b70-91c2-f380a6a32e49.gif)

how's this?

davidmenendez:
visual changes approved by shaun 👍

PR #2874, opened 2023-04-14
fix: user profile image focus v11

davidmenendez:
Contributes to #2839 

improves the focus state of the user profile image. visual change approved by shaun and marwa

profile-focus


lee-chase (changes requested):
Hi David, changing the border width will change the size of the element potentially causing other things to move as a result.

Can we use a box shadow instead to increase the size of the focus indicator.

davidmenendez:
@lee-chase thanks for the suggestion. fixt

lee-chase (approved):

Issue #1776, opened 2022-03-14
About modal: Update component to match new guidelines


tanteresa:
## What will this achieve?
- About modal component in Storybook https://carbon-for-ibm-products.netlify.app/?path=/story/overview-getting-started--page should reflect what has been designed on the PAL site https://pages.github.ibm.com/cdai-design/pal/patterns/about-modal/usage

## How will success be measured?

I can help QA the coded component.

## Additional information
Guidelines: https://pages.github.ibm.com/cdai-design/pal/patterns/about-modal/usage
Sketch file:https://ibm.box.com/s/yrp9ir128xh257i4qtmqvtg09hdxv7nc

lee-chase:
Hello @tanteresa, my assumption here is that the design has changed or the implementation never met the original design. 

How we got here is not as important as how we move forward. Please document in this issue what you wish to be changed. I will mark as blocked awaiting further information.

elycheea:
Hi @tanteresa!

Checking in on this issue. Can you clarify which parts of the implemented AboutModal needs to be updated to match the PAL?

elycheea:
@lizisawizard Can you take a look at this after we wrap up your filtering issues?

lizisawizard:
@tanteresa Coming back to Elysia's comment - I was wondering if you had any guidance on the following questions:
- The documentation says to use the dark version of the modal. Should I switch the storybook to feature only the dark mode of the component with a light mode tab?
- Does anything change with how tabs work on about modal? I couldn't find anything in the documentation
- Can you clarify the color variables you're using for the modal?
- The documentation mentioned that the version number can link to an external page. Should I add that to the component's functionality?

elycheea:
Targeted for after IBM Products v2 release.

PR #2877, opened 2023-04-14
feat(AboutModal): update to match new guidelines v11

glapadre:
Contributes to #1776

Styling and padding updated to match [pal](https://pages.github.ibm.com/cdai-design/pal/patterns/about-modal/usage/). Updated storybook stories to match. 

#### What did you change?
`
packages/cloud-cognitive/src/components/AboutModal/AboutModal.js
packages/cloud-cognitive/src/components/AboutModal/_about-modal.scss
packages/cloud-cognitive/src/components/AboutModal/AboutModal.stories.js
packages/cloud-cognitive/src/components/AboutModal/AboutModal.test.js
packages/cloud-cognitive/src/components/AboutModal/_storybook-styles.scss
packages/cloud-cognitive/src/__tests__/__snapshots__/styles.test.js.snap
examples/carbon-for-ibm-products/AboutModal/src/Example/Example.js
`

Storybook


elycheea:
@thefirstartist Can you do a quick review of the updated AboutModal before we merge it? Here’s the link to the [deploy preview](https://deploy-preview-2877--v11-carbon-for-ibm-products.netlify.app/).

thefirstartist:
@elycheea All goods except when viewing it in dark theme, the footer bar shouldn't be in white color.  ([design spec](https://pages.github.ibm.com/cdai-design/pal/patterns/about-modal/usage#anatomy)).
image

packages/cloud-cognitive/src/components/AboutModal/AboutModal.js
elycheea, line 1: Since version number is called out in the anatomy and something I think we expect to see consistently, it’s probably worth using a prop instead of injecting it as part of the content.

[Figma](https://www.figma.com/file/GqvE8KdjEfxyS6viCHpX6Q/(v11)-Gray-10---IBM-Products?node-id=9478-404995&t=FCGuSHDztrS3FHvf-0) also shows the version number should use `$text-secondary`.

packages/cloud-cognitive/src/components/AboutModal/AboutModal.test.js
```diff
@@ -136,32 +111,16 @@ describe(componentName, () => {
...
-    expect(tabToSelect).toHaveClass(tabSelected);
-  });
-
-  it('renders a version number', () => {
-    renderComponent({
-      additionalInfo: [{ label: tabLabel1, content: versionNumber }],
-    });
-    screen.getByText(versionNumber);
-  });
-
```
elycheea, line 164: We should keep version number and update this test since version number is a specific element of the AboutModals.

packages/cloud-cognitive/src/components/AboutModal/_about-modal.scss
```diff
@@ -73,9 +65,11 @@ $block-class: #{c4p-settings.$pkg-prefix}--about-modal;
   margin-left: $spacing-03;
```
elycheea, line 65: I think this border-left should update the token to `$border-strong` or `$border-strong-01`.

packages/cloud-cognitive/src/components/AboutModal/AboutModal.js
```diff
@@ -52,13 +47,12 @@ export let AboutModal = React.forwardRef(
       closeIconDescription,
       content,
       copyrightText,
-      legalText,
+      generalText,
```
elycheea, line 50: The new guidance for about modal uses the term `General text` to describe the content that goes here. Thoughts on this naming? I feel like this is equivalent to `description` that we use elsewhere. Alternatively, we could possibly have this just as `children`? 🤔

elycheea: Actually, it could serve the same purpose as `content` ... which if products are using example content currently, in the new design, it probably _should_ fall under the `General text` from guidance. So I think `content` might be more appropriate.

elycheea (changes requested):
Looks good overall minus a few tokens. Actually, I think we could remove the story for `About modal with dark theme` since we have the theme selector now.

Do have some thoughts about 
- adding `version` prop
- `generalText` v. `description` v. `content`

but we can discuss this further with others.

glapadre:
After some further review, adding the `versionNumber` as a prop makes sense so I updated that. I also decided to change `generalText` to `content` as it is the most generic fitting term for what it represents.

packages/ibm-products/src/components/AboutModal/AboutModal.stories.js
```diff
@@ -160,49 +166,60 @@ const commonArgs = {
...
       ...commonArgs,
     },
   }
 );
 
-const withAllBodyStoryName =
-  'About modal with links and legal and copyright text';
-export const withAllBody = prepareStory(
-  Template.bind({}, withAllBodyStoryName, true),
+const withLinksStoryName = 'About modal with links';
```
elycheea, line 201: We can actually probably remove the `with links` story since the `links` prop in Storybook will allow you to add/remove links to the story.

In fact, most things you should be able to accomplish in a 1-2 stories if you update the controls to be more like links. But anything that would be a bit more involved to update (e.g. different variations of `content`) **can definitely be addressed later** than just removing redundant stories like links or even “all props set.”

elycheea (commented):
Looks great! Just had a couple comments on prop names and typing but probably better to discuss with a few others.

Also think we could reduce the stories since _most_ things are doable if you leverage the controls. Some changes to consolidate the stories will probably take a little more time though in which case I’d say to merge this once we resolve the prop discussion and address those when you have more time.

elycheea (approved):

Issue #2753, opened 2023-03-21
<AVT1>An explicit ARIA 'role' is not valid for <th> element within a ARIA role 'table' per the ARIA in HTML specification


zhanttbj:
===== ENVIRONMENT =====
Browser: Version 110.0.1
Test Tool Type: IBM Accessbility Checker

===== DEFECT DESCRIPTION ===== 
An explicit ARIA 'role' is not valid for element within a ARIA role 'table' per the ARIA in HTML specification

=====STEPS TO REPRODUCE =====

Login https://dev.console.test.cloud.ibm.com/projects with with IBM ID & password
Check accessibility

===== EXPECTED RESULT =====
IBM 4.1.2 Name, Role, Value
https://www.ibm.com/able/requirements/requirements/#4_1_2

For all user interface components (including but not limited to: form elements, links, and components generated by scripts), the name and role can be programmatically determined; states, properties, and values that can be set by the user can be programmatically set; and notification of changes to these items is available to user agents, including assistive technologies.

===== SCREEN CAPTURE =====
<img width="825" alt="image" src="https://user-images.githubusercontent.com/110212663/226540186-202c2b2f-c239-4054-b4ca-4c2be69aca18.png">

zhanttbj:
This is from https://github.ibm.com/ibmcloud/projects-ui/issues/646, pls help to check and fix, thanks

elycheea:
@Ratheeshrajan Can you take a look into this issue?

zhanttbj:
Hi Team, can we have this issue fixed by end of April? thanks

Ratheeshrajan:
Closing as #2882 and #2883 merged.

PR #2883, opened 2023-04-18
fix(Datagrid): invalid aria role v11

Ratheeshrajan:
Contributes to #2753 

An explicit ARIA 'role' is not valid for elements within an ARIA role 'table' per the ARIA in HTML specification

#### What did you change?
`role: false` added in `<tr>`, `<th>`, `<td>` elements.
`DatagridHeaderRow.js`
`DatagridRow.js`
Open IBM Equal Access Accessibility Checker and scan.

**Before**
Screenshot 2023-04-18 at 8 42 06 AM

**After**
image


janhassel (approved):
Same as in #2882:

> Not a violation per se apparently but also redundant: The `<tbody>` element currently has `role="rowgroup"` added which is already the implicit role of `<tbody>`. Seems to come from `...getTableBodyRows()` in the various `DataTable…Body` files.
> 
> Should this also be removed in this PR?

Ratheeshrajan:
@janhassel removed explicit role from `tbody` elements.

janhassel (approved):
LGTM

matthewgallo (approved):

Issue #2732, opened 2023-03-10
`@carbon/react@^1.16` upgrade fails tests


SimonFinney:
When attempting to upgrade to `@carbon/react@^1.16` in #2731, `AddSelect` and `AddSelectBody` tests failed:

> Expected test not to call console.warn
>
> Warning: The `light` prop for `Dropdown` has been deprecated in favor of the new `Layer` component. It will be removed in the next major release.

<details>
  <summary>Test output</summary>

```
lerna notice cli v5.5.0
lerna info versioning independent
lerna notice filter including "@carbon/ibm-products"
lerna info filter [ '@carbon/ibm-products' ]
lerna info Executing command in 1 package: "yarn run test"
@carbon/ibm-products:  PASS  src/components/EditSidePanel/EditSidePanel.test.js
@carbon/ibm-products:  PASS  src/components/InlineEditV1/InlineEditV1.test.js
@carbon/ibm-products:  PASS  src/components/CreateSidePanel/CreateSidePanel.test.js
@carbon/ibm-products:  PASS  src/components/PageHeader/PageHeader.test.js
@carbon/ibm-products:  PASS  src/components/DataSpreadsheet/DataSpreadsheet.test.js
@carbon/ibm-products:  PASS  src/components/APIKeyModal/APIKeyModal.test.js
@carbon/ibm-products:  PASS  src/components/InlineEditV2/InlineEditV2.test.js
@carbon/ibm-products:  PASS  src/components/OptionsTile/OptionsTile.test.js
@carbon/ibm-products:  PASS  src/__tests__/styles.test.js
@carbon/ibm-products:  PASS  src/components/Toolbar/Toolbar.test.js
@carbon/ibm-products:  PASS  src/components/EditUpdateCards/EditUpdateCards.test.js
@carbon/ibm-products:  PASS  src/components/Card/Card.test.js
@carbon/ibm-products:  FAIL  src/components/AddSelect/AddSelectBody.test.js
@carbon/ibm-products:   ● undefined › handles multi select submit with modifiers
@carbon/ibm-products:     Expected test not to call console.warn()
@carbon/ibm-products:     Warning: The `light` prop for `Dropdown` has been deprecated in favor of the new `Layer` component. It will be removed in the next major release.
[...]

PR #2908, opened 2023-04-20
fix: warnings in add and select tests

davidmenendez:
Contributes to #2732 

in addition to the aformentioned issue, i also noticed some other warning in the console related to missing labels, which i also addressed here as well as a small css issue.


elycheea:
@davidmenendez updated the title for a hopefully more accurate representation than the issue number. 😜

davidmenendez:
@elycheea thank you! fixt

packages/cloud-cognitive/src/components/AddSelect/AddSelectFilter.js
```diff
@@ -112,15 +112,18 @@ export let AddSelectFilter = ({
...
-                  <Dropdown
-                    id={filterOpts.id}
-                    key={filterOpts.id}
-                    titleText={filterOpts.label}
-                    items={filterOpts.opts}
-                    onChange={(value) => onchangeHandler(value, filterOpts.id)}
-                    selectedItem={getSelectedItem(filterOpts.id)}
-                    label={placeholder}
-                  />
+                  <Layer key={filterOpts.id}>
```
lee-chase, line 115: Does the layer need to be inside the map function?

lee-chase (commented):
Ready to approve, just wondered why the Layer is inside the map.

packages/cloud-cognitive/src/components/AddSelect/AddSelectFilter.js
davidmenendez: my understanding about how layer works is that it needs to surround each individual dropdown https://react.carbondesignsystem.com/?path=/story/components-dropdown--with-layer

davidmenendez: actually sorry i think i misunderstood the implementation. making some changes and then pushing an update

elycheea: I don’t think it needs to wrap each individual dropdown. I’d consider it more like a `div` that wraps a set of dropdowns on the same “layer.”

elycheea:
@davidmenendez Also have conflicts — probably from folder rename.

davidmenendez:
@elycheea i moved `<Layer />` to be top level. i think that should resolve it.

elycheea (commented):
I reset the layer since we were seeing Layer 3 within another Layer 3.

image

Also moved the className to the `Layer` component since we were seeing this.
image

elycheea (approved):

Issue #2910, opened 2023-04-21
[Datagrid] unable to use keyboard to interact with clickable rows


erin-hughes:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

When using the useOnRowClick hook to make all the rows in the Datagrid clickable, it is not possible to tab to the rows or to use the keyboard to interact with them in any way.

> Is this issue related to a specific component?

Datagrid

> What did you expect to happen? What happened instead? What would you like to
> see changed?

I'd expect to be able to use the keyboard to tab through the rows in the datagrid, and to be able to fire the provided onRowClick function when pressing Enter, similar to how you can use the keyboard to interact with rows when using the useSelectRows hook. Instead it's not currently possible to interact with the clickable rows using the keyboard at all.

> What browser are you working in?

Chrome version 112.0.5615.121

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

@carbon/ibm-products 2.0.0-rc.26

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

CP4S

## Steps to reproduce the issue

1. Render a Datagrid component with the useOnRowClick hook
2. Use the tab key to try and interact with the rows in the datagrid
3. See that it's not possible to tab into the datagrid

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/hopeful-tereshkova-ygwzwq?file=/src/Example/Example.js

## Additional information

- Screenshots or code
- Notes

matthewgallo:
Hey @erin-hughes 👋 keyboard support for clickable rows has been added, thanks for your patience. It will be available in the next release, which occurs every Tuesday morning.

matthewgallo:
Should be available to try out in `v1.51.0`. Let us know if there are any issues!

PR #2914, opened 2023-04-24
feat(Datagrid): add keyboard support for clickable rows

matthewgallo:
Contributes to #2910 

Adds keyboard support to the `useOnRowClick` hook.

#### What did you change?
`useOnRowClick.js`
Storybook


AlexanderMelox:
Question, should arrow keys make it the focus go up and down in this situation too?

matthewgallo:
Yeah good question, I don't think so because there is not an active focus region with focusable elements inside (such as seen with the `useInlineEdit` hook or the `DataSpreadsheet`).

lee-chase (approved):

AlexanderMelox (approved):

Issue #2871, opened 2023-04-14
InlineEdit V2 changes from V1


lee-chase:
The InlineEdit V2 component differs in the following ways from V1. This may have been dropped deliberately or be regressions.

- [ ] Ellipsis shown when text exceeds bounds 
![image](https://user-images.githubusercontent.com/15086604/232091916-1d48dcc5-e3ec-4591-a248-0ac13523db20.png)

- [x] A size property mimicking the possible values for a Carbon text field (additional sizes added since v1).

- [x] The facility to have edit icon always visible (useful for lone fields).

- [ ] An option to take up only the space needed (inlineEditFullWidth)

- [ ] A light variant (deprecated in Carbon 11)

- [x] The `Invalid` State only shows when the component has focus.

- [x] The ability to adopt the font settings of it's container (as required by PageHeader).

lee-chase:
@elycheea can we have design comment on these feature changes.

sdignum:
- Ellipsis shown when text exceeds bounds. _V2 is correct as it should match what the regular text input field behavior is (which, the example for carbon's text input field, does not show an ellipsis on overflow)._

 - A size property mimicking the possible values for a Carbon text field (additional sizes added since v1). _I can't comment on whether this property is needed. The field width should be able to be determined by the context of the container it sits in so it's an appropriate size for that use case. Not sure if that needs to be a property in the component or not.

 - The facility to have edit icon always visible (useful for lone fields). _I guess that could be useful. I don't see the harm in allowing it._

 - An option to take up only the space needed (inlineEditFullWidth). _Same comment as second bullet. Not sure how that all works...just needs to be able to be a size that fits the various use cases._

 - A light variant (deprecated in Carbon 11). _I don't think that is needed...the bg color being different when in edit mode is desired here._

 - The Invalid State only shows when the component has focus. _Yes, invalid state should remain if not in focus._

- The ability to adopt the font settings of it's container (as required by PageHeader). _Yes, this should be possible as well._

lee-chase:
@sdignum 
- The question is not as to whether an ellipsis should be shown when editing, how the text should appear when not being edited. In V2 the text is truncated, in V1 and ellipsis is shown. The html `input` element does not allow the consumer to choose to show content overflows with an ellipsis or by making it visible.

![image](https://user-images.githubusercontent.com/15086604/233048308-4dbb9e28-ae42-4849-a559-d435562e664f.png)
[...]

PR #2918, opened 2023-04-24
feat: update InlineEditV2

lee-chase:
Contributes to #2871

There are a number of differences between V1 and V2 Inline edit outlined in the issue.  This addresses some of them.

#### What did you change?

- Invalid state shows even without focus
- rename invalidLabel to invalidText inline with Carbon TextInput
- added option to make edit button always visible.
- update snapshots
- added size to v2
- added option to allow InlineEditV2 to inherit type (allows use in PageHeader title).

Storybook.


davidmenendez (approved):
LGTM

PR #2928, opened 2023-04-25
feat: feature flagging in v1 with example

lee-chase:
Contributes to #2747

The same faetures as PR 2748 but without the DataGrid feature flag bit, just update the example component.

#### What did you change?

Merged feature flag examples down from V2.

Storybook

### DIFF
--- a/packages/cloud-cognitive/README.md
+++ b/packages/cloud-cognitive/README.md
@@ -121,14 +121,19 @@ pkg.component.SidePanel = true;
 // Live dangerously: enable all components!
 pkg.setAllComponents(true);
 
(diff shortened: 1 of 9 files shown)


davidmenendez:
still reviewing the code, but i'm seeing this in the preview
Screenshot 2023-04-25 at 2 05 24 PM

davidmenendez (commented):
code looks good to me. i think the storybook toast just needs to be fixed

davidmenendez (changes requested):
code looks good to me. i think the storybook toast just needs to be fixed

lee-chase:
Good catch @davidmenendez an issue with the V10 notification in effect.

Issue #2851, opened 2023-04-12
Fix grid usage in edit and update tearsheets


elycheea:
From the [edit and update](https://github.ibm.com/CDAI-design/pal/issues/1156) migration design review —

- [ ] **Fix grid usage in full page examples in Storybook** (see #2813)
  - Flexbox or CSS grid are fine since consumer is responsible for actual content

elycheea:
@SimonFinney Might be fastest for you since you made the same fix for create tearsheets, but if you’re starting other work, maybe @jagathgj or @davidmenendez can help with this one.

PR #2929, opened 2023-04-25
fix: Issue 2851

davidmenendez:
Contributes to #2851 

i _think_ this covers the issue. i wasn't able to identify any other spots in tearsheet where grid didn't have an accompanying column.

### DIFF
--- a/packages/cloud-cognitive/src/components/EditTearsheet/EditTearsheet.js
+++ b/packages/cloud-cognitive/src/components/EditTearsheet/EditTearsheet.js
@@ -14,6 +14,7 @@ import {
   SideNav,
   SideNavItems,
   SideNavMenuItem,
+  Column,
 } from '@carbon/react';
 import { pkg } from '../../settings';
 import { getDevtoolsProps } from '../../global/js/utils/devtools';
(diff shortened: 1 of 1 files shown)


lee-chase (dismissed):
Assuming we only want 12 of 16 columns at large sizes here @davidmenendez

elycheea:
@davidmenendez Can you resolve conflicts?

elycheea (changes requested):
I think we’re also missing some styles on EditTearsheet. Might be because of some of the differences in [Create versus Edit pattern](https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2813/files#diff-aa17d2f53d1e3b6cf0967476db457b8f4fee94cf512ff26c705c1f6fde91276eR97), but I think we need to add a class on top of the base Carbon grid to keep our spacing in v1.

Deploy preview | v1
----------------|---
image | image

This is at a pretty large monitor size and a bit less noticeable at smaller sizes, but this would result in some unexpected behavior if we pushed as is.

Also looks like we missed a fix in EditFullPage; the section headers are misaligned.

image

davidmenendez:
i updated the columns to use the same widths as the create tearsheet story

elycheea (approved):
Grid and columns should now be updated to match Create in both in EditFullPage and EditTearsheet.

PR #2932, opened 2023-04-26
feat: Ellipsis for inline edit v2 pkg v2

lee-chase:
Contributes to #1649 

Inline edit v2 is missing an ellipsis when not in edit mode and overflowing it's container.

#### What did you change?

Add an ellipsis in when needed and not editing to inline edit v2.

Storybook


lee-chase:
@davidmenendez another PR for you I think

davidmenendez (approved):

PR #2949, opened 2023-05-01
feat(datagrid): filter panel stories v2

AlexanderMelox:
Contributes to #2523 

This adds panel stories (v2)


lee-chase (changes requested):
This appears to try and use a number of V10 icons. Perhaps a merge from V10 has not gone quite right.

AlexanderMelox:
@lee-chase thanks for looking into that. Fixed it in the last commit

matthewgallo (approved):

Issue #2923, opened 2023-04-25
Tearsheet: add first focusable element


elycheea:
In v2, first focusable element in the Tearsheet requires two tabs. In CreateTearsheet, we do allow users to set initial focus. This should be moved to the base Tearsheet.

> i was looking into this issue and it's definitely doable as is, but the problem is with create tearsheet. there's no good way to enable the focus functionality in tearsheet without it impacting the already built in focus functionality in create tearsheet. my recommendation is to **pull out the focus functionality in create tearsheet and create a general focus function in tearsheet**.

_Originally posted by @davidmenendez in https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/2783#issuecomment-1520757311_

elycheea:
@davidmenendez let’s verify if `selectorPrimaryFocus` in Tearsheet resolves the issue and also verify if this is the intended behavior.

@matthewgallo suggested Tearsheet follows an opt in approach for first focus.

davidmenendez:
confirmed that passing `selectorPrimaryFocus` will work. i think it just may need to be displayed in the documentation to ensure this functionality is surfaced to the user. i'll open up a PR to include `selectorPrimaryFocus` in the storybook stories.

davidmenendez:
i'd also like to point out this comment in this issue https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/2452#issuecomment-1397083681

i think it may still be worthwhile to remove or refactor the specific create tearsheet focus functionality in order to resolve this. additional investigation is required.

PR #2950, opened 2023-05-01
fix: surface selectorPrimaryFocus prop in tearsheet storybook

davidmenendez:
Contributes to #2923

`selectorPrimaryFocus` still works as intended for tearsheet, but i think it just needed to be surfaced as an option in storybook / the docs.


lee-chase (approved):
Makes sense to me.

Issue #2604, opened 2023-01-09
Move v11 examples to Stackblitz


elycheea:
## Additional information

- Designs
- Existing code
- etc

lee-chase:
@elycheea this is blocked until we are able to find a resolution or work around to #2734

PR #2962, opened 2023-05-04
fix: examples and some minor changes

lee-chase:
Contributes to #2604 

Update all examples to Vite.

Things to pay closer attention to:
- Package.json changes (description, kewords, name)
- No thumbnails should have changed
- No gallery.config.json files should have changed

NOTE: 
- example-gallery not yet updated
- PageHeader example is still broken
- pkg-prefix is still broken
- Looks like iconDescription with a label when only label is needed
- WebTerminal is still broken

TODO:
- remove iconDescription when using Label
- remove lodash imports (when @carbon/ibm-products updated)

#### What did you change?

All the examples

Ran each one from a directory outside of the monorepo.


matthewgallo (changes requested):
PageHeader `actionBarItems` props need to be updated to address prop type error since you have now removed `actionBarItems[n].iconDescription` but `actionBarItems[n].iconDescription` is still marked as required in the PageHeader prop types.

matthewgallo:
There are also some linting issues in some of the example directories

matthewgallo:
Not sure why some of these checks aren't running 🤔 there is also a conflict @lee-chase

Issue #2884, opened 2023-04-18
v2 release


elycheea:
### Renaming repository to `ibm-products`
- Should we rename the folder in the monorepo to `ibm-products`
- Should we rename the repository to `ibm-products`
- Might break netlify if we rename ... should look into this
- Use new ibm-products Carbon URL

Major release seems like appropriate timing to move away from the legacy name.

### Order of operations
- [x] Rename `cloud-cognitive` folder @matthewgallo 
  - [x] v1 ~#2904~ #3009
  - [x] v2 https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/2905
- [x] Update dependencies #2911 @elycheea 
- [x] “Final release” of v1 from `main` before renaming to `v1` (1.53.1 from new branch)
- [x] Rename branches ~1 week ahead of release @lee-chase 
- [x] Add `v1-ibm-products.carbondesignsystem.com` @matthewgallo
- [x] Update examples (prefixes @lee-chase) #2999 
- [x] Update v2 migration docs (@matthewgallo)
- [x] Rename repository? (~this one tbd~ sounds like @lee-chase)
    - [x] Update more URLs #3004 #3005
- [x] Update CodeSandbox URLs
- [x] Release v2 _(manual release)_ (probably May 17)

_Can be asynchronous from actual release_
- [x] Add v2 Storybook banner (@elycheea) #2994
- [x] Update Storybook URLs (@elycheea) #3012 #3014

PR #2979, opened 2023-05-10
build: Branch main rename build

lee-chase:
Contributes to #2884 

Updates the build steps targeting V1

#### What did you change?

Build related `main` to `main_v1`

Will need to run a test build


matthewgallo (dismissed):
The only change that I think we need to add here is updating `.github/workflows/release.yml` to point to the new `main_v1` branch. Similar to how is done in the `release-v11.yml` file pointing to the `carbon-v11` branch.

lee-chase:
Think you are correct @matthewgallo, I don't think a ci-check:v1 is needed. Should probably test this assumption.

Issue #2976, opened 2023-05-10
About modal API needs review following changes


lee-chase:
Noted that the additionalInfo is still an array, but we are only rendering the first element. 

While it might be useful to cope with and without array the array version should be deprecated if no longer apprpriate

lee-chase:
@glapadre as you recently changed this is it one you can pick up? I'll be. happy to review.

glapadre:
@davidmenendez thanks for jumping on this, I just saw this message!

PR #2982, opened 2023-05-10
fix: about modal node for footer

davidmenendez:
Contributes to #2976

switches the `additionalInformation` prop in about modal to use a node.


elycheea (approved):
Thanks for the quick fix.

Just made a small fix in the stories to re-add controls for the `additionalInfo` so that we can see between no additional info or multiple logos.

elycheea (dismissed):
Thanks for the quick fix.

Just made a small fix in the stories to re-add controls for the `additionalInfo` so that we can see between no additional info or multiple logos.

lee-chase (approved):

PR #2993, opened 2023-05-12
fix: bring config & version numbers up to those in v1

lee-chase:
The build is still failing, probably need to ensure we update lerna and perhaps other dependencies too.


lee-chase:
This sort of fixes things @elycheea 

A bit concerned as to the change to lerna.json, but the feature is deprecated and using `--immutable` causes an error also.

I think V2 as in 2.0.0, probably has to be release manually from the command line.

elycheea (approved):

PR #3020, opened 2023-05-17
fix: use correct ci-check script in release.yml

matthewgallo:
Release yml has an incorrect ci-check script that doesn't exist anymore. This should use the correct one


elycheea:
@allcontributors add @lee-chase for review and infra

Issue #3001, opened 2023-05-15
Nonlinear Reading stories: the examples under Docs are not loading


paul-balchin-ibm:
## What will this achieve?

There is an error rendering Storybook examples under the [Docs tab](https://deploy-preview-2985--v1-carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-novice-to-pro-nonlinearreading-canary--single-level). This occurred because I had not updated the Storybook names in the docs after changing them in the actual stories.

<img src="https://github.com/carbon-design-system/ibm-cloud-cognitive/assets/52505287/baf20bf3-f15a-406d-ac81-8e1b3e093b2c" width="600"/>

## How will success be measured?

The sample renders under the Docs tab will render as expected.

PR #3025, opened 2023-05-17
fix(non-linear): update story refs in docs tab

paul-balchin-ibm:
Contributes to #3001

Storybook examples under the Docs tab are now loading.

#### What did you change?

Updated Storybook story names in `mdx` file to match.

1. Loaded [Storybook for Non-Linear Reading](https://deploy-preview-3002--v1-carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-novice-to-pro-nonlinearreading-canary--single-level) component.
2. Switched to Docs tab.
3. Saw that the examples had loaded correctly.


elycheea:
@paul-balchin-ibm Looks like the docs includes an example under `Embedded`. Is that not the same as “in-series” / “multiple-level” now?

paul-balchin-ibm:
> Looks like the docs includes an example under `Embedded`. Is that not the same as “in-series” / “multiple-level” now?

Good catch. 

Stories updates
1. Removed `defaultProps` as they were overridden in every example.
2. _Wanted to_ remove `args` from each template, but there seems to be a bug where if `args` is missing as a template argument, then in the Docs tab the "Show code" button changes to the "No code available" button. (!?) I left `args` in along with a comment explaining the problem.

Docs updates
1. Updated example titles.
2. Replaced old "Embedded" story with "With Gradient" story.

Issue #2994, opened 2023-05-12
Add Storybook banners


elycheea:
- [ ] Create v1 and v2 banners @Laura-Marshall 
    - PNG or SVG most likely
    - [ ] Animated version to come later
- [ ] Add banners on Storybooks
- Should consider adding links (v1 to v2 and vice versa — e.g. “Looking for the Carbon 10 version?”)

elycheea:
From Laura: https://ibm.box.com/s/qfjfh1t28z6gvy092myy56gar75rz9q9

elycheea:
@allcontributors add @kristastarr for accessibility and code

elycheea:
@allcontributors add @mikeolasov for design

PR #3033, opened 2023-05-18
docs(storybook): add welcome page

elycheea:
@Laura-Marshall for the banner :raised_hands:

Contributes to #2994 
Includes fixes from @lee-chase and @matthewgallo in #2995 and #3015

#### What did you change?
Add welcome page to v1 Storybook.

image

Storybook


elycheea:
@Laura-Marshall @mikeolasov I know we’re working on an animated version of these, but also thinking if we’ll want to consider dark/light mode as well since we hardcoded the link colors so if someone does change the theme to a dark theme, the links don’t become too light.

@Laura-Marshall Also noticed the v1 banner still uses the DataSpreadsheet. 👀

matthewgallo (approved):

Issue #2796, opened 2023-03-30
Update alignment of tooltip in UserProfileImage


elycheea:
From the [user profile image](https://github.ibm.com/CDAI-design/pal/issues/1154) migration design reviews —

Tooltip position on the user profile image has changed between v1 and v2.

v1 | v2
---|---
<img src="https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/8969ac65-5df8-49d7-98ed-6740e9e2b50d" width="200"/> | <img src="https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/a9a459b8-38ef-4ca5-8faa-f32002eb702c" width="200"/>

Looks like there is now a single `align` prop. We can probably provide some control over setting this (possibly even set a default to minimize regression).

Ratheeshrajan:
Closing as #3044 merged.

PR #3044, opened 2023-05-23
fix(UserProfileImage): tooltip position

Ratheeshrajan:
Contributes to #2796 

Tooltip position on the user profile image has changed between v1 and v2.Provide some control over setting `align` prop for `IconButton`

#### What did you change?
`UserProfileImage/UserProfileImage.js`
`UserProfileImage/UserProfileImage.stories.js`

Storybook


packages/ibm-products/src/components/UserProfileImage/UserProfileImage.stories.js
elycheea, line 1: I believe if we set a default in `UserProfileImage.js`, we might not need to for each of these stories. Or we can just set it in the default args if we still need to.

packages/ibm-products/src/components/UserProfileImage/UserProfileImage.js
```diff
@@ -134,14 +135,14 @@ export let UserProfileImage = React.forwardRef(
...
     );
-
     return (
       FillItem &&
       (tooltipText ? (
         <IconButton
           label={tooltipText}
           className={`${blockClass}__tooltip`}
           kind="ghost"
+          align={tooltipAlignment ? tooltipAlignment : 'bottom'}
```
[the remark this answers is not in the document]
Ratheeshrajan: Updated as requested

packages/ibm-products/src/components/UserProfileImage/UserProfileImage.stories.js
Ratheeshrajan: Updated as requested

elycheea (approved):

PR #3050, opened 2023-05-25
fix: private true prevents lock file from updating

matthewgallo:
Removes `"private": true` from storybook-addon-theme package. This unfortunately prevents yarn.lock from updating when it needs. We will need to find another way to prevent this package from publishing in both the `release-v1.yml` and `release.yml` actions.


paul-balchin-ibm (approved):
LGTM

Issue #2922, opened 2023-04-25
Tearsheet content is on third layer, instead of first


janhassel:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

In #2762, tearsheet content was wrapped in `Layer` which is good but unfortunately the layering is currently off.
The tearsheet content should be on the first layer (currently on the third), the tearsheet header and influencer should be on the second.

> Is this issue related to a specific component?

Tearsheet

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Tearsheet layering is off. This causes visual regressions in g90 and g100 themes as well as preventing the usage of any layers inside the tearsheet content by the consumer.

> What browser are you working in?

Chrome 112

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

`@carbon/ibm-products@2.0.0-rc.26`

## Steps to reproduce the issue

1. Open tearsheet story: https://v11-carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-tearsheet-tearsheet--tearsheet
2. Change theme to either g90 or g100
3. Alternatively, inspect in code. The tearsheet content is on `cds--layer-three`, but should be on `cds--layer-one`.

## Additional information

<img width="1144" alt="image" src="https://user-images.githubusercontent.com/28265588/234267664-5ce303a2-5b87-4643-b482-8a89225165ec.png">

jagathgj:
Fixed and merged it. Hence closing.

PR #3051, opened 2023-05-30
fix(tearsheet): layer removed from body

jagathgj:
Contributes to #2922 

Removed layer from Tearsheet Modal body and added background color to Tearsheet Modal header

#### What did you change?
Tearsheetshell and Tearsheet.scss

storybook

### DIFF
--- a/packages/ibm-products/src/components/Tearsheet/_tearsheet.scss
+++ b/packages/ibm-products/src/components/Tearsheet/_tearsheet.scss
@@ -148,12 +148,14 @@ $motion-duration: $duration-moderate-02;
   }
 
   &.#{$block-class}--wide .#{$block-class}__header {
+    background-color: $layer;
     padding: $spacing-06 $spacing-07;
     border-bottom: 1px solid $border-subtle-01;
     margin: 0;
   }
 
   &.#{$block-class}--narrow .#{$block-class}__header {
+    background-color: $layer;
     padding: $spacing-05;
     border-bottom: 1px solid $border-subtle-01;
     margin: 0;
@@ -270,7 +272,7 @@ $motion-duration: $duration-moderate-02;
   .#{$block-class}__main {
     display: flex;
     flex-direction: row;
-    background-color: $layer;
+    background-color: $background;
     grid-column: 1 / -1;
     grid-row: 1 / -1;
   }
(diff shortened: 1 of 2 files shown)


packages/ibm-products/src/components/Tearsheet/_tearsheet.scss
```diff
@@ -270,7 +272,7 @@ $motion-duration: $duration-moderate-02;
   .#{$block-class}__main {
     display: flex;
     flex-direction: row;
-    background-color: $layer;
+    background-color: $background;
```
janhassel, line 275: We need to overwrite this for narrow tearsheets as they use `layer` as their background instead of `background`:

```scss
.#{$block_class}--narrow .#{$block_class}__main {
  background-color: $layer;
}
```

jagathgj: Updated as requested

packages/ibm-products/src/components/Tearsheet/_tearsheet.scss
```diff
@@ -270,11 +272,15 @@ $motion-duration: $duration-moderate-02;
...
     grid-column: 1 / -1;
     grid-row: 1 / -1;
   }
 
+  .#{$block-class}--narrow .#{$block-class}__main {
```
janhassel, line 280: Since this whole style block is nested in `--tearsheet` and the `--tearsheet--narrow` selector is used on the same element, you'll need to use the sass parent selector (`&`) here.

```suggestion
  &.#{$block-class}--narrow .#{$block-class}__main {
```

jagathgj: Thanks Jan. Updated as suggested.

janhassel (dismissed):
LGTM

I would recommend to remove the "light entry field" from the stories as all the light props have been disabled in favor of using layers.
https://github.com/carbon-design-system/ibm-products/blob/7d5b5013039a6d98e3aab8375a7ea249c44ee4ae/packages/ibm-products/src/components/Tearsheet/Tearsheet.stories.js#L137

But I'll defer to @elycheea for that decision and whether or not it should be part of this PR or a separate one.

elycheea:
@janhassel @jagathgj I think we can go ahead and remove the `light` field here since this PR is already fairly scoped. Thank you!

jagathgj:
Thanks @elycheea . I will change it in this PR itself.

elycheea (approved):

PR #3052, opened 2023-05-30
fix: remove storybook-addon-theme from `main_v1` to address failing v1 release action

matthewgallo:
Removes `storybook-addon-theme` package to avoid publishing issues from the v1 release action (as this package already gets published from the v2 release action). All imports from this package pull from npm and not locally so there should not be any breaking changes there.

#### What did you change?
Removed `config/storybook-addon-theme`
Updated some scripts to reflect the removal
Updated yarn.lock to use the latest version of `@carbon/storybook-addon-theme`
Ran build scripts and storybook scripts locally

cc: @lee-chase


lee-chase (approved):
Makes sense to me

Issue #3043, opened 2023-05-22
fix remaining accessibility concerns in add select


davidmenendez:
the following are the remaining accessibility issues in add select that should be addressed as per Jan Hassel

Story: Default-

* row
  * Missing attributes: aria-level, aria-posinset, aria-setsize, aria-selected
  * Radio button SVG has aria-label={rowLabel}  and role={img}. This is unnecessary and causes the label to be read twice by screen readers. Remove the aria-label and role and add aria-hidden="true". (all of this should actually be the default when using a Carbon icon, are you doing any overrides?)
  * Radio button SVG should have cursor: pointer style
  * 
* cell
  * Missing attribute: role="gridcell"

Story: With Hierarchy-

* row
  * Missing attributes: all of the above and aria-expanded
  * Radio button SVG: same points as above
  * The “View children” button is in the tab order and in the a11y tree. It shouldn’t. (tabindex="-1" and aria-hidden="true")
* cell
  * Missing attribute: same as above
  * Arrow left doesn’t work to go to parent hierarchy
  * The first item of the child hierarchy should be focused automatically when drilling down (Tab -> Arrow down to “California” -> Arrow right -> “Los Angeles” should have focus)

PR #3054, opened 2023-05-30
fix: Add select aria updates

davidmenendez:
Contributes to #3043 

just addresses several aria related accessibility issues for add select. the only issue this doesn't address is certain keyboard navigation functionality, which i will address separately.


matthewgallo (changes requested):
Looks good, just need to add `posinset` and `setsize` to cspell ignores. There are some linting errors otherwise

davidmenendez:
@matthewgallo fixes made

matthewgallo (approved):

PR #3055, opened 2023-05-30
feat(SteppedAnimatedMedia): Novice to Pro internal component

jlongshore:
Contributes to #
Novice to Pro components

#### What did you change?
Added the SteppedAnimatedMedia component which will be used in the "Coachmark" and "Interstitial Screen" Novice to Pro components. 

Jest tests and Storybook on top of internal code and design reviews.


packages/ibm-products/src/components/SteppedAnimatedMedia/SteppedAnimatedMedia.js
```diff
@@ -0,0 +1,153 @@
...
+
+// Default values for props
+const defaults = {
+  playStep: 0,
+};
+
+/**
+ * DO NOT USE. This component is for the exclusive use
+ * of other Novice to Pro components.
+ */
```
elycheea, line 42: Pretty minor, but we changed this slightly in the Guidebanner and Carousel PR.

https://github.com/carbon-design-system/ibm-products/blob/main_v1/packages/ibm-products/src/components/Carousel/Carousel.js#L40-L42

packages/ibm-products/src/components/SteppedAnimatedMedia/SteppedAnimatedMedia.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+- [Overview](#overview)
+- [Example usage](#example-usage)
+- [Component API](#component-api)
+
+## Overview
+
+<!-- TODO: Overview. -->
+
+This component is for the internal use of other Novice to Pro components.
```
elycheea, line 20: If you want to keep consistent with what @paul-balchin-ibm had in the Carousel mdx file.

https://github.com/carbon-design-system/ibm-products/pull/3053/commits/df1bfa6d3a7e9a2eae2aa9c1ebc0d7234c05feb5#diff-a5b061b297a9f66494ded70a1b37fcc438507e7c897137e33201d18eb3d71a05R17-R18

packages/ibm-products/src/components/SteppedAnimatedMedia/SteppedAnimatedMedia.js
```diff
@@ -0,0 +1,153 @@
...
+          blockClass, // Apply the block class to the main HTML element
+          className, // Apply any supplied class names to the main HTML element.
+          // example: `${blockClass}__template-string-class-${kind}-n-${size}`,
+          {
+            // switched classes dependant on props or state
+            // example: [`${blockClass}__here-if-small`]: size === 'sm',
+          }
+        )}
+        ref={localRef}
+        role="main"
```
elycheea, line 120: Feel like we should be cautious about using `role="main"`[^1]. Was this for getting past the tests? If so, I’d recommend wrapping `SteppedAnimatedMedia` with a `<main>` tag for just the test since these are components and the landmark tests are a bit tricky otherwise.

[^1]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/main_role

packages/ibm-products/src/components/SteppedAnimatedMedia/SteppedAnimatedMedia.js
```diff
@@ -0,0 +1,153 @@
...
+      />
+    );
+  }
+);
+
+// Return a placeholder if not released and not enabled by feature flag
+SteppedAnimatedMedia = pkg.checkComponentEnabled(
+  SteppedAnimatedMedia,
+  componentName
+);
```
elycheea, line 131: Since this is internal and not meant to be used, we don’t need to flag this as Canary.

packages/ibm-products/src/components/SteppedAnimatedMedia/SteppedAnimatedMedia.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+<Canvas>
+  <Story
+    id={getStoryId(SteppedAnimatedMedia.displayName, 'stepped-animated-media')}
+  />
+</Canvas>
+
+## Code sample
+
+<!-- <CodesandboxLink exampleDirectory="SteppedAnimatedMedia" /> -->
```
elycheea, line 34: We can remove the code samples since this is only meant to be used internally.

packages/ibm-products/src/components/SteppedAnimatedMedia/SteppedAnimatedMedia.js
```diff
@@ -0,0 +1,153 @@
...
+// See https://www.npmjs.com/package/prop-types#usage.
+SteppedAnimatedMedia.propTypes = {
+  /**
+   * Optional class name for this component.
+   */
+  className: PropTypes.string,
+  /**
+   * The file path(s) to json formatted Lottie animation files.
+   */
+  filePaths: PropTypes.arrayOf(PropTypes.string).isRequired,
```
elycheea, line 148: In `EmptyStates`, I believe we use `illustration`, which doesn’t really convey that it’s a path/URL ... `UserProfileImage` uses `image`. We usually try to keep consistent, but since this is internal we can probably change it later without a breaking change. 🤔

elycheea (commented):
Just a few small changes and suggestions for consistency with what @paul-balchin-ibm and I landed on for `Carousel`.

**`role="main"` should be addressed though. 😬

packages/ibm-products/src/components/SteppedAnimatedMedia/SteppedAnimatedMedia.js
jlongshore: Ah no, I believe this is just how the components are scaffolded when they are generated and I left it as is. This might make more sense as a role="img" or "presentation"?

elycheea (approved):
🎉

elycheea:
@all-contributors add @jlongshore for code

PR #3061, opened 2023-05-31
chore: disable codeowners for Security on main

elycheea:
Stop spamming @amtoussam and @jlongshore from getting IBM Products notifications (for now) 😉.

#### What did you change?
Commented out Security package line but decided not to fully remove for some tiny bit of visibility. Also unsure if this will be like actions where removing from `main` will stop defaulting to in `main_v1` based PRs. 👀 

🤐


elycheea:
Checked back on some other PRs, no longer think this is necessary.

Issue #3065, opened 2023-06-01
Datagrid: Batch actions button menu has styling issues


matthewgallo:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

When using Datagrid with batch actions, the bath action toolbar button menu component has styling issues in v2.
> Is this issue related to a specific component?

Yes, Datagrid, more specifically the ButtonMenu in the batch action toolbar
> What did you expect to happen? What happened instead? What would you like to
> see changed?

I expected the styles to not have any issues, there are gaps in the background color because of the `.cds-btn` classes we manually add to one of the inner elements in `ButtonMenu`.
> What browser are you working in?

Chrome
> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

Latest v2
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

N/A
## Steps to reproduce the issue

1. Toggle batch action toolbar from one of the Datagrid stories that uses batch actions.
2. Hover on the "More +" button menu component, and you'll see the style issues

![image](https://github.com/carbon-design-system/ibm-products/assets/10215203/e303dc16-b6d8-41ab-93e4-e3c3ad889c69)

PR #3066, opened 2023-06-01
fix(Datagrid): address `ButtonMenu` style issues in batch action toolbar

matthewgallo:
Contributes to https://github.com/carbon-design-system/ibm-products/issues/3065

This PR fixes the style issues found with the ButtonMenu when used inside of the batch action toolbar of the Datagrid.

#### What did you change?
```
packages/ibm-products/src/components/ButtonMenu/ButtonMenu.js
packages/ibm-products/src/components/ButtonMenu/_button-menu.scss
packages/ibm-products/src/components/Datagrid/_datagrid.scss
packages/ibm-products/src/components/Datagrid/styles/_datagrid.scss
```
Storybook

***This is only an issue in v2

_Before_
image

_After_
image


packages/ibm-products/src/__tests__/__snapshots__/styles.test.js.snap
AlexanderMelox, line 1: Question, does this file need c4p replaced with a sass variable to support if the prefix changes in the future?

packages/ibm-products/src/components/Datagrid/_datagrid.scss
```diff
@@ -33,7 +34,8 @@ $block-class: #{c4p-settings.$pkg-prefix}--datagrid;
     display: none;
   }
 
-  .bx--overflow-menu-options > .c4p--datagrid__row-size-dropdown {
+  .#{c4p-settings.$carbon-prefix}--overflow-menu-options
+    > .c4p--datagrid__row-size-dropdown {
```
AlexanderMelox, line 38: replace c4p with package prefix

AlexanderMelox (changes requested):

packages/ibm-products/src/__tests__/__snapshots__/styles.test.js.snap
matthewgallo: No because this file contains compiled styles.

packages/ibm-products/src/components/Datagrid/_datagrid.scss
matthewgallo: Updated

packages/ibm-products/src/__tests__/__snapshots__/styles.test.js.snap
matthewgallo: Meaning if we changed the prefix in the future, we'll run `yarn test:c4p:snapshot` which will generate an updated snapshot again containing the new prefix (via this [test](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products/src/__tests__/styles.test.js)).

Issue #2975, opened 2023-05-09
*Enhancement Request* for NotificationsPanel Componenent to have option to remove 'Do not Disturb' toggle from panel 


bridgettepepper:
## What package(s) are you using?
**NotificationsPanel Component** - https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-patterns-notifications-notificationspanel--default&args=doNotDisturbDefaultToggled:false;doNotDisturbLabel 

Version - @carbon/ibm-products - 'latest' 

- [X ] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## What did you want to know?

First want to say how much I love the notifications panel component you all have created! I love that I did not have to create my own and use something already there

My enhancement request - would it be possible to have the 'doNotDisturb' toggle / text be optional to remove from the panel without having to alter the component code within the package? Not all applications (including mine) would require this feature out the gate, and being able to easily remove it would be a great feature. 

Example: removing the settings / 'view more' buttons from the component is easily done by commenting out the below lines within  <NotificationsPanel /> :
//onViewAllClick={}
//onSettingsClick={() => console.log('Clicked settings gear button')}

Unfortunately doing the same for 'Do not Disturb' event does nothing - //onDoNotDisturbChange={function noRefCheck(){}}

Thank you for considering this enhancement! Please let me know if any additional information is needed. Keep up the great work :D 

- Bridgette

elycheea:
@Ratheeshrajan Since I’ve already added you to a few notifications related issues. :)

Ratheeshrajan:
closing this as #3102

PR #3102, opened 2023-06-06
fix(NotificationPanel): option to remove dnd

Ratheeshrajan:
Contributes to #2975 

Option to remove 'Do not Disturb' label and toggle from the notification panel by removing `onDoNotDisturbChange`.

#### What did you change?
Checking if there is `onDoNotDisturbChange` func available or not to render 'Do not Disturb'
1. Storybook
2. Comment out `onDoNotDisturbChange: action('Toggled to do not disturb'),` from `NotificationsPanel.stories.js`


lee-chase (approved):

lee-chase (changes requested):
Can we mark the proptype as optional and ensure storybook is capable of displaying this state.

Ratheeshrajan:
> Can we mark the proptype as optional and ensure storybook is capable of displaying this state.

Introduced an optional prop `showDoNotDisturb` 
image
Hope this works.

lee-chase:
Sorry @Ratheeshrajan I think the preferred solution is to do as you did initially and comment the propTypes as follows and not add the additional property. 

```
  /**
   * Optional: function that returns the current selected value of the disable notification toggle
   */
  onDoNotDisturbChange: PropTypes.func,
```

You could also update the propTypes for the comments for any related properties `doNotDisturbLabel` and `doNotDisturbDefaultToggled` to state if/when they are required.

lee-chase (changes requested):
See comment.

Ratheeshrajan:
@lee-chase reverted back to the initial solution.

Issue #2960, opened 2023-05-04
Incorrect color in notification bell when on light theme


lee-chase:
The Carbon UIShell is themeable in V11 so this now looks incorrect.

## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

See the bell, it's not very visible.

![image](https://user-images.githubusercontent.com/15086604/236270082-41153b64-5386-4b95-b5e0-c0711e2aa9f6.png)

lee-chase:
@elycheea should be simple to fix

Ratheeshrajan:
@elycheea, yes, I believe still there are some hardcoded fill in header icons and SVG.

Ratheeshrajan:
Closing as #3103 merged.

PR #3103, opened 2023-06-06
fix(Notification): incorrect icon color light theme

Ratheeshrajan:
Contributes to #2960 
The Carbon UIShell is themeable in V11. So the hardcoded filled icon looks incorrect color in light themes.

#### What did you change?
Hardcoded `fill` removed from <Icon> and SVG. So this will accept the theme-based var `--cds-icon-secondary`.

Storybook


matthewgallo (approved):
👏

PR #3104, opened 2023-06-06
fix(Datagrid): prevent undefined value for isTableSortable prop

elahemirzaee:
Contributes to [ibm-products](https://github.com/carbon-design-system/ibm-products)

when using tearsheet isTableSortable prop is undefined and since it is a required prop it causes error

#### What did you change?
i added default value false for prop

checked it in storybook and my production version of my project


matthewgallo (approved):
Thank you for the contribution @elahemirzaee!

Issue #3107, opened 2023-06-07
import modal state persists after closing


davidmenendez:
this bug was reported in slack after a user noted that they were unable to clear the error state from the modal after closing it.

should be a quick fix.

PR #3108, opened 2023-06-07
fix: reset import modal state on close

davidmenendez:
Contributes to #3107


lee-chase (approved):
Approving but....

Should the handlers not be called like...

```
if (onClose) {
  onClose();
}
```

lee-chase (approved):

Issue #2725, opened 2023-03-09
Datagrid rerenders in specific scenarios in v2


Ratheeshrajan:
some times need to click twice on the button to open the filter panel/flyout. This is noticed in v11.

## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
Carbon for IBM Products v2.0.0-rc.20

## Steps to reproduce the issue

1. Click on the filter flyout/panel button

## Additional information

https://user-images.githubusercontent.com/305492/224034547-f4d8cbea-d7c7-46e7-9c96-adee2e163c13.mov

elycheea:
@AlexanderMelox @Ratheeshrajan Can you help estimate the size for this. Seems like the re-render issue.

elycheea:
Editable cells, actions dropdown, etc are some areas @matthewgallo’s encountered.

PR #3109, opened 2023-06-07
chore(Datagrid): address rerender issues with filter flyout button

matthewgallo:
Contributes to #2725 

I was able to narrow down the re-rendering issues with the `FilterFlyout` button, which caused the trigger to have to be clicked twice before opening the flyout. From what I can tell so far, it stems from the `renderIcon` prop from Carbon's `Button` component. I simply changed from using `Button` with the `renderIcon` prop to `IconButton` and the issue is gone.

I'm going to continue doing some investigating, because things like the ButtonMenu still have this problem.

I also made a small change to the `tabIndex` of the containing element for the Datagrid to remove a console error.

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/FilterFlyout.js
packages/ibm-products/src/components/Datagrid/styles/addons/_FilterFlyout.scss
```
Storybook

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridContent.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridContent.js
@@ -84,7 +84,7 @@ export const DatagridContent = ({ datagridState }) => {
           getTableProps()?.className
         )}
         role={withInlineEdit && 'grid'}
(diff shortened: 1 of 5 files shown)


AlexanderMelox (approved):
Great work on this @matthewgallo 🎉

Issue #2603, opened 2023-01-08
Get click callback in expandable rows of Datagrid?


houhoucoop:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## What did you want to know?
According to the [README](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/1ed3a18d22e3368d11d26270681e815b560cba87/packages/cloud-cognitive/src/components/Datagrid/Datagrid.mdx#expandable-rows), the `Datagrid` supports expandable rows with the use of the `useExpandedRow` hook and able to custom the expanded content with `ExpandedRowContentComponent`, but it seems there is no way to get the click callback function or pass a custom click function when toggling the row?

jagathgj:
Hi @elycheea , There is a possibility to get the click callback function or pass a custom click function without altering the codebase. Below is a hook function to do the same.

const useExpandListener = (hooks) => {
  const getProps = (props, { row }) => {
    return { ...props, onClick: () => { console.log('Row Toggled', row.id); props.onClick(); } }
  }
  hooks.getToggleRowExpandedProps.push(getProps);
};


Consumer can either use this hook or I can add it to codebase. Whats your thoughts on this?

AlexanderMelox:
I do like this idea @jagathgj but I think it's prob easier just having the "handler" callbacks inside the components to avoid confusion. I do understand it tidy's up the codebase but in certain examples like FilterPanel, I have a lot more than just a click, I have an onChange handler on every input and an open and close callback.

ConradSchmidt:
@AlexanderMelox adding a prop would (at some point) defeat the point of the hook system in the datagrid, because if you hard coding these features you remove the flexibility of the hook system.

You could integrate the `useExpandListener` into the repo, even though I'm not convinced that having this feature available will lead to good UX - triggering something else when expanding seems "disconnected" (it's not like clicking on a row).

If you still want to go the route of using a prop, be aware how this would work:
1. one hook which is applied to all rows
2. each row has its own hook

AlexanderMelox:
@ConradSchmidt Sorry when I meant prop I meant the properties that we pass into useDatagrid()

AlexanderMelox:
I would not use the hooks for developer things, where as the hooks in our situation are "plugins" that add extra functionality to the table. 

We've been using this approach before like below 
[...]

PR #3114, opened 2023-06-09
fix(Datagrid): expand listener added

jagathgj:
Contributes to #2603 

Event listener function added to expandable rows.

#### What did you change?
ExpandableRow.stories.js
Datagrid/index.js
useRowExpandListener.js

storybook


davidmenendez (approved):
code wise this looks good to me. not sure if anyone who's more familiar with datagrid wants to chime in though.

packages/ibm-products/src/components/Datagrid/useRowExpandListener.js
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+const useRowExpandListener = (hooks) => {
+  const getProps = (props, { row }) => {
+    return {
+      ...props,
+      onClick: () => {
+        if (row.original.onRowExpand) {
+          row.original.onRowExpand(row);
+        }
```
AlexanderMelox, line 17:
```suggestion
          row?.original?.onRowExpand?.(row);
```

I would make this less bloated

AlexanderMelox:
@jagathgj I think it should be fine to just add it to `useExpandedRow`, the onExpandRow can live in there. I don't see a need to make another hook just to listen to the row click 🤔 Any thoughts on this @matthewgallo

matthewgallo (changes requested):
Agree with @AlexanderMelox, I don't think this necessarily is something that needs it's own hook. There are other extensions which add a listener without the need for a completely new hook.

elycheea:
@jagathgj Sorry for the delay on this one — seems like we keep missing it, but looks like we have a conflict with one of the recent PRs now but once we resolve, @matthewgallo can help re-review this.

matthewgallo (changes requested):
To follow the pattern we've set with other event handlers like this, I think it makes most sense to have `onRowExpand` be passed directly to the `useDatagrid` hook and it returns the row that was expanded. As opposed to this approach which adds it directly on the row within the data.

matthewgallo (approved):

Issue #2232, opened 2022-08-25
Add accessibility review


elycheea:

elycheea:
Pulled from the [Equal Access Toolkit](https://www.ibm.com/able/toolkit/develop/overview)’s develop section.

I do think we need to tune this a bit more for components as well as for our process. I do wonder if it’d make more sense to include Level 1 accessibility in earlier points of the contribution process, although it may be hard to keep track of as our team builds out each component further. 🤔 

Would love to get the design steps added in the PAL repos as well.

---

## Accessibility review

### [Keyboard](https://www.ibm.com/able/toolkit/develop/keyboard)

#### Level 1
- [ ] Properly code the navigation order
  - [ ] Ensure interactive elements are reachable with the keyboard
  - [ ] Where possible, achieve the desired tab order by adjusting the DOM, rather than overriding the tabindex
  - [ ] When overriding the default focus indicator, confirm focus is highly visible
- [ ] Implement established keyboard conventions
  - [ ] Use standard HTML elements where possible, using CSS to alter appearance not behavior
  - [ ] Be familiar with established keyboard conventions
  - [ ] Implement keyboard operation for custom elements

#### Level 2
- [ ] Ensure keyboard access to complex interactions
  - [ ] Make custom media player and animation controls keyboard accessible

#### Level 3
- [ ] Code screen updates predictably
  - [ ] Use “activate” rather than “focus” as a trigger
  - [ ] Use an onchange event without disorienting the user

### [Dynamic updates](https://www.ibm.com/able/toolkit/develop/dynamic-updates)

#### Level 1
- [ ] Provide errors accessibly
  - [ ] Use accessible methods to implement error messages
  - [ ] Implement validation and notification to minimize interruptions

#### Level 2
- [ ] Make content accessible that appears on hover or focus
  - [ ] Ensure <kbd>Esc</kbd> dismisses new content
  - [ ] If content appears on hover, the new content needs to remain visible until dismissed
  - [ ] Moving the pointer away from the trigger should not be the action that dismisses the new content
  - [ ] Ensure custom tooltips and similar hover text can be triggered by keyboard
  - [ ] Discuss alternatives to hover text
  - [ ] Use caution where interactive controls appear in the hover text
- [ ] Code cancellation actions accessibly
  - [ ] Provide accessible ways of cancelling pointer interaction
- [ ] Properly code status messages
  - [ ] Use `role=status` to advise on the success or results of an action
[...]

PR #3126, opened 2023-06-13
docs: update release review template

### LINKED ISSUE #2235: Update release review guidelines and template
Discussing with @matthewgallo, we found some areas where we could clarify or elaborate further in our release review process.

> The UI produced is accessible, responsive, translatable, cross-browser, and responds to the currently set Carbon theme.

This step requires several important checks but wraps them into a single bullet. These should be broken out to capture and set clearer expectations. e.g. _how do we check for translatability_? Accessibility may be broken into a separate review process as well.

> All significant DOM elements have meaningful classes.

Should add mention of following BEM as well.

elycheea:
Closes #2235 and #2232

#### What did you change?
- Add step to check for completed accessibility review
- Break `The UI produced is accessible, responsive, translatable, cross-browser, and responds to the currently set Carbon theme` into multiple steps
- Remove redundant template from release review guidelines

Markdown preview 💁‍♀️


matthewgallo (approved):
🎉

PR #3139, opened 2023-06-15
fix: scope `run` block to step when committing lock file

matthewgallo:
Both release actions seem to be attempting to run after every commit. The screenshot below shows it with Lee's storybook docs PR, but I also noticed it with https://github.com/carbon-design-system/ibm-products/pull/3136 as well.

Screenshot 2023-06-15 at 9 54 45 AM

Not sure if it is happening because of the indentation of the `run` block that commits the lock file, github actions vscode plugin seems to want it indented, as opposed to directly below the step (ie `- name`).

image

#### What did you change?
`release.yml` and `release-v1.yml`
😱 after merging this

### DIFF
--- a/.github/workflows/release-v1.yml
+++ b/.github/workflows/release-v1.yml
@@ -72,7 +72,7 @@ jobs:
         run: yarn
 
       - name: Commit lock file
-      - run: |
+        run: |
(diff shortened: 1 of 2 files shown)


elycheea (approved):
Oops. I probably missed that while I was editing. Thank you for catching that and fixing it!

PR #3142, opened 2023-06-18
fix(Datagrid): Invalid prop value was passed through table toolbar search component

elahemirzaee:
Contributes to [ibm-products](https://github.com/carbon-design-system/ibm-products)

Invalid prop value was passed through table toolbar search component for table

#### What did you change?
i changed the  prop 'size' value to "lg" instead of "xl" because 'xl' is not one of the pre-defined values of 'size' prop and is not acceptable based on [ibm product storybook](https://carbon-for-ibm-products.netlify.app/)
checked it in storybook and my production version of my project


matthewgallo (approved):
Thanks for contributing @elahemirzaee!

Issue #2895, opened 2023-04-19
React 18 support


elycheea:
## Additional information

- Designs
- Existing code
- etc

```[tasklist]
### Tasks
- [ ] Look into how Carbon manages their React support
- [ ] Run some tests on React 16 and 17 to see if we can still use it
```

elycheea:
@lee-chase to add more detail after v2 release.

elycheea:
@lee-chase: To be determined, whether we still need to support React 16? Set in peer dependencies ... which would probably make it a breaking change.

PR #3146, opened 2023-06-21
feat: upgrade main package to react 18

lee-chase:
Contributes to #2895 

Upgrade of main package to React 18 to match Carbon V11.

#### What did you change?
- Upgraded main packages to React 18
- Upgraded React-dnd to latest version
- Updated package settings to ensure 'displayName' is still set as the property is no longer enumerable on a render function.
- Added React 16 and React 17 samples.
- Updated gallery config

Ran storybook and verified React 16 and 17 examples work with React 18 compiled IBM Products


lee-chase:
- Carbon 18 upgrade… This is proving to be a bit more involved than just updating Storybook. As part of the upgrade we moved to a new version of @testing-library/react which has a number of changes. 

- A bunch of stuff seems to be async which means if you are waiting for a non-default state or user event then you need to await the result.
  - act() appears to be needed for some events and re-renders, possibly only ones that cause focus, and  may be related to https://github.com/testing-library/user-event/issues/1114 as I would expect user-event to do this for us.
  - Canary seems to get listed in the errors when the cause is act()
- I have also removed enzyme.
- I have added a react-16 and react-17 example, but cannot update the other examples to react-18 until we release a react-18 version. There appears to be no issue using 16 and 17 with code compiled against 18. Perhaps a cautions RC release would be wise

```
Expected test not to call console.error()
@carbon/ibm-products:     Warning: An update to Tooltip inside a test was not wrapped in act(...).
@carbon/ibm-products:     When testing, code that causes React state updates should be wrapped into act(...):
@carbon/ibm-products:     act(() => {
@carbon/ibm-products:       /* fire events that update state */
@carbon/ibm-products:     });
@carbon/ibm-products:     /* assert on the output */
@carbon/ibm-products:     This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
@carbon/ibm-products: 
@carbon/ibm-products:       46 |           : // Note that Canary is a direct render fn (not a forward-ref) and
@carbon/ibm-products:       47 |             // will ignore the passed props and ref (if any)
@carbon/ibm-products:     > 48 |             Canary.bind(undefined, { componentName: name }))(
@carbon/ibm-products:          |                                                            ^
@carbon/ibm-products:       49 |         // Call it now (after this it will be directly called).
@carbon/ibm-products:       50 |         props,
@carbon/ibm-products:       51 |         ref
.
.
. // THIS APPEARS TO BE THE REAL ISSUE
@carbon/ibm-products:       Warning: An update to Toolbar inside a test was not wrapped in act(...).
@carbon/ibm-products:       When testing, code that causes React state updates should be wrapped into act(...):
@carbon/ibm-products:       act(() => {
@carbon/ibm-products:         /* fire events that update state */
@carbon/ibm-products:       });
@carbon/ibm-products:       /* assert on the output */
```

.github/CONTRIBUTING.md
```diff
@@ -83,10 +83,9 @@ For more information on how this affects components see
 If you’re contributing to our JavaScript code, test your changes by running
 `yarn test`.
 
-New tests are written in
-[React Testing Library](https://testing-library.com/docs/react-testing-library/intro),
-with [Enzyme](https://enzymejs.github.io/enzyme) available for migration
-compatibility, and follow the
+Tests are written in
```
lee-chase, line 86: Enzyme has now been removed

config/jest-config-ibm-cloud-cognitive/index.js
```diff
@@ -74,9 +73,10 @@ module.exports = {
     'e2e',
     'templates',
     '/umd/',
+    '/react-dnd/',
   ],
   transformIgnorePatterns: [
-    '[/\\\\]node_modules[/\\\\].+\\.(js|jsx)$',
+    `/node_modules/(?!react-dnd|dnd-core|@react-dnd)`,
```
lee-chase, line 79: React dnd was failing babel transform when running tests

examples/carbon-for-ibm-products/example-gallery/src/gallery-config/index.js
```diff
@@ -37,6 +37,8 @@ import imageTearsheet from './Tearsheet--thumbnail.png';
 import imageUserProfileImage from './UserProfileImage--thumbnail.png';
 import imageWebTerminal from './WebTerminal--thumbnail.png';
 import imagePrefixExample from './prefix-example--thumbnail.png';
+import imageReact16Example from './react-16-example--thumbnail.png';
```
lee-chase, line 40: Should raise an issue to convert other examples to React 18

packages/core/.storybook/preview.js
```diff
@@ -157,4 +157,4 @@ const globals = {
   [CARBON_THEME_PARAM_KEY]: CARBON_THEMES.g10,
 };
 
-export { argTypes, decorators, globals, parameters, Style };
+export default { argTypes, decorators, globals, parameters, Style };
```
lee-chase, line 160: Could revert if https://github.com/storybookjs/storybook/issues/23604 is fixed

packages/ibm-products/src/__tests__/settings.test.js
```diff
@@ -23,28 +22,31 @@ describe(name, () => {
 
     // dynamic import so we can modify the import on the component before using it
     const { StatusIcon } = await import('../components/StatusIcon');
-    const wrapper = shallow(
+    const { container } = render(
```
lee-chase, line 25: Reworked test to remove enzyme

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.js
```diff
@@ -614,7 +614,7 @@ export let DataSpreadsheet = React.forwardRef(
       const activeCellValue = activeCellFullData
         ? activeCellFullData.row.cells[activeCellCoordinates?.column].value
         : null;
-      setCellEditorValue(activeCellValue);
+      setCellEditorValue(activeCellValue || '');
```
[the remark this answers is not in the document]
lee-chase: The initial value of this is ''

packages/ibm-products/src/components/Datagrid/Datagrid.test.js
```diff
@@ -1058,10 +1062,9 @@ describe(componentName, () => {
         const { container } = render(
           <BasicUsage data-testid={dataTestId} datagridState={null} />
         );
-        expect(errorMock).toBeCalledWith(
-          expect.stringMatching(
-            /^Warning: Failed prop type: The prop `datagridState` is marked as required in `Datagrid`, but its value is `null`./
-          )
+        checkLogging(
```
lee-chase, line 1065: use the common code for the log check

packages/ibm-products/src/settings.js
```diff
@@ -55,7 +55,13 @@ pkgSettings.checkComponentEnabled = (component, name) => {
     // Transfer object properties already assigned (eg propTypes, displayName)
     // then merge in the stub forward-ref which checks the component status
     // when first used.
-    return Object.assign({}, component, forward);
+    // NOTE: React 18 = displayName not iterable on render function
```
lee-chase, line 58: Display name is no longer iterable so need copying manually

packages/ibm-products/src/global/js/utils/test-helper.js
```diff
@@ -173,21 +173,24 @@ export const expectWarnAsync = async (message, test) => {
...
+export const expectMultipleWarn = async (messages, test) => {
   const warn = jest.spyOn(console, 'warn').mockImplementation(jest.fn());
-  const result = test();
+  const result = await test();
+
   expect(warn).toBeCalledTimes(messages.length);
-  messages.forEach((args, index) =>
-    expect(warn).toHaveBeenNthCalledWith(index + 1, ...makeMatcherArray(args))
-  );
+  // TODO: React 18 update - console messages appear to be failing with calls that look like printf props
```
lee-chase, line 181: Log checking turned off and should  be raised as an issue.

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.js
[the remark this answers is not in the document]
matthewgallo: I don't see a problem with this since the initial value is also an empty string

examples/carbon-for-ibm-products/example-gallery/src/gallery-config/index.js
elycheea: https://github.com/carbon-design-system/ibm-products/issues/3363

examples/carbon-for-ibm-products/example-gallery/src/gallery-config/index.js
```diff
@@ -171,6 +173,16 @@ const config = [
     directory: 'prefix-example',
     thumbnail: `url(${imagePrefixExample}`,
   },
+  {
+    label: 'About Modal',
```
matthewgallo, line 177: Update label to indicate React 16

packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.test.js
```diff
@@ -149,28 +151,31 @@ describe('Notifications', () => {
...
       ],
     });
-    const logLink = screen.getByText(/view logs/i);
-    userEvent.click(logLink);
-    expect(logLink);
+    const logLink = screen.getByRole('link');
+    expect(logLink).toHaveTextContent(link.text);
+    expect(logLink).toHaveAttribute('href', link.url);
+    // REACT 18: We were testing clicking of a link, which caused an error to be logged
+    // Error: Not implemented: navigation (except hash changes)
```
elycheea, line 175: Agree with the change, but don’t think we need to keep this comment in here lest it lives on forever.

packages/ibm-products/src/global/js/utils/test-helper.js
elycheea: https://github.com/carbon-design-system/ibm-products/issues/3369

elycheea (commented):
Thanks @lee-chase for taking the lead on this huge effort and for walking us through the changes.

I opened issues anywhere we noted some changes that should be addressed later or investigated further. Had a few small comments on one of the tests we rewrote from enzyme — we actually do this in a few places but would prefer not to introduce it newly if we can help it. If it takes more than a few minutes to address, then let’s open another bug to investigate further.

elycheea:
Let’s merge **after** next Tuesday’s release once we resolve the conflicts.

lee-chase:
An rc is a nice idea, we can just flip the tags on a release and then check it out over a day or two.

P.S. The only tests nerfed now are the console log/error. All other tests restored.

elycheea (approved):

Issue #2974, opened 2023-05-09
[Datagrid, FilterSummary] Filter summary doesn't use filter labels


joyful-these:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

The tags in the filter summary doesn't use the filter labels. 

> Is this issue related to a specific component?

FilterSummary

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The tag bubbles in the filter summary bar should use the labels of the filters, instead of key:value. This is to make them more consistent with the rest of the UI, and easier to read.

For example: A filter called 'Password strength', with the option 'Minor warning' selected.
Current tag appearance in filter summary:
<img width="174" alt="Screenshot 2023-05-09 at 12 15 52 PM" src="https://github.com/carbon-design-system/ibm-cloud-cognitive/assets/108295261/08ea0a90-ad73-45ea-add3-e80e7af67e8d">
Desired tag appearance:
<img width="174" alt="Screenshot 2023-05-09 at 12 40 40 PM" src="https://github.com/carbon-design-system/ibm-cloud-cognitive/assets/108295261/9e298671-2bf1-44cf-9c77-c59afb6e5fe8">

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

1.51.0

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

CP4WAIOps

elycheea:
@AlexanderMelox Were you already looking into this? Not sure if this is a duplicate because I remember us discussing this one at some point.

PR #3171, opened 2023-06-29
feat(FilterSummary): adds custom label render (v1)

AlexanderMelox:
Contributes to #2974 

This PR allows the user to render the label as they please, they can supply their own Title case, or sentence case or do whatever transformation of string they want.


packages/ibm-products/src/components/FilterSummary/FilterSummary.js
```diff
@@ -20,12 +20,13 @@ let FilterSummary = React.forwardRef(
       clearFiltersText = 'Clear filters',
       clearFilters = () => {},
       filters = [],
+      renderLabel = null,
```
paul-balchin-ibm, line 23: Is it necessary to declare it as `null`? If `null` is removed, `renderLabel` will default to `undefined`... same difference.

Or is it a style/standard practice?

AlexanderMelox: In general, I like setting a default so I know when I check in the component below I can ensure it's null or undefined. In reality I did this because `renderLabel` is a function but for the check I need it to be null or undefined.

paul-balchin-ibm (approved):

Issue #3177, opened 2023-07-03
Accessibility tests failing in `main`


elycheea:
## What will this achieve?

Fix tests in CI.

## How will success be measured?

Tests should pass.

## Additional information

- Designs
- Existing code
- etc

https://github.com/carbon-design-system/ibm-products/actions/runs/5446628501/jobs/9908217875?pr=3176

lee-chase:
@tay1orjones is this something Carbon has looked at before?

elycheea:
Should double check if we need to address this for v1 as well before closing.

tay1orjones:
I think y'all have seen by now, but we ran into the same. A new ruleset was published for accessibility-checker on July 1. If you're configured to use `latest`, it'll automatically pull the new ruleset potentially causing failures.

PR #3179, opened 2023-07-04
fix: a11y issues with some forms 

lee-chase:
Contributes to #3177

Some forms are report a11y issues requiring name.

#### What did you change?

- Added `formName` attribute to components with a11y issues.
- Updated stories and tests.

Ran tests to confirm a11y issues not reported


lee-chase:
NOTE: needs a v1 port

elycheea (approved):

Issue #1154, opened 2021-08-13
Setting a custom prefix does not work


matthewgallo:
## What package(s) are you using?

- [X] Cloud & Cognitive

## Detailed description

> Describe in detail the issue you're having.

In code sandbox, we should be able to set a custom prefix in the config.js file. The prefix does get added to all of the class names, however, none of the styles get added when the custom prefix is set.
> Is this issue related to a specific component?

Configuration
> What did you expect to happen? What happened instead? What would you like to
> see changed?

I expected all of the styles to still work when using a custom prefix
> What browser are you working in?

Chrome
> What version of the @carbon/ibm-cloud-cognitive packge are you using?

Latest
## Steps to reproduce the issue

1. Create a code sandbox
2. In the `config.js` file, add this to set the custom prefix `pkg.prefix = "my-custom-prefix";`
3. Now you should be able to see that the styles are broken/do not get applied in the way you would expect

lee-chase:
@matthewgallo did you change the CSS prefix and load the SCSS styles.

https://codesandbox.io/s/examplecomponent-prefix-test-3ss0i?file=/src/_ibm-cloud-and-cognitive.scss

dcwarwick:
setting prefix in code and scss does now work, and we changed the prefix as part of the v1 release.

PR #3183, opened 2023-07-07
chore: always use prefix in dev storybook

lee-chase:
Contributes to #1154 

Propose changing all development storybook stories to use a custom prefix.

#### What did you change?

Core storybook preview and styling to set a prefix when in development mode.

- yarn storybook
- Viewed in storybook and inspected the DOM and noted prefix
- yarn build
- npx serve packages/core/storybook-static and checked the prefix in the DOM

### DIFF
--- a/packages/core/.storybook/preview.js
+++ b/packages/core/.storybook/preview.js
@@ -19,6 +19,7 @@ import React, { useEffect } from 'react';
 import { pkg } from '../../ibm-products/src/settings';
 
 import index from './index.scss';
+import prefixedIndex from './_prefix-index.scss';
 import { getSectionSequence } from '../story-structure';
 import { StoryDocsPage } from '../../ibm-products/src/global/js/utils/StoryDocsPage';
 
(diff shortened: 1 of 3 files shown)


packages/core/.storybook/preview.js
```diff
@@ -38,6 +39,12 @@ const Style = ({ children, styles }) => {
   return children;
 };
 
+let isDev = CONFIG_TYPE === 'DEVELOPMENT'; //  process.env?.NODE_ENV === 'development';
```
matthewgallo, line 42: Don't think you need to use `let` here. Also, where is `CONFIG_TYPE` coming from? 🤔

matthewgallo: We have a `isProduction` variable that could possibly be used from `packages/ibm-products/src/global/js/utils/pconsole.js`.

lee-chase: As far as I can tell Storybook does not pass the NODE_ENV parameter

lee-chase: CONFIG_TYPE is added by storybook I believe

https://github.com/search?q=repo%3Astorybookjs%2Fstorybook%20config_type&type=code

matthewgallo (approved):

Issue #3180, opened 2023-07-06
Datagrid: add docs page content back


matthewgallo:
## What will this achieve?

After the move to Storybook 7, we lost the docs content for the Datagrid. Using the new doc page structure that @lee-chase has put in place, we should be able to add all of it back. Until that work is complete, we can point people to the Carbon for IBM Products v1 storybook which still has all the docs.

https://v1-ibm-products.carbondesignsystem.com/?path=/docs/ibm-products-components-datagrid-datagrid-canary--basic-usage

## Additional information

- Designs
- Existing code
- etc

PR #3187, opened 2023-07-10
chore: add datagrid docs back to storybook

matthewgallo:
Contributes to #3180

Adds Datagrid docs back to storybook using the new custom doc component created by @lee-chase. For any extension that has it's own specific docs, they are now located in it's own doc page since we are utilizing `autodocs` from storybook.

#### What did you change?
Updated doc files for Datagrid stories.
Storybook


davidmenendez (approved):
awesome! thanks for getting that back in order

PR #3198, opened 2023-07-11
fix: update to Carbon 11 compatible versions to latest

davidmenendez:
ran `yarn upgrade:carbon` and fixed the necessary tests


elycheea (approved):
Thank you, @davidmenendez 🙏

PR #3203, opened 2023-07-12
chore(Datagrid): add skeleton story (v2)

matthewgallo:
Contributes to #3120 

Adds a loading/skeleton story for Datagrid as requested during design review.

#### What did you change?
```
packages/core/story-structure.js
packages/ibm-products/src/components/Datagrid/Extensions/Skeleton/Skeleton.stories.js
packages/ibm-products/src/components/Datagrid/utils/getArgTypes.js
```
Storybook

### DIFF
--- a/packages/core/story-structure.js
+++ b/packages/core/story-structure.js
@@ -55,6 +55,7 @@ const s = [
               'c/Datagrid/Extensions/ClickableRow',
               'c/Datagrid/Extensions/EditableCell',
               'c/Datagrid/Extensions/ColumnCustomization',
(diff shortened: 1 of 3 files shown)


elycheea (approved):
I think `Skeleton` probably belongs under the base component rather than as an extension, but I think we can save some of the reorganization for later since we had a [few ideas in the Mural](https://app.mural.co/t/dsagmembers0371/m/dsagmembers0371/1682445571192/c1700d659af7daf306e6355813464f5bd7727236?wid=0-1682473487094).

Issue #3208, opened 2023-07-12
Storybook specific styles don’t reference dev prefix in local environment


elycheea:
## What will this achieve?

With the introduction of dev-prefix on local, we need to figure out a way to update Storybook prefix as well.

Noticed this while working on `ClickableRow` stories.

Storybook | Local
-----------|------
![image.png](https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/3737c36a-77bf-43d5-9c2f-e19765a72542) | ![image.png](https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/ef47fa3d-abf1-4229-8500-3f883610c70f)

Our Storybook works fine, but locally, the prefix still references our `$pkg-prefix` instead.

![image.png](https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/b0328ac3-b058-495d-9786-f0437f330e8c)

## How will success be measured?

`storybook-styles.scss` work on local ... but don’t break in Storybook. :sweat_smile:

## Additional information

- Designs
- Existing code
- etc

elycheea:
Moved to on hold until we can discuss possible solutions further with the rest of the team.

PR #3223, opened 2023-07-13
fix: possible solution for storybook styles custom prefix

matthewgallo:
Contributes to #3208 

Opening to start discussion about a possible solution to fixing `_storybook-styles.scss` from not picking up the custom prefix that we now set in development mode.

Downside to this approach is that every component that has a `_storybook-styles.scss` will need to adopt this approach.

One other option is not using the custom prefix in development at all and simply add an example to the example gallery.

#### What did you change?
Exported `isDev` from `preview.js` to conditionally use a set of storybook styles with the custom prefix.
Storybook

### DIFF
--- a/packages/core/.storybook/preview.js
+++ b/packages/core/.storybook/preview.js
@@ -39,7 +39,7 @@ const Style = ({ children, styles }) => {
   return children;
 };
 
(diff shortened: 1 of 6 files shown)


elycheea:
Closing in favor of #3400

Issue #3226, opened 2023-07-13
Improve how batch action works with other plugins


elycheea:
Batch actions is a commonly used plugin but when used in combination with certain other plugins/hooks, we see some unexpected or undesired behavior.

Currently, when adding [batch actions to clickable rows](https://pages.github.ibm.com/cdai-design/pal/components/data-table/clickable-row-items/usage/#interactions), the active state (blue bar) is not visible. Based on the design guidance, in this context, only selecting the checkbox should select the row, while clicking on the whole row should change the active row but not necessarily affect selection. 
This could be confusing so probably best to create an example and review with our designers. This might need to be a separate issue depending how easily we’re able to get this working.

```[tasklist]
### Tasks
- [ ] https://github.com/carbon-design-system/ibm-products/issues/3334
- [x] Add examples for clickable rows with batch actions
```

Issue #2607, opened 2023-01-11
Datagrid design review: Batch actions


elycheea:
## Design review

- Component epic #1607 #2499
- Designers/reviewers: 
- Links: [PAL guidance](https://pages.github.ibm.com/cdai-design/pal/components/data-table/batchactions/usage) / [Storybook](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary--batch-actions) / [Storybook](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary--select-items-in-all-pages)

---

### Standards

- [ ] All pattern updates/changes/iterations have been discussed with the
      component developer
- [ ] Patterns have been approved by either DSAG or another approving entity

### Pattern and behavior

- [ ] The component behaves according to the guidelines set by the pattern
      maintainer
- [ ] The component’s UI matches the pattern specifications set by the pattern
      maintainer
- [ ] The component’s motion matches the specifications set by the pattern
      maintainer(s)
- [ ] The UI produced is responsive, cross-browser, and responds to the
      currently set Carbon theme.
- [ ] Colors, themes, sizes and additional props are true and correct, aligning
      with Carbon set tokens (unless otherwise specified by the pattern)
- [ ] Paddings/margins/spacings are true and correct, in both desktop and mobile
      views

### Storybook

- [ ] A functioning component renders in Storybook
- [ ] The Storybook displays multiple scenarios that show how the component can
      be used
- [ ] Changing props in the Storybook updates the component

elycheea:
@vsnichols 💪

elycheea:
@kristastarr to review for accessibility.
Feel free to remove any checks that don’t apply.

---

## Accessibility review

### [Keyboard](https://www.ibm.com/able/toolkit/develop/keyboard)

#### Level 1

- [ ] Properly code the navigation order
  - [ ] Ensure interactive elements are reachable with the keyboard
  - [ ] Where possible, achieve the desired tab order by adjusting the DOM,
        rather than overriding the tabindex
  - [ ] When overriding the default focus indicator, confirm focus is highly
        visible
- [ ] Implement established keyboard conventions
  - [ ] Use standard HTML elements where possible, using CSS to alter appearance
        not behavior
  - [ ] Be familiar with established keyboard conventions
  - [ ] Implement keyboard operation for custom elements

#### Level 2

[...]

PR #3228, opened 2023-07-14
fix(Datagrid): add batch actions to clickable row with panel story

matthewgallo:
Contributes to #3226 and #2607

Adds batch actions to the clickable row with panel story to provide a more detailing use case. I also fix the alignment of the batch action buttons in this PR as well.

#### What did you change?
- Added batch actions to clickable with panel story
- Removed styling causing batch action layout issues

Storybook


elycheea (changes requested):
Looks great, but caught one issue where the selected state is still missing.

I’m not seeing `.cds--data-table--selected` on the table row now. Not sure if that’s coming from `DatagridRow` or if it needs to be added in `useOnRowClick` too. At quick glance, it does look like `useSelectRows` also attempts to address this (or a similar issue) so hopefully a simple enough fix from there. 😓

matthewgallo:
Finally got around to adding the selected indicator back @elycheea 😅

elycheea (approved):
🎉  looks good!
Think we also need a v1 PR for this one though!

PR #3236, opened 2023-07-17
feat(InterstitialScreen): Novice to Pro component

jlongshore:
Contributes to:
Novice to Pro components

Added the following components and stories:
InterstitialScreen
InterstitialScreenView

Storybook in addition to internal code and design reviews.


elycheea:
@jlongshore Worried we might have had a merge from `main` instead of `main_v1` here — hence the conflicts. 👀  Not sure if you want to save your changes in n2p and open a new PR / branch. (Don’t delete the branch if you wanna do that as @paul-balchin-ibm learned in the past 😹)

packages/ibm-products/src/components/Carousel/Carousel.js
jlongshore, line 1: This was an issue we found with allowing the scroll event to bubble up beyond the carousel & its items when the carousel's height is causing scrollbar on a parent element. If the cursor was over the cursor the vertical scroll wouldn't work

jlongshore: I call this out because I know its not directly an interstitial screen item... just used internally. Happy to break it out if need be.

packages/ibm-products/src/components/InterstitialScreen/_interstitial-screen.scss
elycheea, line 1: Some differing opinions on this one, but I _usually_ prefer that we type out the classnames since it makes it a bit quicker to locate when bugs arise. We have a few places we do this though so it’s not really a requested change just a thought/discussion.

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreen.mdx
elycheea, line 1: Probably worth doing some brief introduction on `InterstitialScreen` and its subcomponents here since I had to ask you about it directly, haha.

elycheea (changes requested):
Mostly have some structural questions and suggestions and a handful of small catches.
Stories can probably be cleaned up to a few variants (can reduce some overhead). Definitely should add a little bit of docs in the mdx file so we can explain how InterstitialScreen the subcomponents should be used together.

The ComposedModal / full page one might be worth bringing up with the larger group since I’m sure whoever worked on create flows can share some insight if that was some of the reason why we did not use `ComposedModal`.

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreen.js
```diff
@@ -0,0 +1,480 @@
...
+        className={cx(
+          blockClass, // Apply the block class to the main HTML element
+          className, // Apply any supplied class names to the main HTML element.
+          variantClass,
+          isVisibleClass
+        )}
+        ref={ref}
+        {...getDevtoolsProps(componentName)}
+      >
+        <div className={containerClass}>
```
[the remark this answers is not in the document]
jlongshore: Modal variant now uses ComposedModal under the hood.

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreen.stories.js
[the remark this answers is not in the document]
jlongshore: I dig it, I did stay with 'Full screen' rather than 'Full page' as I believe the usage guidelines as well as the prop being called 'isFullScreen'... I like the simplification and agree - users can toggle / modify all this with the controls provided.

packages/ibm-products/src/components/InterstitialScreenViewModule/_interstitial-screen-view-module.scss
```diff
@@ -13,11 +13,35 @@
...
+
+    // &--md {
+    //   margin-top: $spacing-11;
+    //   margin-bottom: $spacing-11;
+    // }
+
+    // &--lg {
+    //   margin-top: $spacing-12;
+    //   margin-bottom: $spacing-12;
+    // }
```
elycheea, line 44: Probably safe to delete these?

```suggestion
```

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreen.test.js
elycheea, line 1: Nice seeing the accessibility checks on each variant. 🔥

elycheea (commented):
Changes are looking good. Small comments left in.

In general, we try to use `getByRole`[^1] over a test ID, but I’d rather not block the PR over something like this. I think if we can remove the one comment, we can go ahead and merge this. (Comments are small, but I always worry we’ll never remove them otherwise. There might be a preferred way to query, but what you have still works.)

[^1]: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-byrole-most-of-the-time

elycheea (approved):
Closes #2892

Issue #3129, opened 2023-06-14
Replace ComboButton and ButtonMenu in v2


lee-chase:
Looks like this

![image](https://github.com/carbon-design-system/ibm-products/assets/15086604/622d4145-32a7-44d4-b756-998c76bdd779)

Looked like this in V1

<img width="450" alt="image" src="https://github.com/carbon-design-system/ibm-products/assets/15086604/c090bbff-1607-4555-a553-853ae7cef1c2">

```[tasklist]
### Tasks
- [ ] Replace internal button menus with new Carbon `MenuButton`
- [ ] Replace combo buttons with Carbon `ComboButton`
- [ ] Deprecate `ButtonMenu` and `ComboButton` (or feature flag)
- [ ] Update tests probably
```

elycheea:
Carbon released their version now (kudos to @janhassel) so I think we should actually deprecate and replace our usage with the new version.

https://react.carbondesignsystem.com/?path=/story/components-combobutton--default

davidmenendez:
going to mark this as temporarily on hold while carbon fixes an issue with combo button
![Screenshot 2023-07-07 at 10 54 23 AM](https://github.com/carbon-design-system/ibm-products/assets/6370760/3acee634-3557-4a5a-854a-8e2d4583254d)

PR #3284, opened 2023-07-21
fix: replaces and deprecates ButtonMenu and ComboButton

davidmenendez:
Contributes to #3129


davidmenendez:
@elycheea regarding `PageHeader`, yes you were right that it was involved. Thanks for the reminder to double check. `PageHeader` actually uses `ButtonSetWithOverflow` which has been updated to use carbons `MenuButton`. so everything should be covered. you can verify it by checking the [before](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-page-header-pageheader--fully-loaded-and-some) and [after](https://react.carbondesignsystem.com/?path=/docs/components-menubutton--overview). as of writing this comment the carbon storybook is broken again, so you can't verify everything. i'll comment again when it is 😅

packages/ibm-products/src/components/Datagrid/Datagrid/DatagridToolbar.js
```diff
@@ -116,7 +110,7 @@ const DatagridBatchActionsToolbar = (datagridState, width, ref) => {
               return null;
             }
             return (
-              <ButtonMenuItem
+              <MenuItem
                 key={`${batchAction.label}-${index}`}
                 itemText={batchAction.label}
                 onClick={(event) => {
```
elycheea, line 116:
```suggestion
              <MenuItem
                key={`${batchAction.label}-${index}`}
                label={batchAction.label}
                onClick={(event) => {
```

I think this is what’s causing the menu items not to render correctly.

image

elycheea (commented):
Looks good minus one issue.

Also, as we discussed earlier today, we probably should come up with some clearer communication on the deprecated components.

The [Security package](https://v1-ibm-products.carbondesignsystem.com/?path=/story/security-components-card-legacy--content) uses a `legacy` tag and a link to the new version.

image

I think we can do something more akin to the feature flagging style.

image

Probably still a warning type message.

> This component has been deprecated in favor of `MenuButton` from Carbon.
> Learn more

davidmenendez:
@elycheea changes made 👍

elycheea:
Seeing a failed test ... looks like it was from @matthewgallo’s PR in #3371.
Guessing we have some minor adjustments to make based on either `DatagridActions` or `DatagridToolbar`. 👀

davidmenendez:
ok the latest changes appear to have fixed the issue. waiting on checks...

elycheea (approved):
🎉

PR #3288, opened 2023-07-24
fix: get pre commit hooks running again

davidmenendez:
i kept noticing that pre commit hooks weren't running locally for me, so i wanted to fix that. i also wanted to clean up some of the package.json commands because it felt like there was some unnecessary stuff in there.


davidmenendez:
@lee-chase when you have a chance can you take a look? i think you might have some knowledge about these hooks and scripts

davidmenendez:
@matthewgallo if you want to take a look at this too that would be appreciated

matthewgallo (approved):
Tested from a new branch created from this one, works great! 🥳

Issue #2980, opened 2023-05-10
`<Tearsheet>` and `<CreateTearsheet>` have different default value for `verticalPosition` props


joshualiu:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

Both `<Tearsheet>` and `<CreateTearsheet>` components from `ibm-products` have `verticalPosition` props.
For `<Tearsheet>`, the default value is `lower` and marked as deprecated, while for `<CreateTearsheet>` the value is `normal`.
https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-components-tearsheet-tearsheet--tearsheet#component-api
https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-patterns-create-flows-createtearsheet--multi-step-tearsheet#component-api

Design doc: https://pages.github.ibm.com/cdai-design/pal/components/tearsheet/usage/#placement says to keep 88px from the top. Which I assume `<CreateTearsheet>` needs to be updated.
## Steps to reproduce the issue
refer to the storybook links above

## Additional information

- Screenshots or code
`<Tearsheet>` : https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/main_v1/packages/cloud-cognitive/src/components/Tearsheet/Tearsheet.js#L87
`<CreateTearsheet>`: https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/main_v1/packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.js#L316

```[tasklist]
### Acceptance criteria
```

elycheea:
@joshualiu Would like to contribute the fix for this? @matthewgallo says it seems like `verticalPosition` is deprecated so should be straightforward to fix. 🙂

joshualiu:
@elycheea Sure! Happy to!

elycheea:
Just documenting that #3865 was a related issue. We did not change the defaults, but when addressing the Storybook issue, did notice that `normal` and `lower` may both be relevant, depending if teams are actually using PageHeader with or without breadcrumbs.

Still, it would still be good fix to have our defaults consistent longer term.

elycheea:
[...]

PR #3320, opened 2023-07-25
fix(CreateTearsheet): Fix for issue #2980 with CreateTearsheet

rakesh-musturi:
Contributes to https://github.com/carbon-design-system/ibm-products/issues/2980

Default value for verticalPosition prop was normal which is changed now to "lower"

What did you change?
Changed the default prop value to "lower"

This is a simple prop change


davidmenendez:
hey @rakesh-musturi thanks for helping out. there may be a slight confusion about what needs to be done. currently `verticalPosition` is being deprecated in the underlying shell component. the change you actually need to make to get rid of any messaging regarding this can be found here https://github.com/carbon-design-system/ibm-products/blob/29b5206096b3b28e6a2c4e14ff4721769c433821/packages/ibm-products/src/components/Tearsheet/TearsheetShell.js#L349

Issue #3227, opened 2023-07-13
Remove sortable styles from Datagrid headers when `useSortableColumns` not in use


elycheea:
@thefirstartist noticed this in #2608, but it looks like we are always adding the class for `__sortableColumn` in header rows even when the `useSortableColumns` hook is not included.

I’m guessing this is due to some default settings that either we or `react-table` set because from what I’m able to tell, `header.canSort` is always true. :eyes:

thefirstartist:
@elycheeaI think we see hover feedback on headers because all columns are resizable. One major flaw of resizable columns is that all column widths are fixed when users resize their browser window. This can be really annoying because now users need to adjust individual columns manually. 

Additionally, the resizable column is not accessible since it cannot be achieved with a keyboard.

elycheea:
@thefirstartist There’s a PR and work to bring keyboard support for resize that we discuss with Mike Gower a while back.

> One major flaw of resizable columns is that all column widths are fixed when users resize their browser window. This can be really annoying because now users need to adjust individual columns manually.

Ahh, I can see that as an issue but not sure what the ideal behavior is otherwise. I think this is an area where we can explore further — as we see more usage, hopefully we’ll gain some ideas how how this could work better.

PR #3322, opened 2023-07-25
fix(Datagrid): remove sort styles hook not in use

Ratheeshrajan:
Contributes to #3227 

Always adding the class for `__sortableColumn` in header rows even when the useSortableColumns hook is not included.

#### What did you change?

`__sortableColumn` is conditionally added when `isTableSortable` is `true`. 

Storybook


davidmenendez:
looks like the lint check failed. not your fault, the pre-commit hook has been broken. i have a fix for it here (https://github.com/carbon-design-system/ibm-products/pull/3288). you just need to run `yarn format` and `yarn lint` to make sure everything is formatted correctly

matthewgallo (approved):

Issue #3184, opened 2023-07-07
The accessibility test is failing on Datagrid component


seonyukim:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.
The DataGrid has an accessibility test violation. When the table is rendered it doesn't take aria-label  for `<table>` and `<th>` .

> Is this issue related to a specific component?
DataGrid
> What did you expect to happen? What happened instead? What would you like to
> see changed?
No accessibility test failure for Datagrid
> What browser are you working in?
Chrome
> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?
"2.1.2"
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
As soon as possible
## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-products/tree/main/examples/carbon-for-ibm-products/SidePanel

## Additional information
![Screenshot 2023-07-07 at 2 28 01 PM](https://github.com/carbon-design-system/ibm-products/assets/14826482/f1bdf30b-e842-4cd4-886a-c0ab5a1c31a7)

- Screenshots or code
- Notes

seonyukim:
Do you mean this is fixed v2? I initially found from V2

davidmenendez:
yes it's fixed in v2. it should be in our latest release

seonyukim:
Yeah!!! Thanks!

Issue #2837, opened 2023-04-06
[Datagrid] using useSortableColumns and useCustomizeColumns hooks together breaks column customization.


xmannyxfreshx1:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description 

> Describe in detail the issue you're having.

Using useSortableColumns and useCustomizeColumns hooks together breaks column customization. Both work fine independently but when trying to use both at the same time, the column customization tearsheet fails to populate the columns. Sorting works independently or combined.

> Is this issue related to a specific component?

Datagrid

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The exception is for both hooks to work in harmony 😄  

> What browser are you working in?

Chrome
> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

`"@carbon/ibm-products": "1.47.0",`
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

CP4WAIOps

## Steps to reproduce the issue

1. Step one:
Load a datagrid and enable either `useSortableColumns` or  hooks and notice they work as expected
2. Step two
Use both hooks at the same time, then notice the columns are not customizable anymore.

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-cloud-cognitive/tree/main/examples/cloud-cognitive/codesandbox

https://codesandbox.io/s/angry-stallman-muehg2?file=/src/Example/Example.js

## Additional information

- Screenshots or code
### Datagrid with both hooks enabled
![Screenshot 2023-04-06 at 10 46 08 AM](https://user-images.githubusercontent.com/13371221/230415680-e0e126c9-92b6-4287-b0de-6be11e2748b6.png)

### Datagrid with just `useCustomizeColumns` enabled
![Screenshot 2023-04-06 at 10 48 22 AM](https://user-images.githubusercontent.com/13371221/230416118-bb4944a0-ae6b-44e5-888f-b21f6467be2b.png)

- Notes

K-Markopoulos:
The `useCustomizeColumns` doesn't work with `useColumnCenterAlign` hook.  If the latter is present and `centerAlignedColumn` is set for a column, this column won't be displayed in the Customize Columns tearsheet.

[...]

PR #3325, opened 2023-07-25
fix: datagrid title

davidmenendez:
Contributes to #3184 

addresses accessibility concerns in `Datagrid` by adding a `title` prop for the `table` component and adding titles to the table headers.


packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHeaderRow.js
```diff
@@ -44,6 +48,8 @@ const HeaderRow = (datagridState, headRef, headerGroup) => (
               header.getHeaderProps().className
             )}
             key={header.id}
+            title={header.Header.props?.title}
+            aria-hidden={!header.Header.props?.title}
```
matthewgallo, line 52: There is some nuance here because of how the structure of the column headers can change based on the hooks/plugins that are being used.

This is something that I've been looking at recently (particularly [here](https://github.com/carbon-design-system/ibm-products/pull/3344)). For example, if you add `useSortableColumns`, with this change all `th` elements will receive `aria-hidden="true"` which is not expected. This is because adding `useSortableColumns` changes `header.Header` (or `colDef.Header`) from just an object to a function.

Can we add a default title attribute to account for these instances? You can see this by looking at the `Sortable Columns` [story in the deploy preview](https://deploy-preview-3325--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary--sortable-columns).

image

matthewgallo (changes requested):
Left a comment regarding the logic used for `aria-hidden` on the column headers and how it is marking column headers to `aria-hidden="true"` incorrectly if there are any hooks/plugins used that manipulate the column headers. Similar to #2837, where the structure of the column headers changes, causing unexpected behavior.

davidmenendez:
@matthewgallo thanks for pointing that out. fixes made. i believe my new approach addresses your concerns

davidmenendez:
just kidding. got it working locally before i pulled the latest and something broke

davidmenendez:
i'm going to close this and start a new PR based on the new code that matt has written to address this problem in https://github.com/carbon-design-system/ibm-products/pull/3344

Issue #3202, opened 2023-07-12
*Enhancement Request* for NotificationsPanel Component ‘link’ object to have the ability to ‘open in new window’ when clicked by user


bridgettepepper:
## What will this achieve?
:*Enhancement Request* for NotificationsPanel Component ‘link’ object to have the ability to ‘pop out into new window’ when clicked by user. 

Currently the 'data' object for the notifications panel has a 'link' object with 'text' and 'url'. However, since the url is encapsulated by an object, i'm unable to add any url attributes such as 'target= '_blank'' so the link, when clicked, will open into a new browser tab. And there is no easy way to do this currently. 

My request is to include a way to allow for links to open into a new window when clicked, by adding a 'target' property to the 'link' object with '_blank' as the allowed string. This should be an optional property for flexibility in case there are situations where link should not open in new window. 

Current Data object structure: 
```
[{
  id: string | number,
  type: 'error' | 'warning' | 'success' | 'informational',
  timestamp: Date,
  title: string,
  description: string,
  link: {
    url: string,
    text: string
  },
  unread: bool,
  onNotificationClick: func
}]

```

Proposed enhancement adding an optional 'target' property to link object of data:
```
[{
  id: string | number,
  type: 'error' | 'warning' | 'success' | 'informational',
  timestamp: Date,
  title: string,
  description: string,
  link: {
    url: string,
    text: string
    target: string -> allowed string '_blank' so link opens in new tab once clicked
  },
  unread: bool,
  onNotificationClick: func
}]
```

## Additional information

- Designs
- Existing code
- etc

punnoosewilson:
Hi @bridgettepepper,
We've added an extra optional object inside the link. So you can include not only the `target` prop but also any of the `Link` props with in that optional object.

I've attached a screenshot below with some sample data to assist you:

![image](https://github.com/carbon-design-system/ibm-products/assets/40118548/2ca9e4ff-9b5e-494e-a344-8c36bf6ec369)

PS: A [PR](https://github.com/carbon-design-system/ibm-products/pull/3326) is raised against this issue. This feature will be available after the merge.

punnoosewilson:
Hi @bridgettepepper,
Implemented much straight forward method for constancy with v1. 
The optional object is no longer required. You can directly give the link-props.

I've attached a sample data screenshot below.

[...]

PR #3326, opened 2023-07-26
feat(NotificationsPanel): include target attribute inside the link

punnoosewilson:
Contributes to #3202 

Enhancement Request for NotificationsPanel Component ‘link’ object to have the ability to ‘open in new window’ when clicked by user

#### What did you change?
- Included target property inside the link object.

256222812-a9ba2cae-8787-4a5f-bc02-e0874e6d92fa

- Storybook

### DIFF
--- a/packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.js
+++ b/packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.js
@@ -346,6 +346,9 @@ export let NotificationsPanel = React.forwardRef(
                 <Link
                   href={notification.link.url}
                   className={`${blockClass}__notifications-link`}
+                  {...(notification.link.target
+                    ? { target: notification.link.target }
+                    : {})}
                 >
                   {notification.link.text}
                 </Link>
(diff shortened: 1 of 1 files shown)


matthewgallo (changes requested):
I think a better option would be to pass in rest props to the `Link` component. This way any other `Link` props that consumers want to pass in, they are able to do so.

Something like this:
```jsx
<LInk
  {...notification.link}
  href={notification.link.url}
  className={`${blockClass}__notifications-link`}
>
  {notification.link.text}
</Link>
```

punnoosewilson:
Hi @matthewgallo,
Thank you for the suggestion.
 
If we spread `notification.link` then unwanted props like 'text and url' will appear inside the link.
 
image
 
'

So I'm planning to incorporate another prop inside the link - 'optional'. So customers can include any of the `Link` props within that new prop - 'optional'.

Screenshot 2023-07-27 at 12 47 46 PM

'

Please let me know what you think. 
If everything is fine, I'll push the changes.

matthewgallo:
Ahh yes, I think we can go with the `optional` property then and spread that. Thanks @punnoosewilson!

punnoosewilson:
@matthewgallo 
I've pushed the changes.

matthewgallo (approved):

matthewgallo:
@punnoosewilson can you also create a PR for v1?

Issue #3163, opened 2023-06-26
Options tile toggle is not working as expected


sdepuy:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

The `onToggle` callback does not get invoked when clicking on the toggle component of the `Options tile` component.

Note that when using the keyboard to focus the toggle element and push the SPACE key, it works as expected.

> Is this issue related to a specific component?

Yes, `Options tile`.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The `onToggle` callback should be called so we can manage the `enabled` state of the component.

> What browser are you working in?

Chrome and FF

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

Using @carbon/ibm-products v2.1.3 or v2.2.0 and @carbon/react v1.31.3.

Note that the @carbon/react `Toggle` underlying component works as expected.

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

Content Service on AWS.

## Steps to reproduce the issue

```
<div>
          <OptionsTile
            enabled={enabled}
            onToggle={setEnabled}
            title="Hardware acceleration"
          />
          <Toggle
            size="sm"
            toggled={enabled}
            onToggle={setEnabled}
            labelA="Hardware acceleration"
            labelB="Hardware acceleration"
          />
</div>
```

The `Options tile` component does not work but the `Toggle` component works fine.

> Please create a reduced test case in CodeSandbox
>
[> https://codesandbox.io/s/github/carbon-design-system/ibm-products/tree/main/examples/carbon-for-ibm-products/SidePanel
](https://codesandbox.io/p/sandbox/github/carbon-design-system/ibm-products/tree/main/examples/carbon-for-ibm-products/OptionsTile?file=%2Fsrc%2FExample%2FExample.jsx%3A1%2C1)

## Additional information

- Screenshots or code
- Notes

rakesh-musturi:
@sdepuy every thing works by justing adding an onClick function to optionTile component. 
https://codesandbox.io/p/sandbox/zen-kapitsa-wqx572
check this one, i have just added to your code and it works.
Will updage the storybook too
cc @elycheea

PR #3330, opened 2023-07-30
fix: update OptionTile story issue 3163

rakesh-musturi:
Contributes to #
#3163 
{{short description}}
On click function was missing from the example

#### What did you change?
added onClick function

Storybook
working example
https://codesandbox.io/p/sandbox/zen-kapitsa-wqx572

### DIFF
--- a/packages/ibm-products/src/components/OptionsTile/OptionsTile.stories.js
+++ b/packages/ibm-products/src/components/OptionsTile/OptionsTile.stories.js
@@ -129,7 +129,7 @@ const TemplateStatic = ({ enabled, ...rest }) => {
   const [liveEnabled, setLiveEnabled] = useState(enabled);
 
   function onToggle(e) {
-    setLiveEnabled(e);
+    setLiveEnabled(!liveEnabled);
     action('onToggle')(e);
   }
 
@@ -141,6 +141,7 @@ const TemplateStatic = ({ enabled, ...rest }) => {
     <OptionsTile
       onToggle={onToggle}
       onChange={onChange}
+      onClick={onToggle}
       {...rest}
       enabled={liveEnabled}
     />


davidmenendez (approved):

packages/ibm-products/src/components/OptionsTile/OptionsTile.stories.js
```diff
@@ -141,6 +141,7 @@ const TemplateStatic = ({ enabled, ...rest }) => {
     <OptionsTile
       onToggle={onToggle}
       onChange={onChange}
+      onClick={onToggle}
```
janhassel, line 144: The `OptionsTile` doesn't accept an `onClick` prop which means this will be spread to the outer `<div>` which would be invalid without additional semantics and creates an unintended interaction.

```suggestion

```

packages/ibm-products/src/components/OptionsTile/OptionsTile.stories.js
```diff
@@ -129,7 +129,7 @@ const TemplateStatic = ({ enabled, ...rest }) => {
   const [liveEnabled, setLiveEnabled] = useState(enabled);
 
   function onToggle(e) {
-    setLiveEnabled(e);
+    setLiveEnabled(!liveEnabled);
```
janhassel, line 132: `onToggle` is called with either `true` or `false` depending on what the next state should be, therefore passing it directly into the `setLiveEnabled` function was appropriate.

```suggestion
    setLiveEnabled(e);
```

janhassel (changes requested):
These updates simulate an unwanted behavior (clicking the OptionsTile header to toggle the enabled state). I think the root issue you're trying to solve is that `onToggle` seems to not be called correctly. I did a bit of digging and think it's due to some issues with how the component currently implements the Carbon toggle.

It should not set `labelA` and `labelB` to empty strings, but use `hideLabel` instead. Also, the `labelText` can be omitted as `aria-labelledby` is already provided.

With these updates the `onToggle` is called as expected in my local setup. Can you give this a try on your end?

```diff
  <Toggle
    id={`${id}-toggle`}
    className={`${blockClass}__toggle`}
    toggled={enabled}
-   labelA=""
-   labelB=""
    aria-labelledby={titleId}
+   hideLabel
    onToggle={onToggle}
    size="sm"
    disabled={isLocked}
-    labelText={title}
  />
```

https://github.com/carbon-design-system/ibm-products/blob/40fe19b4ef32f97f01e8664bbe81fc1e459d0f3f/packages/ibm-products/src/components/OptionsTile/OptionsTile.js#L260..L271

packages/ibm-products/src/components/OptionsTile/OptionsTile.stories.js
rakesh-musturi: Im not clear what do you mean by unintended interactions? And why is onClick not accepted when Toggle component itself has an onClick?

janhassel: The `OptionsTile` component does not have a prop called `onClick`. This means when you pass it anyways, it will be emitted on the outer `<div>` with any other attributes the component doesn't handle. This happens here: https://github.com/carbon-design-system/ibm-products/blob/40fe19b4ef32f97f01e8664bbe81fc1e459d0f3f/packages/ibm-products/src/components/OptionsTile/OptionsTile.js#L243-L244

The interaction pattern for OptionsTile is that clicking on the header will open and close the content of the tile. To enable or disable the tile, the user must click on the toggle directly. Otherwise they wouldn't be able to open or close the tile without unexpected side effects that may come from enabling or disabling an option.

Have you tried my suggestion to update `/packages/ibm-products/src/components/OptionsTile/OptionsTile.js` with the code shared above?

rakesh-musturi: Did not work for me, can you please check the sandbox link which i have commented

rakesh-musturi: https://codesandbox.io/p/sandbox/zen-kapitsa-wqx572

janhassel: In your linked codesandbox, you can see how adding the `onClick` is creating the unwanted interaction I was mentioning. Since it is emitted on the outer `<div>`, whenever you click anywhere on the OptionsTile it toggles between enabled and disabled.

https://github.com/carbon-design-system/ibm-products/assets/28265588/a5a46ec8-cfdc-4030-9db5-73b40ce28131

Instead, it should only switch between enabled and disabled when the user clicks the toggle on the right.
Right now, the `<OptionsTile>` implementation wrongly sets the `<Toggle>` props `labelA` and `labelB` to empty strings. This is the root cause of the issue reported in #3163.

The reason behind this is that the Carbon `<Toggle>` uses a `<button>` under the hood which is set to be visually hidden. Clicking on the generated `<label>` element will trigger the button click and therefore the `onToggle` event. Since the `<OptionsTile>` currently sets the label to empty strings though, there is no clickable target left.

In order to fix this, the `<OptionsTile>` should use `props.hideLabel` for the `<Toggle>` instead as this will influence the underlying markup and cause the `<Toggle>` to render without labels but with a clickable target.

The fix needs to be applied to `/packages/ibm-products/src/components/OptionsTile/OptionsTile.js`:
https://github.com/carbon-design-system/ibm-products/blob/40fe19b4ef32f97f01e8664bbe81fc1e459d0f3f/packages/ibm-products/src/components/OptionsTile/OptionsTile.js#L260-L271
```diff
<Toggle
  id={`${id}-toggle`}
  className={`${blockClass}__toggle`}
  toggled={enabled}
- labelA=""
- labelB=""
  aria-labelledby={titleId}
+ hideLabel
  onToggle={onToggle}
  size="sm"
  disabled={isLocked}
- labelText={title}
/>
```

davidmenendez:
i'm going to close this in favor of https://github.com/carbon-design-system/ibm-products/pull/3381

PR #3331, opened 2023-07-31
feat(NotificationsPanel): include target attribute inside the link - v1

punnoosewilson:
Contributes to #3202 

Enhancement Request for NotificationsPanel Component ‘link’ object to have the ability to ‘open in new window’ when clicked by user. 

#### What did you change?
- Included an extra optional object inside the link. So the consumers can include any of the Link props with in that optional object.

Storybook

PS: PR for v1


davidmenendez (dismissed):
i like this idea, but i think this can be simplified. i don't think a new `optional` prop is required. i think simply doing `{...notification.link}` should be sufficient for passing on any extra props.

punnoosewilson:
> i like this idea, but i think this can be simplified. i don't think a new `optional` prop is required. i think simply doing `{...notification.link}` should be sufficient for passing on any extra props.

Hi @davidmenendez ,
We also thought about the same idea. but it had some problem. 
I've mentioned this inside the PR created for v2.
https://github.com/carbon-design-system/ibm-products/pull/3326#issuecomment-1653077177

That's why we introduced an `optional` prop.

lee-chase:
> > i like this idea, but i think this can be simplified. i don't think a new `optional` prop is required. i think simply doing `{...notification.link}` should be sufficient for passing on any extra props.
> 
> Hi @davidmenendez , We also thought about the same idea. but it had some problem. I've mentioned this inside the PR created for v2. [#3326 (comment)](https://github.com/carbon-design-system/ibm-products/pull/3326#issuecomment-1653077177)
> 
> That's why we introduced an `optional` prop.

I can see how {...notification.link} in would be problematic due to the 'text' property.

You can probably get away with 
`{...{...notification.link, text: null}}`

alternatively use the prepareProps utility function.

`{...prepareProps({}, notification.link, ['text'])}`

punnoosewilson:
Thank you for the suggestion @lee-chase. 
I've incorporated prepareProps utility function.
`{...prepareProps({}, notification.link, ['text', 'url'])}`

@davidmenendez @matthewgallo could you review it?

PS: For the consistency I'm planning to do the same update inside v11.

elycheea (approved):
🎉

elycheea:
@punnoosewilson Don’t forget to add a v2 PR as well!

@all-contributors add @punnoosewilson for code

PR #3348, opened 2023-08-02
fix(Datagrid): add support for other extensions with customizable columns (v1)

matthewgallo:
Contributes to #2837 

This PR adds some logic to `DraggableItemsList` that will allow for use of `useCustomizeColumns` with one or more hooks that manipulate the column header (such as `useSortableColumns` or `useColumnCenterAlign`). This also allowed for refactoring how we get the column title if there is a hook being used which manipulates the column header structure.

We previously attempted to support customizable columns with `useSortableColumns` by passing down a prop called `isTableSortable` but this did not get at the root of the problem which is, the use of one or more extensions which may manipulate the structure of the column header. I added `getNestedTitle` which recursively checks for the title so this should account for instances when one or more hooks are used that deal with the column header.

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Datagrid/addons/CustomizeColumns/CustomizeColumnsTearsheet.js
packages/ibm-products/src/components/Datagrid/Datagrid/addons/CustomizeColumns/TearsheetWrapper.js
packages/ibm-products/src/components/Datagrid/styles/_useColumnCenterAlign.scss
packages/ibm-products/src/components/Datagrid/useDefaultStringRenderer.js
packages/ibm-products/src/components/Datagrid/useSortableColumns.js
```
Storybook, for testing I added `useSortableColumns` and `useColumnCenterAlign` together and on their own with the `useCustomizeColumns` hook, to test this functionality

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
@@ -37,7 +37,6 @@ const Columns = ({
   assistiveTextInstructionsLabel,
   assistiveTextDisabledInstructionsLabel,
   selectAllLabel,
-  isTableSortable,
 }) => {
   const [ariaRegionText, setAriaRegionText] = React.useState('');
   const [focusIndex, setFocusIndex] = React.useState(-1);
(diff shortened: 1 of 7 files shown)


davidmenendez:
looks good but v1 tests are failing

davidmenendez (approved):

PR #3358, opened 2023-08-04
fix(Nav): show active state also for links with a fragment

dbrugger:
Contributes to #

The Nav menu shows an active state if the user is on the page that is on the url of the link.
But this does not work if the link contains a fragment (hash)

#### What did you change?

Changed logic to use path _and_ hash instead of _or_.


lee-chase:
Why switch from displayName to name?

https://legacy.reactjs.org/docs/react-component.html#displayname

dbrugger:
Thanks @lee-chase . I had this fix lying around since before https://github.com/carbon-design-system/ibm-products/commit/6364c8b2c74e2521f48f924e0e92feb4f5c50b3e . will adjust.

lee-chase (approved):

PR #3362, opened 2023-08-09
fix: datagrid accessibility updates

davidmenendez:
Contributes to #3184 

adds a `title` prop for the table and adds `title` attributes to `th` tags if they exist. if they don't exist add's `aria-hidden`


davidmenendez:
@matthewgallo when you have a minute 👀

Issue #3153, opened 2023-06-22
Security releases: publishing under `v1-latest` dist-tag


matthewgallo:
## What will this achieve?

The security package is still getting published under the `v1-latest` dist-tag because it's part of the group of packages published from `release-v1.yml`. What we need to happen is publish `ibm-products` under `v1-latest` and publish security as `latest`. Change will need to be made [here](https://github.com/carbon-design-system/ibm-products/blob/main/.github/workflows/release-v1.yml#L43-L45).

## How will success be measured?

`@carbon/ibm-products: 1.x.x` => `v1-latest` dist-tag
`@carbon/security` => `latest` dist-tag

## Additional information

- Designs
- Existing code
- etc

elycheea:
@lee-chase Mind taking a look at this if you have a moment between the upgrades?

PR #3368, opened 2023-08-14
fix: dist tag latest-v1

lee-chase:
Contributes to #3153 

First step to try and fix this. I suspect extra work is needed to access npm.

#### What did you change?

Added shell script to update dist tags.

Command line

### DIFF
--- a/update-disttags.sh
+++ b/update-disttags.sh
@@ -1,0 +1,36 @@
+
+#  Copyright IBM Corp. 2023, 2023
+#  This source code is licensed under the Apache-2.0 license found in the
+#  LICENSE file in the root directory of this source tree.
+
+# Usage `sh update-disttags.sh <folder> <package>`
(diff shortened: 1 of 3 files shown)


update-disttags.sh
```diff
@@ -0,0 +1,36 @@
...
+# If the package version does not match the `latest` on npm but does match `v1-latest` then
+# the dist-tag `latest` is updated.
+
+FOLDER=$1;
+PACKAGE=$2;
+
+cd $FOLDER;
+VERSION=$(npm pkg get version --workspaces=false | tr -d \");
+cd ../..
+NPN_LATEST=$(npm dist-tag ls $PACKAGE@latest | grep ^latest | cut -d " " -f2);
```
elycheea, line 22: Should this be `NPM` instead of `NPN`?

lee-chase: Good spot/eyesight.

Issue #3328, opened 2023-07-26
[Datagrid Filter Panel] Previously selected filters re-appear after using the 'X' close button in the filter panel.


xmannyxfreshx1:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

Previously selected filters re-appear after closing the fliter panel using the 'X' close button in the top right.
Only happens after clearing filter selections using the 'clear filters' button in the filter tag summary.

> Is this issue related to a specific component?

Datagrid, filter panel with batch updates

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Expected: Closing the filter panel should not re-apply previously selected filters after they have been cleared
> What browser are you working in?

Chrome

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

`"@carbon/ibm-products": "2.4.1",`

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

CP4AIOPs

## Steps to reproduce the issue

1. Load a Datagrid and filter panel with batch updates
2. Open filter panel and select filters, notice that the table is filtered accordingly 
3. While the panel is still open, use the 'Clear filters' ghost button on the right hand side. Notice the filters are cleared as expected.
4. Close the panel using the 'X' button in the top right, notice that the previously selected filters re-appear in the filter summary and in the panel (once re-opened)

> Issue can be reproduced with storybook component
>

https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-panel--panel-batch

## Additional information

- Screenshots or code
- Notes

matthewgallo:
This should be fixed as part of https://github.com/carbon-design-system/ibm-products/pull/3582. Going to close, if there still are any issues after the next release that includes this fix, feel free to reopen!

PR #3384, opened 2023-08-21
fix: remove filter revert on close

rakesh-musturi:
Contributes to #3328 

Previously selected filters re-appear after closing the fliter panel using the 'X' close button in the top right.
Only happens after clearing filter selections using the 'clear filters' button in the filter tag summary.

#### What did you change?
Commented the function revertToPreviousFilters(); which is setting the previous filters even after clearing the filters
Working in storybook

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/hooks/useFilters.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/hooks/useFilters.js
@@ -334,7 +334,8 @@ const useFilters = ({
   const cancel = () => {
     // Reverting to previous filters only applies when using batch actions
     if (updateMethod === BATCH) {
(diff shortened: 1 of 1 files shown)


elycheea (changes requested):
So while this seems to resolve the current bug, I do think this current change has unintended consequences where you are unable to restore the current set of filters in batch mode at all. 

I think it would make more sense to investigate how we’re clearing the filters. Once someone explicitly clears the filters, we may not need to retain the previous filter state.

rakesh-musturi:
@elycheea Got it, will fix this too

matthewgallo:
Hey @rakesh-musturi, this should now be fixed as part of  https://github.com/carbon-design-system/ibm-products/pull/3582. If you still see this behavior after the next release, feel free to reach back out!

rakeshmusturi:

Issue #3242, opened 2023-07-18
Fix alignment in empty state buttons


elycheea:
- [x] Center align buttons in empty state
- [x] Reduce extra whitespace in empty state
    - Expect we can change this to use `display: inline-block` instead of `display: block` 
- [x] Update description color to `$text-secondary`

> 7. Visual things: 
> The “Create new” text is not centered vertically in the button
> The focus state on “View documentation”  looks weird since there is lots of white space on the right of the text as the focus indicator takes up the entire width of the “empty-state__content” container 
> ![Screenshot 2023-07-03 at 8 16 23 AM](https://github.com/carbon-design-system/ibm-products/assets/21047571/2316620d-0900-4b90-bfe2-174719e4081a) 

_Originally posted by @kristastarr in https://github.com/carbon-design-system/ibm-products/issues/3117#issuecomment-1618294932_

> This is looking good! A couple tweaks:
> - [ ] Vertically center align main CTA button text 
> - [ ] Use $text-02 for description text (not text-01)

_Originally posted by @vsnichols in https://github.com/carbon-design-system/ibm-products/issues/3121#issuecomment-1631602128_

PR #3389, opened 2023-08-22
fix(datagrid): empty state fix

jagathgj:
Contributes to #3242 

Empty state button and description styles for datagrid modified

#### What did you change?
color for description(`text-secondary`) and replaced `display: block` to `display: inline-block` for action button

storybook


davidmenendez (approved):

elycheea:
@jagathgj @davidmenendez I think these changes should be implemented on the EmptyState component instead. I believe these are discrepancies in our current implementation so while we’re seeing them in the Datagrid, the origin is probably the actual `EmptyState`!

davidmenendez (changes requested):
yeah i agree with @elycheea it didn't occur to me initially, but yeah these changes should be applied to the empty state component specifically.

jagathgj:
@davidmenendez I have made suggested changes. Please review and let me know if any more changes needed.

elycheea:

packages/ibm-products/src/components/EmptyStates/_empty-state.scss
```diff
@@ -73,7 +74,7 @@ $block-class: #{c4p-settings.$pkg-prefix}--empty-state;
 }
 
 .#{$block-class} .#{$block-class}__action-button {
-  display: block;
+  display: inline-block;
```
elycheea, line 77: I think the `inline-block` should be on instead of the action button, otherwise we still see the overly large focus area.

image


```scss
.#{$block-class}--empty-state .#{$block-class}--empty-state__link {
}
```

elycheea (approved):

Issue #2507, opened 2022-11-30
Datagrid design review: Resizable columns


elycheea:
## Design review

- Component epic #1607 #2499
- Designers/reviewers: Virginia Nicholson
- Links: [PAL guidance](https://pages.github.ibm.com/cdai-design/pal/components/data-table/resizable-columns/usage) / [Storybook](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary) / [Prototype in PR](https://deploy-preview-2464--carbon-for-ibm-products.netlify.app/)

---

### Standards

- [ ] All pattern updates/changes/iterations have been discussed with the
      component developer
- [ ] Patterns have been approved by either DSAG or another approving entity

### Pattern and behavior

- [ ] The component behaves according to the guidelines set by the pattern
      maintainer
- [ ] The component’s UI matches the pattern specifications set by the pattern
      maintainer
- [ ] The component’s motion matches the specifications set by the pattern
      maintainer(s)
- [ ] The UI produced is responsive, cross-browser, and responds to the
      currently set Carbon theme.
- [ ] Colors, themes, sizes and additional props are true and correct, aligning
      with Carbon set tokens (unless otherwise specified by the pattern)
- [ ] Paddings/margins/spacings are true and correct, in both desktop and mobile
      views

### Storybook

- [ ] A functioning component renders in Storybook
- [ ] The Storybook displays multiple scenarios that show how the component can
      be used
- [ ] Changing props in the Storybook updates the component

matthewgallo:
• Explore examples from Google sheets and Excel around keyboard accessibility for resizable columns
• Think about how keyboard focus state works with sorting and other potential focusable elements in the column header

elycheea:
@Ratheeshrajan Matt’s already been involved in the initial design reviews for these issues. While he’s out, I think you’re the best person to help close out these issues.

elycheea:
@vsnichols @johanna-koval Wanted to follow up on some of the comments from the design review.
- Google Sheets and Excel are more of data spreadsheets; but we can see how data tables are similar. I did not find any examples of resize actions in either case though. This does make me think our guidance should include some recommendations on when to choose a `DataSpreadsheet` over a `Data table` though.
- Looked into sortable headers and realize Carbon table now includes sortable so we should update to include their tabbable version; will open a new issue for this change

[...]

Issue #1923, opened 2022-04-15
Data table: Align resizable columns to PAL pattern


johanna-koval:
The current behavior of [row size](https://pages.github.ibm.com/wdp-gov/wkc-shared-ui/?path=/story/datagrid--row-size-dropdown) needs a few adjustments and additions to align to the new [Resizable columns Pal pattern](https://pages.github.ibm.com/wdp-gov/wkc-shared-ui/?path=/story/datagrid--row-size-dropdown)

Related Epic:
https://github.com/carbon-design-system/ibm-cloud-cognitive/issues/1607

Adjustments and additional behaviors

- Resize multiple columns at a time :

- [ ] Hold Shift, and hover over the right side of any column header until a horizontal split cursor appears.

- [ ] To manually define the width that will be applied to all visible columns, drag the column line to set the desired width.

- [ ] To resize all visible columns to fit the content, double-click the column line.

![resizing_2columns](https://user-images.githubusercontent.com/5809654/163623808-8ea76600-7142-4e45-be50-0c6f45e52fe5.gif)

- Resize column to fit its content:

- [ ] Hover over the right side of the column header until a horizontal split cursor appears. Double-click the column line. The column is resized to fit the contents.
![doubleClick](https://user-images.githubusercontent.com/5809654/163623937-7118dbf8-b482-429a-919f-71d3fa8d4f85.gif)

- [x] Display minimum one letter in a column

<img width="541" alt="CleanShot 2022-04-15 at 15 29 57@2x" src="https://user-images.githubusercontent.com/5809654/163623978-4a22713e-5ef8-49af-8523-a348513f897e.png">

- [ ]  Wrap column header titles

![wrappingHeader](https://user-images.githubusercontent.com/5809654/163624168-03443ea7-d707-496b-91ea-a7cc32b0d9e6.gif)

- Accessibility
-Keyboard navigation

- [ ] tab: navigates the user to the different interactive elements

- [ ] space bar: triggers the selected element

- [ ] enter: triggers the selected element

- [ ] left & right arrow keys: resize column left or right

- [ ] Tab order

First interactive element in the body area.
Proceed left to right and top to bottom through the rest of the body elements.
Primary cell
Secondary resize handle
Secondary cell

lizisawizard:
@johanna-koval For the wrapping column header titles, do single word titles like "Name" just behave like before?

johanna-koval:
[...]

PR #3394, opened 2023-08-22
feat(Datagrid): add keyboard support for col resizing

matthewgallo:
Contributes to #1923 and #2507

Took another look at adding keyboard support for resizable columns. This time around i discovered that we can pass a [custom stateReducer](https://github.com/TanStack/table/blob/v7/docs/src/pages/docs/api/useTable.md) to `useTable` that will allow us to manually track the resize events via the `dispatch` function they provide to us in the table state, giving us the ability to add the missing keyboard support.

I kept the same approach of using a slider input and changed the visual representation of the resizer, as the previous version felt a bit clunky.

New focus state of the resizer can be seen below:
image

Resizing with both mouse and key events enforce a minimum width so at least one character from the column header is displayed.

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHeaderRow.js
packages/ibm-products/src/components/Datagrid/styles/_datagrid.scss
packages/ibm-products/src/components/Datagrid/useDatagrid.js
```
Storybook


elycheea (approved):
🔥 works great across row height changes and even when headers are sortable.

Issue #3398, opened 2023-08-24
The CreateFullPage influencer title is in the designs, but it has not been added to the actual implementation


baraaabdelsalam:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

The influencer title is specified in the designs as you can see in the screenshot and the title prop exists but its not being used by the influencer.

![Screenshot 2023-08-24 at 11 17 23](https://github.com/carbon-design-system/ibm-products/assets/22206101/6f8cb385-4195-4e7f-b1d4-a1f7fb460b4c)

> Is this issue related to a specific component?

CreateFullPage

> What did you expect to happen? What happened instead? What would you like to
> see changed?

The title string should appear in the influencer if provided

PR #3399, opened 2023-08-24
fix(create-full-page): add influencer title support

baraaabdelsalam:
Contributes to #3398

The influencer title is specified in the designs and the title prop exists but its not being used by the influencer and therefore it doesn't show up on the UI

#### What did you change?
- changed the title proptype from node to string (for aria-label support)
- passed the title to the createInfluencer
- conditionally render the title in create influencer
- added the padding for the whole influencer area instead of the progress indicator only, based on the design specs
- fixed the styling to work as expected whether the "title" is provided or not

- Added 2 unit tests
- Tested it on storybook and updated the default props to include the title


packages/ibm-products/src/components/CreateInfluencer/CreateInfluencer.js
```diff
@@ -43,8 +43,16 @@ export let CreateInfluencer = ({ className, currentStep, stepData }) => {
...
+    const conditionallyRenderInfluencerTitle = () => {
+      if (title) {
+        return <h3 className={cx(`${blockClass}__title`)}>{title}</h3>;
+      }
+      return null;
+    };
+
     return (
       <div className={`${blockClass}__left-nav`}>
+        {conditionallyRenderInfluencerTitle()}
```
davidmenendez, line 55: i don't think a function needs to be declared for this. 
```
{title && <h3 className={`${blockClass}__title`}>{title}</h3>
```
should be sufficient.

davidmenendez (changes requested):
overall loos good. just had one small comment.

PR #3400, opened 2023-08-24
fix: storybook use of package dev prefix

lee-chase:
Addresses issues from  #3183 

Hard code `c4p` prefixes can go unnoticed during development as the used to match the default. PR #3183 switched to using a custom prefix during development, however a number of stories which made use of the package prefix were left styled incorrectly.

#### What did you change?

- Added an alias tot he storybook webpack build used by SCSS files in storybook and any story files. No component files needed updating.
- Adjusted storybook preview.js and index.scss 
- Prevented the copying of the config-dev.scss file during the build.

Ran storybook in both dev mode and release, checking that the custom prefix was only used in dev mode and that the stories looked correct.


matthewgallo (approved):
👏

PR #3403, opened 2023-08-24
feat(Datagrid): persist col resizing values in local storage

matthewgallo:
Contributes to #1923 and #2507 

Now that we are controlling the column resizing we have the ability to persist the column widths to local storage after a column resize event and when the page refreshes the columns will maintain the resized width instead of reverting to the original width.

#### What did you change?
```
packages/ibm-products/src/global/js/utils/getLocalStorageItem.js
packages/ibm-products/src/global/js/utils/setLocalStorageItem.js
```
Storybook


davidmenendez (changes requested):
i like the idea for sure. my only suggestion is that in order to ensure that there isn't overlap in local storage to use specific datagrid id's. this way the settings are specific to individual tables.

matthewgallo:
Going to close in favor of https://github.com/carbon-design-system/ibm-products/pull/3406

Issue #3418, opened 2023-08-29
Web Terminal Content conflicts with UIShell Content


lee-chase:
In attempting to fix the Web Terminal docs page (where the story is shown) use of a Carbon Content component caused the page to be 100vh + 3rem in height. This should not happen, fix was not applied.

It should be possible to use the Web Terminal in conjunction with UIShell header and content components without the page height/width exceeding the bounds of the screen. The example should show this and reserve vertical space for the example.

elycheea:
Most likely will provide a breaking change ... or incorporate some feature flags to make this work properly with UI shell.

ljcarot:
This looks like it has been resolved.  Going to move to Icebox for now in case it surfaces again.

amal-k-joy:
We can close this as [WebTerminal](https://ibm-products.carbondesignsystem.com/?path=/docs/patterns-webterminal--overview) is available as pure pattern

Issue #3417, opened 2023-08-29
Modal dialogs (About, Create, Edit etc) should have a portal target or be aware of containing CSS


lee-chase:
In certain circumstances (Docs page in storybook). A consumer may have introduced CSS (e.g. transform: scale(1); ) that causes the use of `fixed` and `absolute`  to not be relative to the whole page.

The simplest solution here would be to allow a portal target to be defined, as per Tearsheet and other components, to allow the modal to appear centrally on the page.

Another solution is to contain the elements, but this is much trickier as Carbon sets the glass pane to 100vh 100vw.

lee-chase:
NOTE: @carbon/react uses portal in it stories

PR #3419, opened 2023-08-29
fix: some full screen docs pages

lee-chase:
Some stories are poorly formatted or present badly on the docs page. This is because Storybook has changed the way in which the stories are hosted on the docs page. 

We should as a result try to ensure
- All stories reserve the space they need to present themselves (some stories have no content or only content that does not reserve vertical space.
- Have the stories, not common CSS, specify the size of the viewport on the docs page.

e.g. 
image

#### What did you change?
- Wrapped side panel stories to reserve the full height of the page.
- Removed the 50vh common minimum height in storybook styling.
- Added `layout: 'fullscreen'` to stories which should be full screen.
- Fixed CreateSidePanel, Datagrid, DataSpreadsheet, EditSidePanel, EditTearsheetNarrow, SidePanel and Tearsheet stories.
- Partially fixed ExportModal story (raised #3417)
- Noted WebTerminal has multiple issue with regards to viewing in the docs page. Raising #3418 

Viewed in Storybook

### DIFF
--- a/packages/core/.storybook/index.scss
+++ b/packages/core/.storybook/index.scss
@@ -9,7 +9,7 @@
 // Carbon styles at the end -- this is to obtain a "worst case" for our own CSS,
 // to ensure we are resilient against different CSS loading orders and our
 // styles have the specificity necessary to override Carbon styles when needed.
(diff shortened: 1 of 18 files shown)


matthewgallo (approved):
There is some weird behavior with the entrance/exit of the side panels in the storybook docs preview. Do you think that is something related? I think we can open that as a new issue and have this PR address the layout issues that existed before.

Issue #3197, opened 2023-07-11
[DataGrid] onSort causes first sort in DataGrid to not work


cgerein:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

Passing a custom sort function to `useDataGrid()` with `onSort` causes the first sort you try to do to not change the actual sort state on the column, but the `onSort` function will still be called. This will cause onSort to call with the id and sort order you would expect, but the DataGrid will still show the previous order. Any clicks/changes to column order after the first click works how you would expect it to.

`"@carbon/ibm-products": "2.1.2"`

## Steps to reproduce the issue

1. Pass an `onSort` function to the DataGrid state
```
useDataGrid({ ...state, onSort: () => {} }, ...plugins, useSortableColumns)
```
2. Go to your DataGrid and try to sort a column
3. Depending on your function, it will have been called, but the column state will be the same
4. Clicking the column to sort again will call the `onSort` function again with the same parameters, and will properly set the columns sort state

Here's an example of the issue
https://codesandbox.io/p/sandbox/tender-northcutt-5lw2vc?file=%2Fsrc%2FExample%2FExample.jsx%3A9%2C24

## Additional information

- Screenshots or code
- Notes

davidmenendez:
hey @cgerein i tried to recreate this, but in your example and locally i found that adding `onSort` actually prevents the sort from firing at all. the problem is that when you pass `onSort` you're signifying that you will be handling sorting the column yourself. so when you're callback is just `(a, b) => console.log(a, b)` nothing happens because no sorting has actually taken place.

can you clarify what your expected behavior is?

elycheea:
Maybe be worth pointing the documentation to `react-table` here in name of separating business and component logic.

cgerein:
Hey, thanks for taking a look. I'm still seeing the issue with it in that codesandbox in the description. Essentially what I was seeing there and locally was that the first click you did to sort a column would be called with ASC, but the column state wouldn't change. By that I meant that the column header wouldn't update the sort state.

Once you clicked the header a second time it shows the ascending icon properly, but is also still passing ASC to the `onSort` function. After that second click the column was sorting, or calling the `onSort` with the correct parameters.

My expected behaviour would be that clicking twice on a column header doesn't pass `ASC` to the `onSort` twice.

[...]

PR #3425, opened 2023-08-29
fix: datagrid onsort

davidmenendez:
Contributes to #3197


davidmenendez:
user clarified the behavior. will open a follow up PR when a solution is found

PR #3436, opened 2023-08-31
feat: Modal portalTarget

### LINKED ISSUE #3417: Modal dialogs (About, Create, Edit etc) should have a portal target or be aware of containing CSS

lee-chase:
Fixes #3417 
Modal dialogs were relying on the CSS of the elements they were hosted in. This moves them by default to the body element. 

It can be turned off by setting `default-portal-target-body` to false.

The user can also specify directly `portalTarget` as a property on each component.

#### What did you change?

- Added `usePortalTarget` hook and adjusted all components based on modal dialogs.
- Added note to Readme on how to list flags

Storybook checking the modals worked in the docs page.

NOTE: We should probably add better documentation for flags.


davidmenendez (dismissed):

packages/ibm-products/src/components/APIKeyModal/APIKeyModal.test.js
```diff
@@ -92,21 +92,24 @@ describe(componentName, () => {
...
     await act(() =>
-      click(container.querySelector(`.${carbon.prefix}--btn--primary`))
+      click(modal.querySelector(`.${carbon.prefix}--btn--primary`))
     );
     expect(navigator.clipboard.writeText).toHaveBeenCalledWith(props.apiKey);
     getByLabelText(defaultProps.copyIconDescription);
   });
 
-  it('renders with standard setup', async () => {
+  it.only('renders with standard setup', async () => {
```
matthewgallo, line 112:
```suggestion
  it('renders with standard setup', async () => {
```

PR #3448, opened 2023-09-06
ci(dco): add dco workflow

tay1orjones:
#### What did you change?

This adds the DCO workflow. I also added the DCO token to the repo secrets.

There's not a good way to test actions until they're in main 😕


elycheea (approved):
Thank you 🎉

Issue #2624, opened 2023-01-24
Add intitalStep prop to CreateFullPage Component


ErickBarreraIBM:
## What will this achieve?
This will allow the CreateFullPage component to be opened on a step other than the first step.  Currently if there is a long running task on one of the downstream steps and user refreshes browser or another user connects and loads component they will always be taken to the first step.  This change will allow for the user to be taken to the current "in-progress" step directly.  This is helpful because sometimes we want to lock out user from making any changes in the upstream steps while there is a long running task currently executing.

The CreateTearsheet component already supports this.  Can we please implement a similar pattern in the CreateFullPage component.

This is the property that allows this functionality in the CreateTearsheet component:
https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/f91a72b3d95231f9922efb221734248b45254fa5/packages/cloud-cognitive/src/components/CreateTearsheet/CreateTearsheet.js#L272

## How will success be measured?
Success will be the CreateFullPage component supporting a new property `initialStep` which enables the component to be loaded at a specific step.

## Additional information

The CreateTearsheet component already supports this functionality and has an `initialStep` Property.

swwebster:
In the meantime, we have had to implement a funky UX workaround where we disable upstream inputs and direct the user to click "Next" until they get back to the step with the process in-progress. I can provide a screenshot/mockup of that if desired.

swwebster:
This is still relevant from my perspective. @ErickBarreraIBM I assume this hasn't been fixed on your end?

elycheea:
@swwebster Reopened. We’ve been wrapped up with v2 release and the design jam over the past few weeks. Going to tag this as both v1 and v2.

elycheea:
@jagathgj assigning to you for now but let me know if you aren’t able to look into this in the next two weeks.

swwebster:
Awesome! Thanks @elycheea :)

jagathgj:
Issue fixed hence closing

ErickBarreraIBM:
Thanks everyone!

PR #3453, opened 2023-09-08
feat(createfullpage): initialstep added  for v1

jagathgj:
Contributes to #2624 

InitialStep prop added to CreateFullPage component

#### What did you change?
Prop added, useEffect updated, utilities added

storybook, jest


lee-chase (approved):
Approving as per v2

Issue #3446, opened 2023-09-06
After we update to version 2.8.1, ES module Error happens 


SeonyuK:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.
After updating to version 2.8.0 this error happens
```
Error: require() of ES Module /Users/jeremy/Workspace/ui-management/node_modules/react-dnd/dist/index.js from /Users/jeremy/Workspace/ui-management/node_modules/@carbon/ibm-products/lib/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js not supported.
Instead change the require of index.js in /Users/jeremy/Workspace/ui-management/node_modules/@carbon/ibm-products/lib/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js to a dynamic import() which is available in all CommonJS modules.

```
> Is this issue related to a specific component?
no
> What did you expect to happen? What happened instead? What would you like to
> see changed?
not having the error
> What browser are you working in?

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
2.8.0
> package are you using?

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-products/tree/main/examples/carbon-for-ibm-products/SidePanel

## Additional information

- Screenshots or code
![Screenshot 2023-09-06 at 12 04 18 PM](https://github.com/carbon-design-system/ibm-products/assets/144151347/5bf58523-fb76-44f4-abe8-195146209d96)

- Notes

elycheea:
So it looks like the latest update of [`react-dnd`](https://github.com/react-dnd/react-dnd/releases/tag/v16.0.0) is now pure ESM. I do think we need to be on v16, which is why we update the package when we moved to React 18.

It looks like a fix to continue CJS support was merged but never published, and it’s been some time.

Without this fix, we may need to evaluate alternatives, which could push our stable releases back, or consider dropping support for CJS. 😞

lee-chase:
[...]

PR #3458, opened 2023-09-11
fix(build): temporarily remove drag support in customize cols

matthewgallo:
Contributes to https://github.com/carbon-design-system/ibm-products/issues/3446

Because of the known issues discovered from the latest version of `react-dnd` not supporting commonjs support, we need to remove `react-dnd` from our library. This PR removes `react-dnd` and the dragging functionality of the customize columns Datagrid extension, until we find a solution that has both ESM/commonjs.

#### What did you change?
```
.yarn/cache/hoist-non-react-statics-npm-3.3.2-e7b709e6c1-b153827042.zip
packages/ibm-products/package.json
packages/ibm-products/src/components/Datagrid/Datagrid/DraggableElement.js
yarn.lock
```
Verified customize columns story still loads but does not include dragging behavior and tested that `yarn build` successfully runs


elycheea (commented):
Thanks for the quick turnaround. Might be worth adding a notification on the story though that the extension is being reworked after 2.8.1.

matthewgallo:
Add this to both of the customize columns stories.

image

elycheea (approved):

Issue #3241, opened 2023-07-18
Improve keyboard actions in Datagrid toolbar


elycheea:
### Priority fixes
- [x] Tabbing to previous or next item from settings popover should most likely close popover in toolbar
- [x] Popover should close on Escape
  - Carbon uses `onKeyDown` in Popover Storybook example

### Enhancements
These are good to fix if simple, but are lower priority given that they most likely have a workaround and can be addressed after initial release so that we can prioritize bigger fixes.

- [ ] When opening the popover with enter/space, focus should go to the first item in the popover list (radio)
  - Note: There is a workaround — users have an extra tab / up/down to navigate into the popover options
- [ ] Using arrow keys to navigate radio items should not scroll the full table

--

> 2. When you are able to correctly open the dropdown, you must press tab again - and the faint 1px blue focus state appears around settings, - then press tab a second time to enter the list before you are able to press up/down/L/R to navigate the list.   Focus should be present in the first list item after pressing space/enter to activate the popover.  Also as you press up/down arrows to cycle through the radio button options, the table itself moves up and down behind it.   
> 4. When popover is open and you tab to settings, it closes. However, if popover is open and you shift+tab to Download before interacting with the popover content, the popover remains open and overlaps with the Download tooltip. I’m thinking it should close so that what happens with shift+tab is consistent with what happens with tab.  
![Screenshot 2023-07-03 at 8 16 05 AM](https://github.com/carbon-design-system/ibm-products/assets/21047571/4b84562a-1973-40e7-8f5b-2ce2123d4fa5)
> 
> 5.  When open, the popover should dismiss when Esc is pressed 

_Originally posted by @kristastarr in https://github.com/carbon-design-system/ibm-products/issues/3117#issuecomment-1618294932_

PR #3460, opened 2023-09-11
fix: datagrid toolbar accessibility

davidmenendez:
Contributes to #3241


davidmenendez:

davidmenendez:
recheck

elycheea (approved):
Looks great! I’m guessing we can’t just cherry-pick these to v1 since `IconButton` and `Popover` are v11 only though. 👀 

I’ll also break the enhancement portion of the issue into a separate issue so we can close the issue once v1 is merged.

Issue #3397, opened 2023-08-23
Datagrid filter flyout click fires twice in Storybook


elycheea:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

When clicking on the [filter flyout](https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-flyout--flyout-batch), noticed that the flyout no longer opens. Not sure if this started earlier, but noticed this after the React 18 merge.

Note that after clicking once, you see both onFlyoutOpen and onFlyoutClose were fired in actions.

<img width="1468" alt="image" src="https://github.com/carbon-design-system/ibm-products/assets/1717398/fb7247d6-1b9f-468f-9d09-565b1916c32f">

> Is this issue related to a specific component?

Datagrid filter flyout

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Flyout should open ... (`onFlyoutOpen`) but not close (`onFlyoutClose`) at least for one click.

> What browser are you working in?

Chrome 115

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

v2 — this is not observed in v1

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Navigate to [filter flyout story](https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-flyout--flyout-batch)
2. Click on flyout menu/icon

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-products/tree/main/examples/carbon-for-ibm-products/SidePanel

## Additional information

- Screenshots or code
- Notes

PR #3473, opened 2023-09-13
fix: filter flyout outside click

davidmenendez:
Contributes to #3397 

clicking the flyout button was causing the outside click detector to be called. this refactors the `useOutsideClick` hook to provide the ref and fixes the aforementioned bug by applying the hook correctly to the container.


davidmenendez:
upon closer inspection the only change that really needed to be made was to move the ref higher up and make a small adjustment to the hook

PR #3479, opened 2023-09-15
fix: datagrid row size accessibility v1

davidmenendez:
Contributes to #3241


elycheea (changes requested):
Keyboard looks good but hopefully minor change request — 

Row size dropdown should align right as the other toolbar items do.
image
 
We removed `getDropdownPosition()`, but I think we can probably use CSS to accomplish this?

davidmenendez:
@elycheea updated

elycheea (approved):

Issue #3485, opened 2023-09-20
Datagrid pagination release review fixes


elycheea:
* (DatagridPagination && datagridState.page) || datagridState.rows; should define page and rows in proptypes. also deconstruct
* pagination requires additional string props?
  * `DatagridPagination.js` as an example, but should we also add in `...rest` or specify props for i18n support?
* remove console.log
* copyright
* add test for `updatePagination`

_Originally posted by @davidmenendez in https://github.com/carbon-design-system/ibm-products/issues/3245#issuecomment-1728299331_

elycheea:
@davidmenendez Don’t forget to add the v1 PR for pagination review fixes.

PR #3492, opened 2023-09-21
fix: datagrid pagination review fixes

davidmenendez:
Contributes to #3485 

i chose to omit the console log and testing points because after thinking about it more i think it's ok to leave those since it's for storybook specifically. if anything going over this review has solidified my desire to reorganize these files and folders so it's much more clear what is code built into the component and which isn't for storybook purposes or otherwise.


paul-balchin-ibm:
`DatagridPagination.js`, line 14 has a `console.log`...

elycheea (commented):
Fair points. I think I’m okay with the log for Storybook. I think for `updatePagination()`, I don’t remember if there was a specific scenario for which we wanted to add the test. If @paul-balchin-ibm remembers, I think we should include it. If not, then let’s do another test for coverage later.

packages/ibm-products/src/components/Datagrid/utils/DatagridPagination.js
```diff
@@ -18,6 +18,7 @@ export const DatagridPagination = ({ state, setPageSize, gotoPage, rows }) => {
 
   return (
     <Pagination
+      {...state}
```
matthewgallo, line 21: This is likely not what you meant to do because this attaches all of the datagrid state to the Carbon `Pagination` component which ends up applying most of it as attributes, see below:

image

matthewgallo (changes requested):
See comment about passing `{...state}` to `Pagination` component.

packages/ibm-products/src/components/Datagrid/utils/DatagridPagination.js
davidmenendez: ohh interesting. thanks for the heads up!

matthewgallo (approved):

matthewgallo (approved):

Issue #2978, opened 2023-05-10
Tearsheet: invalid aria-label


wkeese:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

The `Tearsheet#title` property is a ReactNode:

```
  /**
   * The main title of the tearsheet, displayed in the header area.
   */
  title: PropTypes.node
```

But TearsheetShell.js assigns it to `aria-label`.

So when you set `title` to something like `<div>...</div>` it ends up with the obviously invalid:

```
<div aria-label="[object Object]">
```

(It's invalid because the label needs to be human-readable text that describes the dialog.)

There should be some way to specify the `aria-label` or (better yet) `aria-labelledby` for the dialog.

elycheea:
Looks like we’re overriding the arial labels from ComposedModal. We can look into this!

PR #3496, opened 2023-09-22
fix: invalid aria-label

paul-balchin-ibm:
Contributes to #2978

Issue: 
- assigning a node to aria-label like `aria-label={<div>Hello <b>there</b>.</span>}` in React, 
- renders `aria-label="[object Object]"` in HTML.

#### What did you change?
- new function converts node text to string, `aria-label={getNodeTextContent(the same node)}`,
- renders `aria-label="Hello there."`.

- Tested locally.
- Included a test file `getNodeTextContent.test.js`.


paul-balchin-ibm:

packages/ibm-products/src/global/js/utils/getNodeTextContent.test.js
```diff
@@ -0,0 +1,52 @@
+//
+// Copyright IBM Corp. 2021, 2021
```
elycheea, line 2:
```suggestion
// Copyright IBM Corp. 2023, 2023
```

elycheea (commented):
Ooh, interesting approach that does allow us not to add a potentially duplicative prop. 🤔 Separate `aria-label` prop could come up again at some point, but kind of like this approach too. Good to go other than fixing some copyright dates. :P

paul-balchin-ibm:
image

paul-balchin-ibm:
> Separate `aria-label` prop could come up again at some point...

I'm pretty sure there are. I've also seen other console warnings/errors, when running Storybook as `localhost` such as these just for http://localhost:3000/?path=/story/ibm-products-components-datagrid-datagrid-canary--empty-state
- "Received `false` for a non-boolean attribute `role`"
- "React does not recognize the `renderIcon` prop on a DOM element"
- "React does not recognize the `placeHolderText` prop on a DOM element"
- "Invalid argument supplied to oneOfType, expected an instance of array"
- "Invalid arguments supplied to oneOf, expected an array, got 4 arguments"

I think it would be a good exercise to go through the stories and quash these issues. As a an app dev, I don't want to have filter my own errors from those generated by a library's components.

elycheea (approved):

wkeese:
Hi @paul-balchin-ibm - You've fixed this situation:

```
<Tearsheet
title={<div>bill was here</div>}
```

but when there's a level of indirection it doesn't work:

```
  const Title = () => <div>bill was here</div>;
...
<Tearsheet
title={<Title/>}
```

In that case there's no aria-label for the role=dialog node at all.

wkeese:
I've filed https://github.com/carbon-design-system/ibm-products/issues/4013, which addresses having a separate aria-label prop.  That seems like the final solution to this problem.

Issue #3501, opened 2023-09-25
Fix gallery check action


elycheea:
## What will this achieve?

- [ ] Fix the failed check in under `check gallery`
- [ ] Move `check gallery` to a separate step from `validate title`

![Image](https://github.com/carbon-design-system/ibm-products/assets/1717398/b6bf023e-399a-4a47-a6ae-da40aec799b9)

## How will success be measured?

- `Check gallery` is a separate step from `validate title`
- `Check gallery` passes when gallery up-to-date

## Additional information

- Designs
- Existing code
- etc

PR #3502, opened 2023-09-25
chore: update gallery check

elycheea:
Attempt to address #3501 
@lee-chase thoughts on whether `check_gallery` should actually go into CI check or if there’s any reason why these are two different workflows? 🤔 

#### What did you change?
Moved `check_gallery` to a separate step from `validate_title` and added checkout/node steps (same approach that CI.yml uses).

🙏


lee-chase (approved):

elycheea:
@lee-chase gallery check seems to work now 😹 Check just failed so gallery updated now 😅

Issue #3470, opened 2023-09-12
Allow tooltip alignment to be set on row action buttons


elycheea:
## What will this achieve?

Alleviates an issue where hovering over a row action causes a longer/scrolling Datagrid to shift. (May be more of an issue where scrollbars are not set to always show. We can allow product teams to set the tooltip alignment so that this reduces the likelihood of the datagrid shifting as scrollbars appear/disappear.

https://react.carbondesignsystem.com/?path=/docs/components-overflowmenu--overview#customizing-the-tooltip

## How will success be measured?

Product teams can set `align` prop to change tooltip position/alignment.

## Additional information

- Designs
- Existing code
- etc

SeonyuK:
@elycheea @paul-balchin-ibm If there are more than two overflow menus, it should be `left` instead of `top-right` since the overflow menu is always at the end of the row. `top-right` still causes the same issue.

paul-balchin-ibm:
Thanks @SeonyuK. I'll reopen this issue and get to this ASAP.

paul-balchin-ibm:
Merged. A new update should be available next Tuesday.

PR #3503, opened 2023-09-26
fix(DataGrid): set row action button tooltip alignment

paul-balchin-ibm:
Contributes to #3470

Allow developer to set the tooltip alignment of a row action button.

#### What did you change?

The developer can now pass a `align` property to a `rowActions` object. E.g. (see last line in code sample below):

```
rowActions: [
  {
    id: 'delete',
    itemText: 'Delete',
    icon: TrashCan,
    isDelete: true,
    onClick: action('Clicked row action: delete'),
    align: 'top-right', // <-- add this
  },
],
```

The same `align` values as Tooltip and Popover options are available.

- [x] Storybook

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/useActionsColumn.js
+++ b/packages/ibm-products/src/components/Datagrid/useActionsColumn.js
@@ -55,6 +55,7 @@ const useActionsColumn = (hooks) => {
                     >
                       {rowActions.map((action, index) => {
                         const {
+                          align,
                           id,
                           itemText,
                           onClick,
@@ -81,8 +82,8 @@ const useActionsColumn = (hooks) => {
                           >
                             <OverflowMenu
                               {...rest}
+                              align={align || 'top'}
                               renderIcon={icon}
-                              hasIconOnly
                               light
                               iconDescription={itemText}
                               kind="ghost"
@@ -107,6 +108,7 @@ const useActionsColumn = (hooks) => {
                   {!isFetching && rowActions.length > 2 && (
                     <div>
                       <OverflowMenu
+                        align="top-right"
                         size="sm"
                         light
                         flipped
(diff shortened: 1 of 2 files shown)


packages/ibm-products/src/components/Datagrid/useActionsColumn.js
```diff
@@ -81,8 +82,8 @@ const useActionsColumn = (hooks) => {
                           >
                             <OverflowMenu
                               {...rest}
+                              align={align || 'top'}
                               renderIcon={icon}
-                              hasIconOnly
```
paul-balchin-ibm, line 85: Removed. This was the source of the problem where we found a button was rendering inside another button. This wasn't normally visible because the duplicate was always perfectly placed above the origina.

packages/ibm-products/src/components/Datagrid/useActionsColumn.js
```diff
@@ -107,6 +108,7 @@ const useActionsColumn = (hooks) => {
                   {!isFetching && rowActions.length > 2 && (
                     <div>
                       <OverflowMenu
+                        align="top-right"
```
paul-balchin-ibm, line 111: The `rowAction` `align` option only applies when rendering individual buttons (around line 85). 

There is no feature currently available to modify the OverflowMenu's trigger button ( [•••] ) itself. This is a temporary, hard-coded fix to duplicate the feature above.

packages/ibm-products/src/components/Datagrid/useActionsColumn.js
```diff
@@ -81,8 +82,8 @@ const useActionsColumn = (hooks) => {
                           >
                             <OverflowMenu
                               {...rest}
+                              align={align || 'top'}
```
paul-balchin-ibm, line 85: Apply a default `top` value if `align` is null.

elycheea (approved):

elycheea (approved):

Issue #3478, opened 2023-09-15
Investigate alternative drag and drop implementations


elycheea:
## What will this achieve?

Replaces `react-dnd` and restores drag and drop functionality in customize columns Datagrid extension.

## How will success be measured?

- [x] Both ES and CJS builds succeed
- [x] User can reorder columns via `CustomizeColumnsTearsheet`
- [x] Drag and drop follows the styles from the published [design guidance](https://pages.github.ibm.com/cdai-design/pal/patterns/drag-and-drop/usage)
- [x] User can reorder columns through keyboard actions
- [x] Tests are updated and pass
- [x] Determine whether to pursue a library or create our own implementation

## Additional information

- Designs
- Existing code
- etc

#3446

elycheea:
@matthewgallo has been looking into [`react-beautiful-dnd`](https://github.com/atlassian/react-beautiful-dnd). So far seems like a viable option minus the drop zone styling. May need to discuss this with @carrenelloyd or Tomer if we want to pursue this one.

lee-chase:
There are a number of options here, including creating our own. Looking at downloads it would appear that the big players are....

- react-draggable (keyboard - not in scope)
- react-beautiful-dnd (has keyboard options, maintenance not a priority) https://github.com/atlassian/react-beautiful-dnd#%EF%B8%8F-maintenance--support
- react-dnd (keyboard?? maintenance erratic and waiting on updates to be published) 
- dnd-kit (has keyboard access, created because react-beautiful-dnd is no longer maintained).

![image](https://github.com/carbon-design-system/ibm-products/assets/15086604/522bd9bb-7b96-4706-85a1-1aee20b7266f)

elycheea:
From scrum with @lee-chase and @paul-balchin-ibm — we will investigate `react-beautiful-dnd` or `dnd-kit` for the moment to unblock customize columns with the understanding that this could be replaced when we pursue #2042

PR #3513, opened 2023-09-28
feat: with dnd kit

lee-chase:
Contributes to https://github.com/carbon-design-system/ibm-products/issues/3478

Using dnd-kit.

Probably chopped out more than needed to make sense of this, need to understand if it is working and if we've lost any features.

This works really nice, fewer issues than react-beautiful-dnd and less code. Looks correct on scrolling compared to react-beautiful-dnd.

Only niggle(s)
- Keyboard movement is 25px at a time, not sure why as I think it should be `cell height / 2` reading the code, but I must be missing a rounded pixel somewhere. This is noticeable on longer lists.
- Not visible if dragged out of modal (to be expected).

What did you change?
Datagrid customisable columns

Storybook

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,9 +152,6 @@
     ]
   },
   "dependencies": {
-    "@dnd-kit/core": "^6.0.8",
-    "@dnd-kit/sortable": "^7.0.2",
-    "@dnd-kit/utilities": "^3.2.1",
     "stylelint-plugin-carbon-tokens": "2.3.1"
   },
   "packageManager": "yarn@3.6.1"
--- a/package.json
+++ b/package.json
@@ -152,7 +152,9 @@
     ]
   },
   "dependencies": {
-    "react-beautiful-dnd": "^13.1.1",
+    "@dnd-kit/core": "^6.0.8",
+    "@dnd-kit/sortable": "^7.0.2",
+    "@dnd-kit/utilities": "^3.2.1",
     "stylelint-plugin-carbon-tokens": "2.3.1"
   },
   "packageManager": "yarn@3.6.1"
--- a/package.json
+++ b/package.json
@@ -152,6 +152,7 @@
     ]
   },
   "dependencies": {
+    "react-beautiful-dnd": "^13.1.1",
     "stylelint-plugin-carbon-tokens": "2.3.1"
   },
   "packageManager": "yarn@3.6.1"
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -152,6 +152,9 @@
     ]
   },
   "dependencies": {
+    "@dnd-kit/core": "^6.0.8",
```
matthewgallo, line 155: These will need to go in `packages/ibm-products/package.json`

packages/ibm-products/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -38,38 +26,36 @@ const Columns = ({
...
+      setColumnsObject(
+        update(columns, {
+          $splice: [
+            [from, 1],
+            [to, 0, fromCol],
+          ],
+        })
+      );
+    },
+    [columns, setColumnsObject] // [columns, setColumnsObject]
```
matthewgallo, line 47: Don't think you need this comment

matthewgallo (changes requested):
As discussed over slack, I think there is some more work to do in coordinating the drag interaction and the checkbox. The checkbox is now clickable and can toggle between checked and unchecked as expected but if you click, hold, and drag from the checkbox the application breaks.

image

The option of moving the checkbox outside of the drag area completely seems like the best way forward both in terms of the behavior of these two interactions and the most accessible approach as well.

matthewgallo:
The styling of the cursor while dragging should also use `cursor: grabbing;`.

packages/ibm-products/src/components/Datagrid/Datagrid/addons/CustomizeColumns/DraggableItemsList.js
```diff
@@ -13,123 +13,219 @@ import DraggableElement from '../../DraggableElement';
...
+      `${colTitle} grabbed. Current position ${fromIndex + 1} of ${
+        visibleCols.length
+      }.`
+    );
+  };
+
+  const handleDragUpdate = (event) => {
+    const { active, over } = event;
+
+    const fromIndex = visibleCols.findIndex((col) => col.id === active.id);
```
matthewgallo, line 91: I believe it is one of these two lines that is causing an error to be thrown if dragging from the checkbox

matthewgallo (approved):
I believe this is ready now, we can open an issue to further investigate the checkbox keyboard support should we choose to keep it inside of the draggable area.

PR #3527, opened 2023-10-02
fix(Datagrid): refactor expandable implementation

matthewgallo:
Contributes to #3506

This PR refactors how the expandable area is rendered by the Datagrid. Previously the component utilized the `RowRenderer` to simply rerender the entire row plus the expandable area as well. This created issues with a11y because every time a row expanded, the entire row would rerender causing the row expander button to lose focus.

This approach now only renders the expansion area and not the entire row as well, allowing the focus to remain on the row expander button.

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Datagrid/DatagridExpandedRow.js
packages/ibm-products/src/components/Datagrid/Datagrid/DatagridRow.js
packages/ibm-products/src/components/Datagrid/useExpandedRow.js
```
Storybook

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid.test.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid.test.js
@@ -1487,22 +1487,9 @@ describe(componentName, () => {
         .textContent
     ).toEqual(`Content for ${rowNumber}`);
 
-    fireEvent.click(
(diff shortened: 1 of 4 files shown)


davidmenendez (changes requested):
same comments from the V2 PR

PR #3528, opened 2023-10-02
fix: datagrid empty state fixes

davidmenendez:
Contributes to #3243 

mostly just addresses some testing. to help facilitate testing I added specific class names to all the empty state variants, which honestly should have been done from the start haha

there was a point in the issue to mention how to set the error type, but i think there needs to be a bigger discussion about prop types and that documentation for datagrid. so i opted to skip it for now.

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid.test.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid.test.js
@@ -1079,24 +1079,22 @@ describe(componentName, () => {
     const { rerender } = render(<EmptyUsage data-testid={dataTestId} />);
     screen.getByText('Empty State Title');
     screen.getByText('Description test explaining why this card is empty.');
-    expect(
-      document.querySelector(`.${pkg.prefix}--empty-state-type--noData`)
-    ).toBeInTheDocument();
+    expect(screen.getByRole('img')).toHaveClass(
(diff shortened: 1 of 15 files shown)


packages/ibm-products/src/components/Datagrid/Datagrid.test.js
```diff
@@ -1080,55 +1076,27 @@ describe(componentName, () => {
...
   //Empty State
   it('renders an empty table', async () => {
-    render(<EmptyUsage data-testid={dataTestId}></EmptyUsage>);
+    const { rerender } = render(<EmptyUsage data-testid={dataTestId} />);
+    screen.getByText('Empty State Title');
+    screen.getByText('Description test explaining why this card is empty.');
     expect(
-      screen.getByRole('table').getElementsByTagName('tbody')[0].className
-    ).toEqual('c4p--datagrid__empty-state-body');
+      document.querySelector(`.${pkg.prefix}--empty-state-type--noData`)
```
matthewgallo, line 1083: We should avoid `document.querySelector` if possible

davidmenendez: i'll take another look

davidmenendez:
@matthewgallo fixed the tests by adding a `role` to the empty state svgs

matthewgallo (approved):

PR #3359, opened 2023-08-04
docs: add catalog-info.yaml for Backstage

dbrugger:
Contributes to #

Adding a `catalog-info.yaml` for [Backstage](https://backstage.io) to allow us (and others who are interested to do this) to include `@carbon/ibm-products` in our dev portal.

Sibling PR: https://github.com/carbon-design-system/carbon/pull/14393

#### What did you change?

- Added `catalog-info.yaml`
- Moved `README.md` into the docs directory which is required for `techdocs-core` / `mkdocs` to include it. GitHub will still happily show.

#### How did you test and verify your work?

- Tested integration from PR branch
- Preview docs with `npx @techdocs/cli serve`
- Checked that the `README.md` relocation does not cause broken links

PR #3539, opened 2023-10-03
chore: add backstage config updates

matthewgallo:
Replaces https://github.com/carbon-design-system/ibm-products/pull/3359

Accidentally merged some commits into https://github.com/carbon-design-system/ibm-products/pull/3359 so I'm opening this PR with the changes it previously had to add the proper config for backstage.


davidmenendez (dismissed):

matthewgallo:
@dbrugger It looks like your backstage config PR in Carbon only included `catalog-info.yaml`, is that something we can do here as well?

Issue #3324, opened 2023-07-25
Expandable table row: Hover state is missing in expandable details sections


jimmydyer:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

When a user expands a bunch of rows in the Saved Searches table, it's very hard to distinguish where a row starts and stops. The only section that changes color on mouse hover is the parent row. The details when the row is expanded do not change.

> Is this issue related to a specific component?

Yes. Datagrid expandable rows

> What did you expect to happen? What happened instead? 

- When a row is expanded, the entire row and expanded details should have a hover state (color change). Refer to the Batch expandable table component on Carbon's Storybook: https://react.carbondesignsystem.com/?path=/story/components-datatable-expansion--batch-expansion
- Expanded details of a table row do not have a hover state (Carbon compliance and accessibility issue)
- The expanded batch behavior in the Table header area is missing

> What browser are you working in?

Chrome

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

v10

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

IBM Security.
This is already GA.

## Additional information

1. The first image reflects the Datagrid ExpandableRow extension: https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-components-datagrid-datagrid-canary-extensions-expandablerow--docs
2. The second image reflects the component in Carbon: https://carbondesignsystem.com/components/data-table/usage/

![Screenshot 2023-07-24 at 4 00 34 PM](https://github.com/carbon-design-system/ibm-products/assets/54411478/48a53126-02ca-4b7f-87b8-979d9e753b6c)
![Screenshot 2023-07-24 at 4 01 04 PM](https://github.com/carbon-design-system/ibm-products/assets/54411478/2193023b-36c3-4a2f-bd1a-1ec206d377cd)

elycheea:
@davidmenendez Assigning this to you since you’ll be looking at similar code through #3240 on v2 (I think it applies to expanded rows too). Want to note the [discussion we had in Slack](https://ibm-casdesign.slack.com/archives/CQGR0HC05/p1690292165583689) though —

- Expanded rows should show hover state in expanded details
- Nested rows should retain current style (indicator instead of hover on expanded sections).

PR #3550, opened 2023-10-05
fix(Datagrid): expandable row add background on row hover

matthewgallo:
Contributes to #3324 

Adds background color to expanded area when hovering on the expanded row.

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/styles/_useExpandedRow.scss
```
Storybook

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridExpandedRow.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridExpandedRow.js
@@ -15,8 +15,28 @@ const DatagridExpandedRow =
   (ExpandedRowContentComponent) => (datagridState) => {
     const { row } = datagridState;
     const { expandedContentHeight } = row || {};
+
+    const toggleParentHoverClass = (event, eventType) => {
+      if (
+        event &&
+        event.target &&
+        event.target.parentNode.previousElementSibling
+      ) {
+        const parentNode = event.target.parentNode.previousElementSibling;
+        if (eventType === 'enter') {
+          parentNode.classList.add(`${blockClass}__expandable-row--hover`);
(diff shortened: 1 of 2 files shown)


elycheea (changes requested):
If we’re following the row hover behavior from Carbon, I think hover over the expanded area should also show the hover state. Currently only seeing the hover state when hovering over the row itself but not the expanded content.

matthewgallo:
Ah ok, I must have misread the issue. I'll change this ^

matthewgallo:
Ended up taking a similar approach to Carbon for highlighting the parent row of the expanded area, couldn't get `:has(+ .expanded-class)` previous sibling selector to work here.

packages/ibm-products/src/components/Datagrid/Datagrid/DatagridExpandedRow.js
```diff
@@ -15,8 +15,28 @@ const DatagridExpandedRow =
   (ExpandedRowContentComponent) => (datagridState) => {
     const { row } = datagridState;
     const { expandedContentHeight } = row || {};
+
+    const toggleParentHoverClass = (event, eventType) => {
+      if (
+        event &&
+        event.target &&
+        event.target.parentNode.previousElementSibling
```
davidmenendez, line 23: i'd do something like this to make this a little less painful
```
const parentNode = event?.target?.parentNode?.previousElementSibling;
if (parentNode)
...
```

davidmenendez (changes requested):
looks good. just one small comment.

davidmenendez (approved):

davidmenendez (approved):

elycheea (approved):

PR #3558, opened 2023-10-09
fix: Should our CSS use of carbon rem be to-rem

lee-chase:
Moves away from use of `rem` function in favour of `to-rem` the former being a problem for current versions of SASS.


elycheea (approved):

steven1046:
Is there a recommended sass version to downgrade to to mitigate the rem error? I am on 1.47.0

PR #3560, opened 2023-10-09
feat(Datagrid): adds view all feature for checkboxes

AlexanderMelox:
### Adding this feature to add the ability to display multiple checkboxes under a view all button. 

Solution for #2714

If Checkboxes exceed 10 then only show 5 and show the rest after clicking View all

Screenshot 2023-10-09 at 2 44 40 PM

https://github.com/carbon-design-system/ibm-products/assets/12755042/2dcf4bac-8d25-4453-a46e-076d0bc1cccd

Talked to **Shelley Dignum** about the change, and this is a good first step. The true behavior is that is should open a multi add/select.

From Shelley:
Screenshot 2023-10-09 at 2 48 57 PM

#### What did you change?
I changed the render method of checkboxes inside `useFilters`. This checks if the checkboxes are larger than 10, then only display 5 checkboxes and hide the rest collapsed.

Storybook


sdignum:
@AlexanderMelox Three things:

1. The ghost button should sit 16px to the left so that the text of the button aligns with the text above it. 
2. The expanded button label should be "View less" instead of hide all. 
3. Please use the small ghost button (looks like it's a medium in your video)

sdignum:
@AlexanderMelox Apologies on not using clearer language....align the text of the button with the content above it, in this case the checkboxes. So it should be flush with the edge of the filter panel. As seen in the guidance: 
image

AlexanderMelox:
@sdignum Oh okay! When we flush the button to the left, the hover will also be flushed. Just wondering if thats okay. 
Screenshot 2023-10-09 at 4 26 08 PM

An example

sdignum:
@AlexanderMelox yes, that is fine. I know it's a little weird, but it's a general way of dealing with ghost buttons when they're aligned left under other content.  Thanks! And just a reminder or making you sure you saw my note about the "view less" button not needing the number with it.

AlexanderMelox:
> @AlexanderMelox yes, that is fine. I know it's a little weird, but it's a general way of dealing with ghost buttons when they're aligned left under other content. Thanks! And just a reminder or making you sure you saw my note about the "view less" button not needing the number with it.

Yup I saw it, so align left and when View less is visible no number 👍🏻

packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/FilterPanel.js
```diff
@@ -7,30 +7,31 @@
...
+import { FilterContext } from '.';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import { motion } from 'framer-motion';
+import { pkg } from '../../../../../settings';
+import { rem } from '@carbon/layout';
+
 const blockClass = `${pkg.prefix}--datagrid`;
-const componentClass = `${blockClass}-filter-panel`;
+export const componentClass = `${blockClass}-filter-panel`;
```
AlexanderMelox, line 34: All I did was export this constant to be used in the `OverflowCheckboxes.js`. The rest of the changes was the linter reordering things

packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/hooks/useFilters.js
```diff
@@ -176,6 +179,53 @@ const useFilters = ({
...
+      }
+
+      return (
+        <FormGroup {...components.FormGroup}>
+          {filtersState[column].value.map((option) => (
+            <Checkbox
+              key={option.id}
+              {...option}
+              onChange={(isSelected) => {
+                const checkboxCopy = filtersState[column].value;
```
matthewgallo, line 203: The contents of this `onChange` look to be the exact same as the `onChange` in the [new component](https://github.com/carbon-design-system/ibm-products/pull/3560/files#diff-6ec7ccb1e61376029671970528fcd395601380c7d16af759baa69ba992d7c9a6R27-R44), `OverflowCheckboxes.js`. Can we put this in a utility that can be used from both checkboxes?

AlexanderMelox: Good call

packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/OverflowCheckboxes.js
```diff
@@ -0,0 +1,78 @@
...
+  const [showAllItems, setShowAllItems] = useState(false);
+  const firstFiveItems = filtersState[column].value.slice(0, 5);
+  const restOfTheItems = filtersState[column].value.slice(
+    5,
+    filtersState[column].value.length - 1
+  );
+
+  const renderCheckbox = (option) => (
+    <Checkbox
+      key={option.labelText}
```
matthewgallo, line 24: This is causing a bunch of console errors because all of the `dummyCheckboxes` have the same `labelText` which is causing them to all have the same key instead of a unique key.

packages/ibm-products/src/components/Datagrid/Extensions/Filtering/Panel.stories.js
```diff
@@ -40,6 +41,16 @@ export default {
   },
 };
 
+// This is to show off the View all button in checkboxes
+const dummyCheckboxes = Array(25)
+  .fill(null)
+  .map((_, index) => ({
+    id: index,
```
matthewgallo, line 48: Since you're spreading these properties onto the `Checkbox` component, you end up with a prop type error here because the `id` for `Checkbox` should be `string`, but here you're setting it to a number via the `index`.

packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/OverflowCheckboxes.js
```diff
@@ -0,0 +1,78 @@
...
+      {firstFiveItems.map(renderCheckbox)}
+      {showAllItems && restOfTheItems.map(renderCheckbox)}
+      <Button
+        className={`${componentClass}__view-all-button`}
+        kind="ghost"
+        size="sm"
+        onClick={() => setShowAllItems(!showAllItems)}
+      >
+        {showAllItems
+          ? 'View less'
```
matthewgallo, line 62: These are going to need to be passed in so we can support translation, I would give them a default value though so it defaults to `View less` and `View all`.

packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/OverflowCheckboxes.js
AlexanderMelox: Fixed it by making the label text unique

packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/OverflowCheckboxes.js
AlexanderMelox: Fixed 🔧

sdignum:
> @sdignum you can check the latest changes here https://deploy-preview-3560--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-panel--panel-many-checkboxes

@AlexanderMelox Apologies for missing the Github email. I reviewed and it looks good (not sure why all the checkboxes are disabled but assuming that would not be the case for normal experience I'm good with it).

AlexanderMelox:
> > @sdignum you can check the latest changes here https://deploy-preview-3560--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-panel--panel-many-checkboxes
> 
> @AlexanderMelox Apologies for missing the Github email. I reviewed and it looks good (not sure why all the checkboxes are disabled but assuming that would not be the case for normal experience I'm good with it).

The checkboxes are disabled because they are dummy checkboxes that don't do anything so I don't want the code to break because of it. It's there to show the usage

matthewgallo (approved):

Issue #3548, opened 2023-10-05
Fix vertical alignment in larger row heights


elycheea:
When using larger row heights, such as `xl`, text should be aligned to the top.

[Storybook v2](https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-rowheightsettings--default-settings-usage-story) / [v1](https://v1-ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-rowheightsettings--default-settings-usage-story)

---

According to Carbon, the data in the 'Extra large height' cell aligns to the top vertically.

<img width="817" alt="image" src="https://github.com/carbon-design-system/ibm-products/assets/40870771/7601a1bd-c6d5-4948-a1c1-ff59f08920ea">
  
I see that data is center-aligned vertically in the storybook:
<img width="671" alt="image" src="https://github.com/carbon-design-system/ibm-products/assets/40870771/1b9c6288-5085-41fd-ae0a-3f1be41fce10">

_Originally posted by @thefirstartist in https://github.com/carbon-design-system/ibm-products/issues/3508#issuecomment-1747530116_

nikhilgarg26:
hey @elycheea i am new to open source can you guide me on this
and how to reproduce that issue

elycheea:
Hi @nikhilgarg26! I added some additional links to the Storybook. I believe the reference made by the person reporting the issue was from this [story in Carbon](https://react.carbondesignsystem.com/?path=/story/components-datatable-basic--xl-with-two-lines).

nikhilgarg26:
as of i understand the issue i need to make [this ](https://react.carbondesignsystem.com/?path=/story/components-datatable-basic--xl-with-two-lines) vertically center aligned 

![image](https://github.com/carbon-design-system/ibm-products/assets/116938919/b40f90b1-2b7c-417d-937e-5677055d07a7)

nikhilgarg26:
can we you provide the file where the cshnges need to be made

Tanush-J:
I will give this a try

Tanush-J:
So, I was able to produce this behavior,

https://github.com/carbon-design-system/ibm-products/assets/75922615/a47c90a3-628a-4beb-aee2-7c3a5ba80342

is this the behavior expected?

matthewgallo:
Completed in #3709 and #3718

PR #3568, opened 2023-10-10
fix: vertical alignment in larger row heights

Tanush-J:
Contributes to #3548

After this content is top-aligned in `Extra large` row height.

#### What did you change?

Changed the default value of `verticalAlign` property from `'center'` to `'top'` in DatagridContent.js file.

https://github.com/carbon-design-system/ibm-products/assets/75922615/20d660ed-bf2b-40e6-a118-411d02064281

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridContent.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridContent.js
@@ -39,7 +39,7 @@ export const DatagridContent = ({ datagridState, title }) => {
     CustomizeColumnsTearsheet,
     filterProps,
     fullHeightDatagrid,
-    verticalAlign = 'center',
+    verticalAlign = 'top',
     variableRowHeight,
     gridTitle,
     gridDescription,


Tanush-J:

matthewgallo (changes requested):
Thanks for looking at this @Tanush-J! There may be some missing styles, if you look at one of the stories for the Datagrid that includes selectable rows, you'll notice that the checkbox and the row content do not align.

See issue below:
image

You can use the [Batch action story](https://deploy-preview-3568--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary--batch-actions) as an example to view this behavior.

Instead of changing `verticalAlign` default value to `top`, I think it would be better to do this by targeting the xl class, something like:
```scss
.cds--data-table--xl .c4p--datagrid__head,
.cds--data-table--xl .c4p--datagrid__carbon-row {
  align-items: flex-start;
}
```

This could be added to `_RowSizeDropdown.scss`, replacing `cds--data-table` and `c4p--datagrid` with the appropriate sass variables.

matthewgallo:
Yes, you're correct. There will be some more styling changes required given the different variations of this component.

matthewgallo:
Hey @Tanush-J, we had to address this issue in a separate PR. We'd welcome future contributions!

Issue #3455, opened 2023-09-11
Datagrid[ColumnCustomization]: unexpected reordering near locked column


matthewgallo:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

It appears that reordering a column that is near a column that is locked, causes unexpected reordering from the column previous to the dragged column or possible the column after the locked column.
> Is this issue related to a specific component?

Datagrid, column customization
> What did you expect to happen? What happened instead? What would you like to
> see changed?

Expected only the dragged column to be reordered.
> What browser are you working in?

Chrome
> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

Latest
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

N/A

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-products/tree/main/examples/carbon-for-ibm-products/SidePanel

## Additional information

- Screenshots or code
- Notes

https://github.com/carbon-design-system/ibm-products/assets/10215203/e1f2350e-045e-4182-b9e0-325d7a5fb73c

SeonyuK:
@matthewgallo  
After we got the changes, I found a different issue.
Since the columns still include `expender` and  'spacer' fromVisibleIndex, it is different from updatedDragCols than from columns. This causes `can not read properties of undefined `error.

It will be easy to reproduce when dragging unselected ones.


![Screenshot 2023-11-08 at 11 00 37 AM](https://github.com/carbon-design-system/ibm-products/assets/144151347/5e8b1cd3-1108-4d56-a1aa-91b3bffd8abd)
![Screenshot 2023-11-08 at 11 01 40 AM](https://github.com/carbon-design-system/ibm-products/assets/144151347/40125729-c800-4767-9efd-5dc7e0f8a0de)
![Screenshot 2023-11-08 at 11 01 29 AM](https://github.com/carbon-design-system/ibm-products/assets/144151347/3e61586e-1a2b-4ef6-8b15-d1b6e62b9c4f)
[...]

PR #3598, opened 2023-10-13
fix(Datagrid): add disabled/locked col drag functionality back

matthewgallo:
Contributes to #3455 

Adds back the functionality to:
1. Prevent dragging of columns that are disabled
2. Prevent items to be dropped/reordered to positions that are disabled

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Datagrid/DraggableElement.js
packages/ibm-products/src/components/Datagrid/Extensions/ColumnCustomization/ColumnCustomization.stories.js
```
Storybook

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/DraggableElement.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/DraggableElement.js
@@ -32,7 +32,10 @@ const DraggableElement = ({
     setNodeRef,
     transform,
     transition,
-  } = useSortable({ id });
+  } = useSortable({
+    disabled,
+    id,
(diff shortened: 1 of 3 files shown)


elycheea (commented):
Oooh, thanks for fixing this. So this implementing still allows us to drag to the position of a locked column but prevents the reordering from updating? I think ideally, we should prevent dropping into the locked column position, but understand if that’s not possible with dnd-kit now. 🤔

davidmenendez (approved):

elycheea (approved):
Opened a separate issue for improving the interaction so I think this is ready to go. 🎉

PR #3625, opened 2023-10-17
fix(Card): Update _card.scss

paul-balchin-ibm:
~Contributes to #~ Inline edit.

Fix typo. `#{$block-class}...` → `.#{$block-class}...`. Added missing dot at beginning of selector.


matthewgallo (commented):
You'll also need to update the styles snapshot since the `Card` is a released component. `yarn test:c4p:snapshot`

Issue #3586, opened 2023-10-12
Align select all checkbox in header with Carbon implementation 


xmannyxfreshx1:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description
- `TableSelectAll` component in Datagrid does not function as expected in `indeterminate` state (when some table items are selected, but not all)
- When table is in the `indeterminate` state, clicking on the checkbox in the header should deselect the items currently selected. Instead, it selects all.

> Is this issue related to a specific component?
- Datagrid, DatagridSelectAll

> What did you expect to happen? What happened instead? What would you like to
> see changed?
- described above

> What browser are you working in?
- Chrome
> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?
`"@carbon/ibm-products": "2.10.2",`

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
- CP4AIOPS
## Steps to reproduce the issue

1. Load a table with row selection enabled
2. Select some items on the table
3. Notice the checkbox in the header is in indeterminate state
![Screenshot 2023-10-12 at 10 47 20 AM](https://github.com/carbon-design-system/ibm-products/assets/13371221/ea75be44-1d63-4656-8511-d6f2ff565a34)
5. select the checkbox and notice all items are selected. The expected behavior is that the current items are deselected.

> Please create a reduced test case in CodeSandbox
>
> https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-datagrid-datagrid-canary--selectable-row
Expected behavior: https://react.carbondesignsystem.com/?path=/docs/components-datatable-toolbar--overview#selection

## Additional information

- Screenshots or code
- Notes

PR #3632, opened 2023-10-18
chore(Datagrid): uncheck items when select all is indeterminate

matthewgallo:
Contributes to #3586 

The select all checkbox previously always select all items, even when it was in the indeterminate state. The behavior should match what Carbon's DataTable does when select all is indeterminate which would be unselecting all items.

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Datagrid/DatagridSelectAll.js
```
Storybook (Datagrid, Batch actions story)


packages/ibm-products/src/components/Datagrid/Datagrid/DatagridSelectAll.js
```diff
@@ -64,7 +65,15 @@ const SelectAll = (datagridState) => {
       <TableSelectAll
         {...selectProps}
         name={`${tableId}-select-all-checkbox-name`}
-        onSelect={onChange}
+        onSelect={(event) => {
```
davidmenendez, line 68: super minor, but can you move this out into a `onSelectHandler` or something? just to get those ugly if /else out of the jsx 😅

davidmenendez (changes requested):

davidmenendez (approved):

PR #3633, opened 2023-10-18
chore(Datagrid): uncheck items when select all is indeterminate

matthewgallo:
Contributes to #3586 

The select all checkbox previously always select all items, even when it was in the indeterminate state. The behavior should match what Carbon's DataTable does when select all is indeterminate which would be unselecting all items.

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Datagrid/DatagridSelectAll.js
```
Storybook (Datagrid, Batch actions story)


davidmenendez (changes requested):
same comment in v2

davidmenendez (approved):

PR #3634, opened 2023-10-18
chore: toolbar review fixes

davidmenendez:
Contributes to #3442


matthewgallo (approved):

matthewgallo (changes requested):
Feedback discussed over slack with @paul-balchin-ibm:

> It seems like renderItem and ButtonMenuItem are only referenced once.
Would it not make more contextual sense to just return  ButtonMenuItem from directly inside ButtonMenu?

matthewgallo (approved):

paul-balchin-ibm (approved):

Issue #3595, opened 2023-10-13
Storybook (datagrid) console warnings: master list


paul-balchin-ibm:
_Note: I'm listing all these datagrid errors in a single issue:_ 
1. _Sometimes I come back to the same story and a different list of errors appears, even after refreshing the page._ 
2. _Some of these issues will be the same issue but duplicated across multiple stories._ 

_The list below will document which story each error **first** appears._

## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)

## Detailed description

> Describe in detail the issue you're having.

When running Storybook locally, I see console warnings across many Datagrid stories.

http://localhost:3000/?path=/story/ibm-products-components-datagrid-datagrid-canary--basic-usage
- [x] Resolved in #3660
```
Warning: Received `false` for a non-boolean attribute `role`. If you want to 
write it to the DOM, pass a string instead: role="false" or role={value.
toString()}. If you used to conditionally omit it with role={condition && 
value}, pass role={condition ? value : undefined} instead.
```

http://localhost:3000/?path=/story/ibm-products-components-datagrid-datagrid-canary--empty-state
- [x] Resolved #3654
```
Warning: React does not recognize the `placeHolderText` prop on a DOM element. 
If you intentionally want it to appear in the DOM as a custom attribute, spell 
it as lowercase `placeholdertext` instead. If you accidentally passed it from 
a parent component, remove it from the DOM element.
```
- [x] Resolved #3656
```
Warning: React does not recognize the `renderIcon` prop on a DOM element. If 
you intentionally want it to appear in the DOM as a custom attribute, spell it 
as lowercase `rendericon` instead. If you accidentally passed it from a parent 
component, remove it from the DOM element.
```

http://localhost:3000/?path=/story/ibm-products-components-datagrid-datagrid-canary--batch-actions
- [x] Resolved #3658
```
Warning: The `light` prop for `OverflowMenu` is no longer needed and has been 
deprecated. It will be removed in the next major release. Use the Layer 
component instead.
```
- [x] Resolved #3657
```
Warning: React does not recognize the `isDelete` prop on a DOM element. If you 
intentionally want it to appear in the DOM as a custom attribute, spell it as 
lowercase `isdelete` instead. If you accidentally passed it from a parent 
component, remove it from the DOM element.
```

http://localhost:3000/?path=/story/ibm-products-components-datagrid-datagrid-canary--disable-select-row
[...]

PR #3656, opened 2023-10-20
fix: Storybook error "renderIcon is not a DOM element"

paul-balchin-ibm:
Contributes to #3595

#### What did you change?

Removed a prop.


packages/ibm-products/src/components/Datagrid/utils/DatagridActions.js
```diff
@@ -178,7 +178,6 @@ export const DatagridActions = (datagridState) => {
         <RowSizeDropdown {...rowSizeDropdownProps} />
         <MenuButton
           label="Primary button"
-          renderIcon={ChevronDown}
```
paul-balchin-ibm, line 181: Removed.

This should _not_ throw an error. But, by default, `Menubutton's` default value for `renderIcon` is already `ChevronDown`, so removing it doesn't change the render.

matthewgallo (approved):

PR #3657, opened 2023-10-20
fix: Storybook error "isDelete is not a DOM element"

paul-balchin-ibm:
Contributes to #3595

#### What did you change?

Removed a prop.

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid.stories.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid.stories.js
@@ -511,7 +511,6 @@ export const BatchActions = () => {
         id: 'delete',
         itemText: 'Delete',
         icon: TrashCan,
-        isDelete: true,
         onClick: action('Clicked row action: delete'),
       },
     ];


packages/ibm-products/src/components/Datagrid/Datagrid.stories.js
```diff
@@ -511,7 +511,6 @@ export const BatchActions = () => {
         id: 'delete',
         itemText: 'Delete',
         icon: TrashCan,
-        isDelete: true,
```
paul-balchin-ibm, line 514: Removed.

When rendering _one or two_ "row actions", `useActionsColumn` will render `OverflowMenu` "trigger" components, without menu items, as buttons. (They should render `IconButtons` instead, but that's another story). `OverflowMenu` "trigger" components do not accept `isDelete` as props.

When rendering _more than two_ "row actions", `useActionsColumn` will render one `OverflowMenu` trigger and then one menu item per action. Each item may accept `isDelete` as a prop to render an item red when hovered over.

matthewgallo: I've created https://github.com/carbon-design-system/ibm-products/issues/3662 so that we can still support `isDelete`. This could also potentially include some refactoring to use `IconButton`s in the case of 1 or 2 actions.

matthewgallo (approved):

PR #3660, opened 2023-10-23
fix: Storybook console error 'role=boolean'

paul-balchin-ibm:
Contributes to #3595

Fixes console errors in Storybook.

#### What did you change?

Changed value assignment from a `boolean` to the correct `string`.

- [ ] Verified by @lee-chase that I didn't break anything.


matthewgallo (changes requested):
This affected some of the tests, specifically any test that gets elements by role of `row`.

matthewgallo (approved):

Issue #3590, opened 2023-10-12
Storybook console log: "Unknown key 'Style' exported by preview annotation file"


paul-balchin-ibm:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)

## Detailed description

> Describe in detail the issue you're having.

When running Storybook locally (i.e. `http://localhost:3000/...`), I see a persistent console log (not a warning or error log).

```
Unknown key 'Style' exported by preview annotation file
'.../ibm-products/packages/core/.storybook/preview.js'

@ preview.js-generated-config-entry.js:60
```
> Is this issue related to a specific component?

No, this seems to only happen when Storybook initially loads.

> What did you expect to happen? What happened instead? What would you like to see changed?

We should not see this console log at all.

> What browser are you working in?

Chrome, latest version.

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive) package are you using?

v2.11.0

## Steps to reproduce the issue

1. Fork and clone `@carbon/ibm-products` to your local machine.
2. `yarn` to install packages.
3. `yarn storybook` to run Storybook.
4. Load Storybook in your browser if it doesn't do it automatically, i.e. `http://localhost:3000`.
5. Open the browser's dev tools.
6. Select the "Console" tab to view the console log.

PR #3687, opened 2023-10-25
fix: Storybook console log: "Unknown key Style"

paul-balchin-ibm:
Contributes to #3590

Removes an unnecessary console log in Storybook.

#### What did you change?

Removed an component export

- [x] Verified log no longer appeared in browser console.

### DIFF
--- a/packages/core/.storybook/preview.js
+++ b/packages/core/.storybook/preview.js
@@ -156,4 +156,4 @@ const globals = {
   [CARBON_THEME_PARAM_KEY]: CARBON_THEMES.g10,
 };
 
-export default { argTypes, decorators, globals, parameters, Style };
+export default { argTypes, decorators, globals, parameters };


packages/core/.storybook/preview.js
```diff
@@ -156,4 +156,4 @@ const globals = {
   [CARBON_THEME_PARAM_KEY]: CARBON_THEMES.g10,
 };
 
-export default { argTypes, decorators, globals, parameters, Style };
+export default { argTypes, decorators, globals, parameters };
```
paul-balchin-ibm, line 159: Removed "Style" from being exported.

This was originally exported for use by `packages/security` in Carbon 10 to reduce duplication of code. See [details here](https://github.com/carbon-design-system/ibm-products/pull/918/files#diff-e1422435f234d2a0b80157b1a885d47ae643d8f9e79b7251f889c61bdb47ebb0).

"Security" was not ported to Carbon 11, so no longer needs to be exported.

matthewgallo (approved):

PR #3698, opened 2023-10-26
fix: Storybook error "multiForm is not a valid prop for a DOM element"

paul-balchin-ibm:
Contributes to #3697 

#### What did you change?

Removed a prop.


packages/ibm-products/src/components/CreateTearsheetNarrow/CreateTearsheetNarrow.stories.js
```diff
@@ -299,6 +299,5 @@ export const withMultipleForms = prepareStory(WithValidationTemplate, {
   storyName: withMultipleFormsStoryName,
   args: {
     ...defaultStoryProps,
-    multiForm: true,
```
paul-balchin-ibm, line 302: `multiForm` does not appear as a prop anywhere in Carbon for IBM Products.

(There is a `MultiFormEditTearsheet` component. In all likelihood, this component replaced the prop and this `multiForm` prop was missed during the clean up.

matthewgallo (approved):

PR #3700, opened 2023-10-26
Fix 3699 sb error labeltext of type object

paul-balchin-ibm:
Contributes to #3699

#### What did you change?

Modified layout so that the toggle's "label" showed it's tooltip without throwing a console error.

- [x] viewed in Storybook


packages/ibm-products/src/components/CreateFullPage/_storybook-styles.scss
```diff
@@ -58,7 +58,7 @@ $block-class: #{c4p-settings.$pkg-prefix}--create-full-page;
 .#{$story-class}__error--text {
   @include type.type-style('label-01');
 
-  padding-bottom: $spacing-05;
+  margin-bottom: $spacing-05;
```
paul-balchin-ibm, line 61: Changed from `padding`: was pushing the tooltip away from the label.

matthewgallo (approved):

Issue #3721, opened 2023-10-30
Datagrid: Extra, empty column added to table


silvaDominic:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

After upgrading from 2.11.3 to 2.13.0 all instances of our datagrids have an extra, empty column added to them.

> Is this issue related to a specific component?

The datagrid.

> What did you expect to happen?

No change to # of columns.

> What happened instead?

Additional column was added.

> What browser are you working in?

Chrome/FF/Safari

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

2.13.0

> What offering/product do you work on?

[IBM Turbonomic](https://www.ibm.com/products/turbonomic)

> Any pressing ship or release dates we should be aware of?

Blocking development of multiple teams

## Steps to reproduce the issue

1. Use IBM Carbon version 2.11.3
2. Observe datargid columns
3. Upgrade to at least 2.13.0
4. Observe datagrid columns

## Additional information

https://github.com/carbon-design-system/ibm-products/assets/12387334/7c94b90b-838c-4c5a-b500-5e5a65d8b312

_Observe how multiple datagrids now have an extra, empty column._

![Screenshot 2023-10-30 at 12 29 08 PM](https://github.com/carbon-design-system/ibm-products/assets/12387334/8034e255-c568-4afe-93eb-8ebf8f7fe518)

_Column has an empty `div` as content._

I believe the issue was caused in https://github.com/carbon-design-system/ibm-products/pull/3669 which was part of v2.12.1.

matthewgallo:
Fix will be included in tomorrow's release

PR #3722, opened 2023-10-30
fix(Datagrid): semantically hide spacer col

matthewgallo:
Contributes to #3721 

This PR semantically hides the spacer column used to allow the last visible column to be resized. The spacer column was also getting a class name specifically only for sortable columns, that's been fixed as well.

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Datagrid.test.js
packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHeaderRow.js
```
Storybook and added new test to check we render the expected number of column headers.

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHeaderRow.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHeaderRow.js
@@ -143,12 +143,13 @@ const HeaderRow = (datagridState, headRef, headerGroup) => {
                   [`${blockClass}__resizableColumn`]: header.getResizerProps,
                   [`${blockClass}__isResizing`]: header.isResizing,
                   [`${blockClass}__sortableColumn`]:
-                    datagridState.isTableSortable,
+                    datagridState.isTableSortable && header.id !== 'spacer',
                   [`${blockClass}__isSorted`]: header.isSorted,
                 },
                 header.getHeaderProps().className
               )}
               key={header.id}
+              aria-hidden={header.id === 'spacer' && 'true'}
               {...getAccessibilityProps(header)}
             >
               {header.render('Header')}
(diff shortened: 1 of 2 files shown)


packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHeaderRow.js
```diff
@@ -143,12 +143,13 @@ const HeaderRow = (datagridState, headRef, headerGroup) => {
...
                   [`${blockClass}__isResizing`]: header.isResizing,
                   [`${blockClass}__sortableColumn`]:
-                    datagridState.isTableSortable,
+                    datagridState.isTableSortable && header.id !== 'spacer',
                   [`${blockClass}__isSorted`]: header.isSorted,
                 },
                 header.getHeaderProps().className
               )}
               key={header.id}
+              aria-hidden={header.id === 'spacer' && 'true'}
```
paul-balchin-ibm, line 152: I don't understand how `&& 'true'` is supposed to work.

matthewgallo: Basically if it's a spacer column, aria-hidden will be true, otherwise it will be false for all other columns.

matthewgallo: Otherwise the spacer column is included as a column header, for example without this change
```js
screen.getAllByRole('role');
```
this gives back all of the columns including the spacer column.

paul-balchin-ibm: _Right!_ It will assign the string "true" if hidden.id is "spacer".

Mondays.

paul-balchin-ibm (approved):

Issue #3685, opened 2023-10-25
Storybook console warning: This prop syntax ("ariaLabel") has been deprecated


paul-balchin-ibm:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)

## Detailed description

> Describe in detail the issue you're having.

When running Storybook locally, I see a console warning.

http://localhost:3000/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-panel--panel-batch

```
Warning: This prop syntax has been deprecated. Please use the new `aria-label`.
```

> Is this issue related to a specific component?

Yes, http://localhost:3000/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-panel--panel-batch

> What did you expect to happen? What happened instead? What would you like to see changed?

We should not see these console warnings at all.

> What browser are you working in?

Chrome, latest.

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive) package are you using?

v2.11

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
N/A

## Steps to reproduce the issue

1. Fork and clone `@carbon/ibm-products` to your local machine.
2. `yarn` to install packages.
3. `yarn storybook` to run Storybook.
4. Open the browser to each of the URLs listed above.
6. Open the browser's dev tools.
7. Select the "Console" tab to view the console warnings.

paul-balchin-ibm:
Update. This warning is triggered by various Carbon components that have specified specific proptypes as `deprecated`. See Carbon's [search results](https://github.com/search?q=repo%3Acarbon-design-system%2Fcarbon+Please+use+the+new+%60aria-label%60&type=code).

Components that include the text "'This prop syntax has been deprecated. Please use the new `aria-label`.'":
- Dropdown
- CodeSnippet
- OverflowMenu
- TableToolbar

Props using `ariaLabel` should be replaced with `aria-label` or `['aria-label']`.

paul-balchin-ibm:
VS Code helper for the Dropdown component.

![image](https://github.com/carbon-design-system/ibm-products/assets/52505287/6646402d-a460-4489-b1cf-1a84ce273a6b)

PR #3740, opened 2023-11-01
fix: replace prop from "ariaLabel" to "aria-label"

paul-balchin-ibm:
Contributes to #3685

Some Carbon components have deprecated their `ariaLabel` props with the now-standard `aria-label` props.

These fixes remove the console logs created by some components: 
```
Warning: This prop syntax has been deprecated. Please use the new `aria-label`.
```

Note: The replacements of `ariaLabel` with `aria-label` was _not_ global. Only those where where the prop `aria-label` was being rendered as an HTML DOM attribute. Any other components using `ariaLabel` internally, but not as a DOM attribute, remain unchanged.

#### What did you change?

Renamed some prop names from `ariaLabel` to `aria-label`

- [x] Storybook


paul-balchin-ibm:
Closing. The attempt to "search and replace all (40 files)" broke a Datagrid test for no apparent reason. It was easier to surgically fix _only_ those (6) stories that were generating the error. See the new PR #3765.

PR #3773, opened 2023-11-07
feat: new NonLinearReading component

paul-balchin-ibm:
Contributes to #3602

The **non-linear reading** component ([usage](https://pages.github.ibm.com/cdai-design/pal/novice-to-pro/components/nonlinear-reading/usage)) provides a way to explain concepts more broadly and succinctly while allowing users to read more detail as needed without leaving their context.

#### What did you change?

This is a new component.

- [x] Migrated from Carbon v10
- [x] Storybook review
- [ ] Design reviewers: Cameron Calder, @sdignum, Bridget Ndege.

### DIFF
--- a/packages/ibm-products/src/components/NonLinearReading/NonLinearReading.js
+++ b/packages/ibm-products/src/components/NonLinearReading/NonLinearReading.js
@@ -1,0 +1,99 @@
+/**
+ * Copyright IBM Corp. 2023, 2023
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useRef, useState } from 'react';
+
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import { ChevronDown } from '@carbon/react/icons';
+
+import { getDevtoolsProps } from '../../global/js/utils/devtools';
+import uuidv4 from '../../global/js/utils/uuidv4';
+import { pkg } from '../../settings';
+
+// The block part of our conventional BEM class names (blockClass__E--M).
+const blockClass = `${pkg.prefix}--non-linear-reading`;
+const componentName = 'NonLinearReading';
+
+// Default values for props
+const defaults = {
+  theme: 'light',
+};
+
+/**
+ * Use non-linear reading when space is limited to share a
+ * brief, at-a-glance, summary of a concept that may require
+ * more explanation for some users.
+ */
+export let NonLinearReading = React.forwardRef(
+  (
+    { children, className, definition, theme = defaults.theme, ...rest },
+    ref
+  ) => {
+    const [isOpen, setOpen] = useState(false);
+    const contentId = useRef(uuidv4()).current;
+
+    const handleToggle = () => {
+      setOpen((prevState) => !prevState);
+    };
+
+    return (
+      <span
+        {...rest}
+        className={cx(blockClass, `${blockClass}__${theme}`, className)}
+        ref={ref}
+        role="main"
+        {...getDevtoolsProps(componentName)}
+      >
+        {' '}
+        <button
+          type="button"
+          aria-controls={contentId}
+          aria-expanded={isOpen}
+          className={cx(`${blockClass}__keyword`, [
+            isOpen
+              ? [`${blockClass}__keyword-open`]
+              : [`${blockClass}__keyword-closed`],
+          ])}
+          onClick={handleToggle}
+        >
+          {children}
+          <ChevronDown size={16} />
+        </button>{' '}
+        <span id={contentId} className={`${blockClass}__body`} hidden={!isOpen}>
+          {isOpen && definition}
+        </span>{' '}
+      </span>
+    );
+  }
+);
+
+// Return a placeholder if not released and not enabled by feature flag
+NonLinearReading = pkg.checkComponentEnabled(NonLinearReading, componentName);
+NonLinearReading.displayName = componentName;
+
+NonLinearReading.propTypes = {
+  /**
+   * The keyword of the component appears as a pill.
+   */
+  children: PropTypes.node.isRequired,
+
+  /**
+   * Provide an optional class to be applied to the containing node.
+   */
+  className: PropTypes.string,
+
+  /**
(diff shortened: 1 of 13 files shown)


packages/ibm-products/src/components/NonLinearReading/NonLinearReading.mdx
```diff
@@ -0,0 +1,76 @@
+import { Story, ArgsTable, Source, Canvas } from '@storybook/addon-docs';
```
matthewgallo, line 1: These docs should move to `NonLinearReading.docs-page.js` and you can delete this file.

packages/ibm-products/src/components/NonLinearReading/NonLinearReading.js
```diff
@@ -0,0 +1,99 @@
...
+    const handleToggle = () => {
+      setOpen((prevState) => !prevState);
+    };
+
+    return (
+      <span
+        {...rest}
+        className={cx(blockClass, `${blockClass}__${theme}`, className)}
+        ref={ref}
+        role="main"
```
matthewgallo, line 49: I don't think this should have a role of `main` since this likely isn't the primary content of the page.

packages/ibm-products/src/components/NonLinearReading/NonLinearReading.test.js
```diff
@@ -0,0 +1,92 @@
...
+    <NonLinearReading definition={definition} {...rest}>
+      {children}
+    </NonLinearReading>
+  );
+};
+
+describe(componentName, () => {
+  it('renders a component NonLinearReading', () => {
+    renderComponent();
+    expect(screen.getByRole('main')).toHaveClass(blockClass);
```
matthewgallo, line 38: You'll want to update this test to reflect whatever you think is the more appropriate role for this component, instead of `main`.

packages/ibm-products/src/components/NonLinearReading/NonLinearReading.js
paul-balchin-ibm: I'm new to `roles` and I'm open to options.

matthewgallo: [Here's the list](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles). Not sure what feels most appropriate, `definition` possibly?

paul-balchin-ibm: I couldn't find anything that matched perfectly, either.

Unfortunately, this UX doesn't fit any existing, standard component behaviour.

It's not explicit, but it doesn't seem like the definition should be hidden from the user. [digitala11y.com](https://www.digitala11y.com/definition-role/) explains "the definition of a term that is described in the _natural flow of the narrative_ is associated with the term itself using the aria-labelledby attribute".

I was thinking that this component behaved a lot like an inline version of Carbon's Accordion component ([GitHub](https://github.com/carbon-design-system/carbon/tree/main/packages/react/src/components/Accordion)), but it and its subcomponents don't have roles at all, and they're just buttons, divs, and paragraphs. Zero semantic HTML to tie them together, just `aria-controls` associated to `ids`.

I don't see a clear path here.

matthewgallo: In that case, I think for now just omit the role until we can find something that makes more sense. Don't want this to block you.

packages/ibm-products/src/components/NonLinearReading/NonLinearReading.test.js
```diff
@@ -0,0 +1,96 @@
...
+    <NonLinearReading definition={definition} {...rest}>
+      {children}
+    </NonLinearReading>
+  );
+};
+
+describe(componentName, () => {
+  // This component does not follow HTML standards in terms of
+  // functionality, and there is no pre-defined "role" that applies
+  // to this component. "Role" has been removed.
```
elycheea, line 38: Maybe `role="term"`[^1] and `role="definition"`[^2]? But agree with Matt that probably best to review and investigate appropriate role further rather have this block the rest of the PR.

[^1]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/term_role
[^2]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/definition_role

packages/ibm-products/src/components/NonLinearReading/_storybook-styles.scss
```diff
@@ -0,0 +1,36 @@
...
+//
+
+// @use '@carbon/layout/scss/convert' as *;
+@use '@carbon/react/scss/colors' as *;
+// @use '@carbon/styles/scss/motion' as *;
+@use '@carbon/styles/scss/spacing' as *;
+@use '@carbon/styles/scss/theme' as *;
+@use '@carbon/styles/scss/themes';
+@use '@carbon/styles/scss/type';
+// @use '../../global/styles/project-settings' as c4p-settings;
```
elycheea, line 15:
```suggestion
@use '@carbon/react/scss/colors' as *;
@use '@carbon/styles/scss/spacing' as *;
@use '@carbon/styles/scss/theme' as *;
@use '@carbon/styles/scss/themes';
@use '@carbon/styles/scss/type';
```

paul-balchin-ibm: Sorry, missed that file on review.

packages/ibm-products/src/components/NonLinearReading/NonLinearReading.test.js
```diff
@@ -0,0 +1,96 @@
...
+
+  // it('renders a component NonLinearReading', () => {
```
matthewgallo, line 40: You can completely remove this test since the test `forwards a ref to an appropriate node` also checks that the component has the correct block class.

matthewgallo (changes requested):

paul-balchin-ibm:
Closing. 

For someone reason this was marked "waiting for author" and got lost in the shuffle. With all the major SCSS changes that happened in the interim, it was too out of date to fix.

See updated PR #3889.

PR #3781, opened 2023-11-08
fix(DatagridToolbar): fix accessibility issue with focusable hidden buttons

joyful-these:
For https://github.ibm.com/katamari/dev-issue-tracking/issues/50481

282596005-19a0943b-7ca3-4a04-8802-76b044abc89f

When the MenuButton/TableBatchAction is hidden in the DOM, this currently creates an accessibility issue: `Element "button" should not be focusable within the subtree of an element with an 'aria-hidden' attribute with value 'true'`

This fix sets the tabIndex of buttons to 0 or -1 depending on if aira-hidden or not (based on totalSelected).


matthewgallo (approved):
🎉

PR #3783, opened 2023-11-08
chore(AboutModal): cleanup storybook controls

glapadre:
Contributes to #3447 

{{Update storybook title from Node to String}}

#### What did you change?
`packages/ibm-products/src/components/AboutModal/AboutModal.js
packages/ibm-products/src/components/AboutModal/_about-modal.scss
packages/ibm-products/src/components/AboutModal/AboutModal.stories.js`

Storybook

### DIFF
--- a/packages/ibm-products/src/components/AboutModal/AboutModal.js
+++ b/packages/ibm-products/src/components/AboutModal/AboutModal.js
@@ -213,7 +213,7 @@ AboutModal.propTypes = {
    * IBM 8-bar logo represented alongside the IBM Services logotype.
    * Please follow these guidelines to ensure proper execution.
    */
-  title: PropTypes.node.isRequired,
+  title: PropTypes.string.isRequired,
 
   /**
    * Text that provides information on the version number of your product.
--- a/packages/ibm-products/src/components/AboutModal/AboutModal.stories.js
+++ b/packages/ibm-products/src/components/AboutModal/AboutModal.stories.js
@@ -177,19 +177,7 @@ Template.propTypes = AboutModal.propTypes;
 
 const commonArgs = {
   closeIconDescription: 'Close',
-  title: (
-    <>
-      IBM{' '}
-      <span
-        style={
-          // stylelint-disable-next-line carbon/type-token-use
-          { fontWeight: '600' }
-        }
-      >
-        Product name
-      </span>
-    </>
-  ),
+  title: 'Product name',
   version: 'Version 0.0.0',
   copyrightText: 'Copyright © IBM Corp. 2020, 2023',
 };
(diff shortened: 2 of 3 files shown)


packages/ibm-products/src/components/AboutModal/AboutModal.js
```diff
@@ -221,7 +221,7 @@ AboutModal.propTypes = {
    * IBM 8-bar logo represented alongside the IBM Services logotype.
    * Please follow these guidelines to ensure proper execution.
    */
-  title: PropTypes.node.isRequired,
+  title: PropTypes.string.isRequired,
```
matthewgallo, line 224: Do we want to only strictly accept strings? `PropTypes.node` will allow consumers to pass either a string or a node and gives more flexibility. I believe this was discussed a while back during the initial release review for `AboutModal`, [here](https://github.com/carbon-design-system/ibm-products/issues/385).

[Carbon's ModalHeader component](https://react.carbondesignsystem.com/?path=/docs/components-composedmodal--overview) `title` prop is also specified as `PropTypes.node`.

glapadre: As we talked over Slack Huddle, going to move this back to using `node`, and use a dropdown to show _short_ and _long_ titles

elycheea (commented):
I wonder if rather than short/long (or in addition) ... it makes sense to show title `with formatting`? I believe that was the reason we used a node instead of text for the `title` — so that in addition to a short/long text string, we could also show a product title with the formatting?

So maybe something like ...
- short title
- long title
- short title with formatting

Basically, `text`, more `text`, `node`.

glapadre:
> I wonder if rather than short/long (or in addition) ... it makes sense to show title `with formatting`? I believe that was the reason we used a node instead of text for the `title` — so that in addition to a short/long text string, we could also show a product title with the formatting?
> 
> So maybe something like ...
> 
> * short title
> * long title
> * short title with formatting
> 
> Basically, `text`, more `text`, `node`.

We could add an example that is more clear in showing unique formatting with a node, but technically the two examples we have (short and long) are already using the node format by 'bolding' the product name after 'IBM'

Screenshot 2023-11-15 at 10 50 45 AM

We could add something like adding color, italic text, ect.. but we may not even want to recommend doing that. Thoughts?

elycheea:
I should have been clearer! Since I _was_ thinking it made more sense to no formatting (string only short title) and with formatting (node with the span as the current example shows) as examples. But I also see the value of showing a longer title so short/long/formatted were my suggestions. 😅 

It’s very possible that products will prefer to always include the formatting, but thought it was helpful to include the default if you pass just a string to reflect the code.

glapadre:
According to the [documentation](https://pages.github.ibm.com/cdai-design/pal/patterns/about-modal/usage) it says the IBM logo is required, has that changed? I wouldn't want to show a example without it if we always recommend having it. 

`**Product name**: Header text that provides the product name. The IBM Services logo consists of two discrete, but required, elements: the iconic IBM 8-bar logo represented alongside the IBM Services logotype. Please follow these guidelines to ensure proper execution.`

Issue #3716, opened 2023-10-30
Proposal: Publish styles as separate package


lee-chase:
Publishing the styles as an independent package. E.g. `@carbon/ibm-products-styles`

This would mirror the Carbon core pattern and allow the production of WebComponents as part of the Carbon AI project which looks to make use of Tearsheet and SidePanel.

NOTE: WebComponents would not want to load other non-dev-dependencies

elycheea:
Successfully published in latest release [`@carbon/ibm-products-styles@2.17.0`](https://www.npmjs.com/package/@carbon/ibm-products-styles/v/2.17.0).

PR #3799, opened 2023-11-13
feat: publish styles as independant package

lee-chase:
Contributes to #3716

Created a new package by copying the ibm-products package.json and cutting it down to make  `ibm-products-styles`.

It copies the styles from `packages/ibm-products/src` and builds a package from them. 

Note: That creating a copy of the source is in part based on the assumption that something in the package would need to change in order to bump the version. This should be revisited as a final strategy, it would probably be better to have both @carbon/ibm-products and @carbon/ibm-products-styles have the same version with changes to styles or components causing the other to update it's version.

#### What did you change?

New package

Test build and `npm pack` in the `packages/ibm-products-styles` folder to see what would be published.


.github/workflows/release.yml
elycheea, line 1: Let’s put this as a prerelease workflow instead?

lee-chase: @elycheea done sort of

matthewgallo:
I think the build script is failing because `InlineTip` and `SteppedAnimatedMedia` components were merged this morning, I'm guessing their styles need to be moved now.

.github/workflows/release-scheduled.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '30 4 * * TUE' # Release every Tuesday at 4:30am EST
+
+jobs:
+  Release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ./release-inner.yml
```
elycheea, line 12:
```suggestion
      - uses: ./release.yml
```

.github/workflows/release-preminor.yml
```diff
@@ -0,0 +1,14 @@
...
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '30 4 * * TUE' # Release every Tuesday at 4:30am EST
+
+jobs:
+  Release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ./release-inner.yml
```
elycheea, line 12:
```suggestion
      - uses: ./release.yml
```

packages/ibm-products-styles/jest.config.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Copyright IBM Corp. 2021, 2021
```
elycheea, line 2:
```suggestion
/**
 * Copyright IBM Corp. 2023, 2023
```

packages/ibm-products-styles/src/__tests__/test-helper.js
```diff
@@ -0,0 +1,69 @@
+//
+// Copyright IBM Corp. 2021, 2021
```
elycheea, line 2:
```suggestion
//
// Copyright IBM Corp. 2023, 2023
```

elycheea (commented):
Minor changes in comments.

As we discussed previously, we’ll probably want to update our example component and our own contribution docs with these changes too. We can make those in a new issue. I know Jeff and Paul used `generate` pretty religiously for the n2p components so we’ll need to ensure those SCSS files get moved to the styles package correctly.

elycheea:
`yarn generate` works great — might not be clear to someone new that styles are located in a different package (since it use uses the same `src/components/` path regardless) although they can likely figure it out. Can open an issue to update our getting started guides.

image

elycheea (dismissed):

elycheea:
Approved but wait until after next release to merge so we have time to resolve any build-related issues.

elycheea (approved):
Re-approving. ✔️

Issue #3817, opened 2023-11-16
Side panel title can have too little space reserved


lee-chase:
Sometimes the code calculating the space that needs to be reserved for the SidePanel title goes wrong with text spilling outside the title area.

## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Steps to reproduce the issue

1. View here https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-side-panel-sidepanel--slide-in&args=size:md
2. Show the side panel
3. Scroll down to the bottom
4. Switch the size to any other size (e.g. small)
5. Scroll down again and note the state of the title area
![image](https://github.com/carbon-design-system/ibm-products/assets/15086604/1eb572e4-d30b-4637-bce7-f56c6c0f76f7)
6. Refresh and scroll down to see what it should look like
![image](https://github.com/carbon-design-system/ibm-products/assets/15086604/786645cd-074b-409d-92e5-9747405f236c)

PR #3823, opened 2023-11-16
fix: side panel scrolled height

lee-chase:
Closes to #3817

Changes how the height of the side panel title is calculated

#### What did you change?

Stopped changing the container height, instead adding negative margin to the title-text.

Storybook following steps in issue.


matthewgallo (approved):
🎉 tested with various panel size changes, works great!

Issue #3824, opened 2023-11-16
Release Datagrid base component


elycheea:
### Steps
- [x] Move Datagrid to to released components in `package-settings`
   - [x] Set flag to `Datagrid: true`
- [x] Add Datagrid styles to `index-released-only`
- [x] Remove relevant Datagrid feature flags
  - I _think_ it’s just `Datagrid.useInfiniteScroll`, `Datagrid.useActionsColumn`
- [x] Update `carbon.yml`
  - [x] Set Datagrid status from `experimental` to `stable`
  - [x] Remove `--canary` in URLs

#### Repeat for main_v1.
- [x] Move Datagrid to to released components in `package-settings`
   - [x] Set flag to `Datagrid: true`
- [x] Add Datagrid styles to `index-released-only`
- [x] Remove relevant Datagrid feature flags
  - I _think_ it’s just `Datagrid.useInfiniteScroll`, `Datagrid.useActionsColumn`
- [x] Update `carbon.yml`
  - [x] Set Datagrid status from `experimental` to `stable`
  - [x] Remove `--canary` in URLs

The base component of the Datagrid includes the following features
– Table headers
– Clickable row items
– Row action buttons
– Batch actions
– Empty state
– Frozen columns (scrolling)
– Responsiveness (scrolling)
– Column alignment
– Infinite scrolling
– Resizable columns

PR #3839, opened 2023-11-20
chore: datagrid base release

davidmenendez:
Contributes to #3824


elycheea (changes requested):
We should also remove the feature flag for infinite scrolling, but either way, feature flags are still showing in Storybook[^1][^2]. It does look like we also need to remove the feature flag lines from the stories file as well. 👀 

On the bright side, probably better that we are able to release both the v1 and v2 at the same time so let’s see if we can get these changes in for both v2 and v1.

[^1]: [Infinite scroll](https://deploy-preview-3839--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid--infinite-scroll)
[^2]: [Row action buttons](https://deploy-preview-3839--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-extensions-rowactionbuttons--row-action-buttons-usage-story) (all three stories)

davidmenendez:
overflow bug found in this commit https://github.com/carbon-design-system/ibm-products/commit/ad6ce584f4cc897347f2f4231f839c0e68f98d66

i decided to just revert it because i don't remember what the intended behavior matt coded was supposed to be

davidmenendez:
for #3843

elycheea (approved):

PR #3846, opened 2023-11-22
fix: move spacer and actions to the end

davidmenendez:
solves issue of the actions column not floating to the right of the spacer. this PR will check for an actions column and if found will move it to the end after the spacer.


elycheea (dismissed):
Working as expected so far, including when testing customize columns (no longer inserting the last column in between, haha).

I added a little update to show use the first story to show actions inline (not sticky) while keeping the remaining stories as frozen examples. (Went ahead and removed the feature flag to make sure we don’t miss that in #3839

And another to fix tests based on this change. Approving but think we should get @matthewgallo’s approval too before merging.

elycheea (approved):

matthewgallo (approved):
Don't see any issues here either! And nice that we now have an example of non-sticky row actions 🥳

Issue #2761, opened 2023-03-21
Remove link styling in web terminal overflows


elycheea:
From the [web terminal](https://github.ibm.com/CDAI-design/pal/issues/1149#issuecomment-54512278) migration design review —

### Design specs

- [ ] Remove underlines in docs overflow

![image.png](https://images.zenhubusercontent.com/6109a215337f0517b2f0e02f/a56d1e34-3e34-44d5-bcbd-a16e370016d1)

Lower priority — to be addressed after official v2 release.
Overflow with underlines is not consistent with other overflows. While these are actual links, worth working with design to come to a more consistent solution.

elycheea:
Still relevant, but low priority.

MriduMohan:
Hi @elycheea , I've taken up this issue and provided a fix.

PR #3848, opened 2023-11-27
fix(WebTerminal): removed web terminal styling

MriduMohan:
Contributes to #2761 

#### What did you change?
Changed web-terminal.scss file to remove the link-style from the web-terminal documentation links.

storybook


elycheea (dismissed):

MriduMohan:
> Looks good but actually wondering it might be better to include `.#{$block-class}` for consistency with the rest of the component styles here.

Added .#{$block-class} as per the review comment.

elycheea (approved):
🎉

PR #3849, opened 2023-11-27
chore: revert ibm-products-styles paths

elycheea:
Replaces `@carbon/ibm-products-styles` until package is actually published.
This should fix our broken builds. Once published, we can replace the URLs again.


matthewgallo (approved):
🎉

Issue #3731, opened 2023-10-31
<DataGrid>  and <NoDataEmptyState>  Accessibility test failure


SeonyuK:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

- Datagrid and NoDataEmptyState are failing the accessibility tests. 

- In the **Datagrid**

Rule violation: #aria-allowed-role [minor]
    > Ensures role attribute has an appropriate value for the element
    ARIA role should be appropriate for the element
    https://dequeuniversity.com/rules/axe/4.4/aria-allowed-role?application=axeAPI
    ================================================================================
    Node:
    <input draggable="false" role="separator" style="cursor: col-resize;" class="c4p--datagrid__col-resizer-range" type="range" aria-label="Resize column" value="240">
    Fix any of the following:
      ARIA role separator is not allowed for given element

- In the **NoDataEmptyState**

Rule violation: #svg-img-alt [serious]
    > Ensures <svg> elements with an img, graphics-document or graphics-symbol role have an accessible text
    <svg> elements with an img role must have an alternative text
    https://dequeuniversity.com/rules/axe/4.4/svg-img-alt?application=axeAPI
    ================================================================================
    Node:
    <svg xmlns="http://www.w3.org/2000/svg" width="80" height="80" viewBox="0 0 80 80" class="c4p--empty-state__illustration c4p--empty-state__illustration-noData c4p--empty-state__illustration--lg" role="img">
    Fix any of the following:
      Element has no child that is a title
      aria-label attribute does not exist or is empty
      aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
      Element has no title attribute

> Is this issue related to a specific component?

- Datagrid

> What did you expect to happen? What happened instead? What would you like to
> see changed?

- No fail for the accessibility test

> What browser are you working in?

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

- We are using @carbon/ibm-products": "2.13.0

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

- CP4S.

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

> Please create a reduced test case in CodeSandbox
>
[...]

PR #3862, opened 2023-11-29
fix: no data empty state title

davidmenendez:
Contributes to #3731 

missed applying a title in the no data illustration


matthewgallo (changes requested):
You might need to make some small adjustments now as the empty state Datagrid test is failing.

davidmenendez:
@matthewgallo thanks for pointing that out. should be fixed now.

matthewgallo (approved):

Issue #3892, opened 2023-12-04
Allow Datagrid filters to render display labels


elycheea:
![image](https://user-images.githubusercontent.com/1717398/287798069-f6d1689e-d9e6-459d-82cf-0ed5d5690227.png)

I am guessing we will need to change the filtering implementation a bit to improve the way the labels are displayed (sentence case + i18n) — but think business logic should be separated here — can discuss the approach to achieve this separately.

Seems like we are using `accessor`, `column`, and `id` for the filtering + filter summary labels, but may need a way to show displayed labels other than the identifier.

e.g. `passwordStrength: normal` should be displayed as `Password strength: Normal`.

May need a separate issue to address this for date (see comments), but we can use this issue to evaluate first in case they can be contained.

--- 

> ### Filter summary
> - Where's the 'List all' chevron?
> - Label should match sentence case (first letter capitalized)

_Originally posted by @sdignum in https://github.com/carbon-design-system/ibm-products/issues/2500#issuecomment-1343315181_

> - [ ] Date format presented in the Filter panel are mm/dd/yyyy and in the tags formatted as dd/mm/yyyy
![Screenshot 2023-11-08 at 15 23 46](https://github.com/carbon-design-system/ibm-products/assets/101175720/74930f9c-8bed-4e47-9423-b1c86b33f113)

_Originally posted by @My-Money29 in https://github.com/carbon-design-system/ibm-products/issues/2500#issuecomment-1801908211_

matthewgallo:
Actually, I think this scenario may still be possible. I can open a separate issue for this as suggested!

matthewgallo:
https://github.com/carbon-design-system/ibm-products/issues/3899

PR #3897, opened 2023-12-04
chore(Datagrid): add example usage with `renderLabel` property

matthewgallo:
Contributes to #3892

Adds an example demonstrating how to customize the rendering of the filter tag labels.

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Extensions/Filtering/Flyout.stories.js
packages/ibm-products/src/components/Datagrid/Extensions/Filtering/Panel.stories.js
packages/ibm-products/src/components/Datagrid/utils/handleFilterTagLabelText.js
```
Storybook


elycheea (commented):
I would most likely include `handleFilterTagLabelText` in all the filtering stories rather than just the first one or it might look like a bug?

Then probably worth mentioning the `renderLabel` property in the filtering docs since I didn’t see it in the docs page for filtering.

elycheea (approved):

Issue #3801, opened 2023-11-13
Datagrid filter panel: Couple of issues with keeping selected options before applying.


xmannyxfreshx1:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description
> Describe in detail the issue you're having.
- 2 issues found with datagrid filter panel keeping selected options before applying.
  -  While filters are applied, deselect some options in the same section, then refresh table data (i.e. set `isFetching` prop to true then false) you'll notice that the deselected options have been removed without the user clicking apply
  - After clearing filters using the `Clear fliters` button in filters summary, select options in the panel then click cancel. The previously selected items are set. The expectation is that panel goes back to a cleared state.
 ### 1 issue found after applying filters
  - (Higher Severity) After selecting any filter then applying, if the user closes the panel with filter icon button, the selections are removed. This was working as expected in previous versions. (can be reproduced in code sandbox or storybook)
> Is this issue related to a specific component?
- Datagrid Filter panel batch option

> What did you expect to happen? What happened instead? What would you like to
> see changed?
- explained above
> What browser are you working in?
- Chrome
> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?
- ` "@carbon/ibm-products": "2.15.0",`
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?
- CP4AIOps
## Steps to reproduce the issue
- explained above

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/p/sandbox/datagrid-filter-panel-with-mocked-refresh-7gjpqr?file=%2Fsrc%2FExample%2FExample.jsx%3A1%2C1

## Additional information
- Codesandbox linked above is set up to mock an api call that refreshes the table data when the refresh button (in the toolbar) is clicked.

- Screenshots or code
- Notes

xmannyxfreshx1:
@elycheea I added another bug I found around the filter panel not properly keeping the user selections.

> (Higher Severity) After selecting any filter then applying, if the user closes the panel with filter icon button, the selections are removed. This was working as expected in previous versions.

I believe this one is higher priority/severity since a user is very likely to run into it. It also affects a more common way the filter panel would be used.

matthewgallo:
Hey @xmannyxfreshx1, all of the items listed here should now be addressed.

[...]

PR #3902, opened 2023-12-05
fix(Datagrid): keep filters visible when closing panel via filter icon button

matthewgallo:
Contributes to #3801 

There was some logic missing to prevent the filters from being cleared when closing the filter panel via the icon button used to open/close the panel.

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/hooks/useFilters.js
```
Storybook


elycheea (approved):

matthewgallo:
Yes that is correct. I'll open a new issue to specifically address the two lesser severity items.

PR #3948, opened 2023-12-06
chore(ExportModal): add dropdown menu for validExtensions & preformattedExtensions

glapadre:
Contributes to #3944

Add dropdown menu for validExtensions & preformattedExtensions

#### What did you change?
`packages/ibm-products/src/components/ExportModal/ExportModal.stories.js`

Storybook

### DIFF
--- a/packages/ibm-products/src/components/ExportModal/ExportModal.stories.js
+++ b/packages/ibm-products/src/components/ExportModal/ExportModal.stories.js
@@ -44,6 +44,47 @@ export default {
       ),
     },
   },
+  argTypes: {
+    validExtensions: {
+      control: {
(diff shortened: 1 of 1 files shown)


elycheea (commented):
Looking at the current examples in Export modal — it does feel like the three current stories do a good job of conveying the different use cases quite well. I think if we were to do a single story, it might be worth prepopulating some of the validation messages so that you can more readily recreate the same effect you get from the three different stories?

I’m generally a fan of consolidating the stories, but hoping the different states are still easily discoverable so might be worth chatting about good ways to achieve both.

elycheea (approved):

Issue #3814, opened 2023-11-15
Set up `@carbon/ibm-products-community`


elycheea:
## What will this achieve?

Creates a space for product teams to share community or experimental code that may be not be fully maintained by the core IBM Products team. This could also be a good space for work that’s currently going through the DSAG process or new proposals that are being shared.

## Additional information

- Designs
- Existing code
- etc

elycheea:
https://github.com/carbon-design-system/ibm-products/issues/3871#issue-2019118763

elycheea:
Also maybe reach out to @asfordmatt for feedback on this structure/approach as well.

davidmenendez:
per slack conversation with @asfordmatt 

> thanks David, that looks great. I'll take a more detailed look later, but I think we might be okay with having it only in the main branch for v2 / Carbon 11. There are a few products migrating from  the ibm-cloud-cognitive-cdai package to ibm-products  in preparation for moving up to C11. But there is no replacement for IdeHome  so we will look to contribute a C11 version of that component to this new package ibm-products-community

since the community package was merged in the v2 branch i'll go ahead and close this as done for now. if we need to get a v1 version going i think we can open a separate issue for that 👍

PR #3956, opened 2023-12-08
feat: community package init

davidmenendez:
for #3814


davidmenendez:
@matthewgallo & @lee-chase would love y'all's input!

matthewgallo:
Yes! So I was wondering why we need a storybook setup for this package? The other packages (when there were others ie `cdai` and `security`) all were included in our main storybook, from `packages/core` so there is no need to include a storybook config for individual packages. This could clean things up and reduce the config that you've included. A good example to maybe take a look at is `@carbon/ibm-products-styles` since @lee-chase recently created that new package.

davidmenendez:
> Yes! So I was wondering why we need a storybook setup for this package? The other packages (when there were others ie `cdai` and `security`) all were included in our main storybook, from `packages/core` so there is no need to include a storybook config for individual packages. This could clean things up and reduce the config that you've included. A good example to maybe take a look at is `@carbon/ibm-products-styles` since @lee-chase recently created that new package.

oh ok i wasn't aware we had a shared storybook config. that definitely makes sense. the stories here can definitely reside in a section called `community` 👍

davidmenendez:
now that i know i don't need to include storybook i'm going to just redo this because it appears to be more difficult to remove it as opposed to starting fresh

Issue #3903, opened 2023-12-05
N2P guide banner-optional left padding


cameroncalder:
Enhance the N2P guide banner by adding an optional left-side padding control, equipping developers with the option to change the padding from 16 to 32. This will make it so product teams can place the guide banner starting flush to the left edge of the screen and keep the icon aligned w/ the first column of the Carbon grid.

<img width="1350" alt="image" src="https://github.com/carbon-design-system/ibm-products/assets/50155706/664f9f47-9391-4542-bc4d-3bffb67e41e3">

paul-balchin-ibm:
Reopening as we're still waiting for the `v1` fix.

paul-balchin-ibm:
`v1` and `v2` PRs are merged. Closing.

PR #3965, opened 2023-12-12
fix(Guidebanner): allow adding 1rem space on the left

paul-balchin-ibm:
Contributes to #3903

New prop `withLeftGutter`

- [x] Storybook review
- [x] Design reviewers: Cameron Calder, Shelley Dignum, Bridget Ndege.

### DIFF
--- a/packages/ibm-products/src/components/Guidebanner/Guidebanner.js
+++ b/packages/ibm-products/src/components/Guidebanner/Guidebanner.js
@@ -22,6 +22,7 @@ const componentName = 'Guidebanner';
 
 const defaults = {
   collapsible: false,
+  withLeftGutter: false,
   // Labels
   closeIconDescription: 'Close',
   collapseButtonLabel: 'Read less',
@@ -41,6 +42,7 @@ export let Guidebanner = React.forwardRef(
       className,
       collapsible = defaults.collapsible,
       onClose,
+      withLeftGutter = defaults.withLeftGutter,
       // Labels
       closeIconDescription = defaults.closeIconDescription,
       collapseButtonLabel = defaults.collapseButtonLabel,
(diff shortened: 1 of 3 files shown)


packages/ibm-products/src/components/Guidebanner/Guidebanner.js
```diff
@@ -69,8 +71,9 @@ export let Guidebanner = React.forwardRef(
         className={cx(
           blockClass,
           className,
-          [collapsible ? `${blockClass}__collapsible` : null],
-          [isCollapsed ? `${blockClass}__collapsible-collapsed` : null]
+          collapsible && `${blockClass}__collapsible`,
+          isCollapsed && `${blockClass}__collapsible-collapsed`,
```
paul-balchin-ibm, line 75: Extra: simplified code.

elycheea: Haha, is this pulling from your recent convo with @davidmenendez 😆

packages/ibm-products/src/components/Guidebanner/Guidebanner.stories.js
elycheea, line 1: If the grid matters here, I actually think it might make more sense to change the layout style from `padded` to `fullscreen` so you can see the alignment using Carbon Devtools.

image

paul-balchin-ibm: A component's only direct connection to the outside world should be limited to changing its props values. It's like an API: "apply setting to component" and "listen to changes from component" and nothing else. Trying to get a component to react to external features like page or grid settings can result too easily result in false positives.

elycheea: Oh. Let me rephrase — change the _Storybook_ layout setting. My suggestion is not a change to the component itself. I think this is L34 in Stories file but will give a more accurate representation of how it looks on the page. (PageHeader and Datagrid sometimes use this setting instead of the padded style in Storybook.)

image

paul-balchin-ibm: OH! I see what you mean. Sure, I can change that.

elycheea (approved):

packages/ibm-products/src/components/Guidebanner/Guidebanner.js
paul-balchin-ibm: Actually, I had completely missed his comment while making these changes. :)

PR #3974, opened 2023-12-13
fix: add missing N2P component exports

paul-balchin-ibm:
Adds missing N2P components to list of exports

#### What did you change?

```
packages/ibm-products/src/components/index.js
```


matthewgallo (approved):
🚀

matthewgallo (changes requested):
Something seems up with lottie-web in the tests? Can you run them locally and see if you encounter the same thing?

matthewgallo (approved):

PR #3983, opened 2023-12-18
chore: move sass out of loop

lee-chase:
Contributes to #3896 

Refactor some SASS

#### What did you change?

- Removed duplicate generated SCSS for placement.

### DIFF
--- a/packages/ibm-products-styles/src/components/SidePanel/_side-panel.scss
+++ b/packages/ibm-products-styles/src/components/SidePanel/_side-panel.scss
@@ -82,14 +82,14 @@ $action-set-block-class: #{c4p-settings.$pkg-prefix}--action-set;
         @include setPanelSize($size_value);
       }
     }
-    &.#{$block-class}__container-right-placement.#{$block-class}__container--#{$size} {
-      right: 0;
-      border-left: 1px solid $border-subtle-02;
-    }
-    &.#{$block-class}__container-left-placement.#{$block-class}__container--#{$size} {
-      left: 0;
-      border-right: 1px solid $border-subtle-02;
-    }
+  }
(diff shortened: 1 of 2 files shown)


packages/ibm-products-styles/src/components/SidePanel/_side-panel.scss
```diff
@@ -343,30 +326,29 @@ $action-set-block-class: #{c4p-settings.$pkg-prefix}--action-set;
   }
 
   .#{c4p-settings.$carbon-prefix}--btn.#{$block-class}__navigation-back-button {
-    position: fixed;
+    position: relative;
     z-index: 5;
-    top: 0;
-    left: 0;
+    top: calc(-1 * #{$spacing-03});
+    left: calc(-1 * #{$spacing-03});
```
elycheea, line 332: image

Looks like the change in position to be 8px from the end is intentional since I see this consistently with the close button, but looks like we still need to align the icon.

packages/ibm-products-styles/src/components/SidePanel/_side-panel.scss
```diff
@@ -139,15 +129,8 @@ $action-set-block-class: #{c4p-settings.$pkg-prefix}--action-set;
       margin-bottom: 0;
     }
     &.#{$block-class}__on-detail-step .#{$block-class}__collapsed-title-text {
-      top: $spacing-07;
+      top: $spacing-09;
```
elycheea, line 132: Any specific reason we increased this by `1rem`? Seems minor but does also increase the vertical real estate used by the header. 🤔

lee-chase:
@elycheea, it should now just contain the intended changes. Must have had a dirty main branch.

elycheea (approved):

Issue #2689, opened 2023-02-21
[Datagrid]: Add reduce motion to FilterPanel animation


AlexanderMelox:
Add support for reduced motion to FilterPanel.

elycheea:
In this case, thinking no motion (no fade ins) would be the best approach for this.

PR #3994, opened 2024-01-02
feat(Datagrid): add reduced motion support to filter panel

matthewgallo:
Contributes to #2689 

Allows the filter panel in the Datagrid to respond to the user's reduce motion setting.

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/FilterPanel.js
packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/motion/variants.js
```
Storybook, toggled `Reduce motion` in my osx settings to confirm correct behavior

### DIFF
--- a/examples/carbon-for-ibm-products/APIKeyModal/README.md
+++ b/examples/carbon-for-ibm-products/APIKeyModal/README.md
@@ -50,7 +50,7 @@ The simplest way to update a template is as follows.
 2. Update it as necessary.
 3. Test that it works.
 4. Remove the folder `scripts/example-gallery-builder/update-example/templates` and rename your folder to this.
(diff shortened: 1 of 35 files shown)


mikeolasov:
> Works as expected!
> 
> Thoughts for future — I know @mikeolasov has thoughts on reduced motion versus no motion. Might be something we can explore some more separately since the filter panel treatment should probably be consistent with side panel treatment. (e.g. Could have more subtle slide but buttons don’t animate? ... although I’ll leave it to Mike’s expertise.)

For big moves like the filter panel opening I think it's best to have no motion or a very simple ease out for the last ~10% of the move, where the panel snaps to ~90% open on the first frame and eases the rest (just to lead the user's eye).

I'd also like to see a fade-in or cascade fade-in of the filter panel contents, like the regular motion state.

elycheea:
@mikeolasov opened up a followup issue if we want to do more here.

Issue #3812, opened 2023-11-15
Side Panel with size 2xl has action buttons placed in the middle


youda97:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

I am using the Side Panel with size 2xl and two action buttons. The two buttons are randomly placed at the bottom and are not fixed to a particular side, making it look broken.

> Is this issue related to a specific component?

SIde panel

> What did you expect to happen? What happened instead? What would you like to
> see changed?

I would like to see the bottoms fixed a the left or right side of the side panel, or taking up the width like the other side panels

> What browser are you working in?

Chrome

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

"@carbon/ibm-products": "2.15.0"

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

Turbonomic

## Steps to reproduce the issue

https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-side-panel-sidepanel--without-title&args=actions:3;size:2xl

## Additional information

![Screenshot 2023-11-15 at 2 14 35 PM](https://github.com/carbon-design-system/ibm-products/assets/25139161/61d38357-5d28-4524-9603-89a3468c8c43)

PR #3999, opened 2024-01-04
fix(SidePanel): action buttons layout at 2xl (v1)

paul-balchin-ibm:
Contributes to #3812 (v1)

Update actions container to be 100% wide at `2xl`.

#### What did you change?

```
packages/ibm-products/src/components/SidePanel/_side-panel.scss
```

- [x] Storybook


packages/ibm-products/src/components/SidePanel/_side-panel.scss
```diff
@@ -87,6 +86,9 @@ $action-set-block-class: #{$pkg-prefix}--action-set;
         border-right: 1px solid $decorative-01;
       }
     }
+    .#{$block-class}__actions-container {
+      width: 100%;
+    }
```
paul-balchin-ibm, line 91: Before: the actions container is the "same size" as the side panel. 
- this worked when using absolute values like`rems` ("Make the side panel and actions container 20rems wide"), 
- but broke when using relative values like `%` ("Make the side panel 75% the width of the page, and the actions container 75% of the side panel"). This issue only happens where `size` is `2xl`.

After: the actions container is a static 100% the width of the side panel.

PR #4011, opened 2024-01-05
fix: update format and lint

davidmenendez:
@paul-balchin-ibm and i noticed that stylelint wasn't working in vscode.

during this time i noticed that the lint and format commands needed to be updated and fixed. namely to extend to our other repos. i took a lot of these updates directly from [carbon](https://github.com/carbon-design-system/carbon/blob/main/package.json).

before this is merged the `format` and `lint` commands should be tested locally.


matthewgallo (changes requested):
Looks like the linter caught some [stylelint issues](https://github.com/carbon-design-system/ibm-products/actions/runs/7425544799/job/20207417730?pr=4011)

davidmenendez:
> Looks like the linter caught some [stylelint issues](https://github.com/carbon-design-system/ibm-products/actions/runs/7425544799/job/20207417730?pr=4011)

@matthewgallo yeah i figured if we were satisfied with the updates i'd go ahead and fix those linter issues in a follow up commit

matthewgallo (approved):
Tested both `format` and `lint` locally, this will especially be nice to have fixed with future additions to the community package 🎉

PR #4019, opened 2024-01-09
fix: add min width to 2xl size

lee-chase:
Adjusts the 2xl side panel so that it maintains at least the size of lg on smaller screens.

#### What did you change?

For 2xl the width is now the same as the lg, but with a min-width of 75%;

Ran up storybook

@elycheea that PR I suggested, not sure what Camerons github handle is. There are more considerations than the width, which Cameron may wish to consider, but just wanted to be sure this was not a structural fix.


elycheea:
FYI @cameroncalder

Lee noticed that at the smallest screen size, the side panel at `2xl` follows 75% width as is specified in the guidance.

| `lg` at 320 | `2xl` at 320 |
|------------|--------------|
image | image

Lee’s PR changes that so that at the small breakpoints, the side panel fills the full width. The guidance doesn’t mention this particular behavior so Carrie suggested we run it by you as well.

cameroncalder:
This fix in behavior makes sense to me—to have it expand full width at the smaller screen sizes. At the smaller breakpoint, I would suggest we make `lg` and `2xl` follow the same button positioning logic as the smaller variants: 2 buttons or less = side by side and full width, then stack if there are 3. Currently, both `lg` and `2xl` get a bit fussy and inconsistent at the smaller screen sizes.

lee-chase:
> @cameroncalder The change in button position at `2xl` might have just been a change Paul and I missed in another recent PR. Will followup with Lee when I catch him. 😤

So the action set behavior is defined both in the ActionSet component and SidePanel despite it also being used in other components. 

I think a separate PR out to move all the behavior to the ActionSet and use @container queries to update the layout. This will lead to a slight head scratcher in SidePanel where the width of the use of a 1px left border and box-sizing border box (hint the width of the ActionSet on a 2xl SidePanel is 2xl - 1.

Issue #3505, opened 2023-09-26
Investigate screen reader interpreting cells as groupings


elycheea:
5. Table contents when navigating horizontally 
- When using CNTRL+Option+L/R or Up/Down arrows to navigate horizontally through the table with the screen reader, it reads the the contents of a cell as a group, saying the same content three times.  To minimize the amount of times you need to arrow L and R, the content inside the cell shouldn't be inside a group 
![Screenshot 2023-07-02 at 12 33 39 PM](https://github.com/carbon-design-system/ibm-products/assets/21047571/0a6cc3a4-c5a6-430c-a435-e64207df4219)

Current:

_Originally posted by @kristastarr in https://github.com/carbon-design-system/ibm-products/issues/3118#issuecomment-1616746078_

davidmenendez:
@kristastarr when you get a chance can you talk another look at this? i made some adjustments in #3957 that should alleviate some accessibility issues

kristastarr:
I tested "Basic usage" and "Selectable rows" since it looks like these changes are to the header so apply to all tables - are there specific ones you wanted me to test? 

Observations:
- When using CNRTL+OPT+ arrow key to navigate, if you press arrows consecutively without releasing CNTRL+OPT, I see the same behavior as before; when you release CNTRL+OPT with each press of arrow keys, the text content is still announced as a "group" but is only read out once, so that is an improvement 

- The fields with dates are not showing this issue.  Fields with numbers are.  

- It's confusing that column names are announced as "Last Name and Resize Column", I think because of the aria-label for the splitter.  I don't think that was happening before. 

![Screenshot 2023-12-20 at 9 15 13 AM](https://github.com/carbon-design-system/ibm-products/assets/21047571/09911532-b4c6-4e6c-a3e3-d479fd06d17a)

- I have some other observations about the splitter as well - keyboard navigation works great but the screenreader descriptions could be improved - is there an open issue where I should document those?

PR #4046, opened 2024-01-11
fix: datatable cell accessibility adjustment

davidmenendez:
Contributes to #3505

removes the `title` prop which was causing some issues with screenreaders.


matthewgallo:
Don't we need the `title` attribute for when the cells are truncated using an ellipsis?

davidmenendez:
@matthewgallo good question

based on what resources i was able to find on the matter i've concluded that when using `overflow: ellipses` the content is still available to screen readers. i've also confirmed this by testing locally using voice over. so we should be good here.

matthewgallo (approved):

Issue #3617, opened 2023-10-17
Fix alignment of inline loading in save pattern state


elycheea:
Also caught a small alignment issue in the [Saving](https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-patterns-saving-saving--manual) pattern —

<img width="681" alt="image" src="https://github.com/carbon-design-system/ibm-products/assets/1717398/8cd36c39-0a91-4ecb-a518-5364b69927a2">

The loading icon is not vertically centered.

_Originally posted by @elycheea in https://github.com/carbon-design-system/ibm-products/pull/3571#pullrequestreview-1675083624_
            
**Note:** With SCSS changes, you may need to update the styles snapshot. (Can find this in the package.json.)

```
yarn test:c4p:snapshot
```
```[tasklist]
### Acceptance criteria
- [ ] Vertically center the loading icon when saving
```

Pinaka-Pani-18:
I would like to work on this issue, Could you please assign me?

hamzailyas237:
Is this issue still persists ? I would like to work on this.

PR #4047, opened 2024-01-12
fix(saving): saving loader alignment

anamikaanu96:
Contributes to #3617 

{{Fix alignment of inline loading in save pattern state}}

#### What did you change? 
Added a style in packages/ibm-products-styles/src/components/Saving/_saving.scss 
min-block-size was changed to auto from 2rem


packages/ibm-products-styles/src/components/Saving/_saving.scss
```diff
@@ -12,6 +12,7 @@
 @use '../../global/styles/project-settings' as c4p-settings;
 
 $block-class: #{c4p-settings.$pkg-prefix}--saving;
+// $carbon-prefix: 'cds';
```
elycheea, line 15: Looks like you figured out you could use the `carbon-prefix` from settings. :)
We can remove this line now.

elycheea (commented):
Nice! Just need to remove comments we no longer need and to run `yarn format` to address any spacing / line length issues. 🎉

anamikaanu96:

davidmenendez (changes requested):
looks good. just need to remove line 15

elycheea (approved):

PR #4048, opened 2024-01-12
test(Datagrid): beginning of filtering test coverage

matthewgallo:
Contributes to #3942 

Adds test coverage to at least 80% to the following Datagrid files related to filtering:

```
FilterPanel.js
FilterFlyout.js
FilterSummary.js
useSubscribeToEventEmitter.js
useShouldDisableButtons.js
constants.js
```

Most of the changes are in `Datagrid.test.js`, however, I did clean up a few spots where a prop had a default value but was a required prop so it didn't need a default value.

I also fixed a bug in the `FilterSummary` where we never used the provided `ref` from `forwardRef` if one was passed in, instead we accidentally always use our own. Also added tests for the `FilterSummary`.
#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Datagrid.test.js
packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/FilterFlyout.js
packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/FilterPanel.js
packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/hooks/useShouldDisableButtons.js
packages/ibm-products/src/components/Datagrid/Extensions/Filtering/Flyout.stories.js
packages/ibm-products/src/components/FilterSummary/FilterSummary.js
packages/ibm-products/src/components/FilterSummary/FilterSummary.test.js
```
Made sure that all Datagrid tests pass and everything works as expected in Storybook still

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid.test.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid.test.js
@@ -62,6 +62,10 @@ import {
   FilteringUsage,
   filterProps,
 } from './Extensions/Filtering/Panel.stories';
+import {
+  FilteringUsage as FlyoutUsage,
+  filterProps as flyoutProps,
+} from './Extensions/Filtering/Flyout.stories';
 const { click, hover, unhover } = userEvent.setup({
   // delay: null, // prev version
   advanceTimers: jest.advanceTimersByTime,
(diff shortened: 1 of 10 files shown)


packages/ibm-products/src/components/Datagrid/Datagrid.test.js
```diff
@@ -2282,17 +2286,30 @@ describe(componentName, () => {
...
   it('should test basic interactions of filter panel', async () => {
+    const updatedFilterProps = { ...filterProps };
+    // Removing certain properties to test default function parameters in FilterPanel
+    delete updatedFilterProps.panelTitle;
+    delete updatedFilterProps.closeIconDescription;
+    delete updatedFilterProps.updateMethod;
+    delete updatedFilterProps.primaryActionLabel;
+    delete updatedFilterProps.secondaryActionLabel;
+    delete updatedFilterProps.onPanelOpen;
+    delete updatedFilterProps.onPanelClose;
```
davidmenendez, line 2307: would it be easier to just update `updatedFilterProps` to be the props would want and use `...rest` to remove the un-needed props? or since it looks like it's being used a few times just declare a different variable without those particular props? just seeing if it's possible to cut down on redundancy

davidmenendez (commented):
looks good but just had one question before approving

packages/ibm-products/src/components/Datagrid/Datagrid.test.js
matthewgallo: Agreed, I'll update this ^

Issue #3966, opened 2023-12-12
Increase spacing in side panel when used without a subtitle


carrenelloyd:
Referencing a ticket Shaun opened on Design board 
https://github.ibm.com/CDAI-design/pal/issues/590

> ## Overview
> Currently when using the CCS side panel component without including a subtitle the result is very little white space between the title and the content. Essentially 8px padding beneath the title itself.
> 
> Not every scenario will call for a subtitle so proper spacing provisions should be considered if designers opt out of using it.
> 
> If there was a way to implement spacing-05 or spacing-06 beneath so the content below isn't so tight that'd be ideal to improve legibility.
<img width="961" alt="image" src="https://media.github.ibm.com/user/4655/files/7da08551-e23e-4fa9-9197-9483505084f8">

<img width="1726" alt="Screenshot 2023-12-12 at 9 52 10 AM" src="https://media.github.ibm.com/user/4655/files/7fa03865-1fe9-4358-944f-54855f55b858">

Please see [original issue](https://github.ibm.com/CDAI-design/pal/issues/590) for screenshots of desired spacing in context.

elycheea:
@makafsal We might still need a v1 PR for this if you don’t mind opening one to `main_v1` as well. :)

makafsal:
@elycheea Sure.

PR #4050, opened 2024-01-15
fix(side panel): increase title spacing in side panel without subtitle

makafsal:
Contributes to #3966 

Changed padding to spacing-05 when there is no subtitle, if a subtitle is present then padding would be spacing-03

#### What did you change?
`packages/ibm-products-styles/src/components/SidePanel/_side-panel.scss`

```
.#{$block-class}__title-container {
    ...
    padding: $spacing-05;
    ...
    &:has(~ .#{$block-class}__subtitle-text) {
      padding-bottom: $spacing-03;
    }
    ...
```

Manually tested in Storybook


makafsal:

makafsal:
recheck

davidmenendez (approved):

Issue #3819, opened 2023-11-16
Last hovered row stays in that state when you leave the datagrid table using useRowIsMouseOver 


el-moshme:
## Brief description
The issue is shown in the example in docs
https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary--is-hover-on-row
I would expect the row hover behaviour to stop when the mouse is not over the row any longer 

## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)

## Detailed description

> Describe in detail the issue you're having.

> Is this issue related to a specific component?
Datagrid

> What did you expect to happen? What happened instead? What would you like to
> see changed?
When no longer hovering a row, the hover behaviour stops 

> What browser are you working in?
Chrome 

> What version of the @carbon/ibm-products 
> package are you using?
Issue is visible in demo 
https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary--is-hover-on-row

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. [Step one](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-canary--is-hover-on-row)
2. Step two hover over table rows
3. move mouse away from table. 

> Please create a reduced test case in CodeSandbox
Not required, as the example is in the docs. 

## Additional information

- Screenshots or code

elycheea:
@Ratheeshrajan can you help @anamikaanu96 with learning the Datagrid enough to address this one. @anamikaanu96 you can take this issue on _after_ the alignment issue.

PR #4061, opened 2024-01-17
fix(Datagrid): resolve useRowIsMouseOver issue

anamikaanu96:
Contributes to #3819

{{Fix the issue of Last hovered row stays in the datagrid table when not hovering using useRowIsMouseOver}}

#### What did you change? Added 2 variables in the object of packages/ibm-products/src/components/Datagrid/useRowIsMouseOver.js
to track the hovering. Then destructuring it and setting the value of function in packages/ibm-products/src/components/Datagrid/Datagrid/DatagridRow.js


packages/ibm-products/src/components/Datagrid/useRowIsMouseOver.js
```diff
@@ -29,7 +29,11 @@ const useRowIsMouseOver = (hooks) => {
       isMouseOver: row.index === mouseOverRowIndex,
     }));
 
-    Object.assign(instance, { rows: rowsWithMouseOver });
+    Object.assign(instance, {
+      rows: rowsWithMouseOver,
+      mouseHover: true,
```
matthewgallo, line 34: Can you rename this property? We've been prefixing values like this with `with`. For example, see [here](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products/src/components/Datagrid/useStickyColumn.js#L103).

Usually we've just been replacing `use` from the name of the hook with `with`. In this case, `withRowIsMouseOver`.

packages/ibm-products-styles/src/components/Saving/_saving.scss
```diff
@@ -12,6 +12,7 @@
 @use '../../global/styles/project-settings' as c4p-settings;
 
 $block-class: #{c4p-settings.$pkg-prefix}--saving;
+// $carbon-prefix: 'cds';
```
davidmenendez, line 15: if we're not using this we should just remove it

davidmenendez (changes requested):
looks good to me. just had one small comment. not sure if @matthewgallo has anything else since he's the expert haha

matthewgallo:
Looks good to me too, just need to remove the comment like @davidmenendez [suggested](https://github.com/carbon-design-system/ibm-products/pull/4061#discussion_r1463960547)

packages/ibm-products-styles/src/components/Saving/_saving.scss
anamikaanu96: Removed the comment @davidmenendez

davidmenendez (approved):

Issue #4058, opened 2024-01-16
New Component: DelimitedList


paul-balchin-ibm:
# DelimitedList

_Please refer to #4052 for more details about migrating legacy Security components to **Products v2**._

## Brief description

Migrate **DelimitedList** from `Products v1 / Security / DelimitedList` to **Products v2**. 

References: [GitHub](https://github.com/carbon-design-system/ibm-products/blob/main_v1/packages/security/src/components/DelimitedList/DelimitedList.js), [Storybook](https://v1-ibm-products.carbondesignsystem.com/?path=/story/security-components-delimitedlist--default).

For migration purposes:
- New component architecture must meet **Products v2**'s code specifications.
- The design and behavior of version `v2` must match version `v1` exactly.

## Tasks

Before starting work on this epic, please review and complete the following.

- [x] Initial review of design/existing code. Refer to existing component specifications: [GitHub](https://github.com/carbon-design-system/ibm-products/blob/main_v1/packages/security/src/components/DelimitedList/DelimitedList.js), [Storybook](https://v1-ibm-products.carbondesignsystem.com/?path=/story/security-components-delimitedlist--default).

### Working in Carbon for IBM Products package

- [x] Have you reviewed our [contribution guidelines](https://github.com/carbon-design-system/ibm-products/blob/master/.github/CONTRIBUTING.md)?
- [x] Have you noted our [code guidelines](https://github.com/carbon-design-system/ibm-products/blob/master/docs/CODE_GUIDELINES.md)?
- [x] Did you use the CLI to scaffold your new component?

PR #4069, opened 2024-01-18
feat(DelimitedList): new component DelimitedList

paul-balchin-ibm:
Contributes to #4058.

Migrating **DelimitedList** from `Products v1 / Security / DelimitedList`.

`v1` references: [GitHub](https://github.com/carbon-design-system/ibm-products/blob/main_v1/packages/security/src/components/DelimitedList/DelimitedList.js), [Storybook](https://v1-ibm-products.carbondesignsystem.com/?path=/story/security-components-delimitedlist--default).

#### What did you change?

This is a new `v2` component.

Note: after review with the **Products** and **Design** teams, we decided not to migrate the "append count" feature.
- The original code is inconsistent when trying to make it work, see below.
- Component description is "show count on hover", but is coded as "append count to end of list".
- Omitting this is expected to cause the least amount of disruption when other teams migrate.

The `v1` feature is supposed to add a "[count of total items]" to the end of the list, but only when truncated. On review it was determined that this feature didn't always show the "[count]". This has always been a problem trying to use JavaScript to determine the state of a CSS-only state.

To see this inconsistency, follow this `v1` [Storybook link](https://v1-ibm-products.carbondesignsystem.com/?path=/docs/security-components-delimitedlist--default) and switch between the **Canvas** and **Docs** tabs. Reload the page and try again.

Sometimes it works

image

Sometimes it doesn't

image

- [x] Storybook review.
- [x] Design review by Bridget Ndege.


packages/ibm-products/src/components/DelimitedList/DelimitedList.test.js
```diff
@@ -0,0 +1,58 @@
...
+const blockClass = `${pkg.prefix}--delimited-list`;
+const componentName = DelimitedList.displayName;
+
+// values to use
+const className = `class-${uuidv4()}`;
+const dataTestId = uuidv4();
+
+describe(componentName, () => {
+  it('renders a component DelimitedList', async () => {
+    render(<DelimitedList> </DelimitedList>);
```
matthewgallo, line 25: I would render this component via self closing tag since it doesn't require any children ie
```jsx
render(<DelimitedList />);
```
As well for the other tests below.

packages/ibm-products/src/components/DelimitedList/DelimitedList.test.js
```diff
@@ -0,0 +1,58 @@
...
+    expect(screen.getByRole('main')).toHaveClass(blockClass);
```
matthewgallo, line 26: You'll need to do a little refactoring here after removing the role of main in the component.

packages/ibm-products/src/components/DelimitedList/DelimitedList.test.js
```diff
@@ -0,0 +1,58 @@
+/**
```
matthewgallo, line 1: You might want to also add a test for the `delimiter` prop, making sure that the component uses what is passed to the `delimiter`. Same thing for the `truncate` prop.

paul-balchin-ibm:
Updated as per comments. Thanks, Matt.

matthewgallo (approved):
We've historically tried to avoid using `document.querySelectorAll` and `document.querySelector` directly in tests but I think it will be ok as something to revisit (see more [here](https://testing-library.com/docs/queries/about/#manual-queries)).

Issue #3979, opened 2023-12-14
Community package: styles need to be exported


matthewgallo:
## What will this achieve?

A follow up item from https://github.com/carbon-design-system/ibm-products/pull/3970, we need to export all the styles so that the styles for the components in this package can be consumed by anyone using these components. Should be able to follow the approach from `packages/ibm-products`.

PR #4078, opened 2024-01-18
chore: export community styles

davidmenendez:
Contributes to #3979


matthewgallo (approved):
One thing I did notice is that the styles for the `community` package are not imported in our storybook. This is technically outside of the scope of https://github.com/carbon-design-system/ibm-products/issues/3979 so I think it's fine to address in a follow up PR.

davidmenendez:
> One thing I did notice is that the styles for the `community` package are not imported in our storybook. This is technically outside of the scope of #3979 so I think it's fine to address in a follow up PR.

ok cool i'll open up a separate issue and PR

Issue #3499, opened 2023-09-25
Focus not being trapped within tearsheets


renato-souzaf:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

- Carbon for IBM Products v1.57.0

Focus is not being trapped within component. 
Scenario: 

Once the focus hits the primary/secondary button, and the user press the tab key

Expected: the focus should be moved to the next focusable element *within* the component.
Result: the focus get lost in some random element outside the component.

## Steps to reproduce the issue
Issue reproduceable in the docs:

Step 1: navigate to https://ibm-products.carbondesignsystem.com/?path=/docs/ibm-products-patterns-create-flows-createtearsheetnarrow--docs

Step 2: under "example usage -> default", click in "Open CreateTearsheetNarrow"

Step 3: press tab key until the secondary button is focused.

Step 4: press tab key and check the result.

Expected: the focus should be moved to "Topic name" text input
Result: the focus get lost in some random element outside the component.

## Additional information
This happens only when moving foward. If the first focusable element is focused, and the user press shift+tab, then the focus moves to the next element correctly.

philljenkins:
This keyboard focus issue is impacting the CIO's "Accessible workplace" application that is required to be fully accessible because its primary users are IBM employees with disabilities. 

How can I review to ensure the issue is fixed?

References:
Accessible Workplace POC (no back end attached) tearsheet: https://awpoc.wdc1a.ciocloud.nonprod.intranet.ibm.com/#
Carbon for IBM products component: https://pages.github.ibm.com/cdai-design/pal/components/tearsheet/usage

elycheea:
@philljenkins This fix should already be available in both [2.26.0](https://github.com/carbon-design-system/ibm-products/releases/tag/%40carbon%2Fibm-products%402.26.0) and [1.71.0](https://github.com/carbon-design-system/ibm-products/releases/tag/%40carbon%2Fibm-products%401.71.0).

This issue should have been closed. Your teams can get this fix by updating to a version from 2.26.0 or later.

PR #4129, opened 2024-01-22
fix(Tearsheet): implement focus trapping within tearsheets

makafsal:
Contributes to #3499

This is to fix the focus trap issue in the Tearsheet component

#### What did you change?

1. File: `packages/ibm-products/src/components/Tearsheet/TearsheetShell.js`

Implement a 'keydown' event handler `handleKeyDown()` function that handles the 'Tab' key operations. The function checks for focusable elements inside the Tearsheet component and properly navigates the users to the focusable elements in the Tearsheet.

Change in the logic of focus claim function `claimFocus()`

2. File: `packages/ibm-products/src/components/CreateTearsheet/preview-components/MultiStepWithIntro.js`

Add `tabIndex` property conditionally to the `RadioTile` component

Storybook tested in Chrome, Safari, Firefox, and Edge


amal-k-joy (approved):
Code looks fine and seems to be working.

Ratheeshrajan:
@amal-k-joy @makafsal worth checking this utility https://github.com/carbon-design-system/ibm-products/blob/431e041534aa1151e2f3b622e111c97c4591cbfc/packages/ibm-products/src/global/js/utils/wrapFocus.js. This is used in SidePanel

makafsal:
@Ratheeshrajan @elycheea @lee-chase 

I have tried this `wrapFocus()` utility function. Here are my observations from that:

1. When using the `usePortalTarget` hook to render the ComposedModal component, the `onBlur()` event is not firing, so I used the `onKeydown()` event to experiment with the _wrapFocus()_ function. It is working in a few cases, but it is not solving the issue in _stacked narrow tearsheet_ and _tearsheet with error state_.
2. In addition, focusing on dummy 'focus sentinal' elements may confuse users using screen readers.

However, I am trying to improve my current solution to make it more efficient, and reusable.

davidmenendez (dismissed):

packages/ibm-products/src/global/js/hooks/useFocus.js
```diff
@@ -0,0 +1,82 @@
+/**
+ * Copyright IBM Corp. 2020, 2023
```
elycheea, line 2:
```suggestion
 * Copyright IBM Corp. 2024, 2024
```

davidmenendez (approved):

Issue #4132, opened 2024-01-22
ButtonMenu not exported in v1


elycheea:
Reported [via Slack](https://ibm-casdesign.slack.com/archives/CQGR0HC05/p1705691153221679), with some quick investigation from @paul-balchin-ibm @jlongshore and @davidmenendez.

Looks like the `ButtonMenu` was added to Storybook but never exported for use in our index.
```[tasklist]
### Acceptance criteria
- [ ] Export ButtonMenu in [index.js](https://github.com/carbon-design-system/ibm-products/blob/main_v1/packages/ibm-products/src/components/index.js)
```

PR #4134, opened 2024-01-22
fix(ButtonMenu): export as canary

elycheea:
Contributes to #4132

Fixes an issue where we never actually exported ButtonMenu in our v1 when we added it as a canary component. 🙈 

#### What did you change?
Add to index.js and updated `carbon.yml` for good measure

c4p CI checks pass


davidmenendez (approved):

davidmenendez:
ty!

Issue #4033, opened 2024-01-10
<a11y>Element with "img" role has no label or an empty label


xubinnb:
===== ENVIRONMENT =====
Browser: Version 105.0.5195.125 (Official Build) (arm64)
Automatically update Chrome for all users Learn more
Test Tool Type: IBM Accessbility Checker

===== DEFECT DESCRIPTION ===== 
Problem 1: Element with "img" role has no label or an empty label

=====STEPS TO REPRODUCE =====
Go to https://dev.console.test.cloud.ibm.com/projects

Use IBM Equal Access Accessibility Checker to scan

===== EXPECTED RESULT =====
[IBM 1.1.1 Non-text Content](https://www.ibm.com/able/requirements/requirements/#1_1_1)

All non-text content that is presented to the user has a text alternative that serves the equivalent purpose.

===== SCREEN CAPTURE =====
<img width="1276" alt="image" src="https://media.github.ibm.com/user/42629/files/57acec22-33ad-4f7b-919f-c9443c801129">

elycheea:
@xubinnb Can you share what version of our library you are using? I believe @davidmenendez fixed a related issue previously so it’ll help to know if this is a new issue or regression. Thank you!

xubinnb:
@elycheea As Flannan(flannanl@ca.ibm.com)) mentioned, It seems they fix this in their v2 version. But since we are using v1, we need them to port the fix to v1. Thanks.

PR #4137, opened 2024-01-23
fix(Empty state): accessibility defect

amal-k-joy:
Contributes to #4033 

Element with "img" role has no label or an empty label in empty state components

#### What did you change?
Used the existing _illustrationDescription_ prop and set as the aria-label attribute of the SVGs used in Empty States component.

Verified in the storyboard using IBM Equal Access Accessibility Checker.


amal-k-joy:

matthewgallo (changes requested):
Hey @amal-k-joy, since the `illustrationDescription` prop is now required there are a few components within the library that use the `EmptyState` which will need to be updated and passed this new required prop (ie AddSelect, SingleAddSelect, MultiAddSelect, Datagrid, NotificationsPanel).

packages/ibm-products/src/components/EmptyStates/EmptyState.test.js
```diff
@@ -147,7 +147,7 @@ describe(name, () => {
 
   it('should render the NoDataEmptyState', async () => {
     const { container, rerender } = render(
-      <NoDataEmptyState {...defaultProps} />
+      <NoDataEmptyState {...defaultProps} illustrationDescription="Test alt text" />
```
matthewgallo, line 150: Add `illustrationDescription` to the `defaultProps` object so that you don't have to explicitly pass it in for each test. Then you can just render the component like:
```jsx
render(<NoDataEmptyState {...defaultProps} />);
```

amal-k-joy: @matthewgallo  , added the illustrationDescription to defaultProps

matthewgallo (approved):

PR #4141, opened 2024-01-23
chore: Refactor side panel header

### LINKED ISSUE #4065: Review Sidepanel title/header DOM structure for web component
The scroll behavior uses quite a bit of Javascript to achieve the animation on scroll. This can lead to the scroll event reducing the size of the title, which reduces the size of the scrollable content, which in turn causes a scroll event.

This is most obvious in the web component story with an action bar and the title half way scrolled. 

Review the DOM structure to see if it can be refactored to allow the sticky scroll behavior to function with less javascript.

lee-chase:
Closes #4065

As part of the AI Web Component work it became clear that the header of the SidePanel had a fragile structure requiring a lot of javascript to function.

This PR refactors the Header section in a manner that relies more on HTML structure and CSS sticky. In addition it should be simpler to extract as a component.

#### What did you change?

- Removed some of the code monitoring scroll and calculating transforms from Javascript. 
- Added CSS custom properties to allow the sticky animation to function with some many measurements.

NOTE: Web components does not work well with `:has` which Carbon also uses. Added some additional selectors (not used here) to support WebComponents.

Storybook.


matthewgallo (commented):
Just have a few design comments on the new interaction. One piece that sticks out is that the initial title previously began an opacity change immediately upon scrolling and was not visible by the time it reached the top of the panel. The current change keeps the initial title visible longer than expected which makes the overlap between the initial title and the condensed title a bit jarring.

Another thing I noticed is that the border is being displayed below the subtitle in this refactor and not below the title, causing the subtitle to appear to be cutoff during the scroll animation. See screenshot below:
Screenshot 2024-01-24 at 10 12 34 AM

lee-chase:
> Just have a few design comments on the new interaction. One piece that sticks out is that the initial title previously began an opacity change immediately upon scrolling and was not visible by the time it reached the top of the panel. The current change keeps the initial title visible longer than expected which makes the overlap between the initial title and the condensed title a bit jarring.
> 
> Another thing I noticed is that the border is being displayed below the subtitle in this refactor and not below the title, causing the subtitle to appear to be cutoff during the scroll animation. See screenshot below: Screenshot 2024-01-24 at 10 12 34 AM

@matthewgallo 
While this could be changed, it would then be inconsistent with the current behavior when an ActionToolbar is included.

elycheea:
@mikeolasov I tagged you originally but we’re not changing the animation to based on trigger in this specific PR.

lee-chase:
@matthewgallo changed the title (without label) animation speed to match that of the original version.

matthewgallo (approved):
🔥 Such an improvement for this component! That title animation has snowballed over time

Issue #3736, opened 2023-11-01
datagrid filter flyout cancel button does nothing


tkolanko:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

The cancel button is not active by default when opening the filter flyout and clicking the cancel button after setting a filter does not close the panel

## Steps to reproduce the issue

1. Go to https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-datagrid-datagrid-canary-extensions-flyout--flyout-batch
2. Open the filter flyout - cancel is disabled
3. Change a filter
4. Click cancel
5. Nothing happens
## Additional information

- Screenshots or code
- Notes

davidmenendez:
confirmed with @vsnichols that this is the intended behavior. @tkolanko the bug fix has been opened in #4142 and should be merged shortly. thanks for bringing this to our attention.

PR #4142, opened 2024-01-23
fix: close filter flyout on cancel

davidmenendez:
Contributes to #3736


matthewgallo (commented):
I almost think this is also a design question of whether the filter flyout is meant to be closed when you click `Cancel`. I would expect it and think it makes sense but I think I remember it being intended more as a "reset". @sdignum can you add some context here?

vsnichols:
Hi all!

- Cancel is supposed to close the filter panel without applying any changes
- If filters had already been applied previous to you opening the panel, clicking "cancel" in the panel would not clear the previously applied filters
- That is accomplished by clicking the "Clear filters" button in the filter display bar. 

Would it be possible to add a note stating this to the React storybook guidance?

paul-balchin-ibm:
IMO "Cancel" and "reset" mean two different things. 
- Cancel = drop/leave this process.
- Reset = undo my changes.

Is there a Storybook example with a "cancel" feature that we can look at? I can't find one in the Storybook preview.

PR #4144, opened 2024-01-24
fix(Empty state): accessibility defect 4033 v1

amal-k-joy:
Contributes to #4033 

Element with "img" role has no label or an empty label in empty state components

What did you change?
Used the existing illustrationDescription prop and set as the aria-label attribute of the SVGs used in Empty States component.

Verified in the storyboard using IBM Equal Access Accessibility Checker.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -73,6 +73,7 @@
     "coachmarks",
     "colindex",
     "columnheader",
+    "combobox",
     "composability",
     "csstools",
     "ctacol",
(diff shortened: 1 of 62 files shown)


matthewgallo (changes requested):
I think you also hit this issue with the PR for this on the `main` branch but there are some other component that use the EmptyState within our own library that you'll also need to update with the newly required `illustrationDescription` prop.

matthewgallo:
Also, same comment regarding the test as well https://github.com/carbon-design-system/ibm-products/pull/4137#discussion_r1465089184

matthewgallo:
Also, same comment regarding the test as well https://github.com/carbon-design-system/ibm-products/pull/4137#discussion_r1465089184

Issue #3982, opened 2023-12-15
Community governance


davidmenendez:
With the addition of the [community package](https://carbon-for-ibm-products.netlify.app/?path=/docs/community-components-example-examplecommunitycomponent--docs) to Carbon for IBM we need to establish guidelines and documentation for what the purpose of the community package is and how to contribute to it.
```[tasklist]
### Tasks
- [x] Add a README to document the package purpose
- [ ] Add documentation on governance of the package
- [x] Explore what we can adopt from [ibm-cloud-community](https://github.ibm.com/ibmcloud/cloud-pal-community#contribution-ownership-and-statuses)
- [ ] https://github.com/carbon-design-system/ibm-products/issues/4042
- [ ] https://github.com/carbon-design-system/ibm-products/issues/4624
```

```[tasklist]
### Future considerations
- [ ] Define how components may be migrated from `ibm-products-community` to `ibm-products`
- [ ] Define relationship with design process
```

davidmenendez:
<img width="1124" alt="Screenshot 2024-01-04 at 11 25 48 AM" src="https://github.com/carbon-design-system/ibm-products/assets/6370760/ea6ab360-8a3c-48f0-b9be-5a7c6e7fb232">

elycheea:
David meeting with Stephanie and Clayton tomorrow on governance for their PALs.

elycheea:
Closing this since this now overlaps with Labs/Innovation.

PR #4150, opened 2024-01-24
chore: community readme init

davidmenendez:
Contributes to #3982


packages/ibm-products-community/README.md
```diff
@@ -1,3 +1,25 @@
-# IBM Products community
+# Carbon for IBM Products Community
 
-TODO
+## About
+
+The purpose of the Community package is to host experimental components and
```
elycheea, line 5: While true, I think it’s not _only_ to host experimental components but also to enable teams to share code that may not be officially supported or documented through the design system guild.

packages/ibm-products-community/README.md
```diff
@@ -1,3 +1,25 @@
...
+patterns.
+
+Traditionally, components and patterns are defined by design, approved via DSAG,
+developed by the C4P team, and then added to the core library. The Community
+Package represents a different approach to this system. If a product team or
```
elycheea, line 10:
```suggestion
package represents a different approach to this system. If a product team or
```

packages/ibm-products-community/README.md
```diff
@@ -1,3 +1,25 @@
...
+developer believes they have organically developed something that could or
+should be a pattern and wants to share it with the larger Carbon community,.
```
elycheea, line 12: Seems like we didn’t finish this thought.

Maybe something like ...

```suggestion
should be a pattern and wants to share it with the larger Carbon for IBM 
Products community, they can contribute it for others to leverage even as
the design use cases are being defined.
```

elycheea (commented):
This is a beautifully written introduction. Just a few comments/suggestions.

elycheea (approved):

PR #4155, opened 2024-01-24
feat(ButtonMenu): export button menu item component (v1)

matthewgallo:
Follow up to #4132, this PR just exports the `ButtonMenuItem` so it can be used with the `ButtonMenu` in v1.

#### What did you change?
Export `ButtonMenuItem`


elycheea (approved):
I thought I included this but think I pulled it out when I ran into a build issue that turned out to be unrelated. Thank you for fixing it. 😅

PR #4165, opened 2024-01-25
chore: onboard @carbon/ibm-products to ibm telemetry

francinelucca:
Adds config file and dependency necessary to start tracking telemetry data via [IBM Telemetry](https://github.com/ibm-telemetry/telemetry-js)

#### What did you change?
- Updates readme telemetry disclosure
- Removes `@carbon/telemetry` dependency
- Adds @ibm/telemetry-js dependency
- Updates postinstall script to use IBM telemetry command
- Adds [telemetry.yml](packages/ibm-products/telemetry.yml) config file
    - This file was generated using [IBM Telemetry automated config generator](https://github.com/ibm-telemetry/telemetry-js-config-generator/pull/1)
- Include `telemetry.yml` in package.json files config 

#### Review
Please look through [telemetry.yml](packages/ibm-products/telemetry.yml) config file and ensure desired props/values have been captured. For more info see [Telemetry Config Schema](https://github.com/ibm-telemetry/telemetry-config-schema?tab=readme-ov-file#jsx-scope)

N/A


mattrosno:
@elycheea once this is merged, we can discuss strategies for keeping the telemetry.yml file current when you add a prop or prop value that you'd like to track, that isn't in the yml file yet.

That means either running https://github.com/ibm-telemetry/telemetry-js-config-generator again to update the telemetry.yml, or calling that in an automated way to ensure the yml file stays current.

elycheea:
Sounds good @francinelucca @mattrosno — I’m also using this as an excuse to hang out with Francine when our schedules someday align.

elycheea (approved):
:tada:

Looks good. I agree that we should probably add a CI check for this. Will add my usual thoughts outside of review. :)

elycheea:
@mattrosno @francinelucca 

Probably no surprises that I have questions on maintaining this as ...

1. **Assets are deprecated or replaced with new versions.**
    -  For instance, we are currently working on updates to user profile image ➡️  user avatar, HTTP errors ➡️ full page errors. Both are ongoing, but we don’t expect to remove them until our next major release. It may still make sense to track these so we can hopefully see adoption shift to the new one or identify a need for us to do more communication if not.
    - Also, occasionally, our components graduate. Button menu and combo button are now in the Carbon core library. Maybe still better to track for now and filter out only as / if it _actually_ becomes too noisy.
2. **Tracking across multiple major versions.** I assume we just need to add and generate the telemetry file if we also want to track our v1 package.

No blockers or anything — just thoughts — they might not even end up being as much of concerns once we start seeing our metrics 😅.

Thanks @francinelucca @mattrosno @jdharvey-ibm for the awesome work. Excited to start getting more metrics so we can start tailoring/making more informed decisions on how we maintain our library. 😤

francinelucca:
> Probably no surprises that I have questions on maintaining this as ...
> 
> 1. **Assets are deprecated or replaced with new versions.**
>    
>    * For instance, we are currently working on updates to user profile image ➡️  user avatar, HTTP errors ➡️ full page errors. Both are ongoing, but we don’t expect to remove them until our next major release. It may still make sense to track these so we can hopefully see adoption shift to the new one or identify a need for us to do more communication if not.
>    * Also, occasionally, our components graduate. Button menu and combo button are now in the Carbon core library. Maybe still better to track for now and filter out only as / if it _actually_ becomes too noisy.

Telemetry will collect metrics for all components that it finds tied into the instrumented package (in this case @carbo/ibm-products) and those do not need to be defined anywhere in the config file, you are only required to define prop names and values that you explicitly want to collect for. In this case, after a component is no longer part of the library you might want to go in and remove the props that are specific to this component in from the config file as as cleanup step, but telemetry is able to determine by itself what components to collect for based on usage.

> 2. **Tracking across multiple major versions.** I assume we just need to add and generate the telemetry file if we also want to track our v1 package.

Correct, to track different versions you just need to ensure you have a telemetry.yml file published as a part of that version, along with the @ibm/telemetry-js dependency and post install script. You are free to use the same projectId in the config file so long as the package names are the same across versions. See [Onboarding a package to Ibm Telemetry](https://github.com/ibm-telemetry/telemetry-js?tab=readme-ov-file#onboarding-a-package-to-ibm-telemetry). Happy to help with this as well.

mattrosno:
@elycheea to expand on the component renames as we expect that to frequently happen... the filtering in Kibana allows for "or" fields so if you want to see an all time adoption chart, we can easily say "let me see instances of profile image or user avatar for all time".

We're not putting in "lifecycle" logic to track renames and we're not updating past records as names are updated. Our emphasis is allowing you to query the data with a lot of flexibility so if you know about a rename, you can see that data in aggregate by filtering by the old and new names.

Issue #4003, opened 2024-01-04
Review legacy Security COMPONENTS for migration


elycheea:
Related issues:
- **[Review legacy Security OVERVIEW (Read me first)](https://github.com/carbon-design-system/ibm-products/issues/4052)**
- Review legacy Security COMPONENTS for migration `You are here`
- [Review legacy Security PATTERNS for migration](https://github.com/carbon-design-system/ibm-products/issues/4036)
- [Review legacy Security LAYOUTS for migration](https://github.com/carbon-design-system/ibm-products/issues/4037)

_This is a summary report. The details listed for each component is not comprehensive._

More details provided with each component report below.

**"Easy" to migrate.** Has an up-to-date code spec similar to **Carbon for IBM Products**.

- DelimitedList
- ICA

**"Difficult" to migrate.** Must be converted from a `class` to a `function`.

- Nav
- TruncatedList

**"Difficult" to migrate.** There are three versions to be considered before we even start.

- (Data)Decorator

**Not recommended to be migrated.** See each item for details.

- StackedNotification
- StatusIcon
- StringFormatter
- TimeIndicator
- TypeLayout

---

### Decorator and DataDecorator

This needs to be sorted out before we start: is it possible to use any _one_ as a single source of truth?

**Security** Decorator (component): [GitHub](https://github.com/carbon-design-system/ibm-products/blob/main_v1/packages/security/src/components/DataDecorator/Decorator/Decorator.js), [Storybook](https://v1-ibm-products.carbondesignsystem.com/?path=/story/security-components-decorator--default)
- **✓** Follows **Carbon for IBM Products** component architecture.

**Security** DataDecorator (pattern): [GitHub](https://github.com/carbon-design-system/ibm-products/blob/main_v1/packages/security/src/components/DataDecorator/DataDecorator.js), [Storybook](https://ibm-security.carbondesignsystem.com/?path=/story/components-decorator--default)
- **✗** See `This component must be refactored from a Class to a Function`. ([Link](https://github.com/carbon-design-system/ibm-products/issues/4052))
- **✗** Is importing a `Panel` and other components.

**Data Explorer** Decorator (component): [GitHub](https://github.ibm.com/investigate/investigate/blob/develop/src/components/Decorator/Decorator.js), [Storybook](https://pages.github.ibm.com/investigate/investigate/?path=/story/decorator--default)
- **✗** Is tightly-coupled to the CP4S Shell.

---

### DelimitedList

[...]

PR #4170, opened 2024-01-29
feat(Decorator): new component Decorator

paul-balchin-ibm:
Contributes to #4146 

Migrating **Decorator** from `Products v1 / Security / Decorator`.

`v1` references: [GitHub](https://github.com/carbon-design-system/ibm-products/blob/main_v1/packages/security/src/components/DataDecorator/Decorator/Decorator.js), [Storybook](https://v1-ibm-products.carbondesignsystem.com/?path=/story/security-components-decorator--default).

For migration purposes:
- New component architecture must meet **Products v2**'s code specifications.
- ~The design and behavior of version `v2` must match version `v1` exactly.~ Normally this would be true, but there are **three** variations of the Decorator spread across three repos. Each of them have different designs and behaviours, and after a design review not one of them can be considered "best of class". (See #4003 and scroll down to the "Decorator and DataDecorator" to review all variations.) 
- ~A design update is in progress.~ Complete.

image

|-|-|-|
| active | theme | Specify theme as "light" or "dark".
| className | className |  — | 
| href | href |  — | 
| inline | small | Still a `bool`.
| midLineTruncation | truncateValue | Specify as "end", "start", or `{...}`
| noIcon | hideIcon | Name change.
| onClick | onClick |  — | 
| onClickValue | onClickValue |  — | 
| onContextMenu | onContextMenu |  — | 
| score | score |  — | 
| scoreDescription | scoreDescription |  — | 
| scoreThresholds | scoreThresholds |  — | 
| title | valueTitle | Override the default `value` provided to the mouse tooltip.
| type | label | Name change.
| value | value |  — | 
| — | disabled | New state option for button types.
| — | onClickLabel | New callback function.
| ~fitValue~  | (deprecated) | Is now the default behavior.
| ~noBorderRadius~ | (deprecated) | No longer supported.
| ~noType~ | (deprecated) | A decorator (i.e. a value "decorated with" an identifier)<br/> must show a label in the form of an icon, some text, or both.

#### What did you change?

This is a new `v2` component.

- [x] Storybook review.
- [x] Design update by Cameron Calder and Bridget Ndege.
- [x] Design review by Cameron Calder and Bridget Ndege.


packages/ibm-products/src/components/Decorator/Decorator.js
```diff
@@ -0,0 +1,354 @@
+/**
+ * Copyright IBM Corp. 2024, 2024
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useMemo } from 'react';
```
davidmenendez, line 8: i don't think anything here really _needs_ `useMemo`

packages/ibm-products/src/components/Decorator/Decorator.js
```diff
@@ -0,0 +1,354 @@
...
+    const _labelTitle = useMemo(() => {
+      return labelTitle && labelTitle(score, scoreThresholds, magnitude);
+    }, [magnitude, score, labelTitle, scoreThresholds]);
+
+    // If a "midline" truncation has been defined,
+    // then use the formatted midline value generated here,
+    // else refer to the original value.
+    const _value = useMemo(() => {
+      if (truncateValue?.maxLength) {
+        return truncate(
```
davidmenendez, line 77: you can make this parameter signature cleaner by just passing in `(value, truncateValue)` and then destructing the values you need in `truncate`

packages/ibm-products/src/components/Decorator/Decorator.js
```diff
@@ -0,0 +1,354 @@
...
+    const handleOnClickValue = (event) => {
+      onClickValue(event, { score, label, value, magnitude });
+    };
+
+    const handleOnContextMenu = (event) => {
+      onContextMenu && onContextMenu(event, { score, label, value, magnitude });
+    };
+
+    // RETURN DUAL BUTTONS
+    if (onClickLabel && onClickValue) {
```
davidmenendez, line 122: i think a lesson we've learned over time is to try to make our components dumb when possible. while it might make sense to infer what variant the user needs based on props we tend to run into issues where because something isn't set directly users get confused about why they ended up with Y variant when they wanted X. to save us any headache i recommend removing these inferred conditions and just setting a new prop called `variant` or `type` or something and just having the user manually enter `single`, `dual`, etc

packages/ibm-products/src/components/Decorator/constants.js
```diff
@@ -0,0 +1,40 @@
+/**
```
davidmenendez, line 1: i may have left confusing feedback previously.

this file should be named utils since it has utility functions. if you wanted a constants file to maintain things like `const BENIGN = 'Benign'` this would be the place to do it

packages/ibm-products/src/components/Decorator/constants.js
```diff
@@ -0,0 +1,40 @@
...
+    if (score >= thresholds[3]) {
+      return 'Critical';
+    }
+  }
+
+  return 'Unknown';
+};
+
+const truncate = (inputText, maxLength, front, back) => {
+  let truncatedText = inputText;
```
davidmenendez, line 31: whenever possible we should try to avoid unnecessary mutation. we can simplify this to just be something like
```
if (thing) return ${inputText.substring.otherstuff
return inputText;
```

davidmenendez (changes requested):
looking good! just a few comments 💪

packages/ibm-products/src/components/Decorator/Decorator.js
paul-balchin-ibm: I was trying to avoid unnecessary computations on rerenders if props changed, but on reflection Decorators should not change over time. I'll remove them.

packages/ibm-products/src/components/Decorator/Decorator.js
paul-balchin-ibm: I like it.

What better options, if any, are there for enforcing requirements? E.g. if kind="dual buttons", then onClickLabel and onClickValue must be defined as well. Can it be done through PropTypes? Should a console warning/error be logged? Throw an error (seems harsh)?

davidmenendez: we do have a utility for this kind of thing, though honestly i don't think it's super necessary to be this prescriptive. but if you so choose to there's a bunch of examples of it being used through the repo. just search for `.isRequired.if`

packages/ibm-products/src/components/Decorator/Decorator.js
```diff
@@ -0,0 +1,354 @@
...
+import { pkg } from '../../settings';
+
+import { getMagnitude, truncate } from './constants';
+import { DecoratorIcon } from './DecoratorIcon';
+
+const blockClass = `${pkg.prefix}--decorator`;
+const componentName = 'Decorator';
+
+const defaults = {
+  labelTitle: (score, scoreThresholds, magnitude) => {
```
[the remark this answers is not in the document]
paul-balchin-ibm: Done.

packages/ibm-products/src/components/Decorator/Decorator.js
paul-balchin-ibm: Done.

packages/ibm-products/src/components/Decorator/Decorator.js
paul-balchin-ibm: `.isRequired.if` was a nice idea, but VSCode kept throwing an `"if" is not a function` error for my file alone. I verified against other examples, including looking for imports, but found no real difference. ¯\_(ツ)_/¯ 

I dropped it in favour of documentation.

packages/ibm-products/src/components/Decorator/constants.js
paul-balchin-ibm: Done.

davidmenendez (approved):

Issue #4171, opened 2024-01-29
Setup environment to support typescript


matthewgallo:
## What will this achieve?

To better align with Carbon and other products, we will be incorporating typescript into our codebase. In doing so there will be some infrastructure changes to our project that will be required in order to support this, some of these being babel, jest, eslint, storybook configs.

## How will success be measured?

When our project can be built and run with typescript incorporated (`ci-check` passes and storybook still runs as expected).

PR #4172, opened 2024-01-29
build: add support for typescript

matthewgallo:
Contributes to #4171 

This PR includes the initial changes required to support typescript in our monorepo (`ibm-products`, `ibm-products-community`, `core`). In doing so, this involved updating our babel config package (`config/babel-preset-ibm-cloud-cognitive`) which is used across other packages as a shared config, similarly there were updates required to the jest config package (`config/jest-config-ibm-cloud-cognitive`) to support files/components using typescript.

I also discovered after getting everything running via the new build script that we had some circular dependencies, those have been addressed in this PR as well with some refactoring.

I also removed the local copy of `config/typescript-config-carbon` that I had initially in this PR and was able to pull in this config from `node_modules`.

I have migrated `Saving.tsx` to make sure that everything is working as expected since this is one of our smaller components. Another way I tested these changes was by building `ibm-products` and copying the built directories into an example app's `node_modules`.

#### What did you change?
Config related files (babel, eslint, storybook, jest)

Ran new build script and double checked built directories.


config/babel-preset-ibm-cloud-cognitive/index.js
```diff
@@ -1,37 +1,34 @@
 /**
- * Copyright IBM Corp. 2020, 2021
+ * Copyright IBM Corp. 2016, 2024
```
elycheea, line 2:
```suggestion
 * Copyright IBM Corp. 2020, 2024
```

Was this intentionally changed?

matthewgallo: No, this was just accidental 😅

elycheea (dismissed):
Quite a few circular dependencies you had to resolve 🙈
We should definitely take more note going forward.

Also tested locally with a test app and seems good to me, but probably worth waiting until after tomorrow’s release to merge anyway. :)

elycheea (approved):

PR #4179, opened 2024-01-30
feat(Datagrid): update filtering copyright headers, review eslint disable rules, remove filter feature flags (v1)

matthewgallo:
Contributes to #4156 

This PR mostly contains copyright header updates, some files were missing them altogether while some needed to be updated from WKC copyright header to ours. Carbon uses a rollup package (`rollup-plugin-strip-banner`) to enforce copyright header formatting which is included in the [typescript setup](https://github.com/carbon-design-system/ibm-products/pull/4172/files#diff-110b2eef5f7121d5dd4028f07e7ea48a6c7dc4d7861e69fb5e5640773c3a938aR149) PR.

I also removed the filtering feature flags after going through the release review checklist and double checking test coverage.

#### What did you change?
Filtering files, updated copyright headers and added a few comments.

Storybook


elycheea (approved):
🔥 
Same comment — if we’re removing the feature flag, probably worth calling it a feature!

Issue #3624, opened 2023-10-17
Prevent draggable items from being dropped into locked column position


elycheea:
> Oooh, thanks for fixing this. So this implementation still allows us to drag to the position of a locked column but prevents the reordering from updating? I think ideally, we should prevent dropping into the locked column position, but understand if that’s not possible with dnd-kit now. 🤔

_Originally posted by @elycheea in https://github.com/carbon-design-system/ibm-products/pull/3598#pullrequestreview-1677306919_

Ratheeshrajan:
Closing as #4180 & #4182

PR #4182, opened 2024-01-31
fix(DataGrid): customize column prevent disabled drop v1

Ratheeshrajan:
Contributes to #3624 

This should prevent dropping into the locked column position in the customize column modal.

#### What did you change?
 Transform styles are not being updated if the item is in locked condition.
`packages/ibm-products/src/components/Datagrid/Datagrid/DraggableElement.js`

Storybook


Ratheeshrajan:
The accessibility checker requires that the visible text/label match the `area-label`, in my opinion. The addition of the new prop `menuArearLabel` could potentially lead to an accessibility violation. Take a look at this example:
1. Maintain the label on the `Primary button` menu button.
2. Add `menuAreaLabel` something different text
3. Restart the accessibility checker.

In my openion we only need to add `areaLabel={label}` in story.  @matthewgallo could you add your thoughts.

Issue #4091, opened 2024-01-19
compulsoryColumns is deselect/select by select all checkbox in customize datagrid column 


SeonyuK:
## What package(s) are you using?

- [X] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

In the datagrid component, when customizing the columns in the modal the compulsoryColumns can be deselected by the Select all button.

 
> Is this issue related to a specific component?

DataGrid

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Should not be deselected the compulsoryColumns

> What browser are you working in?

Chrome 

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

V2.18.1

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

## Steps to reproduce the issue

1. Step one
Click the customize column button
2. Step two
Uncheck the Select All button 

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-products/tree/main/examples/carbon-for-ibm-products/SidePanel

## Additional information
![Screenshot 2024-01-19 at 3 22 23 PM](https://github.com/carbon-design-system/ibm-products/assets/144151347/e84f1fae-3ea1-46c6-ad8c-c855a3af4900)

- Screenshots or code
- Notes

```[tasklist]
### Acceptance criteria
- [ ] Exclude frozen columns from select all
```

Ratheeshrajan:
Closing as #4191 & #4194 merged.

PR #4191, opened 2024-02-01
fix(DataGrid): column customisation not unselect locked columns

Ratheeshrajan:
Contributes to #4091 

In the datagrid component, when customizing the columns in the modal the compulsory columns can be deselected by the Select all button.

#### What did you change?
Filtered the locked column when select all works.

Storybook


packages/ibm-products/src/components/Datagrid/Datagrid/addons/CustomizeColumns/Columns.js
```diff
@@ -47,6 +47,7 @@ const Columns = ({
     [columns, setColumnsObject]
   );
 
+  const filteredStickyColumn = columns?.filter((item) => !item.sticky)?.map(item=> item);
```
amal-k-joy, line 50: @Ratheeshrajan 
Do we need this .map() after .filter() ?

Ratheeshrajan: Sorry, my mistake. removed

matthewgallo (approved):
Working as expected now and extra `.map` has been removed.

PR #4195, opened 2024-02-01
feat: Action set auto size n stack

lee-chase:
Updated the action set to manage it's own button sizing and stacking.

#### What did you change?

Action set plus callers passing a size property.

Storybook


lee-chase:
@elycheea @matthewgallo happy with this, although had to skip a couple of tests as I do not know how to get styles to apply during a test.

0image

elycheea:
@lee-chase Going to close this PR for now — this probably has extensive changes since those components have been migrated to TypeScript as well. I also recall we were exploring implementing this at the Button level too? 🤔

Issue #2968, opened 2023-05-08
Refactor empty state to reduce redundancy


davidmenendez:
## What will this achieve?

while working on fixing a style issue on the `Tearsheet` component i noticed that all the other variants of tearsheet are using redundant code. we should look to improve the code quality here by doing something similar to how cards or add select works by implementing a base component that they all use. `EmptyStateContent` is kind of like that, but i think it could be further optimized to be more efficient.

davidmenendez:
after looking at this with fresh eyes i still think that a refactor here would be good, but the way i would do it would cause **breaking changes**.

basically i think that we should step away from the named export pattern `NoTagsEmptyState` `NotFoundEmptyState` etc to a variant pattern similar to the way carbon does with `Button`.

my proposal would look more like `<EmptyState variant='notags' />` the benefit here is that we could drastically cut down on boilerplate code. a base component implementation similar to the way cards work would be nice, but unfortunately storybook makes this a bit difficult by not supporting shared proptypes. the proptypes and comments have to be declared in the export in order to propagate the documentation. because of that every single instance of empty state has to redeclare proptypes and create lots of redundant code.

the only thing that's different between each empty state is the illustration. we can take advantage of `Suspense` and `lazy` from react in order to lazy load all the svg's and then conditionally render the svg without loading all the unnecessary svgs.

something like this could be implemented to use a single file
```
const NoTags = lazy(() => import('./notags.svg'));
const NotFound = lazy(() => import('./notfound.svg'));
...
const Empty = ({ variant }) => (
<div className={`${prefix}-${variant}`>
  <Suspense>
    {variant === 'notags' && <NoTags />}
    {variant === 'notfound' && <NotFound />}
  </Suspense
</div>
)
```

@matthewgallo @lee-chase if you have any opinions on this please feel free to add to the discussion!

matthewgallo:
I really like the approach you've outlined. In the [very initial implementation](https://github.com/carbon-design-system/ibm-products/pull/159/files#diff-c73fc69f5569e7296513bfa0f2b5745e3630f9b36fb76d2174d9661fe9f897c3R42) of `EmptyState`, we had a dynamic `require` to load the correct svgs. If I remember correctly, that worked fine locally but when our package was built there were issues loading the svg because of the dynamic require, not sure if it was webpack related or what.

[...]

PR #4197, opened 2024-02-01
feat: empty state v2

davidmenendez:
Contributes to #2968

this also updates the husky config to ignore the examples folder, which causes issues with merge conflicts


packages/ibm-products/src/components/EmptyStates/assets/ErrorIllustration.js
```diff
@@ -195,3 +195,5 @@ ErrorIllustration.propTypes = {
   size: PropTypes.oneOf(['lg', 'sm']),
   theme: PropTypes.oneOf(['light', 'dark']),
 };
+
+export default ErrorIllustration;
```
matthewgallo, line 199: The illustration components already use a named export [here](https://github.com/carbon-design-system/ibm-products/pull/4197/files#diff-f377d5706753b86522e93b86a1bf5f306c49f06683312b2788fdd39e884b6097R20). I don't think we really need to add a default export as well. But maybe there is a reason for it that I missed.

packages/ibm-products/src/global/js/package-settings.js
```diff
@@ -25,6 +25,7 @@ const defaults = {
     Datagrid: true,
     EditInPlace: true,
     EmptyState: true,
+    EmptyStateV2: true,
```
matthewgallo, line 28: This should be `false` until the component is fully reviewed and released.
```suggestion
    EmptyStateV2: false,
```

packages/ibm-products/src/components/EmptyStates/assets/ErrorIllustration.js
matthewgallo: Adding note here so we don't forget, this is required to use `React.lazy`

matthewgallo (approved):

PR #4208, opened 2024-02-02
feat(DescriptionList): new component DescriptionList

jlongshore:
Contributes to [issue 4189](https://github.com/carbon-design-system/ibm-products/issues/4189)

Migrate **TypeLayout** from `Products v1 / Security / TypeLayout` to **Products v2** as `DescriptionList`. 

`v1` References: [GitHub](https://github.com/carbon-design-system/ibm-products/blob/main_v1/packages/security/src/components/TypeLayout/TypeLayout.js), [Storybook](https://v1-carbon-for-ibm-products.netlify.app/?path=/docs/security-components-typelayout--default).

#### What did you change?

- Name change (TypeLayout -> DescriptionList)
- Removed deprecated prop `bordered` as existing boolean prop `border` replaced...

**Prop updates:**
|-|-|-|
| bordered |  — | Deprecated prop removed, use `border` boolean prop |

- [x] Storybook review.
- [x] Tests migrated and added to.
- [x] Design review by Sarah Pedregon.


elycheea:
@jlongshore sorry, you just fixed conflicts and now you get more. 😹  New components opening up and merging faster than we can keep track. 😅  (Probably okay to wait since you’ll likely get another if `BigNumbers` get merged next.)

packages/ibm-products/src/components/DescriptionList/DescriptionList.js
elycheea, line 1: I see what @paul-balchin-ibm’s concerns were in #4003 on just re-exporting `StructuredList`. 🤔  Not sure how much benefit we really gain here since this is really just meant to restyle the `StructuredList`? Maybe easier migration, but then again, this whole component is renamed anyway so probably not.

Gonna tag @paul-balchin-ibm and @matthewgallo to weigh in.

packages/ibm-products/src/components/DescriptionList/utils/enums.js
elycheea, line 1: Probably extrapolate @davidmenendez’s previous comments on just calling this `constants`.

elycheea (commented):
Mostly structural questions ... and I’ve proceeded to tag almost everyone to weigh in 🙈. Probably worth some discussion next standup unless y’all want to duke it out over Slack today. (Unfortunately, I’m calling it early for 🐉 🧧 today.)

davidmenendez:
hmmm yes this is an interesting situation.

on the one hand this is just a slight variant of an existing carbon component. there's not really much custom code here aside from the styling, which makes me feel like this should probably be handled at the product level.

Screenshot 2024-02-12 at 1 18 35 PM
Screenshot 2024-02-12 at 1 18 39 PM


on the other hand this could be an interesting avenue for us to explore in terms of what kind of patterns and components we support. it might make sense for carbon to own the slight variants of it's components, but this also sounds like it could be candidate for the community treatment.

in theory i support this. **if** we decide that we want to host different variants / flavors / knock-offs of carbon components then i think we might just want to do that from the style library. i don't think we need literally just re-export the carbon components.

my only other concern is that it's these kind of slight variants that will lead to straying more off design. if we support this "unauthorized" carbon variant then it could be a slippery slope design wise. i remember once i had a design for a non-existent carbon button variant (danger ghost i believe) and it became a big point of contention between product and carbon. ultimately i submitted it back to carbon. i think you can see what i'm getting at though.

elycheea (approved):

PR #4216, opened 2024-02-06
refactor: added code to check window and document object for nextjs

prathapsimha:
Contributes to #

Due to NextJS compiling at the server-side, there is no access to the window and document objects during compilation. Added conditions to check window and document object.

#### What did you change?
Added conditions to check window and document object.
- /src/components/Datagrid/Datagrid/DatagridSelectAll.js
- /src/components/Datagrid/useInfiniteScroll.js
- /src/components/Datagrid/useSelectRows.js

I was able to build the code after adding these conditions and also working fine in browser.


matthewgallo (approved):

matthewgallo:
Thank you for contributing @prathapsimha! 🎉

PR #4233, opened 2024-02-07
fix: use of not has selectors

lee-chase:
Adjustments for AI which cannot use :not(:has selector across shadow-root.

#### What did you change?

A number of selectors including name corrections for container and header.

Storybook.


matthewgallo (changes requested):
With the removal of the `${blockClass}__container` class, the `_create-side-panel.scss` styles need at least one update that I spotted.

matthewgallo:
[Just one instance to fix](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products-styles/src/components/CreateSidePanel/_create-side-panel.scss#L22) in `_create-side-panel.scss`.

Issue #3856, opened 2023-11-28
Tearsheet: color layering model doesn't work


janhassel:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

Carbon's v11 color layering model doesn't work appropriately in the Tearsheet component.

> Is this issue related to a specific component?

Tearsheet

> What did you expect to happen? What happened instead? What would you like to
> see changed?

- readonly inputs should have no visible background / match the background color
  - They have a regular field background in the Tearsheet
- inputs placed in the right influencer are missing their background / match the panel's background
  - even if they're promoted to the next layer, the styles aren't applied

> What browser are you working in?

Safari 17.0

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

`@carbon/ibm-products@2.11.2`

## Steps to reproduce the issue

https://stackblitz.com/edit/github-11rfxa?file=src%2FExample%2FExample.jsx

## Additional information
 
<img width="941" alt="image" src="https://github.com/carbon-design-system/ibm-products/assets/28265588/525ac88a-c2f0-403c-85e9-59e90905d8e3">

ljcarot:
@elycheea Can you investigate if this ever got merged or should be?

elycheea:
@ljcarot Doesn’t look like it so we should move this back into the backlog.

ljcarot:
Since TearSheet is going through overhaul this year, adding @RichKummer for awareness.

paul-balchin-ibm:
Is this a duplicate of #6698? If so, it was fixed and closed.

janhassel:
@paul-balchin-ibm Not quite. Support for the layering model in Tearsheet has improved, but is still not correct.

The right influencer seems to be fine now.
<img width="945" alt="Image" src="https://github.com/user-attachments/assets/6f283eac-cce8-47a3-9faf-d52a7c405dd2" />

However, the left influencer is missing a `<Layer>` wrapper.

Current implementation (2.67.0):
<img width="940" alt="Image" src="https://github.com/user-attachments/assets/1b3a3b77-2057-469e-8b64-57aaf4ef34eb" />

Intended behavior:
[...]

PR #4245, opened 2024-02-08
fix(TearSheet): color layering model

Ratheeshrajan:
Contributes to #3856 

Carbon's v11 color layering model doesn't work appropriately in the Tearsheet component.

#### What did you change?
`readOnly` inputs should have no visible background / match the background color
inputs placed in the right influencer are fixed background / match the panel's background

Storybook > Tearsheet with influencer


matthewgallo (changes requested):
There seems to be just a small linting issue. Also wondering about the change in the placement of the `influencer` element? The border on the influencer doesn't seem to be there anymore as well.

elycheea:
Closing this for now since we haven’t updated this since February, and I don’t think we should change the influencer position as part of this PR. Feel free to reopen, but given the small changes, think a new PR is also fine.

PR #4249, opened 2024-02-08
feat(StringFormatter): new component StringFormatter

jlongshore:
Contributes to [issue 4224](https://github.com/carbon-design-system/ibm-products/issues/4224)

Migrate **StringFormatter** from `Products v1 / Security / StringFormatter` to **Products v2** as `StringFormatter`. 

`v1` References: [GitHub](https://github.com/carbon-design-system/ibm-products/blob/main_v1/packages/security/src/components/StringFormatter/StringFormatter.js), [Storybook](https://v1-carbon-for-ibm-products.netlify.app/?path=/docs/security-components-stringformatter--default).

#### What did you change?

- Modified structure slightly to match Products' standards

- [x] Storybook review.
- [x] Tests migrated and added to.
- [x] Design review by Cameron Calder.


packages/ibm-products/src/components/StringFormatter/StringFormatter.js
```diff
@@ -0,0 +1,176 @@
...
+};
+
+/**
+ * StringFormatter allows for truncating text while displaying a tooltip
+ * overlay on hover or focus with the entirety of the provided copy.
+ */
+export let StringFormatter = React.forwardRef(
+  (
+    {
+      // The component props, in alphabetical order (for consistency).
```
davidmenendez, line 42: same comment as before about cleaning up comments

packages/ibm-products/src/components/StringFormatter/StringFormatter.js
```diff
@@ -0,0 +1,176 @@
...
+    return (
+      <span
+        className={cx(
+          blockClass, // Apply the block class to the main HTML element
+          className // Apply any supplied class names to the main HTML element.
+        )}
+        ref={ref}
+        {
+          // Pass through any other property values as HTML attributes.
+          ...rest
```
davidmenendez, line 66: prop spreading should always be first in order to avoid the user possibly overwriting them

packages/ibm-products/src/components/StringFormatter/StringFormatter.js
```diff
@@ -0,0 +1,176 @@
...
+  /** Specify the direction of the tooltip. Can be either top or bottom. */
+  tooltipDirection: PropTypes.oneOf(Object.values(StringFormatterAlignment)),
+};
+
+/**
+ * StringFormatterContent
+ * Only used internally
+ */
+
+let StringFormatterContent = React.forwardRef(
```
davidmenendez, line 136: as a general convention we should always separate components into their own files. this seems small enough that i don't think it's strictly _necessary_ to have it's own dedicated file. i think this code could just be lumped back into `StringFormatter`. either way is completely fine.

davidmenendez (changes requested):
looks good! have a few comments on the code and one regarding the style

Screenshot 2024-02-09 at 11 17 21 AM

the `line-height` here seems a bit small? i don't see any design documentation that's accompanying this, so i don't know for certain. this might be nothing, but i just wanted to point out that it looks a bit squished.

jlongshore:
> the `line-height` here seems a bit small? i don't see any design documentation that's accompanying this, so i don't know for certain. this might be nothing, but i just wanted to point out that it looks a bit squished.

Yep, this thing should just inherit the line height, color - that kinda stuff from its parent or be set with a class or something. I think for the stories we just did body-01 type style and a text-primary color

davidmenendez (approved):

Issue #4248, opened 2024-02-08
Datagrid: return original row data to getRowId in selectedRowData


AlessandroRoic:
## What will this achieve?
Enhancement:

In `/Datagrid/addons/stateReducer.js`, the entire row object is passed to getRowId (not the actual data inside original). 
When we have for instance passed to the Datagrid data prop an object like:

``` javascript
{ uuid: '1231312313', startDate: '08/02/2024', ....}
```
and set getRowId to (row) => row.uuid , it does not get the uuid since it's in the `original` property and in this case the selectedRowData returns the latest selected row inside an undefined property like this:

``` javascript
{ undefined: { id: 1, original: {...}, ... } }
```

This does not work well in this case because you need to set up the row id to uuid.

This is the mentioned code:
``` javascript
if (isChecked) {
  return {
    ...newState,
    selectedRowData: {
      ...newState.selectedRowData,
    -->  [getRowId(rowData, rowData.index)]: rowData,
    },
  };
}
```

Related PR to this issue: https://github.com/carbon-design-system/ibm-products/pull/3859

PR #4275, opened 2024-02-08
fix(Datagrid): changes getRowId parameter to original row data

AlessandroRoic:
Contributes to #4248

Changes the first parameter passed to getRowId to the original property of row.
So now instead of passing the whole row object it will pass just the actual data.

#### What did you change?
```
```

Storybook & run unit tests


AlessandroRoic:
Need to close this since I pushed one commit with my ibm git account

PR #4302, opened 2024-02-09
fix(Datagrid): changes getRowId parameter to original row data

AlessandroRoic:
Contributes to #4248

Changes the first parameter passed to getRowId to the original property of row.
So now instead of passing the whole row object it will pass just the actual data.

This is for v1.
#### What did you change?
```
```

Storybook & run unit tests


matthewgallo (approved):

matthewgallo:
Thanks for contributing @AlessandroRoic 🎉

Issue #4043, opened 2024-01-11
Investigate why certain checks are not being captured in CI checks / locally


elycheea:
- style snapshot
- cspell
- formatter
- commitlint

matthewgallo:
Adding this error I got today when running the snapshot since it feels related to this issue
![image (13)](https://github.com/carbon-design-system/ibm-products/assets/10215203/085162ea-c13c-476c-8d80-809ba0f10d6d)

davidmenendez:
also tagging this PR that shouldn't have been allowed to commit due to a cspell error https://github.com/carbon-design-system/ibm-products/pull/4535

PR #4305, opened 2024-02-09
fix: update to spellchecker

davidmenendez:
Contributes to #4043

the current `spellcheck` command is out of date and not working. this fixes it


paul-balchin-ibm (approved):

matthewgallo:
Nice! I ran into an issue earlier today where `cspell` was flagging issues in a file I hadn't even changed 😅

Issue #4017, opened 2024-01-09
Add `size`s for user avatar


elycheea:
`UserAvatar` includes some changes to sizes from the original `UserProfileImage`.

The new `UserAvatar` should support 4 sizes — `sm`, `md`, `lg`, and `xl`.
Size should likely be a required prop based on the original implementation. I also don’t think there’s a logical default in this case since these will be specific to context.

| Size | Old sizing | New sizing | Type token                    | Icon scale |
|------|-----------|------------|-----------------------|-----------|
| `xl`   | 64px         | 64px           | `$heading-04`            | `32`          |
| `lg`   | 32px         | **48px**    |  `$heading-03`           | `24`          |
| `md` | 24px         | **32px**    | `$body-compact-01` | `20`          |
| `sm` | -                | **24px**    | `$label-01`                  | `16`          |

**Color tokens** —
Icons should use `$icon-inverse`
Text should use `$text-inverse`

These are both white on light themes, gray-100 on dark.

**Name to initials**
Since we’re also adding the type tokens here, it may also make sense to go ahead and add support for converting a name to its initials.
The current prop in `UserProfileImage` uses `initials` but suggest using `name` since it takes a full string and converts it to initials.

```
Thomas Watson and Thomas J. Watson should both return TW.
```

```[tasklist]
### Acceptance criteria
- [x] Add type tokens for each size
- [x] Scale icon for each size
- [x] Add support for showing `name` as initials
- [x] Add tests for `size` prop
- [x] Add tests for `name` prop
```

elycheea:
From Cate:

**Carbon 10 tokens used for user profile image**
- Code: Text & and icon colors need to be converted to tokens
    - Icon: Dark theme ($icon-inverse), Light theme ($icon-03)
    - Text: Dark theme ($text-inverse-01), Light theme ($text-04)
- Code: Initials — type scale convert to text tokens 
    - XL 64 ($product-heading-04), L 32 ($body-short-01), M 24 ($label-01)

Missing sizes for the 48px size.

PR #4312, opened 2024-02-12
feat(UserAvatar): implementation of name,size props

anamikaanu96:
Contributes to #4017

Implementation of name and size props to useravatar

#### What did you change?
packages/ibm-products/src/components/UserAvatar/UserAvatar.js
packages/ibm-products/src/components/UserAvatar/UserAvatar.stories.js
packages/ibm-products/src/components/UserAvatar/UserAvatar.test.js
packages/ibm-products-styles/src/components/UserAvatar/_user-avatar.scss

storybook


packages/ibm-products/src/components/UserAvatar/UserAvatar.js
```diff
@@ -58,8 +58,10 @@ export let UserAvatar = React.forwardRef(
       // The component props, in alphabetical order (for consistency).
       backgroundColor,
       className,
+      name,
       /* TODO: add other props for UserAvatar, with default values if needed */
       renderIcon = defaults.renderIcon,
+      size = 'md',
       tooltipText = defaults.tooltipText,
```
elycheea, line 65: I think we can move `size` to defaults.

I missed this in the previous PRs, but we probably should not set a default for `tooltipText` since this is optional. The default here should only be set in the story.

```suggestion
      size = 'md',
      tooltipText = defaults.tooltipText,
```

packages/ibm-products/src/components/UserAvatar/UserAvatar.js
```diff
@@ -70,19 +72,40 @@ export let UserAvatar = React.forwardRef(
...
-        return icons.user['md'];
+        return icons.user[size];
       } else if (renderIcon === Group) {
-        return icons.group['md'];
+        return icons.group[size];
+      } else if (name) {
+        return formatInitials;
       } else {
-        return renderIcon;
+        return icons['user'][size];
```
elycheea, line 108: The corresponding icon size should always be the same based on the avatar size.
Rather than checking for `renderIcon`, I think it makes more sense to just render the icon[^1] that gets passed in using the sizes mapped below.

| Size  | Icon scale |
|------|-----------|
| `xl`   | `32`          |
| `lg`   | `24`          |
| `md` | `20`          |
| `sm` | `16`          |

[^1]: I think we have a few examples in `ButtonMenu`, `ComboButton`, `EditInPlace`/`InlineEdit` and a few other places like if you’d like to see some examples although they don’t include size.

packages/ibm-products/src/components/UserAvatar/UserAvatar.stories.js
elycheea, line 1: One way we like to handle these is by using a select to provide a few examples in the story.

image

In this case, you could add `no icon`, `user`, `group`, and then a custom icon to show a few common examples.

elycheea (changes requested):

packages/ibm-products/src/components/UserAvatar/UserAvatar.js
```diff
@@ -58,8 +58,10 @@ export let UserAvatar = React.forwardRef(
       // The component props, in alphabetical order (for consistency).
       backgroundColor,
       className,
+      name,
       /* TODO: add other props for UserAvatar, with default values if needed */
       renderIcon = defaults.renderIcon,
```
davidmenendez, line 63: i think this needs to be changed. currently there's no way to access the group variant without already passing the group icon, which we then import in this component. given that i think the component needs to be refactored slightly.

the user can just pass in an icon directly. `<UserAvatar renderIcon={Information />`

in the render we can check for a custom icon or provide a default one. and then we can use a size dictionary to access the sizes.

```
size,
renderIcon: Icon,
...
const sizes = {
  sm: 16,
  md: 20
}
...
{Icon ? <Icon size={sizes[size]} /> : <User />
```

davidmenendez (changes requested):
good start! i think we can make some small adjustments to make this code a little more clean

packages/ibm-products/src/components/UserAvatar/UserAvatar.js
anamikaanu96: Done

packages/ibm-products/src/components/UserAvatar/UserAvatar.js
anamikaanu96: Changed based on @davidmenendez suggestion

packages/ibm-products/src/components/UserAvatar/UserAvatar.stories.js
anamikaanu96: Implemented examples for no icon, user and group using select.
For custom icon, how we can implement using select? should we use different prop?

packages/ibm-products/src/components/UserAvatar/UserAvatar.js
```diff
@@ -46,43 +46,51 @@ const componentName = 'UserAvatar';
...
       // The component props, in alphabetical order (for consistency).
       backgroundColor,
       className,
+      name,
       /* TODO: add other props for UserAvatar, with default values if needed */
-      renderIcon = defaults.renderIcon,
-      tooltipText = defaults.tooltipText,
-      tooltipAlignment = defaults.tooltipAlignment,
+      renderIcon,
+      size,
```
davidmenendez, line 58: we should still have a defaults constant and use a default value here. probably `md`

packages/ibm-products/src/components/UserAvatar/UserAvatar.js
```diff
@@ -46,43 +46,51 @@ const componentName = 'UserAvatar';
...
     {
       // The component props, in alphabetical order (for consistency).
       backgroundColor,
       className,
+      name,
       /* TODO: add other props for UserAvatar, with default values if needed */
-      renderIcon = defaults.renderIcon,
-      tooltipText = defaults.tooltipText,
-      tooltipAlignment = defaults.tooltipAlignment,
+      renderIcon,
```
davidmenendez, line 57: since the user isn't passing props directly to the icon this way we can reference this as a component to be consumed directly. i'll follow this up in an additional comment `renderIcon: RenderIcon`

packages/ibm-products/src/components/UserAvatar/UserAvatar.js
```diff
@@ -46,43 +46,51 @@ const componentName = 'UserAvatar';
...
+        .match(/(^\S|\S$)?/g)
+        .join('')
+        .toUpperCase();
     };
     const getItem = (renderIcon) => {
-      if (renderIcon === User) {
-        return icons.user['md'];
-      } else if (renderIcon === Group) {
-        return icons.group['md'];
+      let Iconcomponent = renderIcon;
```
davidmenendez, line 87: this line isn't necessary when we declare `renderIcon` as a component in the props. and since, as i mentioned before, the user isn't passing in any props directly to icon we don't need to return a function. we can just do something like this
```
renderIcon: RenderIcon,
...
const getContent = () => {
  const iconProps = {
    size: iconSize[size],
  };
  if (RenderIcon) {
    return <RenderIcon {...iconProps} />;
  }
  if (name) {
    return formatInitials;
  }
  return <User {...iconProps} />;
};
```

davidmenendez (changes requested):
making good progress! just a few more suggestions. i'd also like to suggest doing something like this to clean up the render portion. instead of `SetItem` i'd do something like this
```
const Avatar = () => (
  <div
    {...rest}
    className={cx(
      blockClass,
      className,
      `${blockClass}--${backgroundColor}`,
      `${blockClass}--${size}`
    )}
    ref={ref}
    role="img"
    {...getDevtoolsProps(componentName)}
  >
    {getItem()}
  </div>
);

if (tooltipText) {
  return (
    <Tooltip
      align={tooltipAlignment}
      label={tooltipText}
      className={`${blockClass}__tooltip ${carbonPrefix}--icon-tooltip`}
    >
      <TooltipTrigger>
        <Avatar />
      </TooltipTrigger>
    </Tooltip>
  );
}

return <Avatar />;
```

packages/ibm-products/src/components/UserAvatar/UserAvatar.js
davidmenendez: makes things even easier!

packages/ibm-products/src/components/UserAvatar/UserAvatar.js
```diff
@@ -46,43 +46,51 @@ const componentName = 'UserAvatar';
...
-        md: <Group size={32} />,
-      },
+    const iconSize = {
+      sm: 16,
+      md: 20,
+      lg: 24,
+      xl: 32,
+    };
+    const formatInitials = () => {
+      if (name.length === 2) {
```
davidmenendez, line 74: just noticed that we should also tweak this a bit. two problems here

1. when a 2 character string is passed in we're not formatting it with `toUpperCase`
2. if a single string is passed in that's greater than 2, like `David` for example, the output is `DA` which i wouldn't expect because no last name was passed in.

i think we can address this and improve the readability by making some changes
```
const formatInitials = () => {
  const parts = name.split(' ');
  const firstChar = parts[0].charAt(0).toUpperCase();
  if (parts.length === 1) {
    return firstChar;
  }
  const lastChar = parts[parts.length - 1].charAt(0).toUpperCase();
  const initials = [firstChar];
  if (lastChar) {
    initials.push(lastChar);
  }
  return ''.concat(...initials);
};
```

this implementation is slightly more readable than the regex and it clearly demonstrates each outcome

elycheea: Agree with toUpperCase but didn’t mind DA if only two letters are passed. I feel like I’ve seen similar approaches before but unsure if this was intended behavior in our case. 🤔

davidmenendez: i'll open up an additional issue to confirm this behavior 👍

davidmenendez (dismissed):
awesome!

davidmenendez:
@elycheea i don't see anything specifically mentioned in the [documentation](https://pages.github.ibm.com/cdai-design/pal/patterns/user-profile-images/usage/#initials) regarding only first names. this takes multiple last names in account, but not just a single one. seems like a funny oversight. good thing Cher isn't a user 😂 

i think this PR is good to go and i'll go ahead and open another issue to verify the single name question 👍

packages/ibm-products/src/components/UserAvatar/UserAvatar.stories.js
```diff
@@ -80,10 +94,12 @@ const Template = (args) => {
  * TODO: Declare one or more stories, generally one per design scenario.
  * NB no need for a 'Playground' because all stories have all controls anyway.
  */
-export const userAvatar = prepareStory(Template, {
+export const Default = prepareStory(Template, {
+  storyName: 'Default',
   args: {
     ...defaultArgs,
     // TODO: Component args - https://storybook.js.org/docs/react/writing-stories/args#UserAvatar-args
+    name: 'thomas j. watson',
```
elycheea, line 102: If we add this, it should pre-select the `no icon` option in Storybook.

```suggestion
    name: 'thomas j. watson',
    renderIcon: 'No icon',
```

image

packages/ibm-products/src/components/UserAvatar/UserAvatar.js
```diff
@@ -108,24 +116,22 @@ export let UserAvatar = React.forwardRef(
...
+        align={tooltipAlignment}
+        label={tooltipText}
+        className={`${blockClass}__tooltip ${carbonPrefix}--icon-tooltip`}
+      >
+        <TooltipTrigger>
+          <Avatar />
+        </TooltipTrigger>
+      </Tooltip>;
+    }
+    return <Avatar />;
```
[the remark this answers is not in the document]
elycheea: We should add this to tests to make sure that we render a tooltip if `tooltipText` is supplied.

packages/ibm-products/src/components/UserAvatar/UserAvatar.stories.js
elycheea, line 1: I just looked this up and learned this but in Storybook 7, defaults no longer get set in the controls, but if we add `args` under `defaultArgs`, I think this gets fixed.

```
argTypes: { ... },
args: {
  size: 'md',
  tooltipAlignment: 'bottom'
},
parameters: { ... }

**Review by elycheea — CHANGES_REQUESTED** [review]

File: packages/ibm-products/src/components/UserAvatar/UserAvatar.js [hunk 10]
```diff
@@ -58,37 +54,48 @@ export let UserAvatar = React.forwardRef(
...
+      }
+      return ''.concat(...initials);
+    };
+    const getItem = () => {
+      const iconProps = { size: iconSize[size] };
+      if (RenderIcon) {
+        return <RenderIcon {...iconProps} />;
+      }
+      if (name) {
+        return formatInitials();
```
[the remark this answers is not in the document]
elycheea: Okay, confirmed with Marion that `<User {...iconProps} />` can be a fallback if neither an icon nor name is provided. Probably not a common scenario, but at least we’ve accounted for it.

packages/ibm-products/src/components/UserAvatar/UserAvatar.js
elycheea: Also found some of the initial discussion (ha ha) that I believe ties to the original implementation. https://github.com/carbon-design-system/ibm-products/issues/69#issuecomment-769895037

It looks like this is to better support use of display names where maybe there is not a first and last name provided or used. Again, probably mostly just a fallback, but I do think we can probably respect the original implementation of this.

(Happy with either the Regex or refactored version of this.)

anamikaanu96: sticked with the refactored version.

packages/ibm-products/src/components/UserAvatar/UserAvatar.js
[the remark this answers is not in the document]
anamikaanu96: Implemented

packages/ibm-products/src/components/UserAvatar/UserAvatar.js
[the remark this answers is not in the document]
anamikaanu96: it was my mistake, while I edited somehow it missed. Tests added

anamikaanu96:
Done all the required changes, @elycheea could you please recheck

elycheea (dismissed):
Approving although I think we still have a few fixes to make with `formatInitials()`.

If you prefer to merge this as is and open a separate PR, feel free to merge this now, and we can open a separate issue to update this specifically.

Even if we keep the refactored version, it should support these scenarios:

If passed the following as `name`

1. `Thomas J. Watson` becomes `TW`
2. `Thomas Watson` becomes `TW`
3. `thomasjwatson` becomes `TH`

In the last scenario, we have support in case a name (or display name) is passed that does not follow the traditional First Name, Last Name.

(See [UserProfileImage](https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-patterns-user-profile-images-userprofileimage--with-initials&args=initials:thomasjwatson))

elycheea (approved):

Issue #3998, opened 2024-01-03
Implement reduced motion for side panels


elycheea:
Typically, we we consider reduced motion, we default to no motion. @mikeolasov shared some ideas on ways we could incorporate low motion alternatives.

> Thoughts for future — I know @mikeolasov has thoughts on reduced motion versus no motion. Might be something we can explore some more separately since the filter panel treatment should probably be consistent with side panel treatment. (e.g. Could have more subtle slide but buttons don’t animate? ... although I’ll leave it to Mike’s expertise.)

For big moves like the filter panel opening I think it's best to have no motion or a very simple ease out for the last ~10% of the move, where the panel snaps to ~90% open on the first frame and eases the rest (just to lead the user's eye).

I'd also like to see a fade-in or cascade fade-in of the filter panel contents, like the regular motion state.

_Originally posted by @mikeolasov in https://github.com/carbon-design-system/ibm-products/issues/3994#issuecomment-1875708385_
```[tasklist]
### Acceptance criteria
- [x] Create prototype with suggested easings with/without fade-in
- [x] Create PR with suggested easings
- [x] Work with Mike on deploy preview
```

elycheea:
@mikeolasov Happy to find time for us to chat about this some more later. We did implement some updates for this in the Datagrid filter.

Issue #4093, opened 2024-01-19
Reduced motion review


elycheea:
- [ ] Organize our components and patterns to understand what our common types and needs are[^1]
- [ ] Create some general classes and principles for motion

Motion for components needs to be subtle and intentional, especially since we don’t know what else is on the page.[^2]
For component behavior, using the Carbon motion easings should generally suffice.

Should scope down ... 

| Component | Motion                                 | Reduced motion | Status             |
|-------------|---------------------------|-----------------|---------------|
| Status icon  | `in-progress` variant only  | No motion          | Done               |
| User avatar | None | | Done 
| Empty state | None | | Done
| Edit-in-place | Some ... | Needs implementation 
| Add and select | Has some, especially in 

Options tile looks good minus a quirky jerk at the close.

**Page header + side panel** need to revisit the way their headers scroll — should not allow scroll to remain incomplete.
Can put an ease on it if triggered by the user scroll rather than completing scroll to a specific position.
Reduced motion should keep ... gentle fade. Need to prevent the crossfade where page header and side panel headers overlap when animation does not complete.

### Probably needs motion design
(Move to separate issue)

| Component | Motion                                 | Reduced motion | Status             |
|-------------|---------------------------|-----------------|---------------|
| Tag set         | WIP
| Cards           | None | — | Needs review 
| Side panel   | Second step needs motion
| Add and select | 

Tearsheet needs work, but may make sense to work with @mikeolasov in the future when we refactor Tearsheet.

[^1]: e.g. Filter panels and side panels or primary/secondary content.
[^2]: https://carbondesignsystem.com/guidelines/motion/overview/

PR #4318, opened 2024-02-12
feat(SidePanel): initial reduced motion work for side panel opacity

matthewgallo:
Contributes to #3998 

Includes reduced motion support for the `SidePanel`, similar to the FilterPanel this update only uses opacity when the panel is opened or closed. Along, with opacity transition and delay for the action set when reduced motion is enabled.

#### What did you change?
```
packages/ibm-products-styles/src/components/SidePanel/_side-panel.scss
packages/ibm-products/src/components/SidePanel/SidePanel.js
packages/ibm-products/src/components/SidePanel/motion/variants.js
```
Storybook

### DIFF
--- a/packages/ibm-products-styles/src/components/SidePanel/_side-panel.scss
+++ b/packages/ibm-products-styles/src/components/SidePanel/_side-panel.scss
@@ -353,7 +353,7 @@ $action-set-block-class: #{c4p-settings.$pkg-prefix}--action-set;
   }
 
   .#{$block-class}__slug-and-close {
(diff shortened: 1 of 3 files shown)


elycheea (approved):
Was going to comment that the overlay fadeout seems a bit more sudden (no motion), but looks consistent with our overlay exit when the side panel closes even with reduced motion turned off. 🤔

This would most likely more of a motion review than _reduced_ motion. May be something we can look into when we split out motion work from #4093.

Issue #4254, opened 2024-02-08
Add Typescript types to AboutModal


matthewgallo:
[Refer to Carbon's doc on adding component types as needed](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/adding-component-types.md).

PR #4320, opened 2024-02-12
feat(AboutModal): add typescript types

AlexanderMelox:
Contributes to #4254 

Adds typescript support to AboutModal

#### What did you change?
```
packages/ibm-products/src/components/AboutModal/AboutModal.js -> packages/ibm-products/src/components/AboutModal/AboutModal.tsx
```

### DIFF
--- a/packages/ibm-products/src/components/AboutModal/AboutModal.tsx
+++ b/packages/ibm-products/src/components/AboutModal/AboutModal.tsx
@@ -5,31 +5,106 @@
  * LICENSE file in the root directory of this source tree.
  */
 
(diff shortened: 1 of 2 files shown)


AlexanderMelox:

matthewgallo:
As far as the Typescript work everything looks good, but I do notice that the scroll gradient is appearing when there is not scrollable content which makes some content not visible and I don't see that in the current changes in `main`. I also noticed there is a blue focus indicator around the `ModalBody` which also seems to be different from `main`.
image

davidmenendez (approved):

PR #4327, opened 2024-02-13
feat(StatusIndicator): new component StatusIndicator

paul-balchin-ibm:
Contributes to #4071

Migrating **StatusIndicator** from `Products v1 / Patterns / StatusIndicator`.

`v1` references: [GitHub](https://github.com/carbon-design-system/ibm-products/blob/main_v1/packages/security/src/components/StatusIndicator/StatusIndicator.js), [Storybook](https://v1-ibm-products.carbondesignsystem.com/?path=/story/security-patterns-statusindicator--default).

#### What did you change?

This is a new `v2` component.

**StatusIndicator prop updates**
|-|-|-|
| _title_ | _title_ | _Unchanged_. |
| — | onRetry | Optional callback function. |
| — | retryLabel | Text label for the retry button. |
| — | showRetry | |
| ~currentIndex~ | (deprecated) |  |
| ~retry: {action, label}~ | (deprecated) | Replaced by `onRetry` and `retryLabel` above |

**StatusIndicatorStep prop updates** (name changed from `StatusStep`).
|-|-|-|
| _description_ | _description_ | _No change._ | 
| incomplete | inactive | * | 
| current | active | * | 
| failed | error | * | 
| complete | finished | * | 
| errorMsg | errorMessage |  | 

\* The new names are in keeping with the `InlineLoading` component.

- [x] Storybook review.
- [x] Design update by Cameron Calder and Bridget Ndege.
- [x] Design review by Shelly Dignum.


packages/ibm-products/src/components/StatusIndicator/StatusIndicator.js
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * Set to `true` to show the retry button.
+   *
+   * `retryLabel` and `onRetry` must also be defined.
+   */
+  showRetry: PropTypes.bool,
+  /**
+   * The title that appears at the top of the list.
+   */
+  title: PropTypes.string,
```
matthewgallo, line 106: It might be nice to change this to `PropTypes.node` so that either a string or node can be passed to this prop. We've encountered a few other cases where the extra flexibility has been needed/requested.

packages/ibm-products/src/components/StatusIndicator/StatusIndicator.js
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * The text for the retry button.
+   */
+  retryLabel: PropTypes.string,
+  /**
+   * Set to `true` to show the retry button.
+   *
+   * `retryLabel` and `onRetry` must also be defined.
+   */
+  showRetry: PropTypes.bool,
```
matthewgallo, line 102: What do you think about only including the `onRetry` prop? If `onRetry` is passed in then you know you want to show the retry button. I don't see any other logic in terms of showing the retry button, could be an easy way to reduce the total number of props.

packages/ibm-products/src/components/StatusIndicator/StatusIndicator.test.js
```diff
@@ -0,0 +1,75 @@
...
+
+  it('has no accessibility violations', async () => {
+    const { container } = renderComponent();
+    expect(container).toBeAccessible(componentName);
+    expect(container).toHaveNoAxeViolations();
+  });
+
+  it('applies className to the containing node', async () => {
+    const { container } = renderComponent({ className: className });
+    const statusIndicator = container.querySelector(
```
matthewgallo, line 52: Same comment about using `screen.getByTestId(dataTestId);`

packages/ibm-products/src/components/StatusIndicator/StatusIndicatorStep.js
```diff
@@ -0,0 +1,97 @@
...
+    ref
+  ) => {
+    return (
+      <li
+        {...rest}
+        className={cx(blockClass, className, `${blockClass}--${status}`)}
+        ref={ref}
+        {...getDevtoolsProps(componentName)}
+      >
+        <div className={`${blockClass}--details`}>
```
matthewgallo, line 46: Small detail, but to match the class name structure of the other classes in your component I think it makes sense for this class to be `${blockClass}__details`.
```suggestion
        <div className={`${blockClass}__details`}>
```

matthewgallo (commented):
Like the composable approach 😄 just had a few small comments/suggestions

packages/ibm-products/src/components/StatusIndicator/StatusIndicator.js
paul-balchin-ibm: Originally, I was thinking the same thing. And then...

Normally, buttons like this are always visible, it's just their state that changes based on conditions: disabled, busy, enabled, etc.

For the status indicator, the expected behavior is that this button is rendered only if any status returns an error. This is atypical.

More intuitive and self-documenting. "Show the button if there's an error".
```
hasError: useState(boolean)

onRetry: () => { do something }
retryLabel: "Retry"
showRetry: hasError
```

Less intuitive. "The button has a function if there's an error" does not imply that the button will be rendered or not.
```
hasError: useState(boolean)

onRetry: hasError ? undefined : () => { do something }
retryLabel: "Retry"
```

paul-balchin-ibm: > could be an easy way to reduce the total number of props

This is a very simple component, and there are only four props, excluding `children` and `className`.

packages/ibm-products/src/components/StatusIndicator/StatusIndicatorStep.js
paul-balchin-ibm: Oops. My bad.

packages/ibm-products/src/components/StatusIndicator/StatusIndicator.js
matthewgallo: Thanks for the detailed explanation here, very helpful especially for other components moving forward 😄

matthewgallo (approved):

matthewgallo:
Closing in favor of https://github.com/carbon-design-system/ibm-products/pull/4419

PR #4333, opened 2024-02-14
fix(Dataspreadsheet): Duplication selection issue #4241

amal-k-joy:
Prevent Cmd (Ctrl) + click from re-adding the same cells to a selection

Contributes to #4241

Clear the selection , if same selection is made again 

#### What did you change?
Created a method that will identify and remove any duplicate selection when any selection is made

Verified locally


matthewgallo:
You can see this in the `Actions` tab in storybook, `selection area change` lists two selections but they are duplicates.
image

matthewgallo (approved):
🎉

PR #4344, opened 2024-02-15
chore: update to node v20

davidmenendez:
Contributes to #4043 

updates the node version to LTS in order to try to try to identify issues with failing builds and why sometimes CI checks don't run properly.

also verified in the [netlify docs](https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript) that the `.nvmrc` file does overwrite anything set in the netlify UI, so we shouldn't need to update anything there.


package.json
```diff
@@ -1,5 +1,8 @@
 {
   "name": "ibm-products",
+  "engines": {
+    "node": "20"
```
matthewgallo, line 4: Not sure if entirely necessary, but should this be `"node": "20.x"`?

matthewgallo (approved):

Issue #4349, opened 2024-02-16
Reduce frequency of example gallery updates


elycheea:
Currently, the example gallery check runs in each PR and requires that each author updates the gallery as updates are made. This increases the size of many PRs. Since this is more of a step to maintain and keep our examples up-to-date, believe this does not need to run with each PR, especially since updates are not available until a release.

```[tasklist]
### Tasks
- [ ] Move gallery check from CI to separate action
```

PR #4352, opened 2024-02-16
chore: add new example gallery action

elycheea:
Contributes to #4349 

This should move the example gallery check from CI run on every PR to a twice-monthly action.
Hoping I did this right, but the idea is using the same logic from the previous CI check, we verify if gallery needs to be updated, if out-of-date, we exit with an error[^1], which we use to determine whether to conditionally create a new PR in the `create-pr` step. 😯

#### What did you change?
Remove `check_gallery` step from `CI.yml`
Add `update-example-gallery.yml` to run the 8th and 22nd of each month (offset from dependency updates _in hopes of_ keeping our PR counts manageable).

🤐🤫

[^1]: https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context


davidmenendez (approved):
LGTM

lee-chase (approved):

matthewgallo (approved):

Issue #3887, opened 2023-12-01
Datagrid: add support for selectable nested rows


matthewgallo:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

There current is not support for selected nested rows in the Datagrid, however there is design guidance for this variant of Datagrid, [see here](https://pages.github.ibm.com/cdai-design/pal/components/data-table/nested-rows/usage/).
> Is this issue related to a specific component?

Datagrid, `useNestedRows` and `useSelectRows`
> What did you expect to happen? What happened instead? What would you like to
> see changed?

Expected to be able to use these two plugins together.
> What browser are you working in?

Chrome
> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

Latest
> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

N/A

elycheea:
### Acceptance criteria

- [ ] Add a story with selectable rows under nested rows folder
- [ ] Selecting parent should not select all children by default

From discussion on refinement call —
Safest route is probably to leave out selection of all children when selecting the parent. This behavior can vary depending on needs. Add select is another example that treats this a bit separately.

matthewgallo:
It appears that `react-table` forces all child rows to be selected when a parent row is selected, there might be a way to work around this but this appears to be the default behavior.

https://github.com/TanStack/table/blob/v7/src/plugin-hooks/useRowSelect.js#L166-L169

PR #4354, opened 2024-02-19
fix(Datagrid): add support for selectrow and nestedrow

Ratheeshrajan:
Contributes to #3887 

Add support for selected nested rows in the Datagrid.

Note: This PR only covers the selectable nested row part of #3887. The second acceptance criteria in [comment](https://github.com/carbon-design-system/ibm-products/issues/3887#issuecomment-1856319771) "Selecting parent should not select all children by default" is [coming from the React table ](https://github.com/TanStack/table/blob/v7/src/plugin-hooks/useRowSelect.js#L154-L177)and it's hard to override. If there is subRows then child rows get selected when parent row is selected.

#### What did you change?

`rowId` selection changed to a combination of `rowId` and `index` so the nested rows will have unique Id with depth

Storybook > With selectable nested rows


matthewgallo (commented):
Looks great, just needs an update to the example gallery, `yarn update-gallery-config`.

matthewgallo (approved):

Issue #3165, opened 2023-06-27
“Clickable row” should retain focus after side panel closes


elycheea:
After side panel closes, focus state should return to originally activated item (row). I think we do something similar in editable cells since we can return to the active cell in those stories so we should just need to expand on this mechanism for this case.

-- 

**Accessibility review:**
keyboard navigation issues found in Clickable row with side panel:
1. After opening and dismissing the side panel from row 3, pressing tab key again will move the focus to the first row of the table. The desired behavior should be moving the focus down to row 4.
<img width="957" alt="image" src="https://github.com/carbon-design-system/ibm-products/assets/40870771/a3166f52-f016-455a-98ef-8e9dc76c0921">

2. While the focus ring appears on the Close button in side panel, pressing the tab key again doesn't do anything. The desired behavior should be moving the focus in the side panel then go back to row 2 in the table. 
<img width="1120" alt="image" src="https://github.com/carbon-design-system/ibm-products/assets/40870771/bad9fa09-7d55-4f4b-9358-6c92f38ff60c">

_Originally posted by @thefirstartist in https://github.com/carbon-design-system/ibm-products/issues/2608#issuecomment-1593737376_

matthewgallo:
I think we could look into supporting an initially selected row via the `initialState` prop, similar to how other pieces of the Datagrid can set an initial state value from this object. [See here for example](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products/src/components/Datagrid/Datagrid.stories.js#L358-L362)

PR #4405, opened 2024-02-22
fix(DataGrid): clickable row retain focus after side panel close V1

Ratheeshrajan:
Contributes to #3165 

After opening and dismissing the side panel from row 3, pressing the tab key again will move the focus to the first row of the table. The desired behavior should be moving the focus down to row 4.

While the focus ring appears on the Close button in the side panel, pressing the tab key again doesn't do anything. The desired behavior should be moving the focus in the side panel and then go back to row 2 in the table.

#### What did you change?

Side panel `onRequestClose` setback the focus to the selected row. 
If the Side panel's last focusable element loses focus then the focus goes to the selected row.
`packages/ibm-products/src/components/Datagrid/Extensions/ClickableRow/ClickableRow.stories.js`

Storybook

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Extensions/ClickableRow/ClickableRow.stories.js
+++ b/packages/ibm-products/src/components/Datagrid/Extensions/ClickableRow/ClickableRow.stories.js
@@ -6,7 +6,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, { useState } from 'react';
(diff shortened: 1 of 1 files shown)


davidmenendez (changes requested):
same comment, looks good but needs a testing

elycheea:
Closing this for now and will move this issue back to backlog since our team has shifted. I think we’d prefer to close this with the v2 PR even if this one does not have current conflicts to maintain parity between versions.

PR #4407, opened 2024-02-22
chore(Datagrid): datagrid column customization fix

Ratheeshrajan:
Contributes to #4244 
1. Copyright header in every source file (js, css, scss etc.) with the appropriate years.
2. All components utilizing motion must include reduced-motion queries for
accessibility purposes.

#### What did you change?

1. Updated Copyright header in 10 files to include the current year.
2. No transition in reduced motion

Files change


packages/ibm-products-styles/src/components/Datagrid/styles/_draggableElement.scss
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright IBM Corp. 2022, 2022
+// Copyright IBM Corp. 2022, 2022, 2023, 2024
```
matthewgallo, line 2: The format/pattern for the years in the copyright headers is year file was created followed by year the file was last edited, so you only will have two years listed.

matthewgallo (changes requested):
See comment regarding the pattern we use for the years listed in the copyright header (tldr `year created, year last updated`) so you'll only end up with two years listed.

packages/ibm-products-styles/src/components/Datagrid/styles/_draggableElement.scss
Ratheeshrajan: Thanks @matthewgallo for the information. So I can update the year last edited as `2023`, right?
Screenshot 2024-02-23 at 10 58 16 AM

matthewgallo: Correct ^^

Ratheeshrajan: @matthewgallo updated the copyright header `last modified year`.

elycheea (approved):

matthewgallo (approved):

PR #4420, opened 2024-02-22
feat(Nav): new component Nav

lily-peng:
Contributes to [#4136](https://github.com/carbon-design-system/ibm-products/issues/4136).

Migrating **Nav** from `Products v1 / Security / Components / Nav`.

`v1` references: [GitHub](https://github.com/carbon-design-system/ibm-products/tree/main_v1/packages/security/src/components/Nav), [Storybook](https://v1-ibm-products.carbondesignsystem.com/?path=/story/security-components-nav--default).

#### What did you change?
This is a new `v2` component.

- `Nav`, `NavList`, and `NavItem` were all changed from React class components to functional components
- Restructured the component js files to all be top-level inside of in its own folders, to maintain consistency with other v2 components
- Consolidated original 2 stories (Default and Long NavItem test case) into one

* [x]  Storybook review.
* [ ]  Unit tests migrated and added to.
* [ ]  Design review.


lily-peng:

bndege2:
Design review:
- Tab navigation looks good
- Colours styles look good
- Text styles look good

Noted inconsistencies in focus and active states & spacing. 
Nav interaction
Nav spacing

bndege2:
https://github.com/carbon-design-system/ibm-products/assets/157747366/8581c0d5-0812-47ba-a237-e4af1bf86d0e

**Nav item 1-2** is triggering **Nav list 2 expanded on page load** to open when clicked

lily-peng:
image

davidmenendez (changes requested):
great start! just a few comments.

the biggest thing for me is on line `122` of `Nav.js` with the usage of `React.Children.map` i think the intention here is to add props on behalf of the consumer, but i think this is unnecessary and as i've been talking about recently we should be wary of using any [legacy react api's](https://react.dev/reference/react/legacy). in this case i think it's better to just expose those props / api's to the consumer and let them handle it. this really should just be as simple as returning `{children}`. i think making this update will be a pretty big improvement.

in `NavItem.js` i don't think we need to worry about passing an element. i don't think that type of specificity is super important. along with in `NavItemLink` where `create.Element` is being used which is part of the legacy api's we should be trying to avoid.

the only other thing i'd say is that it looks like there's a bunch of different options and props with the sub-components, but those controls aren't surfaced in the storybook. it would be super helpful to allow users to options like `disabled`, `activeHref` or `externalLink`

matthewgallo:
Another option could be a render prop, if you need to add props to one of the children. This is used across Carbon ([see Dropdown](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Dropdown/Dropdown.tsx#L188)) and is also one of the recommended alternatives to `cloneElement` and `Children.map` from the [React docs](https://react.dev/reference/react/cloneElement#alternatives).

lily-peng:
The plan is to support the changes and enhancements suggested by David at a later date.

davidmenendez (approved):
LGTM!

matthewgallo:
Happy to pop it into the merge queue! 😄

Issue #2042, opened 2022-06-15
Drag and drop


elycheea:
# Drag and drop

## Brief description

Drag and drop is a common interaction that is widely used for various situations. In Carbon ecosystem, we clarify the use cases, interactions and styles.

## Details

[Link to design](https://pages.github.ibm.com/cdai-design/pal/patterns/drag-and-drop/usage)

### Links to other materials

e.g. sketch files, icons, images, prototypes

## Design Maintainer(s)

- Minji Kim
- Carrie Lloyd

## Tasks

Before starting work on this epic, please review and complete the following.

### Working in Carbon for IBM Products package

- [ ] Have you reviewed our
      [contribution guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/.github/CONTRIBUTING.md)?
- [ ] Have you noted our
      [code guidelines](https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/master/docs/CODE_GUIDELINES.md)?
- [ ] Did you use the CLI to scaffold your new component?

```[tasklist]
### Acceptance criteria
- [ ] https://github.com/carbon-design-system/ibm-products/issues/3490
- [ ] https://github.com/carbon-design-system/ibm-products/issues/3478
```

```[tasklist]
### Pattern examples
- [ ] Vertical list
- [ ] Horizontal list
- [ ] Between lists
```

elycheea:
Still relevant. Was blocked by #1869 and will be prioritized once we clean house post #2884

elycheea:
@lee-chase as discussed, I don’t think we’re trying to sign up to become yet another dnd library unless we think that’s the best route forward. But we can start by exploring if we can provide the wrappers/styling for it at least. 👀

matthewgallo:
First iteration of drag and drop examples included [here](https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-patterns-drag-and-drop-draganddrop--docs). Happy to revisit this issue/create enhancement issues as they arise.

PR #4448, opened 2024-02-28
feat: add drag and drop examples/explorations

matthewgallo:
Contributes to #2042 

First bit of drag and drop examples added to storybook, this will not be something (as of right now) we'll provide components for. Instead, teams can use these examples to guide their own implementation of drag and drop.

#### What did you change?
Created story files in `packages/ibm-products/src/components/DragAndDrop`
Storybook

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -114,6 +114,7 @@
     "rerender",
     "resizer",
     "responsivity",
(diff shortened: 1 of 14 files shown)


davidmenendez (approved):
very cool!

PR #4463, opened 2024-03-05
fix(Datagrid): column customization accesibility violation role listbox

Ratheeshrajan:
Contributes to #4244 

Accessibility violation in datagrid column customization tearsheet `role="listbox"`.
Violation
Form control with "listbox" role has no associated label
image

#### What did you change?
`aria-label` was added to fix the violation.

Storybook


matthewgallo (approved):
I think an improvement here could be to use the `customizeTearsheetHeadingLabel` prop in the `CustomizeColumnTearsheet` and pass that as a prop to `Columns` and use it as the value for the `aria-label`. This way the `aria-label` isn't hard coded, but happy to have that included later as an enhancement to this.

Ratheeshrajan:
@matthewgallo @elycheea `area-label` now updated as prop `customizeTearsheetHeadingLabel`

matthewgallo (approved):

Issue #3803, opened 2023-11-14
Document recommended approach for enabling datagrid batch actions


elycheea:
Notes from batch actions release review:

- There are currently two options for enabling batch actions
  -  Use `batchActions: true`, and `toolbarBatchActions: getBatchActions(),` as properties to `useDatagrid` config options
  - Or provide your own `DatagridBatchActions` component (similar to `DatagridActions` for rendering the table toolbar)
 
I think we need to clean some things up there, I'm leaning towards the first option listed above because the second option doesn't support some of the responsive behavior built into the batch actions (the items move into the menu button as the viewport shrinks as needed).

_Originally posted by @matthewgallo in https://github.com/carbon-design-system/ibm-products/issues/3465#issuecomment-1791001075_
            
```[tasklist]
### Acceptance criteria
- [ ] Update docs to recommend using `batchActions: true`, `toolbarBatchActions:` over providing own `DatagridBatchActions`
```

We will keep `DatagridBatchActions` available but recommend first option so there’s  more consistency with responsiveness.

elycheea:
Scrum November 16 — could support both options for a more composable approach, but maybe after initial base component release.

PR #4472, opened 2024-03-05
docs(Datagrid): add docs around specifics of batch actions

matthewgallo:
Contributes to #3803 

Adds clarifying docs regarding batch actions in Datagrid.

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Datagrid.docs-page.js
```
Storybook, Datagrid docs page

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid.docs-page.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid.docs-page.js
@@ -417,6 +417,30 @@ return <Datagrid datagridState={datagridState} />;
         },
       },
       {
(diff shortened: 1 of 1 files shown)


packages/ibm-products/src/components/Datagrid/Datagrid.docs-page.js
```diff
@@ -416,6 +416,30 @@ return <Datagrid datagridState={datagridState} />;
           `,
         },
       },
+      {
+        description: `Batch actions can be included by providing \`batchActions: true\` and \`toolbarBatchActions: CarbonButtonProps[]\` to the \`useDatagrid\` hook. While passing in a \`DatagridBatchActions\` component will also work it does not have some of the same responsive behavior built-in, thus it is not the recommended approach, as compared to using the \`batchActions\` and \`toolbarBatchActions\` properties.`,
```
elycheea, line 420: Small suggestion for clarity. I think it’s clearer at least. 🤔 

```suggestion
        description: `Batch actions can be included by providing \`batchActions: true\` and \`toolbarBatchActions: CarbonButtonProps[]\` to the \`useDatagrid\` hook. While passing in a \`DatagridBatchActions\` component will also work, it does not have the same responsive behavior built-in, thus we recommend using the \`batchActions\` and \`toolbarBatchActions\` properties when possible.`,
```

elycheea (approved):

Issue #4059, opened 2024-01-16
[Datagrid] Select all checkbox and button should ignore disabled rows


xmannyxfreshx1:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

- When using the Datagrid and selectable rows, there is the ability to disable selection of some rows. When taking advantage of this ability, the select all checkbox and button should only select the rows that are not disabled.

> Is this issue related to a specific component?

- Datagrid, selectRows hook

> What did you expect to happen? What happened instead? What would you like to
> see changed?

- Currently, the select all checkbox and button will select all rows, even if some rows have row selection disabled. A workaround is to disable the select all checkbox but then we lose the ability to select all other available rows.
- The expectation I had was that the select all checkbox would only select all selectable rows and omit the disabled ones.
> What browser are you working in?

- Chrome

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

2.19.2

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

> Please create a reduced test case in CodeSandbox
>
> https://codesandbox.io/s/github/carbon-design-system/ibm-products/tree/main/examples/carbon-for-ibm-products/SidePanel

## Additional information

- Screenshots or code
- Notes

elycheea:
@makafsal Are you looking into this? If not, I think @Ratheeshrajan may be able to pick this up once Patterns behinds given that he just addressed this for customize columns.

PR #4473, opened 2024-03-05
fix: datagrid select all toggle rows not disabled

davidmenendez:
Contributes to #4059


matthewgallo (approved):
👏 Working as expected

Issue #4359, opened 2024-02-19
Fix scroll gradient in modals


elycheea:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

Scroll gradient is not positioned correctly in the `TagSet` overflow modal. Scroll gradient is not present in AboutModal but believe that it is intended to be.

TagSet overflow | AboutModal
----------------|-----------
<img width="573" alt="image" src="https://github.com/carbon-design-system/ibm-products/assets/1717398/c5ae13a5-5dab-4803-b628-cc8d7eb752ea"> | <img width="743" alt="image" src="https://github.com/carbon-design-system/ibm-products/assets/1717398/f34dda8c-beda-49bd-9a82-fe0423bc4ce2">

> Is this issue related to a specific component?

`TagSetModal`, `AboutModal`.
These likely should just be using [`hasScrollingContent`](https://react.carbondesignsystem.com/?path=/docs/components-composedmodal--overview#overflow-content) from the Carbon composed modal.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

Scroll gradient should be positioned to the bottom of the TagSet modal (likely a result of not having buttons in the `TagSetModal`.) `AboutModal` should also use scrolling content when content is long.

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?
v2.26.0

## Steps to reproduce the issue

1. Navigate to [TagSet story](https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-tag-set-tagset--hundreds-of-tags)
2. Click or activate the overflow (+xxx) to open the modal
3. Modal should open with the scrolling overflow floating

1. Navigate to [AboutModal story](https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-patterns-about-modal-aboutmodal--about-modal&args=content:3)
2. Set content to `long content`
3. Modal does not display a scrolling overflow

## Additional information

- Screenshots or code
- Notes

PR #4478, opened 2024-03-06
fix(TagSet): modal gradient

elycheea:
Contributes to #4359

Fixes modal scroll position. This was chilling in my branches since we discovered the about modal drama from last week. 😅 

#### What did you change?
~Adjust fade positioning although it may make more sense to use `hasScrollingContent`. Opening this PR as is for now since there’s a bug open on `hasScrollingContent` so may be better to wait to actually update this.~

Refactored to use `hasScrollingContent` from the Carbon modal, which requires the addition of an `aria-label`. Tried to introduce it in a way that will not affect anyone already using the `TagSet` or `TagOverflow`.

Storybook, CI


davidmenendez (approved):

AlexanderMelox (approved):
Looks good to me

Issue #4086, opened 2024-01-19
Create new `TagOverflow`component


elycheea:

elycheea:
@szinta This initial implementation leverages `TagSet`. I _think_ we discussed opening a followup ticket to remove the dependency on the original version so we can deprecate it in the future. Is that correct? Want to double check before I add that as another issue.

szinta:
That's right @elycheea. We need one new ticket to remove the dependency on Tagset component.

PR #4529, opened 2024-03-11
refactor: flatten story structure and story titles (components section)

matthewgallo:
Contributes to #3986 

First part of storybook simplification, the majority of the changes in this PR include removing the `getStoryTitle` utility for stories in `IBM Products/components`. I'll wait to do the same for `IBM Products/patterns` to reduce the number of changes required to review since it's a substantial amount.

This removes a lot of abstraction that I don't think is necessary and can make longer term maintenance more difficult.

In addition, I refactored how the `Canary` tag is applied in the side bar items in a way that no longer requires manipulating the story title. As well, as refactoring how we sort our stories (`.storybook/preview.js`).

#### What did you change?
Story files for `IBM Products/Components` section
#### How did you test and verify your work?
Storybook

PR #4535, opened 2024-03-12
feat(TagOverflow): new component TagOverflow

szinta:
Contributes to #4086 

New TagOverflow component to extend the functionality of existing TagSet component to other types of UI elements such as UserAvatar, custom react components etc.

#### What did you change?

- Generated new component files using `yarn generate TagOverflow`
- Generated TagOverflow Examples files using `yarn update-gallery-config`
- Updated the component files to implement TagOverflow functionality

- [x] Storybook


packages/ibm-products/src/components/TagOverflow/TagOverflow.mdx
elycheea, line 1: Based on the changes from #4529 we can change the Story from using `id` to `of`.

elycheea (commented):
Some initial comments based on the changes that Matt has been doing on Storybook refactor. :)

packages/ibm-products/src/components/TagOverflow/TagOverflow.stories.js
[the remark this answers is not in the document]
szinta: Done

packages/ibm-products/src/components/TagOverflow/TagOverflow.mdx
szinta: Done

packages/ibm-products/src/components/TagOverflow/TagOverflow.js
```diff
@@ -0,0 +1,266 @@
...
+// Carbon and package components we use.
+import { Tag, Tooltip } from '@carbon/react';
+import { TagSet } from '../TagSet';
+import { useResizeObserver } from '../../global/js/hooks/useResizeObserver';
+
+// The block part of our conventional BEM class names (blockClass__E--M).
+const blockClass = `${pkg.prefix}--tag-overflow`;
+const componentName = 'TagOverflow';
+
+// NOTE: the component SCSS is not imported here: it is rolled up separately.
```
davidmenendez, line 27: i'm sure a lot of these comments were added automatically, but we should remove any comments that aren't explicitly necessary.

packages/ibm-products/src/components/TagOverflow/TagOverflow.js
```diff
@@ -0,0 +1,266 @@
...
+      items,
+      itemTemplate,
+      maxVisible,
+      // Collect any other property values passed in.
+      ...rest
+    },
+    ref
+  ) => {
+    // type of Tags elements within the component
+    const tagType = 'blue';
```
davidmenendez, line 60: i don't know if the design docs explicitly require only blue tags, but if not i would use this as a default and add a color value to the item object for further user customization

packages/ibm-products/src/components/TagOverflow/TagOverflow.js
```diff
@@ -0,0 +1,266 @@
...
+            prev.push(cur);
+          } else {
+            maxReached = true;
+          }
+        }
+        return prev;
+      }, []);
+    }, [itemRefs, overflowRef, containerWidth, items]);
+
+    const ItemComponent = ({ templateProps }) => {
```
davidmenendez, line 126: typically i would expect to see any other additional component in it's own separate file for organizational purposes. also i like `itemTemplate` but i think something like `tagComponent` is slightly more straightforward.

you could just keep this here as a simple function. something like

```
const getCustomComponent = (item) => {
  const { className, id, ...other } = item;
  return createElement(tagComponent, {
    ...other,
    className: cx(`${blockClass}__item`, className),
    ref: (node) => itemRefHandler(id, node),
  });
};
```

packages/ibm-products/src/components/TagOverflow/TagOverflow.js
```diff
@@ -0,0 +1,266 @@
...
+      }
+    };
+
+    const getVisibleItems = useCallback(() => {
+      if (!itemRefs.current) {
+        return items;
+      }
+
+      const map = getMap();
+      const overflowContanerWidth =
```
davidmenendez, line 101: small typo. should be `overflowContainerWidth`. that should probably have been caught by the spellchecker!

packages/ibm-products/src/components/TagOverflow/TagOverflow.js
```diff
@@ -0,0 +1,266 @@
...
+          {
+            // switched classes dependant on props or state
+            // example: [`${blockClass}__here-if-small`]: size === 'sm',
+          }
+        )}
+        ref={containerRef}
+        role="main"
+        {...getDevtoolsProps(componentName)}
+      >
+        {!!visibleItems?.length &&
```
davidmenendez, line 184: since this variable is initiated with an empty array i think you can simplify this to just `visibleItems.length > 0`

packages/ibm-products/src/components/TagOverflow/TagOverflow.js
```diff
@@ -0,0 +1,266 @@
...
+                    <Tag className={`${blockClass}__item--tag`} type={tagType}>
+                      {item.label}
+                    </Tag>
+                  </Tooltip>
+                </div>
+              );
+            }
+          })}
+        <span className={`${blockClass}__indicator`} ref={overflowRef}>
+          {!!overflowItems?.length && (
```
davidmenendez, line 211: same as above `overflowItems.length > 0`

packages/ibm-products/src/components/TagOverflow/TagOverflow.js
```diff
@@ -0,0 +1,266 @@
...
+            <TagSet
+              tags={overflowItems}
+              allTagsModalTitle="All tags"
+              containingElementRef={overflowRef}
+              allTagsModalSearchLabel="Search all tags"
+              allTagsModalSearchPlaceholderText="Search all tags"
+              showAllTagsLabel="Show all tags"
+            ></TagSet>
```
davidmenendez, line 219: for any element declaration that doesn't have a child you can just use the `<TagSet />` notation

davidmenendez (changes requested):
looks great! just have a few comments. reach out if you have any questions!

packages/ibm-products/src/components/TagOverflow/TagOverflow.js
szinta: There are no defined guidelines for this. I did it that way because all of the [TagOverflow guidance ](https://pages.github.ibm.com/cdai-design/pal/components/tag-overflow/usage)screenshots have blue tags. I totally agree with you. We should provide an option for the user to customise the tag types.

packages/ibm-products/src/components/TagOverflow/TagOverflow.js
szinta: Made the changes as suggested. Thank you for optimising it nicely.

szinta:
@davidmenendez I have implemented all the review comments. Please have a look.

packages/ibm-products/src/components/TagOverflow/constants.js
```diff
@@ -0,0 +1,25 @@
...
+  green: 'Green',
+  gray: 'Gray',
+  'cool-gray': 'Cool-Gray',
+  'warm-gray': 'Warm-Gray',
+  'high-contrast': 'High-Contrast',
+  outline: 'Outline',
+};
+
+// Default value for the type prop of Tag elements within the component
+export const defaultTagType = 'blue';
```
elycheea, line 25: I don’t think we need to set the default to blue. I think we can use the same defaults from Carbon here. I think the guidance includes different examples but don’t believe it meant for them to default to blue. 🤔 I can double check with @Laura-Marshall though.

davidmenendez: yeah i agree with this. i think it's fine to just get rid of this since tags already come with a default state

davidmenendez:
looks good! thanks for making those changes. i think the only thing left is that small point that elysia pointed out regarding the default tag color 👍

packages/ibm-products/src/components/TagOverflow/constants.js
szinta: The logic for assigning 'blue' as the default Tag type has been removed. If type is not provided in item prop, tag will be rendered in Carbon's default state of 'grey'.

szinta:
> looks good! thanks for making those changes. i think the only thing left is that small point that elysia pointed out regarding the default tag color 👍

Done. Please have a look

szinta:
> @sintaibm when you get a chance can you also provide your environment details? i was thinking about that cspell issue and i don't know why it wasn't caught on your initial commit. the typo should have prevented the commit from happening. this is something i've been trying to investigate, so if you could provide information about your environment like node version, yarn version, OS, etc that would be super helpful!

I think the pre-commit hooks aren't working for me.  I don't see any commitlint / lint-staged logs or errors while committing my changes. To see the cspell errors, I had to manually run the command `yarn spellcheck`. I could see approximately 115 cspell errors in 31 files, indicating that these pre-commit hooks are not working for some other developers as well.

Please find my environment details below:
Node version - v18.19.0
Yarn - 4.0.2
OS - macOS Sonoma 14.3.1

davidmenendez (approved):
LGTM!

@elycheea if you have any other notes feel free to dismiss my approval

szinta:
@all-contributors please add @szinta for code

Issue #4476, opened 2024-03-06
Separate Decorator into multiple components


elycheea:
**Issue**

Technically, the `Decorator` is a small component (one icon, one label, and one value), but can behave differently depending on the variation chosen (non-interactive, link, single and dual buttons). 

As a single component, this requires a lot of if/else statements in code, one very long SCSS file, and some props that are not applicable depending on which variant is chosen. To make sense of this, a lot of supplementary documentation must be written to explain how it all works. This is complicated to implement and doesn't scale well.

**Solution**

It's been decided to break down the `Decorator` into smaller, more focused components. This means code, props, and documentation can be written exclusively for each component and its function.

Components
- `DecoratorBase` (internal, serves as a common code base for all the other Decorators)
- `Decorator` (non-interactive)
- `DecoratorLink` (as a single link)
- `DecoratorSingleButton` (as a single button)
- `DecoratorDualButton` (where the label and value each behave as button)

PR #4542, opened 2024-03-13
fix(Decorator): breakout single Decorator into many smaller Decorators

paul-balchin-ibm:
Contributes to #4476

**Issue**

The single Decorator has complex and changing `prop` relationships depending in the `kind` selected. This also requires complicated documentation to explain the different relationships.

**Solution**

Create an internal DecoratorBase component, and build out sub components where their props are specific to each type.

Convert monolithic Decorator into smaller, more focused "sub types": 
- Decorator
- DecoratorLink
- DecoratorSingleButton
- DecoratorDualButton

#### What did you change?

Too many files to list.

- [x] Storybook
- [x] Testing


davidmenendez (changes requested):
this looks great!

just one tiny nitpick: to avoid unnecessary repetition you can remove all the `defaults` in the variants and just use a single one in `DecoratorBase`

paul-balchin-ibm:
> ...remove all the `defaults` in the variants and just use a single one in `DecoratorBase`.

Great idea. Thanks for the details in the PMs  🎉

davidmenendez (approved):
woot

PR #4555, opened 2024-03-14
Side panel focus trap 4228

makafsal:
Contributes to #

Side panel focus issue

#### What did you change?
User `useFocus` hook

Storybook


elycheea:
@makafsal We get this DCO issue on occasion. 👀 Your commit history shows two IDs (you can use `git show <commit-hash>` to check). One is your public GitHub account — the other is your IBM email. I think you can try to add your IBM address to your profile if it’s not already there.

If that doesn’t work, then you can either amend the commits (which sometimes feels like a “dirtier” approach) or create a new PR with the changes using the right ID.

Issue #4228, opened 2024-02-07
Fix focus trap issues in side panel


elycheea:
While addressing #3499, @makafsal noted that we will be able to reuse the `useFocus()` hook to add focus trap in the side panel as well.

PR #4559, opened 2024-03-15
fix(sidepanel): resolve focus trap issue

makafsal:
Contributes to #4228 

Fix focus trap issue in side panel

#### What did you change?
Used `useFocus` hook

Storybook

### DIFF
--- a/packages/ibm-products/src/components/SidePanel/SidePanel.js
+++ b/packages/ibm-products/src/components/SidePanel/SidePanel.js
@@ -18,10 +18,9 @@ import { moderate02 } from '@carbon/motion';
 import { getDevtoolsProps } from '../../global/js/utils/devtools';
 import { allPropTypes } from '../../global/js/utils/props-helper';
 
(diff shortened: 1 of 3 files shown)


packages/ibm-products/src/components/SidePanel/SidePanel.js
```diff
@@ -393,23 +367,19 @@ export let SidePanel = React.forwardRef(
...
-          bodyNode: innerContentRef.current,
-          startTrapRef,
-          endTrapRef,
-          currentActiveNode,
-          oldActiveNode,
-        });
+    useEffect(() => {
+      if (open) {
+        setTimeout(() => {
+          const primeFocusEl = document.querySelector(selectorPrimaryFocus);
```
matthewgallo, line 373: We should add a check here to make sure `document` is defined (ie in the case of SSR apps like next.js).

makafsal: Done.

matthewgallo (approved):

Issue #4323, opened 2024-02-13
Prevent PageHeader using title as trailing breadcrumb


richardpilot:
## What will this achieve?
I would like to prevent the `PageHeader` component using the title prop as the trailing breadcrumb on scroll. The component already allows you to use the `isCurrentPage` prop in the PageHeader `breadcrumbs` prop to achieve the same effect and it would allow the dev to provide a more concise title to appear when it's a breadcrumb. A suggestion could be to see whether `isCurrentPage` is already specified before performing this behaviour.

A good example of this use case in the the PageHeader's own storybook

![image](https://github.com/carbon-design-system/ibm-products/assets/13081745/2cb27755-7008-4d44-94a9-7dd6a3e56bfe)
![image](https://github.com/carbon-design-system/ibm-products/assets/13081745/0f7500dd-2a44-4b99-a50c-3ad7c103d0ae)

## How will success be measured?

When either an alternative collapsed title can be provided or the isCurrentPage in breadcrumbs are honoured and the title prop not used.

## Additional information

```[tasklist]
### Acceptance criteria
- [ ] Add prop for custom breadcrumb title
- [ ] Add a test
```

davidmenendez:
hey @richardpilot a fix was merged last week, which you should be able to see here https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-page-header-pageheader--fully-loaded-and-some&globals=viewport:basic

PR #4562, opened 2024-03-15
fix: adds short text option for pageheader

davidmenendez:
Contributes to #4323 

adds an optional `shortText` prop that can be used instead of label when a breadcrumb is particularly long. also some slight cleanup.


packages/ibm-products/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -356,6 +356,11 @@ BreadcrumbWithOverflow.propTypes = {
        */
       label: PropTypes.node,
 
+      /**
+       * An optional title label for extra long breadcrumbs
+       */
+      shortText: PropTypes.string,
```
elycheea, line 362: Since this is explicitly about the breadcrumb label maybe something more like `shortTitle`, `shortLabel`, or even `shortName`?

```suggestion
      shortText: PropTypes.string,
```

packages/ibm-products/src/components/PageHeader/PageHeader.stories.js
```diff
@@ -341,6 +341,7 @@ const title = {
   },
   'Long title with icon': {
     text: 'A very long page title which will almost certainly have to be truncated at some point',
+    shortText: 'A shortened title for nav',
```
elycheea, line 344: Placement here is a bit odd. Setting this as part of `title` does make sense, but I feel like I’d look for examples for this more explicitly. (I’d realistically probably look in breadcrumbs for `breadcrumb with shortened title` or something like that, but it might not make sense for how we’d actually set this.

I do think a separate option under title like `Long title with short name` might be a good alternative though.

elycheea (changes requested):
Some general naming/structure thoughts.

We should add a test for the shortened title. Also not seeing the new prop in `BreadcrumbWithOverflow` story.

davidmenendez:
@elycheea thanks for the comments. i didn't realize we exposed `BreadcrumbWithOverflow` which exposed a problem with my fix. will continue to work on this.

davidmenendez:
@elycheea ready for re-review

elycheea (commented):
Oh, interesting to split into two props. Any specific reason why we approached it this way? I think it’s a little additional work on the consumer require two props to leverage a shorter title?

packages/ibm-products/src/components/PageHeader/PageHeader.stories.js
```diff
@@ -338,6 +338,7 @@ const title = {
   },
   'Long title with icon': {
     text: 'A very long page title which will almost certainly have to be truncated at some point',
+    shortTitle: 'A shortened title for nav',
     loading: false,
     icon: (props) => <Bee size={24} {...props} />,
```
elycheea, line 343: Looks good. While this works, I think a separate variant like “With `shortTitle` in breadcrumbs” would be easier to capture this behavior.

If we want to reuse this one, I think “Long title with icon and shortTitle” might be better. Then to help with discoverability, we can change the title text? 🤔 

```suggestion
  'Long title with icon and shortTitle': {
    text: 'A very long page title with a short version in breadcrumbs; this will almost certainly be truncated at some point',
    shortTitle: 'A shortened title for nav',
    loading: false,
    icon: (props) => <Bee size={24} {...props} />,
```

davidmenendez:
@elycheea sorry i missed that last comment.

fixt 👍

elycheea:
@davidmenendez Haha, I think my comment was for PageHeader but latest change was in BreadcrumbWithOverflow? 👀 Small change either way though! You might be able to commit the suggestion as is if it makes sense to you too though.

davidmenendez:
@elycheea apologies for the mix up! changes pushed

packages/ibm-products/src/components/PageHeader/PageHeader.stories.js
```diff
@@ -336,9 +336,9 @@ const title = {
...
   },
-  'Long title with icon': {
-    text: 'A very long page title which will almost certainly have to be truncated at some point',
-    shortTitle: 'A shortened title for nav',
+  'Long title with iconLong title with icon and shortTitle': {
```
elycheea, line 339:
```suggestion
  'Long title with icon and shortTitle': {
```

elycheea: 😅

davidmenendez: 😱

davidmenendez:
@elycheea fixt 4 real

packages/ibm-products/src/components/PageHeader/PageHeader.js
```diff
@@ -81,7 +81,6 @@ export let PageHeader = React.forwardRef(
       tags,
       title,
       withoutBackground,
-      breadcrumbOverflowTooltipAlign = defaults.breadcrumbOverflowTooltipAlign,
```
elycheea, line 84: Don’t think we should be removing this here? This and the `overflowTooltipAlign` below were both recently added from @makafsal’s PR #4907.

packages/ibm-products/src/components/PageHeader/PageHeader.js
```diff
@@ -518,21 +531,14 @@ export let PageHeader = React.forwardRef(
...
                           noTrailingSlash={!!title}
                           overflowAriaLabel={breadcrumbOverflowAriaLabel}
-                          breadcrumbs={
-                            breadcrumbs && breadcrumbItemForTitle
-                              ? breadcrumbs.concat(breadcrumbItemForTitle)
-                              : breadcrumbItemForTitle
-                              ? [breadcrumbItemForTitle]
-                              : breadcrumbs // breadcrumbs may be null or undefined
-                          }
-                          overflowTooltipAlign={breadcrumbOverflowTooltipAlign}
```
elycheea, line 533: Let’s keep this line since these were just added to fix another issue.

elycheea (commented):
@davidmenendez _One more thing_ since it looks like we unintentionally revert some changes that @makafsal just merged in #4907.

davidmenendez:
@elycheea thanks for catching that. updated

elycheea (approved):

makafsal:
@elycheea @davidmenendez  My changes are there, looks good 👍

devadula-nandan (approved):

Issue #4477, opened 2024-03-06
New Component: FilterPanel


lily-peng:
# FilterPanel
_Please refer to #4052 for more details about migrating legacy Security components to **Products v2**._
## Brief description
Migrate **FilterPanel** from `Products v1 / Security / FilterPanel` to **Products v2**.

References: [GitHub](https://github.com/carbon-design-system/ibm-products/tree/main_v1/packages/security/src/components/FilterPanel), [Storybook](https://v1-ibm-products.carbondesignsystem.com/?path=/story/security-patterns-filterpanel--default).

For migration purposes:

* New component architecture must meet **Products v2**'s code specifications.
* The design and behavior of version `v2` must match version `v1` exactly.

## Tasks

Before starting work on this epic, please review and complete the following.

- [ ] Initial review of design/existing code. Refer to existing component specifications: [GitHub](https://github.com/carbon-design-system/ibm-products/tree/main_v1/packages/security/src/components/FilterPanel), [Storybook](https://v1-ibm-products.carbondesignsystem.com/?path=/story/security-patterns-filterpanel--default).

### Working in Carbon for IBM Products package

- [x] Have you reviewed our [contribution guidelines](https://github.com/carbon-design-system/ibm-products/blob/master/.github/CONTRIBUTING.md)?
- [x] Have you noted our [code guidelines](https://github.com/carbon-design-system/ibm-products/blob/master/docs/CODE_GUIDELINES.md)?
- [ ] Did you use the CLI to scaffold your new component?

```[tasklist]
### Tasks
- [x] Label (required by Checkbox)
- [x] Checkbox
- [x] Checkbox with overflow
- [x] Group
- [x] Accordion and AccordionLabel
- [ ] Search
```

PR #4563, opened 2024-03-15
feat(FilterPanel): new component with label and checkbox

paul-balchin-ibm:
Contributes to #4477

Migrating **FilterPanel**, **FilterPanelLabel**, and **FilterPanelCheckbox** from `Products v1 / Patterns / FilterPanel`.

FilterPanelLabel and FilterPanelCheckbox are two of the eight(-ish) FilterPanel subcomponents. We will be splitting FilterPanel's feature set across multiple PRs to make it easier to manage.

`v1` references: [GitHub](https://github.com/carbon-design-system/ibm-products/blob/main_v1/packages/security/src/components/FilterPanel/FilterPanel.js), [Storybook](https://v1-ibm-products.carbondesignsystem.com/?path=/story/security-patterns-filterpanel--default).

#### What did you change?

This is a new `v2` component.

**FilterPanel prop updates**

| Old prop   | New prop  | Change             |
| :--------- | :-------- | :----------------- |
| children   | children  | No change          |
| className  | className | No change          |
| title      | title     | No change          |
| filterData | —         | Deprecated in `v1` |

**FilterPanelLabel prop updates**

| Old prop       | New prop  | Change                                                                                                    |
| :------------- | :-------- | :-------------------------------------------------------------------------------------------------------- |
| className      | className | No change                                                                                                 |
| count          | count     | No change                                                                                                 |
| —              | labelText | New                                                                                                       |
| title          | title     | No change                                                                                                 |
| children       | —         | Replaced by `labelText`                                                                                   |
| countClassName | —         | Removed. Developer can use top-level `className` and <br/>inheritance to target specific elements via CSS |
| countLabel     | —         | Removed                                                                                                   |

**FilterPanelCheckbox prop updates**

| Old prop   | New prop  | Change                           |
| :--------- | :-------- | :------------------------------- |
| className  | className | No change                        |
| count      | count     | No change                        |
| —          | id        | New, required by Carbon Checkbox |
| —          | labelText | New, required by Carbon Checkbox |
| title      | title     | No change                        |
| countLabel | —         | Removed                          |


packages/ibm-products/src/components/FilterPanel/FilterPanelLabel/FilterPanelLabel.js
```diff
@@ -0,0 +1,92 @@
...
+    ref
+  ) => {
+    return (
+      <span
+        {...rest}
+        className={cx(blockClass, className)}
+        ref={ref}
+        {...getDevtoolsProps(componentName)}
+      >
+        <span className={`${blockClass}__text`} title={labelTitle}>
```
davidmenendez, line 46: title shouldn't be necessary here

packages/ibm-products/src/components/FilterPanel/FilterPanelLabel/FilterPanelLabel.js
```diff
@@ -0,0 +1,92 @@
...
+          {label}
+        </span>
+
+        <span
+          className={`${blockClass}__count`}
+          aria-label={setCountAriaLabel(count)}
```
davidmenendez, line 52: aria-label shouldn't be necessary here

packages/ibm-products/src/components/FilterPanel/FilterPanelCheckbox/FilterPanelCheckbox.js
```diff
@@ -0,0 +1,89 @@
...
+  ) => {
+    return (
+      <Checkbox
+        {...rest}
+        className={cx(blockClass, className)}
+        labelText={
+          <FilterPanelLabel
+            count={count}
+            setCountAriaLabel={setCountAriaLabel}
+            labelTitle={labelTitle || labelText}
```
davidmenendez, line 38: since the underlying `title` attribute is unnecessary we should be able to just stick to one prop value here

davidmenendez (changes requested):
looks good! just some small notes on the usage of `aria-label` and `title`

packages/ibm-products/src/components/FilterPanel/FilterPanelLabel/FilterPanelLabel.js
paul-balchin-ibm: The filter panel (and the labels) have very limited horizontal space. It's easy for database-driven labels to be long enough to get cut off. A `title` would at least allow a user to hover over the label and see the full text.

packages/ibm-products/src/components/FilterPanel/FilterPanelLabel/FilterPanelLabel.js
paul-balchin-ibm: Agreed.

packages/ibm-products/src/components/FilterPanel/FilterPanelLabel/FilterPanelLabel.js
paul-balchin-ibm: Also, we are passing the label text as `children`, which allows the dev to use `React.node`. Should we switch to label as a proper prop and limit it to "string"?

davidmenendez (dismissed):

packages/ibm-products/src/components/FilterPanel/FilterPanelLabel/FilterPanelLabel.js
elycheea: >The filter panel (and the labels) have very limited horizontal space. It's easy for database-driven labels to be long enough to get cut off. A title would at least allow a user to hover over the label and see the full text.

This has been a theme that’s come up in discussion with @kristastarr on the Datagrid reviews too. I think we can keep as is for this first PR even though I don’t think this is preferred behavior based on what I’ve started reading.

Since it applies to Datagrid as well, I think we should open this as further issue/discussion.

paul-balchin-ibm: All good points. 

On further research, I found the following:
 - `FilterPanelLabel` is used primarily by the checkbox, group, and accordion subcomponents.
 - `FilterPanelCheckbox` specifically returns Carbon's `Checkbox` as-is.

I thought I would standardize on how Carbon's `Checkbox` does it, since we're using it's "label" feature anyways, using the following props:
 - `labelText` is a `node` (required, allows any string or markup)
 - `title` is a `string` (optional, string only, because any markup will show as [Object object] in the tooltip)

packages/ibm-products/src/components/FilterPanel/FilterPanelLabel/FilterPanelLabel.js
paul-balchin-ibm: Done.

packages/ibm-products/src/components/FilterPanel/FilterPanelCheckbox/FilterPanelCheckbox.js
paul-balchin-ibm: Done.

davidmenendez (approved):

Issue #4192, opened 2024-02-01
ExportModal allows filename with blanks to be exported and also has no maxLength restriction


naitikwrk:
## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [] Legacy/CDAI
- [] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

Ideally, it would be nice if we can pass props to TextInput so that we can control properties like invalid, invalidText, maxLength etc. so that ExportModal behaviour can be customised as per consuming app.

naitikwrk:
@elycheea yes. I have the fix on my local fork. I'm waiting for approval from my team before i can contribute here.

davidmenendez:
@naitikwrk thanks for opening up this issue and opening a PR. i noticed that this is for v1. would you be interested in also opening a PR in the v2 (main) branch? i assume we want to include this bug fix in v2 @elycheea

naitikwrk:
@davidmenendez sure

PR #4566, opened 2024-03-17
fix(ExportModal): prevent saving with empty filename

naitikwrk:
Also allow passing additional props to the text input.

Contributes to #4192

#### What did you change?
- Added blank spaces validation on primary button disability check.
- Added additional `inputProps` prop that is passed on to the text input component for allowing more control over it.

- Added related unit tests.


davidmenendez (changes requested):
this looks great!

one small comment. i like the idea of just having a single object for all the standard carbon inputs. i think we should try to phase out the redundant props while we're at it. so let's start by adding to the props documentation that we're going to be deprecating those props at some point as a way to slowly get users to migrate to using `inputProps` instead of `inputLabel` etc

davidmenendez (approved):
actually please disregard my previous comment. i didn't realize that this was targeting v1. i don't think we need to worry about updating the props to address my concerns in the v1 branch.

Issue #4230, opened 2024-02-07
Customize columns Tearsheet subtitle style


dsav-eng:
## Title line template: 
Datagird's Customize columns tearsheet subheading style

## What package(s) are you using?

- [x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)
- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

> Describe in detail the issue you're having.

User is not able to view a full text of tearsheet's subtitle

> Is this issue related to a specific component?

Tearsheet used for datagrid's column customatization
<img width="1512" alt="Screenshot 2024-02-07 at 10 09 26 AM" src="https://github.com/carbon-design-system/ibm-products/assets/61568562/c94cae22-0a70-4885-8e95-1cf04ea0ef7a">

> What did you expect to happen? What happened instead? What would you like to
> see changed?

User should be able to view a full text of tearsheet's subtitle

> What browser are you working in?

Chrome

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

2.17.1

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

Turbonomic

## Steps to reproduce the issue
The issue could be viewed at Carbon's storybook - [Datagirs's column customatization story](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-datagrid-extensions-columncustomization--column-customization-usage-story)

```[tasklist]
### Acceptance criteria
- [ ] Add tooltip to Tearsheet description
- [ ] Update tests as needed
```

elycheea:
Believe this actually results from the Tearsheet. If we want to preserve the original design which overflows after two lines, we can probably add a `title` attribute or tooltip for now.

Ratheeshrajan:
@elycheea we need to make the changes in the Tearsheet component to get the title added, hope it's fine.
`packages/ibm-products/src/components/Tearsheet/TearsheetShell.js`

elycheea:
Was expecting the change in Tearsheet so no worries @Ratheeshrajan!

Did a lot more reading on tooltip versus `title` attribute and think `Tooltip` is the better approach[^1][^2].

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#accessibility_concerns
[^2]: https://www.24a11y.com/2017/the-trials-and-tribulations-of-the-title-attribute/

[...]

PR #4571, opened 2024-03-18
fix(Tearsheet): description title show full text when overflow

Ratheeshrajan:
Contributes to #4230 

If customize column tearsheet `Description`  overflows two lines add `title` attribute to show the full text.

#### What did you change?
Added title to tearsheet description.
`src/components/Tearsheet/TearsheetShell.js`
image

Storybook


elycheea:
Been doing some reading on tooltip v. overflow thanks to recent discussions on them in filter panel (@paul-balchin-ibm @davidmenendez) and accessibility audits that @kristastarr have been in. Tooltip is probably the better approach for this upon further reading[^1][^2]. Sorry for the rework, but I’ll also update the notes in the issue. 😓 

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#accessibility_concerns
[^2]: https://www.24a11y.com/2017/the-trials-and-tribulations-of-the-title-attribute/

Ratheeshrajan:
@elycheea title replaced with tooltip.

davidmenendez:
@elycheea @Ratheeshrajan i posted a [comment in the issue](https://github.com/carbon-design-system/ibm-products/issues/4230#issuecomment-2018735907) but i missed some of discussion going on here.

basically my opinion is that from an accessibility and UX standpoint that i don't think there's any reason to limit the text here and hide it through means of tooltip or title or anything.

i think the code looks great and it's a great solution if tooltip is the route we decide to take. @elycheea if you have any strong feelings one way or the other let me know!

packages/ibm-products/src/components/Tearsheet/TearsheetShell.js
```diff
@@ -115,6 +116,23 @@ export const TearsheetShell = React.forwardRef(
...
+    const tooltipTriggerRef = useRef(null);
+    const [tooltipText, setTooltipText] = useState(null);
+
+    useEffect(() => {
+      if (open) {
+        if (
+          tooltipTriggerRef?.current?.scrollHeight >
+          tooltipTriggerRef?.current?.clientHeight
+        ) {
+          setTooltipText(tooltipTriggerRef?.current?.innerHTML);
```
davidmenendez, line 129: i don't think it's necessary to set the text here. you should just be able to use the `description` prop in the tooltip label. i would probably just change this to be a boolean like `useTooltip` instead.

davidmenendez (changes requested):
one comment

packages/ibm-products/src/components/Tearsheet/TearsheetShell.js
elycheea: This is actually a lot like StringFormatter now. I think we should figure out how we want to handle utilities though.

davidmenendez (approved):

davidmenendez:
@Ratheeshrajan please review the merge conflicts when you have time 👍

elycheea:
Closing this for now and will move this issue back to backlog/triage for now. 

@Ratheeshrajan you’re welcome to reopen any of these PRs later if you’d like to address the conflicts — they’re probably quite a bit since we migrated several components to Typescript since this initial PR was opened. Also want to close these and move the issues back to the backlog in the meantime since you’re also on other projects now too. :)

PR #4573, opened 2024-03-18
fix: filterable column counts

lee-chase:
Filterable column count includes the non-filterable spacer column.

#### What did you change?

Filterable column counts to account for this.

Storybook


matthewgallo (approved):

matthewgallo:
@lee-chase can you also open a v1 PR for this as well?

Issue #4576, opened 2024-03-18
cleanup OptionsTile


davidmenendez:
tangentially related to #4281 

while looking at updating `OptionsTile` to typescript i noticed that this particular component is a bit outdated and could use some looking at and cleanup.

I attempted to do this myself, but after speaking with @janhassel it became clear that this wasn't a quick job. so someone with more time to dedicate to this effort should be put in charge.

to whomever decides to tackle this, feel free to reference https://github.com/carbon-design-system/ibm-products/pull/4582 which is a good starting point and already has some great feedback in it.

davidmenendez:
closing in favor of #6936 which address some of the same concerns but also has the added benefit of being more recent, relevant, and detailed.

PR #4582, opened 2024-03-19
fix: options tile refactor

davidmenendez:
Contributes to #4576 

refactors `OptionsTile` to bring it more into line with our current code standards. does a ton of cleanup and also addresses some accessibility concerns.

my initial commit does not include any updates to the styles and removes the current animation, which doesn't look great. you can clearly see jumpy behavior when opening and closing the tile. maybe someone else might want to tackle that specifically or i can continue to do it.

I think that one of the main things to be mindful of is usage of the `details` element. it's definitely handy to have a native accordion style element that internally handles the `open` and `onToggle` state. the caveat here though is that we can't specify our own `open` state because of how it works under the hood. basically it'll cause an infinite loop of state changes if you try to use them together and mange the state separately. there could be work arounds if we need to specifically expose an `onToggle` callback prop, but for the initial commit i've opted to not manage the `open` state with the exception of using `defaultOpen`.

also the testing has not been updated. i wanted to get feedback initially before updating tests.

### DIFF
--- a/packages/ibm-products/src/components/OptionsTile/OptionsTile.js
+++ b/packages/ibm-products/src/components/OptionsTile/OptionsTile.js
@@ -61,26 +61,26 @@ export let OptionsTile = React.forwardRef(
     };
 
     const getIcon = () => {
+      if (locked) {
+        return Locked;
+      }
       if (invalid) {
         return WarningFilled;
       }
       if (warn) {
         return WarningAltFilled;
       }
-      if (locked) {
-        return Locked;
-      }
     };
 
     const getText = () => {
+      if (locked) {
+        return lockedText;
+      }
       if (invalid) {
         return invalidText;
       }
       if (warn) {
         return warnText;
-      }
-      if (locked) {
-        return lockedText;
       }
       return summary;
     };
@@ -196,8 +196,7 @@ OptionsTile.propTypes = {
    */
   lockedText: PropTypes.string,
   /**
-   * Provide a function which will be called each time the user
-   * interacts with the toggle.
+   * Callback function for the built Carbon toggle element.
    */
   onToggle: PropTypes.func,
   /**
--- a/packages/ibm-products/src/components/OptionsTile/OptionsTile.js
+++ b/packages/ibm-products/src/components/OptionsTile/OptionsTile.js
@@ -1,23 +1,15 @@
 /**
- * Copyright IBM Corp. 2021, 2022
+ * Copyright IBM Corp. 2021, 2024
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
 
-// Import portions of React that are needed.
-import React, { useState, useRef } from 'react';
-
-// Other standard imports.
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
-
 import { getDevtoolsProps } from '../../global/js/utils/devtools';
-import uuidv4 from '../../global/js/utils/uuidv4';
 import { pkg } from '../../settings';
-import { useControllableState } from '../../global/js/hooks';
-
-// Carbon and package components we use.
 import { Layer, Toggle } from '@carbon/react';
 import {
   ChevronDown,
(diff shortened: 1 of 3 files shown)


elycheea:
@janhassel Do you mind taking a quick peek at this as well?

packages/ibm-products/src/components/OptionsTile/OptionsTile.js
```diff
@@ -1,291 +1,143 @@
...
+    const renderTitle = () => {
+      const Icon = getIcon();
+      const text = getText();
 
       return (
         <div className={`${blockClass}__heading`}>
-          <h6 id={titleId} className={`${blockClass}__title`}>
-            {title}
-          </h6>
+          <h6 className={`${blockClass}__title`}>{title}</h6>
```
janhassel, line 94: Ideally, this would not use a fixed heading level as it cause accessibility issues. I'm aware this issue existed before already, but a refactor would be a great opportunity to address this.

I suggest to use Carbon's `<Section>` and `<Heading>` here.

janhassel, line 94: It look like `props.id` is not used anywhere even though the toggle's `aria-labelledby` refers to it.

```suggestion
          <h6 id={id} className={`${blockClass}__title`}>{title}</h6>
```

packages/ibm-products/src/components/OptionsTile/OptionsTile.js
```diff
@@ -303,96 +155,70 @@ export let OptionsTile = React.forwardRef(
...
-
+  defaultOpen: PropTypes.bool,
+  /**
+   * Designates if the toggle is toggled by default
+   */
+  defaultToggled: PropTypes.bool,
+  /**
+   * Optional id
+   */
+  id: PropTypes.string,
```
janhassel, line 181: This can't be optional since the Toggle refers to it via `aria-labelledby`.

janhassel (commented):
@davidmenendez Since you mentioned that the PR is not finalized yet, let me know if some of these points are already planned to be addressed.

- There seems to have been a regression with the animation. In an [earlier deploy](https://deploy-preview-3381--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-options-tile-optionstile--options-tile) there is no jumping visible. Let's aim to fix this regression instead of removing it altogether.
- The behavior is currently not aligned with the design specifications. If the toggle is included and turned off, the description line in the summary element should be hidden.
- Could you elaborate on the problem with the current implementation leveraging `useControllableState`?
  - `props.defaultOpen` and `props.defaultToggled` make a lot of sense to me and should be added to the `useControllableState` call.
- The PR in the current state introduces breaking changes. Is there a process in place for deprecation and backwards compatibility for this library ? @elycheea

davidmenendez:
> @davidmenendez Since you mentioned that the PR is not finalized yet, let me know if some of these points are already planned to be addressed.
> 
> * There seems to have been a regression with the animation. In an [earlier deploy](https://deploy-preview-3381--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-options-tile-optionstile--options-tile) there is no jumping visible. Let's aim to fix this regression instead of removing it altogether.

you're right, i don't see the jumpyness in that preview. my intent wasn't to strictly remove it, but on this first pass to not include it since there was clearly something else affecting it. whether or not i do it now or have someone else look at it in a follow up was a proposal i was making.

> * The behavior is currently not aligned with the design specifications. If the toggle is included and turned off, the description line in the summary element should be hidden.

i will address that.
> * Could you elaborate on the problem with the current implementation leveraging `useControllableState`?
>   
>   * `props.defaultOpen` and `props.defaultToggled` make a lot of sense to me and should be added to the `useControllableState` call.

my initial thought process was wondering if this level of granularity necessary for the internal state. on second thought though i think i agree that it was probably premature of me to remove it. it is always better to allow the user to have that kind of control. i will go ahead and reuse it at your suggestion.

> * The PR in the current state introduces breaking changes. Is there a process in place for deprecation and backwards compatibility for this library ? @elycheea

this i'm not 100% sure about. i think we've tried not to introduce major version changes in favor of using v2 versions of the component. if we need to give this a v2 treatment that's totally doable.

thanks for the feedback @janhassel very helpful!

davidmenendez:
@janhassel apparently i was able to alleviate the issue i was seeing initially with the `onToggle` `open` infinite loop with `useControllableState` so that was nice

janhassel:
@davidmenendez I looked further into what you were saying about the infinite loop. The underlying issue seems to be that the `<details>` element realistically cannot be controlled as the `toggle` event is fired after the element changes its state.

In this case I think it would be best to make the opening and closing mechanism uncontrolled and just provide a `defaultOpen` prop similar to what you had in the beginning. It should not be used in a `useEffect` then though.

I'd leave the `useControllableState` hook alone. If there's anything that needs to be updated, it should be contributed back Carbon itself, otherwise we'll have an inconsistent DX.

Irt the animation: I strongly believe an implementation-only refactor should not come with visual regressions, therefore I see including the animation as a requirement for the refactor's release. Ultimately, it's your call obviously @elycheea.

Tying into that is the story about the release itself. If it'll be flagged as a v2 component or leverage a feature flag system similar to what Carbon is building out right now, the above point would only apply after the new implementation is considered the recommendation and out of experimental/unstable/canary.

That brings up the question: can the current API be kept despite the needed fix of controlled/uncontrolled? Personally, I think so: if we add `props.defaultOpen` we could still support `props.open` as it realistically does the same thing and was just advertised wrongly. `props.open` could then carry a deprecation warning and a note to use `props.defaultOpen` instead.
To address the other state (state of the toggle), the approach could look similar: add `props.defaultEnabled` which is then used in the `useControllableState` hook to give the option of an uncontrolled implementation. When using `props.enabled`, the toggle can be controlled as before (no deprecation needed).

Let me know if I'm missing something.

davidmenendez:
@janhassel thanks for the feedback. i definitely agree with what you're saying. i think i may have been a bit overzealous thinking i could quickly clean things up, but you're right. a refactor shouldn't include a visual regression even if it's already broken.

i'm going to go ahead and close this PR and let someone with more time to dedicate take a look at this. i'll open up a separate issue to address fixing the animation bug as well. i'll just stick with the initial goal of moving this to typescript 👍

Issue #4260, opened 2024-02-08
Add Typescript types to CreateTearsheetDivider


matthewgallo:
As we add Typescript support, want to call out a few helpful tips from the Carbon TypeScript guide that is largely relevant to our library as well.

### Changing the extension via git

When changing the file extension — **do not simply rename the file**. **Instead, change the extension via `git`.**

For example,

```
git mv packages/ibm-products/src/components/ComponentName/ComponentName.js packages/ibm-products/src/components/ComponentName/ComponentName.tsx
```

### Do not add types to internal components or functions
- Leave internals as `.js`
- Add a [jsdoc type annotation](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html#providing-type-hints-in-js-via-jsdoc) of `/** @type any */`
- [Enable errors](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html#ts-check) by adding `// @ts-check` to the first line in the file.

### Testing changes
A few different options for testing changes —

- At the bottom of a file, write a dummy component that uses the component that you’re converting to ensure that you can still pass all appropriate props.
- Take a storybook example for a component, copy and paste it into the bottom of the `.tsx` file and validate whether or not it accepts the props as you’ve defined them.

```[tasklist]
### Tasks
- [ ] Change extension to `.tsx` **via git** (**do not simply rename**, see above)
- [ ] Copy the PropTypes definitions above the component definition
- [ ] Retool the PropTypes to be a TypeScript interface
- [ ] Fix errors as they appear
- [ ] Do NOT add types to internal components or functions that are not exported as part of the public API
- [ ] Test your changes (see above)
```

### Resources
Carbon’s guide on [adding component types](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/adding-component-types.md)

PR #4584, opened 2024-03-20
refactor(CreateTearsheetDivider): convert to typescript

szinta:
Contributes to #4260 

Adds TypeScript types to CreateTearsheetDivider component.

#### What did you change?

`packages/ibm-products/src/components/CreateTearsheet/CreateTearsheetDivider.tsx`

- [X] Storybook
- [X] Validated by pasting the component at the bottom of the .tsx file


packages/ibm-products/src/components/CreateTearsheet/CreateTearsheetDivider.tsx
```diff
@@ -1,19 +1,28 @@
...
+
+interface CreateTearsheetDividerProps {
+  /** Specifies an optional className to be added to the tearsheet divider */
+  className?: string;
+}
+
 export let CreateTearsheetDivider = forwardRef(
-  ({ className, ...rest }, ref) => {
+  (
+    { className, ...rest }: PropsWithChildren<CreateTearsheetDividerProps>,
```
matthewgallo, line 23: I think a better option instead of `PropsWithChildren` might be to just use `React.FC<CreateTearsheetDividerProps>` since this component doesn't accept any children.

matthewgallo (changes requested):

packages/ibm-products/src/components/CreateTearsheet/CreateTearsheetDivider.tsx
matthewgallo: It would look something like this:
```jsx
export let CreateTearsheetDivider: React.FC<CreateTearsheetDividerProps> = forwardRef(
  ({
    className,
    ...rest,
  },
  ref: ForwardedRef<HTMLSpanElement>
)
```

matthewgallo (approved):

PR #4594, opened 2024-03-21
feat(FilterPanel): added checkbox with overflow

paul-balchin-ibm:
Contributes to #4477

Migrating **FilterPanelCheckboxWithOverflow** from `Products v1 / Patterns / FilterPanel`.

`v1` references: [GitHub](https://github.com/carbon-design-system/ibm-products/blob/main_v1/packages/security/src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.js), [Storybook](https://v1-ibm-products.carbondesignsystem.com/?path=/story/security-patterns-filterpanel-filterpanelcheckboxwithoverflowmenu--default).

#### What did you change?

This is a new `v2` subcomponent to `FilterPanel`.

**FilterPanelCheckboxWithOverflow prop updates**

| Old prop              | New prop          | Change                                           |
| :-------------------- | :---------------- | :----------------------------------------------- |
| children              | children          | No change                                        |
| className             | className         | No change                                        |
| (count)                     | count             | Made explicit\*                                  |
| (id)                     | id                | Made explicit\*                                  |
| (labelText)                     | labelText         | Made explicit\*                                  |
| (onChange)                     | onChange          | Made explicit\*                                  |
|                       | overflowMenuProps | New object `{}` specific to `OverflowMenu's` component props\** |
| (title)                      | title             | Made explicit\*                                  |
| open                  | —                 | Removed. Use `overflowMenuProps{open}` instead         |
| overflowMenuAriaLabel | —                 | Removed. Use `overflowMenuProps{['aria-label']}` instead         |
| selectorPrimaryFocus  | —                 | Removed. Use `overflowMenuProps{selectorPrimaryFocus}` instead         |

\* These props were "assumed" in the `v1` component and have been explicitly declared in `v2`.

\** The same props in the default **FilterPanelCheckbox** are duplicated in **FilterPanelCheckboxWithOverflow**. To avoid duplicate prop names between two or more internal components, only key/value pairs defined in `overflowMenuProps` will be assigned specifically to the internal `OverflowMenu` component.

 - [x] Storybook
 - [x] Tests


packages/ibm-products-styles/src/components/FilterPanel/_filter-panel-checkbox-with-overflow.scss
```diff
@@ -0,0 +1,65 @@
...
+@use '@carbon/styles/scss/theme' as *;
+@use '@carbon/styles/scss/type';
+
+// The block part of our conventional BEM class names (blockClass__E--M).
+$block-class: #{c4p-settings.$pkg-prefix}--filter-panel-checkbox-with-overflow;
+$carbon: #{c4p-settings.$carbon-prefix};
+
+// BUG FIX: hide small (unrelated?) rectangle that appears when showing
+// the OverflowMenu, but doesn't seem to belong or serve a function.
+html
```
matthewgallo, line 26: Even though this is a work around, I don't think we should be setting styles on all overflow menus. This will set styles on overflow menus anyone has on a page, even those outside of the `FilterPanel`.  Can we scope this to at least overflow menus inside of the `$block-class`?

packages/ibm-products-styles/src/components/FilterPanel/_filter-panel-checkbox-with-overflow.scss
```diff
@@ -0,0 +1,65 @@
...
+
+// Position the menu button via its wrapper, not the button itself.
+.#{$block-class} .#{$carbon}--overflow-menu__wrapper {
+  position: absolute;
+  top: 0;
+  right: 0;
+}
+// Size the menu button.
+.#{$block-class} .#{$block-class}__overflow-button {
+  width: to-rem(23px);
```
matthewgallo, line 50: Are the following values 23px to account for a 1px border somewhere?

packages/ibm-products-styles/src/components/FilterPanel/_filter-panel-checkbox-with-overflow.scss
```diff
@@ -0,0 +1,65 @@
...
+  height: to-rem(23px);
+  background-color: $field;
+  block-size: to-rem(23px);
```
matthewgallo, line 53: There appears to be a mix of logical properties here and non-logical. `height` and `block-size` for example. I don't think you need both. If you're going to sticky with logical properties, you could also replace `width` with `inline-size`.

packages/ibm-products/src/components/FilterPanel/FilterPanelCheckbox/FilterPanelCheckbox.test.js
```diff
@@ -44,6 +45,14 @@ describe(componentName, () => {
     expect(screen.getByRole('checkbox')).toBeInTheDocument();
   });
 
+  it('renders a count', async () => {
+    const filterPanelLabelClass = `${pkg.prefix}--filter-panel-label`;
+    const { container } = renderComponent({ count: count });
```
matthewgallo, line 50: Just semantics but just passing count feels a bit cleaner
```suggestion
    const { container } = renderComponent({ count });
```

packages/ibm-products/src/components/FilterPanel/FilterPanelCheckboxWithOverflow/FilterPanelCheckboxWithOverflow.js
```diff
@@ -0,0 +1,181 @@
...
+          className={cx(`${blockClass}__checkbox`)}
+          count={count}
+          id={id}
+          labelText={labelText}
+          title={title}
+        />
+
+        {showMenuButton && (
+          <OverflowMenu
+            open={overflowMenuProps?.open ? true : false}
```
matthewgallo, line 101: This should also work

```suggestion
            open={overflowMenuProps?.open}
```

packages/ibm-products/src/components/FilterPanel/FilterPanelCheckboxWithOverflow/FilterPanelCheckboxWithOverflow.test.js
```diff
@@ -0,0 +1,126 @@
...
+  render(
+    <FilterPanelCheckboxWithOverflow
+      data-testid={dataTestId}
+      id={id}
+      labelText={labelText}
+      {...{ ...rest }}
+    >
+      <OverflowMenuItem
+        itemText="Option 1"
+        // onClick={(event) => {
```
matthewgallo, line 38: Remove if no longer needed.

packages/ibm-products/src/components/FilterPanel/FilterPanelCheckboxWithOverflow/FilterPanelCheckboxWithOverflow.test.js
```diff
@@ -0,0 +1,126 @@
...
+    screen.getByTestId(dataTestId);
+  });
+
+  it('forwards a ref to an appropriate node', async () => {
+    const ref = React.createRef();
+    renderComponent({ ref: ref });
+    expect(ref.current).toBeInTheDocument();
+  });
+
+  // it('adds the Devtools attribute to the containing node', async () => {
```
matthewgallo, line 119: Are you able to add this test back in? We try to include this for every component.

matthewgallo (changes requested):
Just spotted a few things, also I believe that `FilterPanelCheckboxWithOverflow` also needs to be exported in `packages/ibm-products/src/components/index.js` alongside the other FilterPanel components.

packages/ibm-products-styles/src/components/FilterPanel/_filter-panel-checkbox-with-overflow.scss
paul-balchin-ibm: I was using the height of one of the other elements as a reference. Seems like changing it to 24px (`$spacing-06`) works fine.

packages/ibm-products-styles/src/components/FilterPanel/_filter-panel-checkbox-with-overflow.scss
paul-balchin-ibm: My bad. My career-long workaround jumped in without thinking :)

matthewgallo (approved):

Issue #4597, opened 2024-03-22
DataSpreadsheet refactor: Refactor handlers into separate utilities or hooks


amal-k-joy:

sstrubberg:
@amal-k-joy, @matthewgallo tells me you completed this work? Can you confirm?

amal-k-joy:
> @amal-k-joy, @matthewgallo tells me you completed this work? Can you confirm?

@sstrubberg , This is completed and the code is merged.

PR #4598, opened 2024-03-22
fix(Dataspreadsheet): refractor larger event handlers to seperate util

amal-k-joy:
Contributes to #4597 

Refactor handlers into separate utilities

#### What did you change?
Created a common util file (comonEventHandlers.js) and moved larger event handlers to that file and called from the respective file.
Verified locally


davidmenendez (approved):
wonderful

PR #4600, opened 2024-03-25
refactor(Toolbar): add Typescript types

amal-k-joy:
Contributes to #4274 

Add Typescript types to Toolbar

#### What did you change?
Change extension to `.tsx` **via git** 
Copy the PropTypes definitions above the component definition
Retool the PropTypes to be a TypeScript interface
Fixed the type errors
Tested locally using storybook


packages/ibm-products/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,149 @@
...
+// cspell:words Focusable focusable
+import { getFocusableElements as _getFocusableElements } from '../../global/js/utils/getFocusableElements';
+import { pkg } from '../../settings';
+
+const { checkComponentEnabled, prefix } = pkg;
+const blockClass = `${prefix}--toolbar`;
+
+interface ToolbarProps {
+  /** Provide the content of the `ToolbarGroup` */
+  children: ReactNode;
```
matthewgallo, line 31: Not necessary to include `children` since you've included the generic type `PropsWithChildren`

packages/ibm-products/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,149 @@
...
+  vertical?: boolean;
+}
+
+const ToolbarContext = createContext<ToolbarContextType>({});
+
+/** Toolbars are a collection of action items that organize a program’s interaction patterns into a series of closely related commands. */
+let Toolbar = forwardRef(
+  (
+    { children, className, vertical, ...rest }: PropsWithChildren<ToolbarProps>,
+    r: React.Ref<any>
```
matthewgallo, line 50:
```suggestion
    ref: React.Ref<HTMLDivElement>
```

packages/ibm-products/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,149 @@
...
+  ) => {
+    const focusableElements = useRef();
+
+    const getFocusableElements = useCallback(
+      (): any => focusableElements.current,
+      [focusableElements]
+    );
+
+    const _ref = useRef<any>();
+    const ref = r || _ref;
```
matthewgallo, line 60: In these cases where we use the forwarded ref or a local one we do something like this:
```jsx
const localRef = useRef<HTMLDivElement>();
const ref = ref || localRef;
```

I know this is more of a naming thing but would be good to keep a consistent approach across components.

packages/ibm-products/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,149 @@
...
+/** Toolbars are a collection of action items that organize a program’s interaction patterns into a series of closely related commands. */
+let Toolbar = forwardRef(
+  (
+    { children, className, vertical, ...rest }: PropsWithChildren<ToolbarProps>,
+    r: React.Ref<any>
+  ) => {
+    const focusableElements = useRef();
+
+    const getFocusableElements = useCallback(
+      (): any => focusableElements.current,
```
matthewgallo, line 55: Any alternatives to `any` here?

packages/ibm-products/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+    const _ref = useRef<any>();
+    const ref = r || _ref;
+
+    const [focus, setFocus] = useState();
+
+    useEffect(() => {
+      focusableElements.current = _getFocusableElements(
+        ref?.['current']
+      ) as any;
```
matthewgallo, line 67: Any alternatives to `any` here?

matthewgallo (commented):
Just have a few small questions. Also, if you used git to rename the file it will make the diff a bit easier to read (ie `git mv packages/ibm-products/src/components/ComponentName/ComponentName.js packages/ibm-products/src/components/ComponentName/ComponentName.tsx`).

packages/ibm-products/src/components/Toolbar/Toolbar.tsx
amal-k-joy: removed the children

packages/ibm-products/src/components/Toolbar/Toolbar.tsx
amal-k-joy: Renamed as you advised

packages/ibm-products/src/components/Toolbar/Toolbar.tsx
amal-k-joy: Changed any to HTMLElement[] as this is a util function defined in global and it returns array of elements.

packages/ibm-products/src/components/Toolbar/Toolbar.tsx
amal-k-joy: Changed any to HTMLElement[] as this is a util function defined in global and it returns array of elements.

packages/ibm-products/src/components/Toolbar/Toolbar.tsx
amal-k-joy: Made the changes as suggested

amal-k-joy:
> Just have a few small questions. Also, if you used git to rename the file it will make the diff a bit easier to read (ie `git mv packages/ibm-products/src/components/ComponentName/ComponentName.js packages/ibm-products/src/components/ComponentName/ComponentName.tsx`).

I used git to rename the file , but still it showed lot of diff. May be that is because I did yarn format after that.

matthewgallo (approved):

PR #4603, opened 2024-03-25
chore: convert optionstile to typescript

davidmenendez:
Contributes to #4281

### DIFF
--- a/packages/ibm-products/src/components/OptionsTile/OptionsTile.tsx
+++ b/packages/ibm-products/src/components/OptionsTile/OptionsTile.tsx
@@ -1,23 +1,18 @@
 /**
- * Copyright IBM Corp. 2021, 2022
+ * Copyright IBM Corp. 2021, 2024
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
  */
 
-// Import portions of React that are needed.
-import React, { useState, useRef } from 'react';
+import React, { useState, useRef, ReactNode } from 'react';
 
-// Other standard imports.
 import PropTypes from 'prop-types';
 import cx from 'classnames';
-
 import { getDevtoolsProps } from '../../global/js/utils/devtools';
 import uuidv4 from '../../global/js/utils/uuidv4';
 import { pkg } from '../../settings';
 import { useControllableState } from '../../global/js/hooks';
-
-// Carbon and package components we use.
 import { Layer, Toggle } from '@carbon/react';
 import {
   ChevronDown,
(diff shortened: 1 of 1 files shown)


packages/ibm-products/src/components/OptionsTile/OptionsTile.tsx
```diff
@@ -26,60 +21,132 @@ import {
   WarningFilled,
 } from '@carbon/react/icons';
 import * as carbonMotion from '@carbon/motion';
+import { CarbonIconType } from '@carbon/icons-react/lib/CarbonIcon';
```
matthewgallo, line 24: Cool, didn't know there was a `CarbonIconType`

packages/ibm-products/src/components/OptionsTile/OptionsTile.tsx
```diff
@@ -26,60 +21,132 @@ import {
...
+  /**
+   * Provide a text explaining why the OptionsTile is in locked state.
+   */
+  lockedText?: string;
+
+  /**
+   * Provide a function which will be called each time the user
+   * toggles the open state of the OptionsTile.
+   */
+  onChange?: (value: any) => void;
```
matthewgallo, line 71:
```suggestion
  onChange?: (value: boolean) => void;
```

This might be an option to avoid having to use `any` as it looks like this `onChange` only returns a boolean of whether the options tile is open or not.

packages/ibm-products/src/components/OptionsTile/OptionsTile.tsx
```diff
@@ -26,60 +21,132 @@ import {
...
+
+  /**
+   * Provide a function which will be called each time the user
+   * interacts with the toggle.
+   */
+  onToggle: (value: any) => void;
```
matthewgallo, line 77: Same suggested as above as the onToggle prop returns a boolean as well
```suggestion
  onToggle: (value: boolean) => void;
```

packages/ibm-products/src/components/OptionsTile/OptionsTile.tsx
```diff
@@ -26,60 +21,132 @@ import {
...
   ) => {
     const [prevIsOpen, setPrevIsOpen] = useState(open);
     const [closing, setClosing] = useState(false);
 
     const [isOpen, setIsOpen] = useControllableState({
-      value: open,
       defaultValue: open || null,
-      onChange: (value) => onChange(value),
+      name: 'OptionsTile',
+      onChange: (value: any) => onChange?.(value),
```
matthewgallo, line 144:
```suggestion
      onChange: (value: boolean) => onChange?.(value),
```

packages/ibm-products/src/components/OptionsTile/OptionsTile.tsx
```diff
@@ -26,60 +21,132 @@ import {
...
       invalid,
       invalidText,
       locked,
       lockedText,
-      onChange = defaults.onChange,
+      onChange,
       onToggle,
       open,
-      size = defaults.size,
+      size = 'lg',
```
matthewgallo, line 128: Are we able to revert back to the previous approach with default props with the `defaults` object? I also accidentally removed this in the `SidePanel` PR but you should be able to add it back without problems 🤞

matthewgallo (approved):

Issue #3766, opened 2023-11-06
Productive card - stacked body content enhancement


Laura-Marshall:
On Carbon for IBM Products, there is guidance on the stacked body content card, which isn't available in the code. The requirement is that there would be multiple clickable sections within the tile. 

[Link to design ](https://pages.github.ibm.com/cdai-design/pal/components/card/productive/usage#action-zones)

![image](https://github.com/carbon-design-system/ibm-products/assets/53219208/f59d85ba-393c-450b-9a64-fcdcac583274)

I am aware that teams in the past have edited this component to create this behaviour on homepages etc, for clickable content. Personally in my product, we have quite imminent need for this. As its included in the guidance, it feels right that the component needs this across v10 and v11 if possible!

![image](https://github.com/carbon-design-system/ibm-products/assets/53219208/88c0f778-381e-4021-9d97-5c965261126b)

```[tasklist]
### Tasks
- [ ] Check if we can already do this with the current version of productive cards
- [ ] Add new story for stacked productive cards
- [ ] If needed, evaluate how to incorporate the stacked features in a composable way
```

elycheea:
Moving to Icebox — this would probably make the most sense to explore after #5258

amal-k-joy:
This should be possible with the [new composable card](https://github.com/carbon-design-system/ibm-products/issues/9308) which is under development. In the new version the body section is more like an open slot where adopter can render any content.
 Do you think you need an example implementation?

amal-k-joy:
Closing this as this automatically handles with new version. Please feel free to re-open if you think this is relevant.

PR #4617, opened 2024-03-27
feat(Card): productive card stacked body tiles

Ratheeshrajan:
Contributes to #3766 

The requirement is that there would be multiple clickable sections within the tile.

#### What did you change?

- Added a new story for `Stacked` type productive card
- Updated Card component to accept stacked-type body component

Storybook


packages/ibm-products/src/components/Card/Card.js
```diff
@@ -193,7 +194,7 @@ export let Card = forwardRef(
         className: cx(`${blockClass}__header-body-container`, {
           [`${blockClass}__clickable`]: clickable,
         }),
-        ...(clickable && clickableProps),
+        ...(clickable || (stackedBody && clickableProps)),
```
davidmenendez, line 197: so in this notation it's spreading `clickableProps` if `clickable` is `true`. how you actually want to write this is if either `clickable` or `stackedBody` is `true`. so it should look like
```
...((clickable || stackedBody) && clickableProps)
```
this notation can look confusing if you're unfamiliar with it. i hope this helps clarify!

packages/ibm-products/src/components/Card/Card.js
```diff
@@ -268,7 +271,11 @@ export let Card = forwardRef(
         <div className={`${blockClass}__content-container`}>
           <div {...getHeaderBodyProps()}>
             <CardHeader {...getHeaderProps()} />
-            <div {...getBodyProps()}>{children}</div>
+            {stackedBody ? (
```
davidmenendez, line 274: it doesn't look like this if statement is doing anything different in the else

packages/ibm-products/src/components/ProductiveCard/ProductiveCard.stories.js
```diff
@@ -205,6 +208,93 @@ Clickable.args = {
...
+    var ampm = hours >= 12 ? 'PM' : 'AM';
+    hours = hours % 12;
+    hours = hours ? hours : 12;
+    minutes = minutes < 10 ? '0' + minutes : minutes;
+    var strTime = hours + ':' + minutes + ' ' + ampm;
+    return strTime;
+  }
+  return (
+    <>
+      {stackedCardTiles.map((tile, idx) => (
```
davidmenendez, line 250: i think we need to change this implementation slightly.

since this is a feature the component is offering we shouldn't be asking the user to have to contribute any of the `blockClass` elements. what we should do here is instead of passing this content to `children` we should have a prop that accepts an array of items that we can iterate on and pass the class name to directly.

so instead of passing this to `children` lets add a new prop like `stackedItems`. and then something like `stackedItems={[Item, Item2, Item])` and then the user would pass whatever content and structure they want.

davidmenendez (changes requested):
great start! i think we can make some changes to the implementation here to make this more composable and user friendly.

the only other thing i want to mention is that in the storybook the `UserAvatar` isn't really visible. we should fix that to look like more like design with more contrast.

Screenshot 2024-04-02 at 9 51 50 AM

davidmenendez:
@Ratheeshrajan any update here?

elycheea:
Closing this for now and will move this issue back to backlog/triage for now. These files also have significant changes since we’ve migrated to Typescript and introduced additional card variants.

PR #4633, opened 2024-03-28
refactor(TearsheetNarrow): added typescript types

devadula-nandan:
Contributes to #4502 

migrated the component to typescript.
I have used ` as const` in defaults values to fix the ts error when referencing defaults object values as props

#### What did you change?
```
packages/ibm-products/src/components/Tearsheet/TearsheetNarrow.tsx
```

local dev env by passing different types of props to the component, and verifying ts errors
storybook renders as expected


packages/ibm-products/src/components/Tearsheet/TearsheetNarrow.tsx
```diff
@@ -29,13 +29,107 @@ import {
...
+   */
+  className?: string;
+
+  /**
+   * The accessibility title for the close icon (if shown).
+   *
+   * **Note:** This prop is only required if a close icon is shown, i.e. if
+   * there are a no navigation actions and/or hasCloseIcon is true.
+   */
+  closeIconDescription?: closeIconDescriptionTypes;
```
matthewgallo, line 70: I like this approach for conditionally required props 💯

matthewgallo (approved):

PR #4677, opened 2024-04-02
refactor: use npx over yarn for commitlint

matthewgallo:
I have repeatedly found that the pre-commit hook fails when it hits the `commitlint` step. See the following screenshot:
Screenshot 2024-04-02 at 4 04 22 PM

Replacing `yarn commitlint` with `npx commitlint` fixes this issue for me locally. Not sure if others have hit this as well, but hoping this will prevent any future errors with commitlint and husky.

#### What did you change?
Updated `.husky/commit-msg`
Tested this on a separate branch with these changes and the pre-commit runs as expected

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn commitlint --edit $1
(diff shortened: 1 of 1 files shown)


matthewgallo:
🙃 now this seems to be an intermittent occurrence, I'll close for now

PR #4678, opened 2024-04-03
refactor(EditInPlace): add typescript types

anamikaanu96:
Contributes to #4498 

converted packages/ibm-products/src/components/EditInPlace/EditInPlace.js to packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx

#### What did you change? 
packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx

### DIFF
--- a/packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx
+++ b/packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx
@@ -5,7 +5,13 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, { useState, useEffect, forwardRef, useRef } from 'react';
+import React, {
+  useState,
+  useEffect,
+  forwardRef,
+  useRef,
+  PropsWithChildren,
+} from 'react';
 import { IconButton } from '@carbon/react';
 import cx from 'classnames';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx
```diff
@@ -27,7 +33,101 @@ const defaults = {
...
+   * Can be passed either as one of tooltip options or as an object specifying cancel, edit and save separately
+   */
+  tooltipAlignment?: AlignPropType | Shape;
+  /**
+   * current value of the input
+   */
+  value: string;
+}
+
+export let EditInPlace = forwardRef<any, EditInplaceProps>(
```
matthewgallo, line 130: Can you replace `any` with `HTMLDivElement`?

packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx
```diff
@@ -27,7 +33,101 @@ const defaults = {
...
+  /**
+   * label for cancel button
+   */
+  cancelLabel: string;
+  /**
+   * By default the edit icon is shown on hover only.
+   */
+  editAlwaysVisible?: boolean;
+  /**
+   /**
```
matthewgallo, line 62: You have an extra `/**` on this line and the indentation is tad bit off. 🤔 I would have expected the precommit to have caught the indentation but running `yarn format` should fix it.

anamikaanu96: pre-commit hook is running for me but not sure why its not catching this.  Also I noticed that when I do yarn format, its not formatting the tsx file. So I am doing format document.

Resolved 👍

matthewgallo (approved):

packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx
matthewgallo: Ahh good catch and I see @szinta opened a PR to fix this as well 🥳

PR #4746, opened 2024-04-04
Build: pre-commit hooks not working in local env

szinta:
Contributes to #4043 

Currently, Husky is configured using a 'prepare' script.  'Prepare' works for npm and older versions of Yarn, but Yarn 2+ doesn't support it. So I have added a postinstall script for configuring Husky.

#### What did you change?

1. Added `postinstall` script in package.json to fix pre-commit hooks issue
2. Updated `format` script to include tsx files to fix the issue metioned in https://github.com/carbon-design-system/ibm-products/pull/4678#discussion_r1551158510

Local env

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "lint:js": "eslint packages --fix",
     "lint:styles": "stylelint '**/*.scss'",
     "prepare": "husky install",
+    "postinstall": "husky",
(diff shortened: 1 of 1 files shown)


davidmenendez (approved):
THANK YOU FOR SOLVING THIS!

PR #4751, opened 2024-04-04
refactor(Tearsheet): convert tearsheet shell to typescript

matthewgallo:
Contributes to #4687 

Converts the `TearsheetShell` to use TypeScript as well as `Wrap` which is used within the `TearsheetShell`.
There is still one TypeScript error but is dependent on https://github.com/carbon-design-system/ibm-products/pull/4676 before it is possible to fix.

#### What did you change?
```
packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
packages/ibm-products/src/global/js/utils/Wrap.tsx
```
Locally within vscode


packages/ibm-products/src/components/Tearsheet/TearsheetNarrow.tsx
```diff
@@ -67,7 +63,7 @@ interface TearsheetNarrowProps extends PropsWithChildren {
    * **Note:** This prop is only required if a close icon is shown, i.e. if
    * there are a no navigation actions and/or hasCloseIcon is true.
    */
-  closeIconDescription?: closeIconDescriptionTypes;
+  closeIconDescription?: string;
```
elycheea, line 66: Do we still need this if we include `CloseIconDescriptionTypes` in like 122-124?

In that article @devadula-nandan recently shared[^1], it seemed like we might not need to?

[^1]: https://dev.to/maissenayed/conditional-react-props-with-typescript-43lg

elycheea (approved):
🎉

PR #4801, opened 2024-04-10
feat(ConditionBuilder): First PR with generated files and basic configs, components

amal-k-joy:
Contributes to #4577 

This is the very first PR to implement condition builder. PR has only basic changes and generated files.

#### What did you change?
Scaffold new component files
List new components with `canary` flags under the Components(?) section of Storybook
Ensure new component(s) support `className` prop
Added translation of text support
Localhost


packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.js
```diff
@@ -0,0 +1,142 @@
...
+ */
+
+export let ConditionBuilder = React.forwardRef(
+  (
+    {
+      // The component props, in alphabetical order (for consistency).
+
+      // children /* TODO: remove if not needed. */,
+      className,
+      inputConfig = inputData,
```
davidmenendez, line 58: the sample data should just be passed in through storybook. no need to import it here.

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.js
```diff
@@ -0,0 +1,83 @@
...
+const blockClass = `${pkg.prefix}--condition-builder`;
+
+function ConditionBuilderContent() {
+  const { rootState, setRootState } = useContext(ConditionBuilderContext);
+  const [isConditionbuilderActive, setIsConditionbuilderActive] =
+    useState(true);
+  const conditionBuilderRef = useRef();
+
+  useEffect(() => {
+    if (!rootState?.groups || rootState?.groups.length == 0) {
```
davidmenendez, line 19: i think you can simplify this to `rootState?.groups?.length === 0`

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.js
```diff
@@ -0,0 +1,83 @@
...
+  return (
+    <div ref={conditionBuilderRef}>
+      {isConditionbuilderActive && (
+        <Button
+          className={`${blockClass}__condition-builder`}
+          renderIcon={(props) => <Add size={16} {...props} />}
+          iconDescription={'Add Condition'}
+          kind="ghost"
+          size="sm"
+          onClick={() => onStartConditionbuilder()}
```
davidmenendez, line 51: no need to pass it is as an anonymous function `onClick={onStartConditionbuilder}`

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContext/DataTreeContext.js
```diff
@@ -0,0 +1,186 @@
+import React, { createContext, useState } from 'react';
+import PropTypes from 'prop-types';
+//sentene variant
+export const sampleDataStructure_sentence = {
```
davidmenendez, line 4: anything that is considered sample or for the purposes of the storybook demo should be moved into a separate file to ensure a separation of user and product code. a folder called `assets` or something similar would be a great place to put these.

packages/ibm-products/src/components/ConditionBuilder/sampleInput.tsx
```diff
@@ -0,0 +1,241 @@
+import {
```
davidmenendez, line 1: same comment as above about moving this into a separate folder that contains sample / storybook assets.

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.js
```diff
@@ -0,0 +1,83 @@
...
+
+  const onStartConditionbuilder = () => {
+    //when add conditon button is clicked.
+    setIsConditionbuilderActive(false);
+    setRootState(emptyState); //here we can set an empty skeleton object for an empty condition builder,
+    //or we can even prepopulate some existing builder and continue editing
+  };
+
+  return (
+    <div ref={conditionBuilderRef}>
```
davidmenendez, line 43: this div should have a block classname associated with it

davidmenendez (changes requested):
great start!

aside from a few comments and cspell errors this is looking good.

one of the main things that concerns me right now is that everything is hard coded. we should have no hard coded strings in our components. all of that information should be coming directly from the user. the `add condition` button is a good example of something easy and direct that can be set up as a prop. same with the `iconDescription` prop.

if `dataConfigs.js` is supposed to be for demo purposes for the storybook then that's totally fine and should be moved to the demo assets folder that i mentioned in another comment. i just want to point out that the sooner all those strings are decoupled from the actual implementation the easier it'll be as development continues.

amal-k-joy:
> great start!
> 
> aside from a few comments and cspell errors this is looking good.
> 
> one of the main things that concerns me right now is that everything is hard coded. we should have no hard coded strings in our components. all of that information should be coming directly from the user. the `add condition` button is a good example of something easy and direct that can be set up as a prop. same with the `iconDescription` prop.
> 
> if `dataConfigs.js` is supposed to be for demo purposes for the storybook then that's totally fine and should be moved to the demo assets folder that i mentioned in another comment. i just want to point out that the sooner all those strings are decoupled from the actual implementation the easier it'll be as development continues.

I have created and added a text translation logic that will replace any hardcoded texts. Later we need internationalisation as well. Dataconfig is not for demo purpose , that is where all the configurations like list of operators based on the property type resides(which will make sense in subsequent PRs). I have added translation support there as well.

All other sample datas I have moved to asset folder and passed via storybook as you advised.

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.js
amal-k-joy: modified as advised

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.js
amal-k-joy: modified as advised

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.js
amal-k-joy: modified as advised

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContext/DataTreeContext.js
amal-k-joy: Moved to assets folder. Sample input is the inputConfig expected as input prop. SampleData is the samples of data tree structure we are building . All updates like add condition , modify condition , remove condition will update the rootState object and renders the component  accordingly

packages/ibm-products/src/components/ConditionBuilder/sampleInput.tsx
amal-k-joy: modified as advised

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.js
amal-k-joy: modified as advised

davidmenendez (approved):
two last comments:

1. translation / strings issue has been brought up and there's a lot of discussion surrounding the best way to handle it. personally i think it should just be a prop that accepts an object of strings similarly to how you have translations.

```
<ConditionBuilder strings={{ if: 'if', else: 'else', ... }} />
```

nothing has to be done now because the team still wants to discuss the best way to implement this

2. minor thing but, by convention, anything that isn't a component shouldn't have an uppercase file name. so like `Translations.js` should just be `translations.js`

i'll go ahead and approve this for now so you don't get held up with development. please address number 2 in a follow up and we'll continue to refine how we want the strings / translations to be handled

Issue #4750, opened 2024-04-04
Datagrid loading bars 3 to 1


csamarchi:
# Name of component or pattern
Datagrid

## Brief description
Would like to alter the loading bars inside the Datagrid table to be 1 instead of default 3.

### Links to other materials
Currently:
![Screenshot 2024-04-04 at 12 59 24 PM](https://github.com/carbon-design-system/ibm-products/assets/42501494/80158484-792f-4b08-b4a9-a97898b9d56c)

Goal:
![Screenshot 2024-04-04 at 11 00 57 AM](https://github.com/carbon-design-system/ibm-products/assets/42501494/edbb352c-b33c-49bd-b96b-0f648cf3e034)

PR #4802, opened 2024-04-10
feat(datagrid): added feature to control number of skeleton rows

devadula-nandan:
Contributes to #4750 

isFetching prop now accepts a number and a boolean.
if its boolean and true, it shows 3 row skeletons by default.
if it is a number, it shows the number of row skeletons.

#### What did you change?
packages/ibm-products/src/components/Datagrid/useSkeletonRows.js

in Datagrid.stories.js at line 615 in `Skeleton` story, i have set the isFetching prop to `true`, `false`, `0`, `a random number`.
And got empty states for `false` and `0`
and `3` skeletons for passing true, and expected number of skeletons for the provided number

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridContent.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridContent.js
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corp. 2022, 2023
+ * Copyright IBM Corp. 2022, 2024
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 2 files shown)


devadula-nandan:
i need a second opinion on this 🤔🤔

devadula-nandan:
docs not updated yet.

elycheea (commented):
As discussed on previous team huddle, we’ll update this to use two separate props for the rows and `isFetching` state.

devadula-nandan:
Thanks for the review @elycheea the updates are made.

packages/ibm-products/src/components/Datagrid/Datagrid/DatagridContent.js
```diff
@@ -252,6 +252,7 @@ DatagridContent.propTypes = {
       PropTypes.func,
     ]),
     isFetching: PropTypes.bool,
+    skeletonRows: PropTypes.number,
```
elycheea, line 255: This is good, but think more consistent naming with Carbon might be better if we’re introducing a new prop. The DataTable uses `rowCount`, but think we should still specify skeleton in this case.

```suggestion
    skeletonRowCount: PropTypes.number,
```

elycheea (approved):

Issue #4803, opened 2024-04-10
Explore github action for moving issue to project and applying a specific field option


matthewgallo:
## What will this achieve?

When we give an issue a particular label, we want it to automatically be added to our project and also be assigned a particular field option, for example a label of `TypeScript` should be added to the project and put in the `Area` field of `TypeScript`.
<img src="https://github.com/carbon-design-system/ibm-products/assets/10215203/77ef38de-634b-40ff-98cc-76549ff391df" width="400" alt="screenshot showing how this github action could work" />

## How will success be measured?

When "TypeScript" label is applied to an issue, that issue gets added to the project and assigned to the "TypeScript" project area field.

## Additional information

- Designs
- Existing code
- etc

```[tasklist]
### Tasks
- [ ] https://github.com/carbon-design-system/ibm-products/issues/4876
```

matthewgallo:
Prototype of action [here](https://github.com/matthewgallo/ibm-products-vite-template/blob/main/.github/workflows/add-label-move-to-area.yml). I'll look at porting this over to our repo this sprint.

PR #4872, opened 2024-04-11
feat: create action for triaging labeled issues

matthewgallo:
Contributes to #4803 

Adds the prototype action I created [here](https://github.com/matthewgallo/ibm-products-vite-template/blob/main/.github/workflows/add-label-move-to-area.yml) into our repo. This action triages labeled issues by adding them to our project board and applying the specified field/field option values. In this case, issues labeled `area: typescript` will be moved into "Area: 🟦 TypeScript" in our project.

Discussed with @tay1orjones about potentially including the reusable workflow in a [shared repo](https://github.com/carbon-design-system/.github) for other teams to use as well, but want to make sure it works here first.

If we want to add this same thing for other labels we should be able to easily by updating `.github/workflows/triage-labeled-issues.yml`. The [Graphql API Explorer](https://docs.github.com/en/graphql/overview/explorer) really helped with being able to see the response and to make sure that the `query` was properly structured.

#### What did you change?
Created two new workflow files
```
.github/workflows/move-issue-to-project-and-set-fields.yml // more generic reusable workflow file
.github/workflows/triage-labeled-issues.yml
```
[In test repo](https://github.com/matthewgallo/ibm-products-vite-template/tree/main/.github/workflows)

### DIFF
--- a/.github/workflows/triage-labeled-issues.yml
+++ b/.github/workflows/triage-labeled-issues.yml
@@ -21,3 +21,11 @@ jobs:
       field_option: '🗺 Migration'
       project_number: '65'
     secrets: inherit
+  move-a11y-issues:
+    if: ${{contains(github.event.label.name, format('type{0} ally ♿', ':'))}}
+    uses: ./.github/workflows/move-issue-to-project-and-set-fields.yml
+    with:
+      field: 'Area'
+      field_option: '♿️ Accessibility'
+      project_number: '65'
+    secrets: inherit
--- a/.github/workflows/triage-labeled-issues.yml
+++ b/.github/workflows/triage-labeled-issues.yml
@@ -22,7 +22,7 @@ jobs:
       project_number: '65'
     secrets: inherit
   move-a11y-issues:
-    if: ${{contains(github.event.label.name, format('type{0} ally ♿', ':'))}}
+    if: ${{contains(github.event.label.name, format('type{0} a11y ♿', ':'))}}
     uses: ./.github/workflows/move-issue-to-project-and-set-fields.yml
     with:
       field: 'Area'
--- a/.github/workflows/triage-labeled-issues.yml
+++ b/.github/workflows/triage-labeled-issues.yml
@@ -1,0 +1,23 @@
+name: Triages labeled issues by adding them to a specific project field
+
+on:
+  issues:
+    types: [ labeled ]
+
+jobs:
+  move-ts-issues:
+    if: ${{contains(github.event.label.name, format('area{0} typescript', ':'))}}
+    uses: ./.github/workflows/move-issue-to-project-and-set-fields.yml
+    with:
+      field: 'Area'
+      field_option: '🟦 Typescript'
+      project_number: '65'
+    secrets: inherit
+  move-migration-issues:
+    if: ${{contains(github.event.label.name, format('area{0} migration ➡️', ':'))}}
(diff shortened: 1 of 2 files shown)


.github/workflows/triage-labeled-issues.yml
```diff
@@ -0,0 +1,23 @@
...
+  move-ts-issues:
+    if: ${{contains(github.event.label.name, format('area{0} typescript', ':'))}}
+    uses: ./.github/workflows/move-issue-to-project-and-set-fields.yml
+    with:
+      field: 'Area'
+      field_option: '🟦 Typescript'
+      project_number: '65'
+    secrets: inherit
+  move-migration-issues:
+    if: ${{contains(github.event.label.name, format('area{0} migration ➡️', ':'))}}
```
matthewgallo, line 17: Not sure how that arrow is going to work but that's how the emoji pasted when I got it from the label 🤔

elycheea: Fwiw, it did show up correctly in my email. So _probably_ fine? 😁

matthewgallo: Ah interesting, I've found that some emojis render differently in vscode vs github so maybe there is some kind of mapping that happens. Learning a lot about emojis with this PR 😂

.github/workflows/triage-labeled-issues.yml
```diff
@@ -0,0 +1,31 @@
...
+    uses: ./.github/workflows/move-issue-to-project-and-set-fields.yml
+    with:
+      field: 'Area'
+      field_option: '🗺 Migration'
+      project_number: '65'
+    secrets: inherit
+  move-a11y-issues:
+    if: ${{contains(github.event.label.name, format('type{0} ally ♿', ':'))}}
```
elycheea, line 25: Should this one actually be a11y and not ally?

```suggestion
    if: ${{contains(github.event.label.name, format('type{0} a11y ♿', ':'))}}
```

.github/workflows/triage-labeled-issues.yml
```diff
@@ -0,0 +1,31 @@
...
+      field_option: '🟦 Typescript'
+      project_number: '65'
+    secrets: inherit
+  move-migration-issues:
+    if: ${{contains(github.event.label.name, format('area{0} migration ➡️', ':'))}}
+    uses: ./.github/workflows/move-issue-to-project-and-set-fields.yml
+    with:
+      field: 'Area'
+      field_option: '🗺 Migration'
+      project_number: '65'
```
elycheea, line 22: Not sure if it’s worth storing this as an environment variable[^1]. I think adding more labels to our workflows will be at least somewhat manual so maybe lower value.

[^1]: https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow

.github/workflows/triage-labeled-issues.yml
matthewgallo: Yes 👍

elycheea (approved):

.github/workflows/triage-labeled-issues.yml
matthewgallo: I did try, but got an error about using environment variables for [workflow_call inputs](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callinputs).
image

Issue #4800, opened 2024-04-10
PageHeader breadcrumb tooltip displays underneath Carbon UI shell header


modean:
## What package(s) are you using?

```
"@carbon/colors": "11.21.0",
"@carbon/ibm-products": "2.34.0",
"@carbon/layout": "11.21.0",
"@carbon/react": "1.54.0",
"@carbon/styles": "1.54.0",
"@carbon/themes": "11.34.0",
"@carbon/type": "11.26.0",
"react": "18.2.0",
"react-dom": "18.2.0",
"react-router-dom": "6.22.3"
```

## Detailed description

> Describe in detail the issue you're having.

The `PagerHeader` component provides a breadcrumb trail that is supposed to be using an ellipsis for situation where the browser real estate is narrow and thus the breadcrumb can't be displayed.

In such situation the ellipsis is accompanied by a tooltip. The tooltip unfortunately opens using the position `top` and thus it opens underneath the Carbon UI shell `<header />`. This contradicts the added value of the tooltip.

> Is this issue related to a specific component?

Yes it is related to the `PageHeader` component combined with the Carbon UI shell and there `HeaderContainer`.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

I did expect the tooltip to be visually rendered in a way where it can be read. For instance under the ellipsis and not atop of the ellipsis.

> What browser are you working in?

Chrome 123.0.6312.105 (Official Build) (arm64)
Firefox 124.0.2 (64-bit)

> What version of the @carbon/ibm-products (or @carbon/ibm-cloud-cognitive)
> package are you using?

"@carbon/ibm-products": "2.34.0"

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

n/a

## Steps to reproduce the issue

1. Create a Carbon UI shell sample that makes use of the `HeaderContainer` component.
2. Create a `PageHeader` component that has a breadcrumb
3. Resize the browser window, so that it renders certain breadcrumb trail elements as an ellipsis.
4. Hover over the ellipsis to trigger the tooltip to be displayed.

## Additional information

https://github.com/carbon-design-system/ibm-products/assets/659775/b2bacd7a-c87b-4162-8a15-d79044f4fe4e

modean:
Note that the Carbon `<header />` has a `z-index` of `8000` where the `<SidePanel />` has a `z-index` of `9000`.

makafsal:
[...]

PR #4907, opened 2024-04-16
fix(pageHeader): PageHeader breadcrumb tooltip displays underneath Carbon UI shell header

makafsal:
Contributes to #4800 

Use the new `overflowTooltipAlign` to align tooltip

#### What did you change?
Added a new property `overflowTooltipAlign` to the component `BreadcrumbWithOverflow`.

Now we can align the tooltip from `PageHeader` component like this:
image

Storybook

### DIFF
--- a/packages/ibm-products/src/components/PageHeader/PageHeader.js
+++ b/packages/ibm-products/src/components/PageHeader/PageHeader.js
@@ -521,6 +521,7 @@ export let PageHeader = React.forwardRef(
                               ? [breadcrumbItemForTitle]
                               : breadcrumbs // breadcrumbs may be null or undefined
                           }
+                          overflowTooltipAlign="right"
                         />
                       ) : null}
                     </Column>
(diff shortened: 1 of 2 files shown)


packages/ibm-products/src/components/PageHeader/PageHeader.js
```diff
@@ -528,6 +528,7 @@ export let PageHeader = React.forwardRef(
                               ? [breadcrumbItemForTitle]
                               : breadcrumbs // breadcrumbs may be null or undefined
                           }
+                          overflowTooltipAlign="right"
```
elycheea, line 531: Probably better not to hardcode this. While currently this may be the only one that fully displays at this moment, there may be cases in the future where this is addressed and teams are able to set a different position.

makafsal: `overflowTooltipAlign={breadcrumbOverflowTooltipAlign}`

packages/ibm-products/src/components/PageHeader/PageHeader.js
```diff
@@ -972,12 +978,13 @@ PageHeader.propTypes = {
...
    * The PageHeader is hosted in a Carbon grid, this value is passed through to the Carbon grid narrow prop
    */
-  narrowGrid: PropTypes.bool,
+  narrowGrid: PropTypes.bool
   /**
    * Content for the navigation area in the PageHeader. Should
    * be a React element that is normally a Carbon Tabs component. Optional.
-   */
-  navigation: PropTypes.element, // Supports Tabs
+   */,
```
elycheea, line 985: This little comma snuck out of line 981. 😅

elycheea (approved):

PR #4913, opened 2024-04-16
fix: Use tableHeight prop from react table as a fallback

tom-youd:
Contributes to # (no issue yet, just spotted trying to use the infinite scroll hook)

There is a property from the react-table state called `tableHeight` which I set in my useDatagridState but gets overwritten by the useParentDimensions hook because tableHeight is also defined here but can be `undefined` 

#### What did you change?

Simple one line change to fallback to the react-table prop if tableHeight in the useParentDimensions hook is undefined

Only tested locally so far and works as expected for me


tom-youd:

tom-youd:
recheck

davidmenendez (approved):

Issue #4758, opened 2024-04-05
Remove TagOverflow dependency on TagSet


elycheea:
Based on earlier discussions when first @szinta share out the `TagOverflow` prototype, the current implementation uses `TagSet` as a dependency. We’d like to refactor this so we can eventually deprecate `TagSet` in favor of a more flexible `TagOverflow`.

PR #4929, opened 2024-04-17
feat(tagoverflow): remove TagOverflow dependency on TagSet

szinta:
Contributes to #4758

In TagOverflow component, we were using Tagset to display the details of overflow items. This PR removes this Tagset component dependency from TagOverflow.

#### What did you change?

1. Removed Tagset component from Tagoverflow
2. Copied the Overflow popover and Overflow modal components to TagOverflow
3. Added the styles of Overflow popover and modal in TagOverflow.scss

Storybook


packages/ibm-products/src/components/TagOverflow/TagOverflowModal.js
```diff
@@ -0,0 +1,134 @@
...
+          className={`${blockClass}__search`}
+          labelText={searchLabel}
+          placeholder={searchPlaceholder}
+          onChange={handleSearch}
+          size="lg"
+        />
+      </ModalHeader>
+      <ModalBody className={`${blockClass}__body`} hasForm>
+        {filteredModalTags.map(({ label, ...other }, index) => (
+          <Tag {...other} filter={false} key={`all-tags-${index}`}>
```
davidmenendez, line 108: as a convention it's recommended to not use an index like this as a key

https://react.dev/learn/rendering-lists#why-does-react-need-keys

the tags already have an `id` attribute so i would just continue to use that

packages/ibm-products/src/components/TagOverflow/TagOverflowPopover.js
```diff
@@ -0,0 +1,198 @@
...
+                    }
+                    return (
+                      <li
+                        className={cx(`${blockClass}__tag-item`, {
+                          [`${blockClass}__tag-item--default`]:
+                            overflowType === 'default',
+                          [`${blockClass}__tag-item--tag`]:
+                            overflowType === 'tag',
+                        })}
+                        key={index}
```
davidmenendez, line 120: same comment regarding index for key

packages/ibm-products/src/components/TagOverflow/TagOverflowModal.js
```diff
@@ -0,0 +1,134 @@
...
+  searchPlaceholder,
+
+  // Collect any other property values passed in.
+  ...rest
+}) => {
+  const [filteredModalTags, setFilteredModalTags] = useState([]);
+  const [search, setSearch] = useState('');
+  const renderPortalUse = usePortalTarget(portalTargetIn);
+
+  useEffect(() => {
```
davidmenendez, line 52: the code here looks good, but you don't actually need to use an effect here. since the search isn't coming from another source you can just convert this to be a getter function. something like

```
const getFilteredItems = () => {
...
```

and then you could either call `getFilteredItems().map` in the return or set a variable `const filteredItems = getFilteredItems()`

packages/ibm-products/src/components/TagOverflow/TagOverflowModal.js
```diff
@@ -0,0 +1,134 @@
...
+  };
+
+  return renderPortalUse(
+    <ComposedModal
+      {
+        // Pass through any other property values as HTML attributes.
+        ...rest
+      }
+      containerClassName={`${blockClass}__container`}
+      className={cx(className, `${blockClass}`)}
```
davidmenendez, line 88: template strings are necessary here

packages/ibm-products/src/components/TagOverflow/TagOverflowPopover.js
```diff
@@ -0,0 +1,198 @@
...
+    const localRef = useRef();
+    const overflowTagContent = useRef(null);
+
+    useClickOutside(ref || localRef, () => {
+      if (popoverOpen) {
+        setPopoverOpen(false);
+      }
+    });
+
+    const handleShowAllTagsClick = (ev) => {
```
davidmenendez, line 55: i think as a convention we tend to use `evt` for event variable names. i would update references to events to just be `evt`

packages/ibm-products/src/components/TagOverflow/TagOverflowPopover.js
```diff
@@ -0,0 +1,198 @@
...
+          <Tag
+            onClick={() => setPopoverOpen(!popoverOpen)}
+            className={cx(`${blockClass}__trigger`)}
+          >
+            +{overflowTags.length}
+          </Tag>
+          <PopoverContent>
+            <div ref={overflowTagContent} className={`${blockClass}__content`}>
+              <ul className={`${blockClass}__tag-list`}>
+                {overflowTags
```
davidmenendez, line 98: to improve readability i think we should try to extract logic from the render as much as possible. so i would move this out of the render into it's own filtered variable before mapping

packages/ibm-products/src/components/TagOverflow/TagOverflow.stories.js
```diff
@@ -218,13 +225,15 @@ export const ManyTags = Template.bind({});
 ManyTags.args = {
   containerWidth: 500,
   items: tags,
+  ...overflowAndModalStrings,
```
davidmenendez, line 228: can we update the name here to highlight that this uses the modal? it took me longer than i'd like to admit to figure out how to get the modal to display, so including it in the name here would be helpful. maybe like `TagsWithOverflowModal` or something

davidmenendez (changes requested):
great start! just have a few comments

packages/ibm-products/src/components/TagOverflow/TagOverflowModal.js
szinta: Replaced index with id attribute.

packages/ibm-products/src/components/TagOverflow/TagOverflowPopover.js
szinta: Replaced index with id attribute

packages/ibm-products/src/components/TagOverflow/TagOverflow.stories.js
szinta: Updated story name as you suggested. I have updated other story names as well to make it more meaningful.

packages/ibm-products/src/components/TagOverflow/TagOverflowPopover.js
szinta: Done

packages/ibm-products/src/components/TagOverflow/TagOverflowModal.js
szinta: Done

packages/ibm-products/src/components/TagOverflow/TagOverflowModal.js
szinta: Removed

packages/ibm-products/src/components/TagOverflow/TagOverflowPopover.js
szinta: Done

davidmenendez (approved):

amal-k-joy (approved):

PR #4933, opened 2024-04-17
feat(Datagrid): passes state to onApply

AlexanderMelox:
Passes important state information to onApply so developer using it has access to the changes 

#### What did you change?
`packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/FilterPanel.js`


packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/FilterPanel.js
```diff
@@ -117,7 +117,11 @@ const FilterPanel = ({
   const apply = () => {
     setAllFilters(filtersObjectArray);
     // From the user
-    onApply();
+    onApply({
```
matthewgallo, line 120: Since you're making some updates here with `onApply`, can you add some info about it in the docs?

matthewgallo (changes requested):

packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/FilterPanel.js
AlexanderMelox: Added ✔️

matthewgallo (approved):

Issue #4966, opened 2024-04-22
window not defined in remix apps


davidmenendez:
### Package

Carbon for IBM Products

### Description

window not defined in remix apps is causing errors

### Component(s) impacted

notifications panel

### Browser

Chrome

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

current

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Product/offering

c4p

### CodeSandbox or Stackblitz example

NA

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

PR #4968, opened 2024-04-22
fix: check for document for matchmedia

davidmenendez:
Contributes to #4966 

* adds some utility functions to check for `window` and `document`
* adds a hook to check users reduced motion preferences 
* updated `NotificationsPanel` to use the new `usePrefersReducedMotion` hook to avoid SSR issues


davidmenendez:
for some reason using `usePrefersReducedMotion` breaks in `SidePanel`. additional time should be spent here to take a look at the code.

matthewgallo (changes requested):
Can the options in the callback passed to `mockImplementation` be stored in a separate file and imported since it looks like the same options are used across multiple test files? Really like that we have a shared hook for this now 🎉

davidmenendez:
> Can the options in the callback passed to `mockImplementation` be stored in a separate file and imported since it looks like the same options are used across multiple test files?

good call. i was thinking about this in passing, but didn't bother to try to implement it because i was worried something like this might adversely affect other tests. i'll give it a shot.

davidmenendez:
@matthewgallo as i had initially suspected trying to include the match media mock in the setup breaks a whole bunch of tests. i think for now we leave it as be for the sake of getting the original fix out and then we open another issue to fix the jest setup specifically.

matthewgallo (approved):

PR #4975, opened 2024-04-24
refactor(InterstitialScreenViewModule): add typescript types

anamikaanu96:
Contributes to #4484 

Converted packages/ibm-products/src/components/InterstitialScreenViewModule/InterstitialScreenViewModule.js to packages/ibm-products/src/components/InterstitialScreenViewModule/InterstitialScreenViewModule.tsx


packages/ibm-products/src/components/InterstitialScreenViewModule/InterstitialScreenViewModule.tsx
```diff
@@ -33,10 +33,27 @@ const componentName = 'InterstitialScreenViewModule';
 // Default values should be provided when the component needs to make a choice
 // or assumption when a prop is not supplied.
 
+interface InterstitialScreenViewModuleProps extends PropsWithChildren {
```
matthewgallo, line 36: This component doesn't use `children` so don't think you have to extend `PropsWithChildren` generic type.

matthewgallo (changes requested):
See comment regarding use of `PropsWithChildren` (not required for this component).

PR #4984, opened 2024-04-25
chore: stylelint 16 update

lee-chase:
This is not a viable pull for two reasons

1. The `styelint-plugin-carbon-tokens` is alpha to verify it works.
2. carbon-stylelint-config pulls in `stylelint-a11y` which is out of date. In my local projects I switched to `"@double-great/stylelint-a11y"`

https://github.com/orgs/carbon-design-system/projects/65


matthewgallo:
We are going to revisit this via #6374

PR #4987, opened 2024-04-26
build(storybook): update from webpack to vite to build storybook

### LINKED ISSUE #4946: Investigate removal of `storybook-addon-carbon-theme` from current storybook configuration
## What will this achieve?

Removing the current storybook addon theme switcher (`storybook-addon-carbon-theme`) would allow us to further optimize our storybook performance since we build it every time we start the storybook server. We have some options including the removal of it completely and implementing a similar setup to Carbon (as seen [here](https://github.com/carbon-design-system/carbon/blob/main/packages/react/.storybook/preview.js#L113-L122)) or we could investigate keeping our current `storybook-addon-carbon-theme` but not building it as part of the storybook startup script.

When we don't have to build `storybook-addon-carbon-theme` when starting storybook server.

## Additional information

- Designs
- Existing code
- etc

### LINKED ISSUE #4986: Move from webpack to vite for storybook config
## What will this achieve?

We'll have much better performance moving from webpack to vite to build our storybook and I think it will also make our move to storybook v8 a lot easier as well.

Refactoring `packages/core/.storybook/main.js` to use vite

## Additional information

- Designs
- Existing code
- etc

matthewgallo:
Closes #4986 
Closes #4946

This PR moves our storybook builds from `webpack` to `vite`. This includes significant performance improvements for both startup times and build times.

Moving to `vite` required renaming all story files to use `.jsx` and appending `?inline` to the imported style file for each story that has one (this is the only way to include a default import of a style file in `vite`, [see here for more details](https://vitejs.dev/guide/features#disabling-css-injection-into-the-page)).

We were able to remove several `babel` and `webpack` packages from `packages/core/package.json` with this change. Another part that is included in this PR is removing the `storybook-carbon-addon-theme`. It's been replaced with a similar approach to Carbon, theme can be changed from top toolbar.

#### What did you change?
Storybook config files
Renamed all story files from `.stories.js` to `.stories.jsx`

Verified that running storybook locally works and storybook can also be built as expected (`yarn storybook:build`).

Storybook startup after move to `vite` **~1 sec**
Screenshot 2024-04-26 at 10 50 11 AM


davidmenendez (dismissed):
🎉

davidmenendez (approved):

PR #4993, opened 2024-04-29
Non linear reading

anamikaanu96:
Contributes to #4290 

Converted packages/ibm-products/src/components/NonLinearReading/NonLinearReading.js to packages/ibm-products/src/components/NonLinearReading/NonLinearReading.tsx


packages/ibm-products/src/components/NonLinearReading/NonLinearReading.tsx
```diff
@@ -19,19 +19,45 @@ import { pkg } from '../../settings';
...
-// Default values for props
-const defaults = {
-  theme: 'light',
-};
+type Theme = 'light' | 'dark';
+interface NonLinearReadingProps extends PropsWithChildren {
+  /**
+   * The keyword of the component appears as a pill.
+   */
+  children: ReactNode;
```
szinta, line 27: When you extends from PropsWithChildren, you get the children prop internally typed. So If we are using **PropsWithChildren** then there is no need to explicitly specify **children** type.

anamikaanu96: yea I know but if its internally typed, the children become optional prop. Here the children is mandatory, thats why I specified it explicitly.

szinta: In that case, it is fine explicitly specify children type and you can skip extending from  PropsWithChildren. I think any one approach is fine.

matthewgallo (approved):

PR #5024, opened 2024-04-30
chore(release): release workflow

annawen1:
Adds release workflow that runs with manual trigger. Keeping the current release yml files in place until we have tested and written docs for this new release process 😄


matthewgallo (approved):
So excited to have our new process for releases coming to life 🥳 

Just another note about the `dist-tag` naming in this PR being `next` and in our current prerelease action it's named `canary`. I think it's definitely something we can address as we go if we find it necessary.

elycheea (dismissed):

Issue #4809, opened 2024-04-10
Get Started: Implementation


elycheea:
[Figma specs](https://www.figma.com/file/z5GvCED9gBr1b71soW3fIb/Get-started-pattern?type=design&node-id=567%3A19898&mode=design&t=WgjXCuaKLjv4aDj6-1)

Bridget still looking into positioning for status indicator (feels similar to radio button) so expect this to change over time.

The smaller cards will never include image.
May need to tie a minimum sizes to variant or conditional image based on size? 🤔 

These should use use productive card under the hood. The image and numbered variants can likely build off the default/base version. 

![Get started card -specs (1)](https://github.com/carbon-design-system/ibm-products/assets/15822070/ca606022-1a46-4934-908f-eb4dbb49299d)

### Acceptance Criteria

- Determine whether to create new Get Started component
- Should use clickable variant of productive card
- Add gradient mixin
- Add support for default variant
- Add support for icon (renderIcon) using crossroads as default
- Add support for status indicator (3 states — see [Checklist](https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-novice-to-pro-checklist--task-states))
- Add support for metadata section
- Add under Components/Card in Storybook (tbd)
- Add support for `image` — decorative (empty alt)
- At small screen sizes, `image` not displayed
- Add support for `number` variant

elycheea:
Create follow up issue to update remaining novice-to-pro components to use the mixin (currently see gradient repeated in ~8 files).

sstrubberg:
#### Questions

- Can we implement this using Logical Properties? (where necessary)
- Does this need any accessibility testing written?
- E2E testing?
- Typescript support out of the gate?

elycheea:
> * Can we implement this using Logical Properties? (where necessary)
+1 — we’re _starting_ to look into them and adopt them in some of the newer work but haven’t really added it as part of acceptance criteria yet. #3556 

> * Does this need any accessibility testing written?
We do a very baseline `toBeAccessible()` in scaffolding but not sure how much additional testing we need. Definitely open to pulling in some rigor if Dotcom devs have any input. :)

> * E2E testing?
Currently we’re only doing e2e for avt. But agree we should include this story as part of this initial implementation instead of adding it retroactively.

> * Typescript support out of the gate?
Probably depends on Cards which haven’t been migrated to `.tsx` yet.

PR #5054, opened 2024-05-01
feat(get-started): new component

ariellalgilmore:
Contributes to #4809 

Adds new getStarted component

#### What did you change?

- Added new Get Started component
- updated internal Card component to handle GetStarted updates without effecting other Cards

...still waiting on mobile updates from design

- added simple tests to verify default props and component rendering
- used storybook to check against figma design


bndege2:
@ariellalgilmore @elycheea 
- For all variants- metadata text sizing update to $label-01
- For variant with image: the image is not a 1:1 ratio. Please note that when the card with image sits on top of a card without an image, the image size should align with the card without the image width (see layout image example)
Get started layout 1

- Attached new placeholder svgs to use for the card with image

bndege2:
card with image placeholder 2
card with image placeholder

ariellalgilmore:
@bndege2 just pushed updates! The cards are flexible in size, so if the user puts them in the container like you show above then they will fit in that container

bndege2:
@ariellalgilmore updates look good!

packages/ibm-products/src/components/GetStarted/GetStarted.js
```diff
@@ -0,0 +1,101 @@
...
+
+import { Card } from '../Card';
+const componentName = 'GetStarted';
+
+/**
+ * GetStarted a card with icon, number, and media variants
+ */
+export let GetStarted = React.forwardRef(({ ...rest }, ref) => {
+  return (
+    <Card getStarted ref={ref} {...rest} {...getDevtoolsProps(componentName)} />
```
elycheea, line 23: I saw your comment that we used the internal `Card` instead of `ProductiveCard` — were we unable to get the actions working using `ProductiveCard`?

I was thinking originally something like ...

```suggestion
    <ProductiveCard ref={ref} {...rest} {...getDevtoolsProps(componentName)}>
      <div className={`${blockClass}__header`}>
        {sequence && (
          <div className={`${blockClass}__sequence`}>{sequence}</div>
        )}
        {status && (
          <div className={`${blockClass}__status`}>
            ...
          </div>
        )}
      </div>
      <div className={`${blockClass}__footer`}>
        <div className={`${blockClass}__metadata`}>
          ...
        </div>
      </div>
    </ProductiveCard>
```

... but I’m guessing this was a bit more complicated because of the positioning of sequence/status before the title/label and other media elements. 🤔

Still, anything we can keep specific to just the `GetStarted` component will likely help reduce complexity at the internal `Card` level longer term.

packages/ibm-products/src/components/GetStarted/GetStarted.mdx
```diff
@@ -0,0 +1,52 @@
...
+  <Story of={stories.Default} />
+</Canvas>
+
+## Example usage
+
+### with Pictogram
+
+<Canvas>
+  <Story of={stories.withPictogram} />
+</Canvas>
```
elycheea, line 26: Not seeing a difference in default and pictogram so wonder if we need to have these listed separately?

packages/ibm-products/src/components/GetStarted/GetStarted.stories.jsx
```diff
@@ -0,0 +1,103 @@
...
+Default.args = {
+  ...defaultProps,
+  pictogram: (props) => <Crossroads size={32} {...props} />,
+};
+
+export const withPictogram = Template.bind({});
+withPictogram.args = {
+  ...defaultProps,
+  pictogram: (props) => <Crossroads size={32} {...props} />,
+};
```
elycheea, line 83: Since this is the same as the default, I don’t know if we really need a separate story for this case?

packages/ibm-products/src/components/Card/Card.js
```diff
@@ -41,10 +42,14 @@ export let Card = forwardRef(
       // The component props, in alphabetical order (for consistency).
       actionIcons = defaults.actionIcons,
       actionsPlacement = defaults.actionsPlacement,
+      metaData = defaults.actionIcons,
```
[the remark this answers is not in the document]
elycheea: I do like the thought of reusing what we have, but also don’t think the metadata really counts as “actions” semantically? 🤔

elycheea (commented):
Overall looks great but I have some concerns about the structure and where we’re setting props. I’m guessing there’s a few limitations because of the way the internal Card is set up, but I feel like where we can, we should keep `GetStarted` specific props on the `GetStarted` component itself.

Happy to discuss these further too since I’m sure there’s a lot more complexity to my suggestions than they appear on paper. Might also be worth pulling in @davidmenendez for his thoughts since he worked on the initial card variants. :)

packages/ibm-products/src/components/GetStarted/GetStarted.stories.jsx
ariellalgilmore: removed withPictogram!

packages/ibm-products/src/components/Card/Card.js
```diff
@@ -41,10 +42,14 @@ export let Card = forwardRef(
       // The component props, in alphabetical order (for consistency).
       actionIcons = defaults.actionIcons,
       actionsPlacement = defaults.actionsPlacement,
+      metadata = defaults.actionIcons,
```
davidmenendez, line 45: is there any specific reason why `metadata` is required and we can't just use `actionIcons`?

packages/ibm-products/src/components/Card/Card.js
```diff
@@ -77,6 +84,7 @@ export let Card = forwardRef(
     },
     ref
   ) => {
+    getStarted ? (actionIcons = metadata) : '';
```
davidmenendez, line 87: i haven't come across this notation before. it's very interesting. personally i'm not a fan of mutating existing variables like this. _if_ we need this new `metadata` prop and we can't just use `actionIcons` i would rather see a getter function used here instead to instantiate the variable that contains the icons. something like `const getIcons = () => getStarted ? metadata : actionIcons`

packages/ibm-products/src/components/Card/Card.js
```diff
@@ -211,7 +229,7 @@ export let Card = forwardRef(
       primaryButtonText,
       primaryButtonDisabled,
       description,
-      hasActions: hasActions,
+      hasActions: hasActions && actionsPlacement === 'top',
```
davidmenendez, line 232: just for consistency sake i think you could move this near line 93 where `hasFooterActions` is set and just set something like `hasHeaderActions`

packages/ibm-products/src/components/Card/Card.js
```diff
@@ -259,16 +279,32 @@ export let Card = forwardRef(
...
           <div className={`${blockClass}__pictogram`}>
             <Pictogram />
           </div>
         )}
+        {getStarted && sequence && (
+          <div className={`${blockClass}__sequence`}>{sequence}</div>
+        )}
+        {getStarted && status && (
+          <div className={`${blockClass}__status`}>
+            {status === 'incomplete' ? <Incomplete /> : ''}
```
davidmenendez, line 295: two things

1. you should be able to shorten these by just using the `&&` notation so you don't need the empty strings.
2. i see that there's a `none` value for status to indicate a null or empty state. i think you can forgo that and just make the status either `complete` or `incomplete`

packages/ibm-products/src/components/Card/Card.js
```diff
@@ -259,16 +279,32 @@ export let Card = forwardRef(
...
+            {status === 'complete' ? <CheckmarkOutline /> : ''}
+          </div>
+        )}
         <div className={`${blockClass}__content-container`}>
           <div {...getHeaderBodyProps()}>
-            <CardHeader {...getHeaderProps()} />
-            <div {...getBodyProps()}>{children}</div>
+            <div>
```
davidmenendez, line 301: div with no classname is scary 😱

packages/ibm-products/src/components/Card/CardFooter.js
```diff
@@ -72,12 +75,23 @@ export let CardFooter = ({
         </Button>
       )}
       {hasActions && <div className={`${blockClass}__actions`}>{actions}</div>}
+      {FooterActionIcon && (
+        <div className={`${blockClass}__action-icon`}>
+          {disabled ? (
+            <Error size={16}></Error>
```
davidmenendez, line 81: you should be able to use the shorthand notation for consistency sake `<Error />` `<FooterActionIcon />`

davidmenendez (changes requested):
this looks really cool! aside from a few code comments i just wanted to request a specific story for the disabled state. it seems drastic enough that i would hate to see it missed since it's only available as a toggle currently.

packages/ibm-products/src/components/Card/Card.js
ariellalgilmore: when discussing with the PLG group, actionIcons was an icon you would take action on like in the other cards, but in this case it's more describing the card... so we thought metadata was a better prop name, but code-wise almost the same thing

ariellalgilmore:
@davidmenendez thanks for the review!! updated the getStarted component with your suggestions! I know we are still discussing card, but wanted to update here in the meantime!

davidmenendez (dismissed):
awesome. thanks for making those updates!

packages/ibm-products/src/components/Card/Card.js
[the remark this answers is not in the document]
elycheea: Refactor issue tracked in #5258

davidmenendez (approved):

matthewgallo:
@all-contributors add @ariellalgilmore for code

PR #5085, opened 2024-05-06
feat(conditionBuilder): issue fixes and review changes

amal-k-joy:
Contributes to #4577 

Issue fixes and review comments changes

#### What did you change?

- Introduced `id` to each conditions and groups. Will update/remove the condition based on id rather than index. Also used key as id instead of index for .map().
- Refactor the code to change the `state` prop to group={group} or condition={condition} to  make it more readable 
- Issue fixes and more refactoring

storybook


packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.js
```diff
@@ -31,53 +31,38 @@ const ConditionBuilderContent = ({
...
+  const onChangeHandler = (updatedGroup, groupId) => {
     /**
      * This method is triggered from inner components. This will be called every time when any change is to be updated in the rootState.
      * This gets the updated group as argument.
      */
     setRootState({
       ...rootState,
-      groups: rootState.groups.map((group, gIndex) =>
-        groupIndex === gIndex ? updatedGroup : group
+      groups: rootState.groups.map((group) =>
```
davidmenendez, line 58: since we're only updating one object per change we don't need to use map here to iterate over every item in the array. it would be more efficient to get the index of the group in the array. consider this code
```
    const groupIdx = rootState.groups.findIndex((o) => o.id === groupId);
    const groups = [
      ...rootState.groups.slice(0, groupIdx),
      updatedGroup,
      ...rootState.groups.slice(groupIdx + 1),
    ];
    setRootState({
      ...rootState,
      groups,
    });
```
with code like this we find the index and replace it directly without the need to map over everything else.

packages/ibm-products/src/components/ConditionBuilder/ConditionGroupBuilder/ConditionGroupBuilder.js
```diff
@@ -33,29 +35,52 @@ const ConditionGroupBuilder = ({
...
   };
 
-  const onChangeHandler = (updatedConditions, conditionIndex) => {
+  const onChangeHandler = (updatedCondition, conditionId) => {
     onChange({
-      ...state,
-      conditions: state.conditions.map((condition, cIndex) =>
-        conditionIndex === cIndex ? updatedConditions : condition
+      ...group,
+      conditions: group.conditions.map((condition) =>
```
davidmenendez, line 41: similar comment about updating the array without the usage of `map`

packages/ibm-products/src/components/ConditionBuilder/ConditionGroupBuilder/ConditionGroupBuilder.js
```diff
@@ -33,29 +35,52 @@ const ConditionGroupBuilder = ({
...
-        ...state.conditions.slice(0, conditionIndex + 1),
-        {
-          property: undefined,
-          operator: '',
-          value: '',
-          popoverToOpen: 'propertyField',
-        },
-        ...state.conditions.slice(conditionIndex + 1),
-      ],
+      ...group,
```
davidmenendez, line 59: you had it correct before! using the similar approach as mentioned
```
    const newIdx = conditionIndex + 1;

    const newCondition = {
      property: undefined,
      operator: '',
      value: '',
      popoverToOpen: 'propertyField',
      id: uuidv4(),
    };

    onChange({
      ...group,
      conditions: [
        ...group.conditions.slice(0, newIdx),
        newCondition,
        ...group.conditions.slice(newIdx + 1),
      ],
    });
```

davidmenendez (changes requested):
this looks great! thank you for reviewing my notes. just one small note regarding the `onChange` / update mechanism

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.js
amal-k-joy: modified by using slice().

packages/ibm-products/src/components/ConditionBuilder/ConditionGroupBuilder/ConditionGroupBuilder.js
amal-k-joy: modified by using slice()

packages/ibm-products/src/components/ConditionBuilder/ConditionGroupBuilder/ConditionGroupBuilder.js
amal-k-joy: modified by using slice()

amal-k-joy:
> this looks great! thank you for reviewing my notes. just one small note regarding the `onChange` / update mechanism

Made the changes as suggested

davidmenendez (approved):
awesome!

Issue #4985, opened 2024-04-26
[Feature Request]: Allow specifying more floating menu CSS selectors on Tearsheet component family


liunate:
### The problem

When using `<Tearsheet>`, we have floating element attached outside of the tearsheet DOM element, and `<Tearsheet>` does not allow moving focus onto the floating element.

### The solution

In the `<TearsheetShell>` internal component, selectively destructure `selectorsFloatingMenus` from `rest`, and join the existing hardcoded `selectorsFloatingMenus` property value:
```tsx
// TearsheetShell.js
const { selectorsFloatingMenus, ...restWithoutSelectorsFloatingMenus } = rest;

<ComposedModal
  { 
     ...restWithoutSelectorsFloatingMenus
  }
  selectorsFloatingMenus={[
    `.${carbonPrefix}--overflow-menu-options`,
    `.${carbonPrefix}--tooltip`,
    '.flatpickr-calendar',
    `.${bc}__container`,
     // join `selectorsFloatingMenus` here
]}
```

### Links to other materials

carbon `selectorsFloatingMenus` 
https://react.carbondesignsystem.com/?path=/docs/components-composedmodal--overview#focus-management

### Owner/main maintainer(s)

Nate(me)

### Business priority

High Priority = pressing release

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

liunate:
I can work on this if this is considered a possible enhancement 😉

PR #5092, opened 2024-05-07
feat(TearsheetShell): specify additional floating menu selectors

liunate:
Contributes to #4985 

Allow `<TearsheetShell>` to focus on elements matching CSS selectors specified

#### What did you change?

Pass thru and join `<TearsheetShell selectorsFloatingMenus>` value to the underlying carbon `<ComposedModal selectosFloatingMenus>`

I add the `selectorsFloatingMenus` property on the one of the story render property in `Tearsheet.stories.jsx` such like:

```jsx
      <div ref={ref}>
        <Tearsheet
          selectorsFloatingMenus={['#helloMan']}
        >
          {mainContent}
        </Tearsheet>
        <div id="helloMan" style={{position: 'fixed', left: '16rem', top: '16rem', zIndex: 9999, backgroundColor: 'green'}}>
            <h1> hello </h1>
            <input type='text' defaultValue="abc" />
          </div>
      </div>
```

Without the `#helloMan` being specified, the opened tearshet does not allow focusing on the input. When the selector is specified, the input can be focused and type inside.


liunate:

makafsal:
@liunate Code looks fine to me. But could you please add a storybook for this to demonstrate the feature?

packages/ibm-products/src/components/Tearsheet/TearsheetShell.stories.jsx
```diff
@@ -103,8 +105,8 @@ const Template = ({ influencer, open: _open, slug, ...args }, context) => {
   return (
     <div ref={ref}>
       <Button onClick={() => setOpen(true)}>
-        {beenOpen ? 'Reopen the' : 'Open the'} context.component.componentName
-      </Button>{' '}
+        {`${beenOpen ? 'Reopen' : 'Open'} the ${context.component.displayName}`}
+      </Button>
```
liunate, line 109: I think these part could be rewritten in tidier way like this.

liunate: `context.component.componentName` is no longer value, so I replaced it to show the correct component name. In this case here, the `TearsheetShell`

packages/ibm-products/src/components/Tearsheet/TearsheetShell.stories.jsx
```diff
@@ -140,3 +142,72 @@ export const NoAttributesSet = Template.bind({});
...
+          labelText={`This is a floating DOM element with classname "${className}" added outside of the tearsheet DOM element`}
+          placeholder="Focus and enter something"
+        ></TextInput>
+      </div>,
+      document.body
+    )
+  );
+}
+
+export const FocusOnFloatingMenu = ({ open: _open, ...args }, context) => {
```
liunate, line 174: Here is the recording showcasing the new property in action.

https://github.com/carbon-design-system/ibm-products/assets/38996291/6730b015-88fb-4f72-becb-12ad06391708

liunate: It works fine in story `docs` as well

liunate:
Not sure why it [failed](https://github.com/carbon-design-system/ibm-products/actions/runs/9184608468/job/25257212646?pr=5092#step:6:129) at the `ci-check:test:c4p` which succeeded in my local env. Any idea?

matthewgallo:
Looks like something is timing out. I just triggered it to run again

matthewgallo:
Hey @liunate, you just have one conflict to resolve in `TearsheetShell.stories.js`.

packages/ibm-products/src/components/Tearsheet/TearsheetShell.stories.jsx
```diff
@@ -179,6 +181,75 @@ NoAttributesSet.args = {
 };
 
 export const ReturnFocusToOpenButton = ReturnFocusTemplate.bind({});
-NoAttributesSet.args = {
+ReturnFocusToOpenButton.args = {
```
liunate, line 184: I believe this story template to specify the `.args` was wrong, so updated this to `ReturnFocusToOpenButton`.

liunate:
I just have this PR branch synced again with `main`, see if the changes now looks good for you?

elycheea (commented):
Hey Nate! Thank you for taking the time to contribute this! I noticed that we only show the docs and examples within the `TearsheetShell`, but probably should include `selectorsFloatingMenu` in the prop table for `Tearsheet` and `TearsheetNarrow`.

I believe that’s what Afsal was also referring to for including an example. I’ll bring this up in our next team huddle to verify, but I would guess that including the documentation in our Storybook docs[^1] and prop table is higher priority than the specific example for it. :)

[^1]: https://react.carbondesignsystem.com/?path=/docs/components-composedmodal--overview#focus-management

liunate:
Tks 👍 

The reason I didn't add examples to the enhanced components such as `Tearsheet` and `TearsheetNarrow` is because I feel the features that are commonly seen in both components might be better explained in their common base component story, which is the `TearsheetShell`. 

It also avoids duplicate demonstrations that might increase the user's cognitive load and distract them from the unique features of those enhanced components.

makafsal (dismissed):
Looks fine for me

elycheea (commented):
@liunate I think let’s get the `selectorsFloatingMenus` prop added to `Tearsheet.tsx` and `TearsheetNarrow.tsx` so they appear in the Storybook props table.

liunate:
@elycheea done!

elycheea (approved):
謝啦！ LGTM!

makafsal (approved):
🚀

PR #5097, opened 2024-05-07
feat(Coachmark): convert to .tsx

emyarod:
Contributes to #5095

This PR converts `Coachmark` to TypeScript. Will unblock #4494 

one thing to note is the hard coded drag handle string in `CoachmarkDragbar.tsx`, this will need updating once our i18n strategy for strings is figured out.

one other thing is the `overlayRef` prop type declaration in `Coachmark.tsx`, it may need a second look

#### What did you change?

- file extension
- TS interface
- Coachmark enums

storybook build and type checks


matthewgallo (approved):
Nice work!

PR #5127, opened 2024-05-08
test(NotificationsPanel): add default state avt test

annawen1:
Closes #5059 

#### What did you change?
Created `e2e/components/NotificationsPanel/NotificationsPanel-test.avt.e2e.js`

#### How did you test and verify your work?
`yarn avt`

PR #5159, opened 2024-05-13
test(NotificationsPanel): revert avt test

annawen1:
Reverts carbon-design-system/ibm-products#5127

Need to fix the violations that are occuring


kennylam (approved):
LGTM!

PR #5171, opened 2024-05-14
test(WebTerminal): add default state AVT check

### LINKED ISSUE #5090: Add default state AVT check for WebTerminal

anamikaanu96:
Closes #5090 

Created `e2e/components/WebTerminal/WebTerminal-test.avt.e2e.js`


elycheea (changes requested):
I think this is another case where we need to click on the web terminal in the global header in order to test this one properly.

There’s a few examples this in recent AVT PRs from Anna, Afsal, and Nandan.

https://github.com/carbon-design-system/ibm-products/pull/5127#discussion_r1595592118

cspell.json
```diff
@@ -173,6 +173,7 @@
     "unauthorizedemptystate",
     "userprofileimage",
     "uuidv",
-    "webfonts"
+    "webfonts",
+    "webterminal"
```
elycheea, line 177: This can be removed now since we added `avt.e2e.js` to our cspell ignores.

packages/ibm-products/src/components/WebTerminal/WebTerminal.stories.jsx
```diff
@@ -44,14 +44,16 @@ const Template = (args) => {
 
       <WebTerminal {...args}>
         <div className="example-terminal">
-          <p>Connection successful.</p>
-          <p>
-            DISCLAIMER: This is not a real terminal, you would pass your own
-            terminal component into the children of the WebTerminal component.
-          </p>
+          <main>
```
elycheea, line 47: I don’t think we need to add `<main>` here. The content not within landmark test should be ignored from Playwright config. (I double checked with @matthewgallo!)

If anything, I think `WebTerminal` could potentially benefit from using `<aside>` but I’m not sure that’s a change we should make without more investigation on how this one actually gets applied. 🤔

packages/ibm-products/src/components/WebTerminal/WebTerminal.tsx
```diff
@@ -176,7 +176,7 @@ export let WebTerminal = React.forwardRef(
         }}
         onAnimationEnd={onAnimationEnd}
       >
-        <header className={`${blockClass}__bar`}>
+        <header aria-label="Web terminal bar" className={`${blockClass}__bar`}>
```
elycheea, line 179: We shouldn’t hardcode this so we may need to add a new prop for this since it doesn’t look like we have any other label this could reference.

elycheea (changes requested):

packages/ibm-products/src/components/WebTerminal/WebTerminal.stories.jsx
anamikaanu96: Yes you are right, the content not within landmark test should be ignored from Playwright config.
I added <main> to resolve the error in Equal Access Accessibility Checker, its not causing issue in AVT.
Now I resolved it by adding role="contentinfo".

elycheea: In this case, not failing in AVT should be enough. 🙂 Our Storybook generally does not reflect actual page structure since we primarily demonstrate just the component interactions itself and exclude the rest of the page-level structure and elements.

The `WebTerminal` actually might be one that should have its own landmark (probably `<aside>` as I mentioned before), but I think that should be captured in a separate issue.

Let’s remove `role="contentinfo"`[^1] for now. This role probably isn’t the best choice in for the `WebTerminal` and could mask other issues in the future. We are currently ignoring this in from Playwright so we don’t need to add these roles unnecessarily even if you see the failure in EAC.

[^1]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/contentinfo_role

cspell.json
```diff
@@ -84,6 +84,7 @@
     "columnheader",
     "combobox",
     "combobutton",
+    "contentinfo",
```
elycheea, line 87: We can remove this line now.

packages/ibm-products/src/components/WebTerminal/WebTerminal.tsx
```diff
@@ -279,4 +287,9 @@ WebTerminal.propTypes = {
    * Optionally pass if the web terminal should be open by default
    */
   isInitiallyOpen: PropTypes.bool,
+
+  /**
+   * Specifies aria label for Web terminal
+   */
+  webTerminalAriaLabel: PropTypes.string.isRequired,
```
elycheea, line 294: I think to minimize impact to developer experience, we should set a fallback in [L185](https://github.com/carbon-design-system/ibm-products/pull/5171/files#diff-92f66e28307c3809b4bf6bd1fd6ac70644ed33b32846e25267fd36bb91d0f6e4R185) or set a default prop for the `webTerminalAriaLabel`.

Requiring this could cause errors for anyone using this component.

elycheea (commented):
One more set of small changes. I think we should set a default or fallback for the aria label rather than requiring it. This will minimize impact to anyone using this component.

Should also reduce the repeated labels in stories and tests if you set it too. 🙂

elycheea (dismissed):

makafsal (approved):

e2e/components/WebTerminal/WebTerminal-test.avt.e2e.js
```diff
@@ -0,0 +1,35 @@
...
+      component: 'WebTerminal',
+      id: 'ibm-products-patterns-web-terminal-webterminal--default',
+      globals: {
+        carbonTheme: 'white',
+      },
+    });
+
+    await page.getByLabel('Web terminal').click();
+    const modalElement = page.locator(
+      `.dev-prefix--${pkg.prefix}--web-terminal`
```
annawen1, line 26: We don't need the `dev-prefix` anymore since we've set the `pkg.prefix` to include it when we run the avt tests
```suggestion
      `.${pkg.prefix}--web-terminal`
```

annawen1 (changes requested):
Looks good, just one change with the prefix!

annawen1 (approved):

matthewgallo (approved):

PR #5249, opened 2024-05-16
feat(TagOverflow): overflow tooltip variants

### LINKED ISSUE #4090: Include variants and content for tag overflow tooltips
The tooltip of the tag overflow should display content from the item labels.

There are [two variants for this](https://pages.github.ibm.com/cdai-design/pal/components/tag-overflow/usage/#styling) — “read-only tags” and “interactive”/filterable tags.

### Design notes
- When rendering read-only text, content should wrap if it does not fit the toggletip width (288px)
- When rendering `tag`, content should truncate if the whole tag does not fit the toggletip width

### Development notes
Relevant props from original `TagSet`.
- `overflowAlign`
- `overflowType`
- `showAllTagsLabel`
- `overflowClassName`

```[tasklist]
### Acceptance criteria
- [x] Add support to modify toggletip position using `overflowAlign`
- [x] Add support for rendering text or tags as `overflowType` inside the toggletip
- [x] Add `showAllLabel` if more than 10 tags are show in the overflow
- [x] Add support for `overflowClassName` for the toggletip popup
- [ ] Review accessibility using the [IBM Accessibility Checker browser extension](https://github.com/IBMa/equal-access)
- [ ] Add and update tests for each prop or scenario
```

szinta:
Closes #4090 

Support overflow content variants. Variants are "read-only tags" and "interactive tags."

#### What did you change?

1. TagOverflowPopover and TagoverflowModal components updated to render read-only or interactive tags based on overflowType prop
2. Alignment of both Additional content Popover and Truncation tooltip are set based on  overflowAlign prop.
3. Text wrap/truncation  applied as per design recommendations 

Storybook


packages/ibm-products/src/components/TagOverflow/TagOverflowModal.js
```diff
@@ -104,11 +105,16 @@ export const TagOverflowModal = ({
...
       </ModalHeader>
       <ModalBody className={`${blockClass}__body`} hasForm>
-        {getFilteredItems().map(({ label, id, ...other }) => (
-          <Tag {...other} filter={false} key={id}>
-            {label}
-          </Tag>
-        ))}
+        {getFilteredItems().map(({ label, id, filter, ...other }) => {
+          if (overflowType !== 'tag') {
+            filter = false;
```
davidmenendez, line 110: we should never be mutating data like this. the conditional should be set as a variable (preferably) or done in the prop
```
const isFilterable = overflowType !== 'tag';
filter={isFilterable}
// OR
filter={overflowType !== 'tag'}
```

packages/ibm-products/src/components/TagOverflow/TagOverflowPopover.js
```diff
@@ -102,29 +102,35 @@ export const TagOverflowPopover = React.forwardRef(
...
               <ul className={`${blockClass}__tag-list`}>
-                {getOverflowPopoverItems().map((tag) => {
-                  const tagProps = {};
-                  if (overflowType === 'tag') {
-                    tagProps.type = 'high-contrast';
+                {getOverflowPopoverItems().map(
+                  ({ tagType, onClose, ...other }) => {
+                    const tagProps = { type: tagType, ...other };
+                    if (overflowType === 'tag') {
+                      tagProps.type = 'high-contrast';
```
davidmenendez, line 109: same comment above regarding variable mutation. it would be preferable to just compute this value in the prop instead of overwriting anything.

_if_ we wanted to take this approach however i would say that it should be handled in the `getOverflowPopoverItems` function. so instead of putting this logic here, it could look more like
```
const getOverflowPopoverItems = () => {
  return overflowTags.filter( ... ).map( { code to add or change properties } );
};
```

davidmenendez (changes requested):

davidmenendez:
additionally, this isn't specifically related to the PR, but i wanted to make a note of this. the `getFilteredItems` could be optimized using the array filter method. consider something like this:
```
const getFilteredItems = () => {
  if (open && search) {
    const filteredTags = allTags.filter((tag) => {
      return tag.label
        ?.toLocaleLowerCase()
        ?.includes(search.toLocaleLowerCase());
    });
    return filteredTags;
  }
  return allTags;
};
```
i didn't see `data-search` in the props anywhere, so i omitted it from this code haha. but yeah this code is slightly nicer because it doesn't rely on pushing items to other arrays. the filter method returns a shallow copy already!

packages/ibm-products/src/components/TagOverflow/TagOverflowModal.js
szinta: Done

packages/ibm-products/src/components/TagOverflow/TagOverflowPopover.js
szinta: Since we already use a map method to render the array elements, I'd rather not introduce another map within getOverflowPopoverItems. So, like you mentioned in the previous comment, I used variables for the conditionals rather than mutating the object.

szinta:
> additionally, this isn't specifically related to the PR, but i wanted to make a note of this. the `getFilteredItems` could be optimized using the array filter method. consider something like this:
> 
> ```
> const getFilteredItems = () => {
>   if (open && search) {
>     const filteredTags = allTags.filter((tag) => {
>       return tag.label
>         ?.toLocaleLowerCase()
>         ?.includes(search.toLocaleLowerCase());
>     });
>     return filteredTags;
>   }
>   return allTags;
> };
> ```
> 
> i didn't see `data-search` in the props anywhere, so i omitted it from this code haha. but yeah this code is slightly nicer because it doesn't rely on pushing items to other arrays. the filter method returns a shallow copy already!

Thanks for suggesting this optimisation. I have updated the code.

davidmenendez (approved):

amal-k-joy (approved):

PR #5251, opened 2024-05-16
fix: apply some sass fixes to prevent storybook from crashing

matthewgallo:
Closes #

{{short description}}


matthewgallo:
This was cause because of a local cache issue 😄

Issue #4911, opened 2024-04-16
[Datagrid: Datagrid does not remember current selection and expansion state when re-rendering


flannanl:
### Package

Carbon for IBM Products

### Description

When using `Datagrid` with `useNestedRows` and `useSelectRows`, any selection or expansion state done in the rows are not preserved after re-rendering.

### Component(s) impacted

Here is a snippet that can show the issue. With this snippet, if you try to select or expand a row within the 5-sec refresh, you will see all the selection and expansion states you have done are lost after re-rendering. I do have a current workaround at the moment, but this is a hack of using some of the internal functions available in `datagridState` to preserve the selection and expansion states.

```
import React, { useEffect, useMemo, useState } from 'react';
import {
  Datagrid,
  useDatagrid,
  useNestedRows,
  useSelectRows,
} from '@carbon/ibm-products';
import { makeData } from './utils/makeData';

const defaultHeader = [
  {
    Header: 'Row Index',
    accessor: (row, i) => i,
    sticky: 'left',
    id: 'rowIndex', // id is required when accessor is a function.
  },
  {
    Header: 'Age',
    accessor: 'age',
  },
  {
    Header: 'First Name',
    accessor: 'firstName',
  },
  {
    Header: 'Last Name',
    accessor: 'lastName',
  },
  {
    Header: 'Visits',
    accessor: 'visits',
  },
];

export const Example = () => {
  const columns = React.useMemo(() => [...defaultHeader], []);
  const [data, setData] = useState(makeData(10));

  useEffect(() => {
    setTimeout(() => {
      setData((pastData) =>
        pastData.map((d) => ({
          ...d,
          age: Number(d.age) + 10,
        }))
      );
    }, 5000);
  }, []);

  const rows = useMemo(() => {
    const parentRows = [];
    for (let i = 0; i < data.length; i = i + 2) {
      parentRows.push({
        ...data[i],
        subRows: [data[i + 1]],
      });
    }
    return parentRows;
  }, [data]);

  const datagridState = useDatagrid(
    {
      columns,
      data: rows,
    },
    useSelectRows,
    useNestedRows
  );

  return <Datagrid datagridState={{ ...datagridState }} />;
};

export default Example;
```

### Browser

Chrome, Safari

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

v1.72.2

### Severity

[...]

PR #5261, opened 2024-05-17
feat(Datagrid): add default expansion state to `useNestedRows`

### LINKED ISSUE #5260: [Datagrid]: Support specifying initial states for newly added rows in Datagrid
### Package

Carbon for IBM Products

### Description

Based on https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-components-datagrid-nestedrows--docs#with-initially-expanded-nested-rows, there is currently a way to specify the initial expansion state for nested rows. However, this requires the data or rows to be ALL available before very first time rendering the Datagrid.

In our use case, this is impossible because we fetch data at a later time in paginations. So as we render the Datagrid, we want it to be in loading state and then as data comes in, we will start adding rows to the Datagrid. Because of this, we will never have enough data / info to populate the initial state we want.

Secondly, the initial state currently seems to be based on the actual row IDs, which is generated by Datagrid. There is no way for us to know the actual row ID ahead of time before we render the Datagrid table.

We would like to have a few enhancements and behaviour change:

1) One of our other use cases is that after doing certain actions from Datagrid table, we will create a new nested row, which we want it to be expanded by default. Today, we have no such way to do so. Therefore we would like to have a way to specify an `initialState` for this new row to tell Datagrid to expand or select it at re-render time. After rendering with the new row(s), the `initialState` should be thrown away and let the users to manage the states afterward.
2) We would like the `initialState` to use the `id` in the `data` array instead. This is the ID we know ahead. So Datagrid should use the `data.id` whenever possible and then fallback to the actual row id if it is missing.
3) We want a way to specify the `initialState` of multiple rows as they being injected to the table. This is similar to (1), so if (1) is solved, I believe this will be solved as well.
4) Today, the `initialState` only deals with expansion state, it will be good if it can also deal with selection state.

### Component(s) impacted

IBM Projects UI: https://cloud.ibm.com/projects

### Browser

Chrome, Safari, Firefox, Microsoft Edge

1.72.2

### Severity

### Product/offering

IBM Cloud Projects

N/A

_No response_

_No response_

### Code of Conduct

matthewgallo:
Closes #5260

This PR provides a way to have rows expanded by default (when they're not included as part of `initialState.expandedRowIds`). By adding `defaultExpanded: true` to the row, the `useNestedRows` hook/plugin will toggle the expansion state so that if it is a new row and should be expanded by default, this is now possible.

There is still an issue opened from @flannanl around expansion states being not kept after rerendering, but there is a [separate issue open for this](https://github.com/carbon-design-system/ibm-products/issues/4911).

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/useNestedRows.js
```
Storybook

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/useNestedRows.js
+++ b/packages/ibm-products/src/components/Datagrid/useNestedRows.js
@@ -18,7 +18,7 @@ const useNestedRows = (hooks) => {
     useEffect(() => {
       const { rows } = instance;
       const defaultExpandedRows = rows.filter(
-        (row) => row.original.defaultExpanded
+        (row) => row?.original?.defaultExpanded
       );
       if (defaultExpandedRows?.length) {
         defaultExpandedRows.map((defaultExpandedRow) => {
--- a/packages/ibm-products/src/components/Datagrid/useNestedRows.js
+++ b/packages/ibm-products/src/components/Datagrid/useNestedRows.js
@@ -15,6 +15,24 @@ const blockClass = `${pkg.prefix}--datagrid`;
 const useNestedRows = (hooks) => {
   useNestedRowExpander(hooks);
   const useInstance = (instance) => {
+    useEffect(() => {
+      const { rows } = instance;
+      const defaultExpandedRows = rows.filter(
+        (row) => row.original.defaultExpanded
+      );
+      if (defaultExpandedRows?.length) {
+        defaultExpandedRows.map((defaultExpandedRow) => {
+          if (
+            !defaultExpandedRow.isExpanded &&
+            !defaultExpandedRow.hasExpanded
+          ) {
+            defaultExpandedRow?.toggleRowExpanded?.();
+            defaultExpandedRow.hasExpanded = true;
+            return;
+          }
+        });
+      }
+    }, [instance, instance.rows]);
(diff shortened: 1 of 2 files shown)


packages/ibm-products/src/components/Datagrid/useNestedRows.js
```diff
@@ -15,6 +15,24 @@ const blockClass = `${pkg.prefix}--datagrid`;
 const useNestedRows = (hooks) => {
   useNestedRowExpander(hooks);
   const useInstance = (instance) => {
+    useEffect(() => {
+      const { rows } = instance;
+      const defaultExpandedRows = rows.filter(
+        (row) => row.original.defaultExpanded
```
emyarod, line 21:
```suggestion
        (row) => row?.original?.defaultExpanded
```

do these need to be optional?

emyarod (approved):
looks good to me!

packages/ibm-products/src/components/Datagrid/useNestedRows.js
matthewgallo: Yes, good catch. Safer to add the optional chaining

emyarod (approved):
looks good to me! reapproving after latest updates

sangeethababu9223 (approved):

Issue #5270, opened 2024-05-20
[DataSpreadsheets] Needs null checks to prevent crashes 


AustinGitHub:
### Package

Carbon for IBM Products

### Description

Needs null checks

![Screenshot 2024-05-20 at 1 11 51 PM](https://github.com/carbon-design-system/ibm-products/assets/10100397/f859efab-d10e-4e3e-bbcd-7edf7764b527)

### Component(s) impacted

Data spreadsheets

### Browser

Chrome

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

2.38

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Product/offering

IBM 

### CodeSandbox or Stackblitz example

N/A

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

AustinGitHub:
I will fork the repo and make the changes myself because this is something that we need for DataStage

AustinGitHub:
Null check has been merged in, thanks!

PR #5271, opened 2024-05-20
fix(DataSpreadsheet): Add null check to `setActiveCellContent`

AustinGitHub:
Add null check

Closes # https://github.com/carbon-design-system/ibm-products/issues/5270

#### What did you change?

I added null check 

I verified using comparing how the logic works for when cell is being deleted in `handleCellDeletion.js` it is then set to `null` so I added that same check here.


packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
```diff
@@ -291,7 +291,11 @@ export let DataSpreadsheet = React.forwardRef(
                   activeCellCoordinates?.column
                 ]
               : null;
-          setActiveCellContent(activeCellFullData.render('Cell'));
+          if (activeCellFullData) {
+            setActiveCellContent(activeCellFullData.render('Cell'));
+          } else {
+            setActiveCellContent(null);
+          }
```
AustinGitHub, line 298: What is odd is not sure why we allow `activeCellFullData` to be `null` in this condition block here but then try to do a `.render('cell')` on it. Seems a use case was missed here. So my fix should patch that, thanks!

Screenshot 2024-05-20 at 1 39 49 PM

AustinGitHub:

AustinGitHub:
recheck

elycheea:
@AustinGitHub The commit history shows two different IDs. Usually when we see this happen, it’s because the user of the commit is tied to your internal GitHub account. You can check the email associated with these commits with `git show <commit-hash>`. You can either amend the commits (which can sometimes feel dirtier) or create a new PR with the changes (usually easier if changes are small).

> image
> image

Also no need to add yourself to the contributors file. We usually use the allcontributors bot to add you both to the `.contributorsrc` and the actual contributor table.

AustinGitHub:
thanks, I will try to fix it

AustinGitHub:
I made a new PR https://github.com/carbon-design-system/ibm-products/pull/5305 thanks

PR #5305, opened 2024-05-21
fix(DataSpreadsheet): added null check

AustinGitHub:
Add null check

Closes # https://github.com/carbon-design-system/ibm-products/issues/5270

#### What did you change?

I added null check 

I verified using comparing how the logic works for when cell is being deleted in `handleCellDeletion.js` it is then set to `null` so I added that same check here.


AustinGitHub:

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
```diff
@@ -291,7 +291,11 @@ export let DataSpreadsheet = React.forwardRef(
                   activeCellCoordinates?.column
                 ]
               : null;
-          setActiveCellContent(activeCellFullData.render('Cell'));
+          if (activeCellFullData) {
+            setActiveCellContent(activeCellFullData.render('Cell'));
+          } else {
+            setActiveCellContent(null);
+          }
```
makafsal, line 298: I think we can replace this with the Nullish coalescing operator (??) and optional chaining (.?).
`setActiveCellContent(activeCellFullData?.render('Cell') ?? null);`

makafsal (dismissed):
Looks fine

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
AustinGitHub: Sure

AustinGitHub: Done, thanks!

emyarod (approved):
looks good to me!

AustinGitHub:
Hi @makafsal can you re-review so this PR can be merged in, thanks!

AustinGitHub:
Thanks Matthew!

AustinGitHub:
Coming back to this PR because we moved up package upgrade to 2.40 and I think the original null check I had should've been applied instead. I am seeing this issue on our table where when we edit it is setting it to empty (which is not what is expected)


https://github.com/carbon-design-system/ibm-products/assets/10100397/4669c2a6-4425-4b25-8cd7-52f9c192219c



I checked the code import and the issue is it is doing this where it sets it to false because of the condition rather than the actual render object. 

https://github.com/carbon-design-system/ibm-products/assets/10100397/592f2508-0d70-4fea-8d40-ddead1abad88

I will open an issue for this and make a fix since this is regression

Issue #5272, opened 2024-05-20
DataGrid: Custom class for header not being applied on header element


AustinGitHub:
![Screenshot 2024-05-20 at 3 23 54 PM](https://github.com/carbon-design-system/ibm-products/assets/10100397/e98b8d66-bacc-4b04-be8f-985d7612fbde)

Get's applied on the cell instead of the header cell 

![Screenshot 2024-05-20 at 3 22 12 PM](https://github.com/carbon-design-system/ibm-products/assets/10100397/abba9788-0093-4d58-89cc-f80e17f1d7a1)

See header doesn't have this class 

![Screenshot 2024-05-20 at 3 24 01 PM](https://github.com/carbon-design-system/ibm-products/assets/10100397/fa045317-5553-409d-94cc-055dbc5a631a)

Code 
<img width="1048" alt="Screenshot 2024-05-20 at 3 24 36 PM" src="https://github.com/carbon-design-system/ibm-products/assets/10100397/c9c1db69-103f-47bc-b65f-426069dbf967">

PR #5366, opened 2024-05-29
fix(datagrid): pass classname from header props directly

annawen1:
Closes #[5272](https://github.com/carbon-design-system/ibm-products/issues/5272)

When passing a custom classname to the header props, the classname was coming through as `undefined` and not being applied to the column header `<th>` element. It comes through as `undefined` as `header.getHeaderProps()` doesn't return any `className` field.

#### What did you change?

Reference the `className` from `header` object directly.

Tested in the `Basic usage` story where we are passing in the custom classname `test-column` in the story

Screenshot 2024-05-29 at 9 25 32 AM

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHeaderRow.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHeaderRow.js
@@ -193,7 +193,6 @@ const HeaderRow = (datagridState, headRef, headerGroup) => {
             ...headerProps
           } = header.getHeaderProps();
 
-          console.log('header props', headerProps);
           const resizerProps = header?.getResizerProps?.();
           return (
             <TableHeader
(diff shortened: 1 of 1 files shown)


packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHeaderRow.js
```diff
@@ -190,14 +190,15 @@ const HeaderRow = (datagridState, headRef, headerGroup) => {
           const {
             // eslint-disable-next-line no-unused-vars
             role,
-            className: headerClassName,
             ...headerProps
           } = header.getHeaderProps();
+
+          console.log('header props', headerProps);
```
emyarod, line 196:
```suggestion
```

emyarod (dismissed):
looks good to me!

davidmenendez (approved):

emyarod (approved):
looks good to me!

PR #5372, opened 2024-05-29
fix(notification-panel): remove role on decorative image

### LINKED ISSUE #4940: `NotificationsPanel` a11y violations
## What will this achieve?
Remediate accessibility violations on `NotificationsPanel` component

Pass **IBM Equal Access Accessibility Checker** scan

## Additional information
1. The scan shows `Accessible name does not match or contain the visible label text` for each notification in the panel
image (74)
2. The scan shows `Element with "img" role has no label or an empty label` for the empty state image
image (75)

ariellalgilmore:
Closes #4940

#### What did you change?

removed role="img" on empty since decorative to pass accessibility tests

Ran the accessibility checker and confirmed it wasn't there anymore

### DIFF
--- a/packages/ibm-products/src/components/EmptyStates/assets/NotificationsIllustration.js
+++ b/packages/ibm-products/src/components/EmptyStates/assets/NotificationsIllustration.js
@@ -32,7 +32,6 @@ export const NotificationsIllustration = ({ theme, size, alt, ...rest }) => {
         `${blockClass}__illustration-notification`,
         `${blockClass}__illustration--${size}`,
       ])}
-      role="img"
     >
       <title>{alt}</title>
       {theme === 'dark' ? (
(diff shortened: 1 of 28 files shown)


szinta (dismissed):
LGTM

packages/ibm-products/src/components/EmptyStates/assets/NotificationsIllustration.js
```diff
@@ -33,6 +33,7 @@ export const NotificationsIllustration = ({ theme, size, alt, ...rest }) => {
         `${blockClass}__illustration--${size}`,
       ])}
       role="img"
+      aria-label="Illustration of an empty notification state"
```
elycheea, line 36: Rather than hardcoding the `aria-label`, we can probably either reuse the `alt` in L38 or use `aria-labelledby`.

ariellalgilmore: oh that showed me we just actually needed something in the `title`, so we can use `illustrativeDescription` and it also solved the fix

elycheea (commented):
Thought we would have to remove the title/alt too but seems like just removing role was enough.

Let’s just update the PR title to reflect the change.
(FYI, we removed `role="img"` since the image is decorative in this context.)

elycheea (approved):

PR #5398, opened 2024-06-04
fix(Datagrid): implemented nestedRow dynamic width

### LINKED ISSUE #4972: Overlapping columns in nested rows — Reduce width instead of margin in first column of nested row
### Package

Carbon for IBM Products

### Description

Currently, to account for the first column of a nested row being smaller due to it being indented, the `margin-right` property is reduced by 50px (and more for further nesting). This however doesn't account for text/content for the column being more than the available space and leads to content overlapping with the next column (This seems to especially be an issue on smaller screens.

Switching to reducing the width of the cell instead would stop this overlapping and allow text-breaking etc. To do this I think there would need to be a way to get the width of this cell (either by passing the value set on the parent column, or getting the width of the cell itself, and reducing this by the same amount as is currently being set by `margin-right`. 

image

image

### Component(s) impacted

Datagrid

### Browser

Firefox

v.2.35.0

### Severity

### Product/offering

Cloud Pak For Integration (Platform UI)

https://ibm-products.carbondesignsystem.com/?path=/docs/ibm-products-components-datagrid-nestedrows--docs

- Create Nested rows dataGrid
- Observe column overlap on small screen size or in dev tools

_No response_

### Code of Conduct

devadula-nandan:
Closes #4972 

made the width dynamically change, instead of giving margin-right = -50 to fix overlapping issue.
also if any cell is resized below 50px (minimum width), the difference is added to margin right to fix the layout of the row

#### What did you change?
packages/ibm-products/src/components/Datagrid/useNestedRows.js

storybook

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/useNestedRows.js
+++ b/packages/ibm-products/src/components/Datagrid/useNestedRows.js
@@ -60,15 +60,6 @@ const useNestedRows = (hooks) => {
       // reduce the "first cell"s width to compensate added (left) margin
       const isFirstCell =
         instance.columns.findIndex((c) => c.id === cell.column.id) === 0;
-      // if (
-      //   isFirstCell &&
-      //   cell.row.depth > 0 &&
-      //   parseInt(props.style.width, 10) <= 100
-      // ) {
-      //   console.log(
-      //     -32 * cell.row.depth - 18 + parseInt(props.style.width, 10) - 50
-      //   );
-      // }
       return [
         props,
         {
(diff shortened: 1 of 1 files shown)


packages/ibm-products/src/components/Datagrid/useNestedRows.js
```diff
@@ -64,11 +64,21 @@ const useNestedRows = (hooks) => {
         props,
         {
           style: {
-            marginRight: `${
+            marginRight:
+              isFirstCell &&
```
amal-k-joy, line 68: Hey ,
Looks good and seems to be working.
Can you add some comments regarding the logic and figures used here for calculation?

amal-k-joy (changes requested):

packages/ibm-products/src/components/Datagrid/useNestedRows.js
devadula-nandan: done some refactors for readability, and added comments

devadula-nandan:
@amal-k-joy added comments as per discussion

amal-k-joy (approved):
Good work

IgnacioBecerra (approved):
LGTM!

Issue #5386, opened 2024-05-31
[Datagrid]: Validator always gets called even though cell is not in edit state


AustinGitHub:
### Package

Carbon for IBM Products

### Description

https://github.com/carbon-design-system/ibm-products/assets/10100397/5378fcb3-b2bf-4ee2-80c7-49bf361a2f58

### Component(s) impacted

Datagrids

### Browser

Chrome

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

2.40

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Product/offering

IBM DataStage

### CodeSandbox or Stackblitz example

N/A

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

PR #5412, opened 2024-06-04
fix(Datagrid): Call validator only when `inEditMode` is true to avoid unnecessary calling

AustinGitHub:
Closes https://github.com/carbon-design-system/ibm-products/issues/5386

#### What did you change?

Added condition check for `inEditMode` so `validator` doesn't get called when it isn't needed. 

Storybook 

https://github.com/carbon-design-system/ibm-products/assets/10100397/62df0d42-e087-48f6-99ea-4325a62a5c03


matthewgallo (approved):
This is great, thanks for contributing @AustinGitHub! 🎉

annawen1 (approved):

PR #5447, opened 2024-06-10
fix: address bug in update gallery script

### LINKED ISSUE #5423: Fix `update-gallery-config` script
`yarn update-gallery-config` helps us keep our example galleries in Stackblitz and CodeSandbox up-to-date. (Thank you, @lee-chase.)

Recently, both @matthewgallo and I have started to see errors when running the gallery update script — which fails to run entirely.

```
TypeError: fs.fmSync is not a function
```

> Image

Looking in Node.js file system docs[^1], this function isn’t listed at all so need to investigate and test to ensure our examples can still be updated.

[^1]: https://nodejs.org/api/fs.html

matthewgallo:
Closes #5423 

This change fixes a bug in the example gallery update script. It previously attempted to call `fs.fmSync` which does not exist rather than `fs.rmSync` as intended.

#### What did you change?
```
scripts/example-gallery-builder/update-example/index.js
```
Applied this change and was able to successfully run `yarn update-gallery-config` which failed previously

### DIFF
--- a/scripts/example-gallery-builder/update-example/index.js
+++ b/scripts/example-gallery-builder/update-example/index.js
@@ -118,7 +118,7 @@ const updateExample = (name, config) => {
 
     const example_node_modules = join(rootPath, examplePath, 'node_modules');
     if (fs.existsSync(example_node_modules)) {
-      fs.fmSync(example_node_modules, { recursive: true, force: true });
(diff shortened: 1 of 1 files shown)


matthewgallo:
Credit goes to @elycheea! 🥳

annawen1 (approved):

PR #5467, opened 2024-06-11
fix(sidepanel): close/ai placement

### LINKED ISSUE #5466: Alignment of close/AI button is incorrect if the placement prop is left
The alignment of close and AI icon is showing as incorrect if the placement prop is left hence it is invisible.

Screenshot 2024-06-11 at 2 04 33 PM

anamikaanu96:
Closes #5466 

 To make the positions of close/ai, i changed the position of class `dev-prefix--c4p--side-panel__slug-and-close` from `fixed` to `absolute`.


matthewgallo:
@lee-chase I remember you changed the positioning back to `fixed` from `absolute` in [this PR](https://github.com/carbon-design-system/ibm-products/pull/4426). Do you recall why? Was it the fact that when positioned absolute it disappeared on scroll? I don't see that behavior here with @anamikaanu96's changes.

lee-chase:
> @lee-chase I remember you changed the positioning back to `fixed` from `absolute` in [this PR](https://github.com/carbon-design-system/ibm-products/pull/4426). Do you recall why? Was it the fact that when positioned absolute it disappeared on scroll? I don't see that behavior here with @anamikaanu96's changes.

The issue is when `reduced motion` is turned on. Currently there appears to be numerous problems with reduced motion on, with or without the changes.

matthewgallo:
@anamikaanu96 there is some weird behavior as @lee-chase mentioned when reduced motion is turned on with the side panel close icon. Can you try to take a look at this?

matthewgallo (approved):
Tested with placement of panel on left and right and with and without reduced motion, not seeing any layout issues that were noticed before. Lets just keep an eye out for anything here!

Issue #5330, opened 2024-05-23
[DataSpreadsheet]: Hitting enter on a cell does new line instead of entering edit mode on same line


AustinGitHub:
### Package

Carbon for IBM Products

### Description

https://github.com/carbon-design-system/ibm-products/assets/10100397/31e79048-8557-402c-9e9d-8c142027ebc5

### Component(s) impacted

Data Spreadsheets

### Browser

Chrome

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

2.39.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Product/offering

IBM

### CodeSandbox or Stackblitz example

https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-data-spreadsheet-dataspreadsheet--data-spreadsheet&globals=viewport:basic

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues
```[tasklist]
### Tasks
- [ ] Pressing enter should cause the current cell to enter edit mode and not apply new line, unless the cell is already in edit mode
```

AustinGitHub:
Will take a look at this today

PR #5479, opened 2024-06-12
fix(Dataspreadsheets): don't append new line when we hit enter

AustinGitHub:
Closes #

https://github.com/carbon-design-system/ibm-products/issues/5330

#### What did you change?

If previous state wasn't in edit mode, then don't append the new line. I checked the storybook before this change and there wasn't a way to do new line through any commands that I found. So this doesn't regress anything. Thanks!

Storybook

https://github.com/carbon-design-system/ibm-products/assets/10100397/62512d95-c05b-4839-8a5b-9444de88cc40


AustinGitHub:
Hi @kennylam @ariellalgilmore can you help review this PR? Thanks!

davidmenendez (approved):

Issue #5484, opened 2024-06-12
[Dataspreadsheet]: Drag-drop columns doesn't update selection highlight


AustinGitHub:
### Package

Carbon for IBM Products

### Description

![Screenshot 2024-06-12 at 1 56 39 PM](https://github.com/carbon-design-system/ibm-products/assets/10100397/69f3afae-be99-453a-abd5-937d730f294e)

### Component(s) impacted

Dataspreadsheets

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

2.42

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Product/offering

IBM

### CodeSandbox or Stackblitz example

https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-data-spreadsheet-dataspreadsheet--large-dataset-spreadsheet&globals=viewport:basic

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

AustinGitHub:
Another issue I see on this is the cursor still is in grab selection even though mouse up occurred.

Issue #5262, opened 2024-05-17
[DataSpreadsheet]: SelectionAreaChange does not capture every event + Selection mouse move gets stuck when remove onclick


AustinGitHub:
### Package

Carbon for IBM Products

### Description

https://github.com/carbon-design-system/ibm-products/assets/10100397/650e7460-a4a3-4070-9c3d-72aa8f68fa35

When I release the click hold on the selection sometimes when I move my mouse around, it still thinks it is in selection mode is is highlighting other cells. 

I would expect the `selection area change` to receive update when we are going from cell to cell, seems this is not the case.

### Component(s) impacted

DataSpreadsheets

### Browser

Chrome

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

2.38.0

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Product/offering

IBM DataStage

### CodeSandbox or Stackblitz example

https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-data-spreadsheet-dataspreadsheet--data-spreadsheet&globals=viewport:basic

### Steps to reproduce the issue (if applicable)

See Video. 

For the selection being active when mouse is no longer being drag held, this occurs sometimes but pretty common. 

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

AustinGitHub:
Will take a look at this today

AustinGitHub:
![Screenshot 2024-06-12 at 12 52 34 PM](https://github.com/carbon-design-system/ibm-products/assets/10100397/e2449eb0-3833-4855-b68f-364487d49f17)

Issue is that the selection did change but the comparison in the code is incorrect

AustinGitHub:
For this issue "Selection mouse move gets stuck when remove onclick"
This is because the `onMouseUp` event seems to only be recorded if we do mouse up on the cell that we initially clicked down on. When we click and drag then do mouse up, it does not get recorded. Debugging this part of this issue.

elycheea:
@all-contributors add @AustinGitHub for code

PR #5420, opened 2024-06-05
chore(deps): bump braces and micromatch

elycheea:
Closes #5404

#### What did you change?
braces@3.0.2 ➡️ braces@3.0.3
micromatch@4.0.5 ➡️ micromatch@4.0.6

#### How did you test and verify your work?
Mend passes in CI 🤞

PR #5485, opened 2024-06-12
fix(Dataspreadsheet): selection area behavior fixed

AustinGitHub:
Closes #

https://github.com/carbon-design-system/ibm-products/issues/5262
https://github.com/carbon-design-system/ibm-products/issues/5484

#### What did you change?

I made it so we check if there is a column we are holding and dragging so the mouse up behavior can work as intended. This also fixes the issue where when we do mouseUp we don't persist the `headerCellHoldActive` of true when it should be false. 

I also fixed the issue where the callback function was not logging all the selections when the selection area changed. There is another condition we need to also check for. I checked original behaviors and I am not seeing any regressions. Thanks!

Storybook 

https://github.com/carbon-design-system/ibm-products/assets/10100397/86f5f60c-2ca1-4203-9a60-644f0e999907


AustinGitHub:
Going to try and add fix for this https://github.com/carbon-design-system/ibm-products/issues/5484 in here as well. I was debugging separate and realized it would be easier to add in this PR.

AustinGitHub:
https://github.com/carbon-design-system/ibm-products/assets/10100397/7e4f1455-778b-4d7b-abf0-0af1e5bec52c

I also fix this issue https://github.com/carbon-design-system/ibm-products/issues/5484

matthewgallo (changes requested):
Nice! One thing I notice is that the selection area if you select an entire column and then reorder it, the selection area is lost. Can you take a look at this so that if you have a column selected (selection area exists for the entire column) it is still included after moving the column?

AustinGitHub:
Sure

matthewgallo:
https://github.com/carbon-design-system/ibm-products/pull/5420 was merged which fixes the Mend check that was failing.

matthewgallo (approved):

AustinGitHub:
oh, I had only resolved the code conflict

davidmenendez (approved):

matthewgallo (approved):

PR #5500, opened 2024-06-13
feat(dataspreadsheet): optional custom component to row header

### LINKED ISSUE #5253: [Feature Request]: DataSpreadsheets: Add way to add custom actions to row index cell
### The problem

Need a way to be able to add this dropdown action here 

Screenshot 2024-05-16 at 10 22 44 AM

### The solution

To the dataspreadsheet component, add a way to pass through an option which can contain elements so we can be able to display something like that. 

This will make it easier for users to customize what they want to add there next to the index. This will satisfy our needs and anyone else's. Thanks!

### Design link

_No response_

### Links to other materials

_No response_

### Owner/main maintainer(s)

IBM

_No response_

### Product/offering

_No response_

### Business priority

None

### Code of Conduct

AustinGitHub:
Closes #5253 

#### What did you change?

Allow users to pass through their own custom component to the data spreadsheet table on column 0. So user can pass through icons, button, menu list, etc. 

This component is also directional so user can specify `Left` or `Right` to determine which side they want the action to be. 

I also created new story for this feature. 

Storybook

https://github.com/carbon-design-system/ibm-products/assets/10100397/83454d2a-8883-4273-95ae-75092c5a0000

The width of the column 0 also scales with the number we put in to render so there shouldn't be any overflow issues. 

https://github.com/carbon-design-system/ibm-products/assets/10100397/ddf9117a-d5db-4622-97c1-3b31bea2cbf3


AustinGitHub:
Please let me know if there is better way to name the variables and descriptions. I was unsure on those, thanks!

davidmenendez:
@AustinGitHub thanks for the submission. i'll need to confirm with design that this enhancement is acceptable.

davidmenendez:
oh ok i see it under the [usage guidelines](https://pages.github.ibm.com/cdai-design/pal/components/data-spreadsheet/style). i'll go through the code now 👍

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.stories.jsx
```diff
@@ -96,6 +97,29 @@ const Template = ({ ...args }) => {
...
+    </OverflowMenu>
+  );
+
+  return (
+    <DataSpreadsheet
+      columns={columns}
+      data={data}
+      onDataUpdate={setData}
+      rowNumberingComponentPosition="Left"
+      rowNumberingComponent={buildComponent}
```
davidmenendez, line 116: i think `rowNumberingComponent` is a tad bit confusing. i think that because the documentation specifies usage of overflow menu that maybe we do an implementation where the API is called `rowActions` which accepts an array of actions that can be mapped over internally.

davidmenendez (changes requested):

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.stories.jsx
AustinGitHub: So does the documentation specify it has to be overflow menu? I have it set so user provides custom component so it could be icon, button, or overflow menu.

davidmenendez: Screenshot 2024-06-17 at 1 02 33 PM

it's a bit hard to find because it's nested under an accordion, but it seems like it's mostly targeted towards overflow usage.

actually looking at this more i think maybe this should be set in the `columns` API. something like

```
const columnData = [
  {
    Header: 'Row Index',
    headerActions: [{ label: 'delete', onClick: () => {} }, ...],
    accessor: (row, index) => index,
...
```

let me pull in @matthewgallo on this since he's much more familiar with this component and i'd like to get his take on this.

AustinGitHub: Sure, thanks

davidmenendez: after talking it over with @matthewgallo we think that a new prop called `renderHeaderOverflow` would work for this. the only main change we'd like to see is that this overflow is also accessible in the column header as per the usage guidelines. the prop could also provide a callback that distinguishes between a row and column header in case a separate set of actions are required for rows and columns. does that make sense?

AustinGitHub: I see what you mean, but for the column, this is already customizable for the user to do. See how I am currently doing it for my code 

Screenshot 2024-06-17 at 11 40 48 AM

And then I provide my overflow menu item in here

Screenshot 2024-06-17 at 11 41 09 AM

AustinGitHub: Result
Screenshot 2024-06-17 at 11 44 27 AM

AustinGitHub: Hi @davidmenendez I discussed with Matt on this with some back and forth to get understanding. 

Currently, with column headers user can already provide custom component to it as with the example I provided above with how we have this in our code right now. 

For row headers on the row indexing numbering column there is no user control on this as it is hardcoded in. I explained to Matt how it is easier for Developer to provide their own custom component such as the overflow menu and if that menu has action events, it is easier for us to handle these action events rather than having to try and get them sorted through carbon products.

The changes I will be making is renaming the components to something more understandable like `renderRowHeader` and `renderRowHeaderDirection`. I'll also double check with the carbon guidelines, but storybook does have it's own section showing user how to utilize this feature and example is going to be using the overflowmenu. 

In future, if this is wanting to be changed around then that is fine but for now since column headers is already customizable, rather than making this not customizable and forcing it to be a pure string and having separate way of adding component. It is easier for now to leave it as is to avoid regressions. 

Any thoughts on this? 

I will update the PR in a bit with renaming the variables. Thanks

davidmenendez: this sounds good to me. if @matthewgallo has any questions or concerns i'll leave it up to him to post them here.

matthewgallo: Sounds like a plan for now, this way we don't interfere with any existing functionality, column headers already being able to be customized. 👍

AustinGitHub: Thanks everyone, will leave a comment once I have the PR updated.

AustinGitHub:
Updated the variable names, thanks!

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheetBody.tsx
```diff
@@ -628,6 +648,10 @@ export const DataSpreadsheetBody = forwardRef(
         const row = rows?.[index];
         if (rows && rows.length) {
           prepareRow?.(row);
+          let uniqueRowHeaderComponent: any[] | undefined = undefined;
+          if (typeof renderRowHeader === 'function') {
```
davidmenendez, line 652: this check is already defined in `hasCustomRowHeader`. in the render i would just do `{hasCustomRowHeader && renderRowHeader(index}`

davidmenendez (changes requested):
looking good! just have one small comment. also, if you could include a unit test for this new feature that would be much appreciated.

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
```diff
@@ -1011,6 +1040,15 @@ DataSpreadsheet.propTypes = {
...
 
+  /**
+   * Component next to numbering rows
+   */
+  renderRowHeader: PropTypes.func,
+
+  /**
+   * Component next to numbering rows
+   */
+  renderRowHeaderDirection: PropTypes.string,
```
matthewgallo, line 1051: Can you change this to `PropTypes.oneOf(['left', 'right']);`?

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
```diff
@@ -118,6 +123,16 @@ interface DataSpreadsheetProps {
    */
   onSelectionAreaChange?: () => void;
 
+  /**
+   * Position of the custom row numbering component
+   */
+  renderRowHeaderDirection?: string;
```
matthewgallo, line 129: Can you update this as well? `'left' | 'right';`?

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheetBody.tsx
AustinGitHub: Fixed thanks! I had to add null check on the function, but was able to get rid of these lines and do singular line. Thanks!

matthewgallo:
Hey @AustinGitHub, one of the main pieces of design feedback is that the overflow should be square (32x32).

AustinGitHub:
Sure, I will take a look today

AustinGitHub:
Hi @matthewgallo since this is an optional component it is up to user to defined the width and height. I added the change but since this is optional component, it would be up to dev that provides component to add their own stylings to the component. But for storybook I put it to 32x32. Thanks!

davidmenendez (approved):
LGTM

matthewgallo (approved):

PR #5506, opened 2024-06-14
fix(dataspreadsheet): drag drop when we have overflow

### LINKED ISSUE #5222: Data Spreadsheet: Can't drag and drop column when there is overflow on the table
### Package

Carbon for IBM Products

### Description

https://github.com/carbon-design-system/ibm-products/assets/10100397/f44bf335-ebf8-409d-be24-c1445adb62a5

### Component(s) impacted

Data spreadsheets

### Browser

Chrome

2.38.0

### Severity

### Product/offering

IBM

https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-data-spreadsheet-dataspreadsheet--with-many-columns&globals=viewport:basic

See video attached

_No response_

### Code of Conduct

AustinGitHub:
Closes #5222

#### What did you change?

I added code logic so we can have the table auto-scroll when we are dragging left and right. While this isn't the smoothest implementation, this is better than nothing and would appreciate if we can at least have some initial implementation and then if carbon wants to adjust later that would be fine. But right now having no functional use case is going to be something customers will point out to us. Thanks!

Storybook

https://github.com/carbon-design-system/ibm-products/assets/10100397/9c32420f-a592-4228-984b-0bc1fb458efc


matthewgallo (changes requested):
Thanks for working on this @AustinGitHub. One thing I'm noticing is that you can endlessly scroll to the right. I think the expected behavior would be if you reach the end of the `listContainer` element to stop any scrolling.

https://github.com/carbon-design-system/ibm-products/assets/10215203/71974e12-036a-4b77-bb64-70494fe4cb70

AustinGitHub:
Thanks! I will take a look today

AustinGitHub:
https://github.com/carbon-design-system/ibm-products/assets/10100397/a13b0695-6d59-4fd3-82f3-05bad4c4b92f

I fixed to best I can for that issue. I see the endless scrolling not happening anymore and in some cases it doesn't scroll but in other cases I do notice it does a little but but not endless. I don't have a lot of time to do a deep dive into trying to get an absolute fix, but I think this is manageable for now.

packages/ibm-products/src/components/DataSpreadsheet/utils/moveColumnIndicatorLine.js
```diff
@@ -30,20 +30,53 @@ export const moveColumnIndicatorLine = ({
   const selectionAreaOrigin = ref.current.querySelector(
     `[data-matcher-id="${matcherId}"]`
   );
+
+  const listContainer = ref.current.querySelector(
+    `.${blockClass}__list--container`
+  );
+
+  let scrollSpeed = 10; // Scrolling speed
```
matthewgallo, line 38: One more thing, since these variables are not re-assigned a new value can you change from `let` to `const`?

AustinGitHub: Sure I can do that

matthewgallo (dismissed):

devadula-nandan (approved):
🎉

packages/ibm-products-styles/src/components/DataSpreadsheet/_data-spreadsheet.scss
```diff
@@ -267,10 +267,6 @@ $header-cell-background: $layer-accent-01;
     width: $spacing-01;
     background-color: $background-brand;
   }
-  .#{$block-class}__th--active-header,
-  .#{$block-class}__td-th--active-header.#{$block-class}__td {
-    background-color: $background-selected-hover;
-  }
```
devadula-nandan, line 273: any reason why we remove this style? its not highlighting the cell coordinates like before.
screenshot from production
Screenshot 2024-06-18 at 11 31 56 AM

AustinGitHub: Yes, this class is the reason this bleed-through is occurring

Screenshot 2024-06-17 at 11 17 56 PM

devadula-nandan: adding this background to cell container might solve this issue, could you check it once?
Screenshot 2024-06-18 at 12 13 33 PM

devadula-nandan: we can apply default background color to the parent, as in screenshot, so we can avoid compromising on this feature?

Screenshot 2024-06-18 at 12 31 06 PM

AustinGitHub: Thanks, I will take a look later today

AustinGitHub: I have fixed this, thanks!

Screenshot 2024-06-18 at 8 52 46 AM

matthewgallo (approved):

devadula-nandan (approved):
🎉

PR #5527, opened 2024-06-17
fix(dataspreadsheet): Change condition check (fix regression)

### LINKED ISSUE #5526: [dataspreadsheet]: condition check is returning out boolean instead of render object
### Package

Carbon for IBM Products

### Description

Condition is doing this from what I am seeing, it is changing the value to false instead of keeping the object

https://github.com/carbon-design-system/ibm-products/assets/10100397/50f6c7e5-099f-47f1-9a23-59b3e6914046

The issue it is causing, the editing values are returning undefined due to the above condition. 
Note: The `(Null)` is custom code on our end, ignore that

https://github.com/carbon-design-system/ibm-products/assets/10100397/8571ba18-9c11-41d3-9451-51aab0fe4123

### Component(s) impacted

Dataspreadsheets

### Browser

Chrome

2.40

### Severity

### Product/offering

IBM

N/A

I tried to reproduce in storybook but I think because this is coming through as a package import on our local code env, it can't really be that reproducible. 

_No response_

### Code of Conduct

AustinGitHub:
Closes #5526 

#### What did you change?

Fix condition check, see issue videos on why this condition check changed

Story book, original use case still works. Won't know absolute fix until next version release since issue seems to be coming through due to how the importing of the component is happening.

### DIFF
--- a/packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
+++ b/packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
@@ -299,7 +299,11 @@ export let DataSpreadsheet = React.forwardRef(
                   activeCellCoordinates?.column
                 ]
               : null;
-          setActiveCellContent(activeCellFullData?.render('Cell') ?? null);
+          if (activeCellFullData) {
+            setActiveCellContent(activeCellFullData.render('Cell'));
+          } else {
+            setActiveCellContent(null);
+          }
         }
         if (
           (activeCellCoordinates && activeCellCoordinates?.row === 'header') ||


packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
```diff
@@ -299,7 +299,11 @@ export let DataSpreadsheet = React.forwardRef(
                   activeCellCoordinates?.column
                 ]
               : null;
-          setActiveCellContent(activeCellFullData?.render('Cell') ?? null);
+          if (activeCellFullData) {
```
AustinGitHub, line 302: @matthewgallo Based on the videos here in the issue https://github.com/carbon-design-system/ibm-products/issues/5526 do you know if this is just a typescript issue? Because it looks like the null check originally does not work as intended when we import it through our packages on our code. I'm not that familiar with TypeScript outside the basics, not sure on the syntax specific differences there are. Can you let me know if you have any understanding? Thanks!

Would like to get this PR merged in as soon as possible if this is a true fix for what I am seeing on our local env. This is actually a regression in 2.40 for us though on storybook it wasn't able to be replicated, like I mentioned in the issue it looks like due to how it gets imported to the packages, the null checks get constructed weirdly. Thanks!

matthewgallo: Sure, I'll take a look.

AustinGitHub: I'll make a video and send it over on slack after lunch. Thanks!

matthewgallo (approved):

emyarod (approved):
looks good to me!

PR #5546, opened 2024-06-19
fix(datagrid): check tableId before triggering reset of filters

### LINKED ISSUE #4523: Clear Filter clears filters from all tables when multiple DataGridComponents are rendered at the same time
- [ ] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive) - Yes

## Detailed description
- Use Carbon Tab Panels . Add two panels .
- In each of the Panel render the DataGrid component  each having a unique id.
- Apply filters in both the tables.
- Now click on Clear filters in one of the table.
- This would clear filters from both the DataGrid Table .
-  Also if both the DataGrid tables have any common columns  and if filter is applied for that column in both tables , the click on X (to clear individual filter ) on one table would clear the filter for respective column in another table.

DataGrid Filters

Since both are 2 different table , though they are rendered at the same time . The clear filter should only clear filter from one table and not both .

Chrome and Safari

> package are you using?
 "@carbon/ibm-products": "2.30.0",
  "@carbon/react": "1.52.0",

- Screenshots or code
image

annawen1:
Closes #4523 

When there are multiple instances of Datagrid used, clicking the "clear filters" in one DataGrid instance causes all DataGrid filters to be reset instead of the specific instance.

This PR passes the `tableId` of the DataGrid to its Filter Context and is then compared and must match before resetting the filters.

#### What did you change?

- Passing in the `tableId` to the `FilterProvider` so the context is associated with the table id. Then also passing in the tableId to the event emitter for comparison in the `reset()` callback

```
components/Datagrid/Datagrid/Datagrid.tsx
components/Datagrid/Datagrid/DatagridContent.tsx
components/Datagrid/Datagrid/addons/Filtering/FilterFlyout.js
components/Datagrid/Datagrid/addons/Filtering/FilterPanel.js
components/Datagrid/Datagrid/addons/Filtering/FilterProvider.js
components/Datagrid/Datagrid/addons/Filtering/hooks/useFilters.js
cspell.json
```

I duplicated the Datagrid components in the `FilteringUsage.js` file that the Filtering story utilizes and tested within storybook.

https://github.com/carbon-design-system/ibm-products/assets/54281166/3af2ad3f-0763-49e5-9c09-d1acbf506caf


matthewgallo (changes requested):
💯 this is great!

Can you also update the `useSubscribeToEventEmitter` hook in `FilterPanel.js` as well to include these same updates?

annawen1:
> 💯 this is great!
> 
> Can you also update the `useSubscribeToEventEmitter` hook in `FilterPanel.js` as well to include these same updates?

Oop! yeah good catch!

matthewgallo (approved):

makafsal (approved):
Looks good.

PR #5575, opened 2024-06-24
fix(options-tile): fix styling of summary and chevron when nesting OptionsTile

### LINKED ISSUE #5560: [OptionsTile]: Nested options tile does not render chevron and summary correctly
### Package

Carbon for IBM Products

### Description

When nesting an OptionsTile into a parent OptionsTile, the nested options tile exhibits the following problems:
- When the nested options tile is collapsed, its summary is not visible. The expected behavior is that when a nested options tile is collapsed, that the summary is visible.

- When the nested options tile is collapsed, its chevron is pointing upwards (instead of downwards). Since the chevron does not change between collapsed and expanded, the animation between the two states is also not visible.
  
image

The screen-shot above has been created using the following code:

```jsx
<OptionsTile
  summary="This is the parent"
  title="Parent options tile"
>
  <OptionsTile
    summary="Summary of the nested options tile"
    title="Nested options tile"
  >
    <p>
      Lorem ipsum dolor sit amet consectetur adipisicing elit.
    </p>
  </OptionsTile>
</OptionsTile>
```

### Component(s) impacted

OptionsTile

### Browser

Chrome, Safari

2.43.1

### Severity

### Product/offering

IBM z

https://stackblitz.com/edit/github-pjtyuj?file=src%2FExample%2FExample.jsx

_No response_

_No response_

### Code of Conduct

reviewthis:
Closes #5560

#### What did you change?

change the selectors for the CSS rules applicable to summary and chevron to use BEM modifiers, instead of looking at parent elements to determine the open and closing states.

The effects of the fix are visual, so (manually) compared how the chevron looks like and behaves during opening/closing, verified that the summary is visible on a nested options tile when it is collapsed.

### DIFF
--- a/packages/ibm-products-styles/src/components/OptionsTile/_options-tile.scss
+++ b/packages/ibm-products-styles/src/components/OptionsTile/_options-tile.scss
@@ -163,20 +163,17 @@ $block-class: #{c4p-settings.$pkg-prefix}--options-tile;
   overflow: hidden;
 }
 
-.#{$block-class}:not(.#{$block-class}--closing)
-  > details[open]
-  .#{$block-class}__summary,
+.#{$block-class}__summary.#{$block-class}__summary--open:not(.#{$block-class}__summary--closing),
 .#{$block-class}__summary--hidden {
   height: 0;
   margin-top: 0;
   opacity: 0;
 }
 
-.#{$block-class}:not(.#{$block-class}--closing)
-  > details[open]
-  .#{$block-class}__chevron {
+.#{$block-class}__chevron.#{$block-class}__chevron--open:not(.#{$block-class}__chevron--closing) {
   transform: rotate(180deg);
 }
+
 
 .#{$block-class} > details[open] .#{$block-class}__content {
   padding-top: $spacing-03;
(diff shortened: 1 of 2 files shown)


reviewthis:

reviewthis:
recheck

matthewgallo:
Hey @reviewthis, I believe we've seen issues with the DCO signatures when the user is different between the commit and the user opening the PR.

elycheea:
@reviewthis If you check the commit history above, it looks like the change was implemented from another account. That user ID is the one that needs to sign the DCO, but also if that’s incorrect, it may be easier to open a new PR, ensuring your changes are tied to the correct user / email ID.

reviewthis:
@elycheea @matthewgallo Thanks for looking into this PR. I created it with my original gh account (@reviewthis), where I seem to have changed the username at some point in the past. In retrospect, seems not to have been a good idea.

I would close this PR, and create a new one using my new account (@markusnosse), hoping that this time things will work out.

markusnosse:
Same PR from "new" account: #5609

Issue #4032, opened 2024-01-10
<a11y>The ARIA role 'separator' is not valid for the element <input>


xubinnb:
===== ENVIRONMENT =====
Browser: Version 105.0.5195.125 (Official Build) (arm64)
Automatically update Chrome for all users Learn more
Test Tool Type: IBM Accessbility Checker

===== DEFECT DESCRIPTION ===== 
Problem 1: The ARIA role 'separator' is not valid for the element
Problem 2: An element with ARIA role 'separator' does not have the required ARIA attribute(s): 'aria-valuenow'

=====STEPS TO REPRODUCE =====
Go to https://dev.console.test.cloud.ibm.com/projects

Use IBM Equal Access Accessibility Checker to scan

===== EXPECTED RESULT =====
[IBM 4.1.2 Name, Role, Value](https://www.ibm.com/able/requirements/requirements/#4_1_2)
For all user interface components (including but not limited to: form elements, links, and components generated by scripts), the name and role can be programmatically determined; states, properties, and values that can be set by the user can be programmatically set; and notification of changes to these items is available to user agents, including assistive technologies.

===== SCREEN CAPTURE =====
<img width="1268" alt="image" src="https://media.github.ibm.com/user/42629/files/8fe62374-3e2d-4455-833c-e36809d3c46c">
<img width="1276" alt="image" src="https://media.github.ibm.com/user/42629/files/d4ac4c6f-dae8-41e6-8030-316e7ce20ca2">
<img width="1268" alt="image" src="https://media.github.ibm.com/user/42629/files/eb0369c5-a97d-4bbc-9572-d27a42ef9d64">
<img width="1270" alt="image" src="https://media.github.ibm.com/user/42629/files/5078d216-db7c-4655-ab1c-0047deb14db6">
```[tasklist]
### Acceptance criteria
- [ ] Investigate alternative element for `input`
```

davidmenendez:
hey @xubinnb this should have been fixed in a previous commit https://github.com/carbon-design-system/ibm-products/pull/4072

please update to the latest version of carbon for ibm products and let us know if this problem still persists! thanks.

elycheea:
@SeonyuK did some investigation and we may not have resolved this issue. Even though we removed `role="separator"` in our implementation, it’s still visible in Storybook and in latest versions of our package. The separator role seems to be getting set via `react-table`[^1].

Suggest we explore options from my previous comment for this one rather than just deleting the role.

[...]

PR #5583, opened 2024-06-24
fix: remove unnecessary title attributes from datagrid

### LINKED ISSUE #5351: DataGrid : Option to disable/enable  datagrid tooltip
### Package

Carbon for IBM Products

### Description

By default the DataGrid adds  tooltip for the table header and table body.
In situations when the column content is not a plain text but a JSX element , the tooltip displays incorrect value 

We can either
- Disable the tooltip for the required column or Provide a custom string whenever a cell value is a JSX element /component

Screenshot 2024-05-28 at 8 14 34 PM

### Component(s) impacted

DataGrid 

### Browser

Chrome, Safari, Microsoft Edge

"@carbon/ibm-products": "2.40.0",

### Severity

### Product/offering

Infrastructure 

https://codesandbox.io/p/sandbox/datagrid-clear-fliter-issue-forked-plpt64?file=%2Fsrc%2FExample.js%3A161%2C7

_No response_

_No response_

### Code of Conduct

davidmenendez:
Closes #5351 

this PR removes a bunch of unnecessary `title` attributes that were strung throughout the `Datagrid` component that was causing accessibility issues. [current documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#accessibility_concerns) tells us there's only a few valid instances in which to use `title` and none of the instances here meet the criteria.

> Outside of potential uses on input elements, there are notable instances where the title is the recommended attribute for accessibility user experiences.

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridContent.tsx
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridContent.tsx
@@ -58,7 +58,6 @@ export const DatagridContent = ({
     fullHeightDatagrid,
     verticalAlign = 'center',
     variableRowHeight,
-    gridTitle,
     gridDescription,
     useDenseHeader,
     withInlineEdit,
@@ -237,7 +236,6 @@ export const DatagridContent = ({
           [`${blockClass}__with-pagination`]: DatagridPagination,
           [`${blockClass}__dense-header`]: useDenseHeader,
         })}
-        title={gridTitle}
         description={gridDescription}
       >
         <DatagridToolbar
@@ -307,7 +305,6 @@ DatagridContent.propTypes = {
     useDenseHeader: PropTypes.bool,
     withInlineEdit: PropTypes.bool,
     verticalAlign: PropTypes.string,
-    gridTitle: PropTypes.node,
     gridDescription: PropTypes.node,
     page: PropTypes.arrayOf(PropTypes.object),
     rows: PropTypes.arrayOf(PropTypes.object),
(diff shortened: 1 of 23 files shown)


packages/ibm-products/src/components/Datagrid/Datagrid/DatagridContent.tsx
```diff
@@ -237,7 +236,6 @@ export const DatagridContent = ({
           [`${blockClass}__with-pagination`]: DatagridPagination,
           [`${blockClass}__dense-header`]: useDenseHeader,
         })}
-        title={gridTitle}
```
matthewgallo, line 240: This prop isn't a `title` attribute but an actual prop for the `TableContainer` from `@carbon/react` which displays an optional title for the table.
Screenshot 2024-06-24 at 3 46 00 PM

davidmenendez: oooh i see. ok. my mistake. will revert those changes.

davidmenendez:
@matthewgallo thanks for catching that title issue. i went ahead and reverted that commit and only removed instances related to `title`.

i also added an additional small fix that adds `role: undefined` to cells and rows to alleviate violations that were showing up in the accessibility checker while testing

Screenshot 2024-06-24 at 3 26 42 PM

took us down from 209 to 59 violations. progress 😅 

Screenshot 2024-06-24 at 3 25 14 PM
Screenshot 2024-06-24 at 3 25 19 PM

davidmenendez:
the last commit i added removed some additional problematic roles. this gets our violation count down to 32 💪

davidmenendez:
actually went ahead and addressed #4032 in the last commit because the fix was basically the same

matthewgallo (approved):

PR #5587, opened 2024-06-25
test(tearsheet): implement AVT coverage for complex functionalities

### LINKED ISSUE #5521: Add AVT coverage for Tearsheet (wide) states: Open/close, disabled, determinate, focus trap, keyboard nav
- Tearsheet buttons can be disabled (not sure if this needs to be tested separately if it’s based on the button?)
- Tearsheet has a unique stacking behavior (we tagged as “determinate” for lack of a better name/category for now)

makafsal:
Closes #5521 

#### What did you change?
1. Focus trap and focus return to launcher button


ariellalgilmore (approved):
LGTM! just wondering if we should test for using the `esc` when closing the tearsheet

elycheea (commented):
Good point, @ariellalgilmore.

@makafsal since the actual keyboard actions and tabbing sections can get quite long, I think it’d be helpful to add comments[^1] to help us navigate the tests.

[^1]: [Carbon’s `Modal` AVT tests](https://github.com/carbon-design-system/carbon/blob/main/e2e/components/Modal/Modal-test.avt.e2e.js)

elycheea:
FYI, noticed the AVT tests aren’t actually running for `Tearsheet` in our CI. Checked with @matthewgallo and @davidmenendez and I think we may just need to add some more avt runners.

ariellalgilmore (approved):

elycheea (approved):

PR #5605, opened 2024-06-26
chore(issue-template): added better description and default to "none" for severity of bugs

jeffchew:
This is an update to the bug issue template so that it will no longer default to "Severity 1" for all issues reported. This also borrows from Carbon core to point to the descriptions of severities as well.


matthewgallo (approved):
🎉

sangeethababu9223 (approved):

PR #5619, opened 2024-06-27
fix(DataSpreadsheet): address issue with active cell value

### LINKED ISSUE #5617: [DataSpreadsheet]: Something is hardcoded for Column Index 1 to not read the proper value datatype
### Package

Carbon for IBM Products

### Description

On our end 

https://github.com/carbon-design-system/ibm-products/assets/10100397/d464eb1f-be17-4a61-9aa0-28e6feb8082f

One correlation I can make is in all the stories the column index 1 is all `Row Index` with integer datatype. I think there is something hardcoded on carbon products side to where when we have Column Index 1 if it isn't integer it will return undefined. 

This issue came from the conversion of `js` to `tsx`. So maybe some strict rule is being applied in `tsx` that cause this issue for us. Hi @amal-k-joy I saw you worked on the conversion for this component. I am going to investigate and see if I can fix, but just wanted to let you know in case if maybe you have understanding of what is occurring. Thanks!

### Component(s) impacted

Dataspreadsheets

### Browser

Chrome

2.39.0

### Suggested Severity

### Product/offering

IBM DataStage

https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-data-spreadsheet-dataspreadsheet--large-dataset-spreadsheet-custom&globals=viewport:basic

I need to investigate further, but I believe because in the story it is Row Index, there is some datatype detection issue being made. 

The storybook can sort of be replicated to my video I posted above, for example when we double click into row index, we see that it shows empty. That is similar to what I see on my implementation except it should treat this column as a string rather than an integer column. 

We will need this to be in patch as soon as possible. If my understanding is correct, then we shouldn't be expecting user to always have Column index 1 as an integer column but treat it as any type of column they want, string or integer, etc. I am not 100% sure on this yet which is why I need to investigate. 

### Code of Conduct

matthewgallo:
Closes #5617

This change properly sets the cell editor value so that when entering edit mode for cells within the first column (index 0) the correct value is applied.

#### What did you change?
```
packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
```
Storybook

### DIFF
--- a/packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
+++ b/packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
@@ -641,12 +641,9 @@ export let DataSpreadsheet = React.forwardRef(
             ]
           : null;
 
-      let activeCellValue;
-      if (activeCellFullData && activeCellCoordinates?.column) {
-        activeCellValue = activeCellFullData
-          ? activeCellFullData.row.cells?.[activeCellCoordinates?.column]?.value
-          : null;
-      }
+      const activeCellValue =
+        activeCellFullData?.row?.cells?.[Number(activeCellCoordinates?.column)]
+          ?.value;
 
       setCellEditorValue(activeCellValue || '');
       if (cellEditorRulerRef?.current) {


emyarod (approved):
looks good to me!

davidmenendez (approved):

Issue #5620, opened 2024-06-27
[DataSpreadsheet]: When drag/drop of columns, no callback can be detected for dev


AustinGitHub:
### Package

Carbon for IBM Products

### Description

When I drag/drop a column I see that `onDataUpdate` is not getting this response so how is used supposed to trigger the update state?

I set up callback here in local env, data update does not get callback happening. Dev is not able to know that data got changed 

https://github.com/carbon-design-system/ibm-products/assets/10100397/a05f086f-16ab-4e0d-bb45-5a3cc49b9703

### Component(s) impacted

Data Spreadsheets

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

2.44.0-rc.1

### Suggested Severity

None

### Product/offering

IBM DataStage

### CodeSandbox or Stackblitz example

N/A

### Steps to reproduce the issue (if applicable)

Since you need specific call back function implemented to see it, not able to share demo. I put video above though of this callback not working

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

AustinGitHub:
I am looking into fixing this. This is necessary for devs to be able to detect this change

AustinGitHub:
Ok, I see why it was like this when implemented, but on DataStage implementation, the reason we need this is because we are keeping track of the headers inside of an array so we need to know if the headers change order we need to update on our end for mapping purposes since user can manipulate the column metadata at any point so if the column headers are out of order then we can't accommodate user changes accordingly

PR #5621, opened 2024-06-27
fix(Dataspreadsheet): add callback function for after drag/drop ends

AustinGitHub:
Closes https://github.com/carbon-design-system/ibm-products/issues/5620

#### What did you change?

I added a callback method so user can know what new column headers are after drag/drop is done

Storybook

https://github.com/carbon-design-system/ibm-products/assets/10100397/00c0e6d4-42de-4b5d-aaed-fb34c6bc7b99


davidmenendez (changes requested):
code looks good. only thing i'd like to see changed is to rename the prop to something like `onColDrag` or something just to continue to maintain the `on` naming convention for handlers.

AustinGitHub:
Added the fix, thanks!

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
```diff
@@ -285,6 +294,46 @@ export let DataSpreadsheet = React.forwardRef(
...
+      }
+    }, [
+      previousState?.selectedHeaderReorderActive,
+      currentColumns,
+      headerCellHoldActive,
+      columns,
+      activeCellContent,
+      onColDrag,
+    ]);
+
```
amal-k-joy, line 336: Hey @AustinGitHub ,
Code looks fine, but I was thinking of this scenario. 
Let initial order of columns be `orderDefault`. 
When colums are dragged , order changed from `orderDefault` to `order 1` , then changed to `order 2`. Then again dragged back to `orderDefault`. At this last drag scenario , will this callback get triggers?
@davidmenendez Can you share your thoughts? Is this a required scenario ?

amal-k-joy (changes requested):

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
AustinGitHub: Let me check

AustinGitHub: You're right, it doesn't get triggered in this use case. I will add a fix for this

AustinGitHub:
https://github.com/carbon-design-system/ibm-products/assets/10100397/7624e79a-3189-4b0f-9b3c-dac3199671e2

Here is the fix

davidmenendez (approved):
@amal-k-joy good catch!

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
```diff
@@ -296,34 +297,41 @@ export let DataSpreadsheet = React.forwardRef(
...
 
       if (
         !previousState.selectedHeaderReorderActive &&
         !headerCellHoldActive &&
         currentHeaders.length > 0 &&
-        pastHeaders.length > 0 &&
-        JSON.stringify(currentHeaders) !== JSON.stringify(pastHeaders)
+        pastColumns.length > 0 &&
+        JSON.stringify(currentHeaders) !== JSON.stringify(pastColumns)
       ) {
```
amal-k-joy, line 318: Hey @AustinGitHub ,
Thanks for the changes and those looks good. But I am seeing many duplicate lines of code in the `if` block and `else if` block. Is it possible to do some refactor?

amal-k-joy (changes requested):

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
AustinGitHub: Sure

AustinGitHub: I fixed it, thanks!

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
```diff
@@ -285,6 +295,50 @@ export let DataSpreadsheet = React.forwardRef(
...
+        setPastColumns(currentHeaders);
+      }
+
+      if (
+        !previousState.selectedHeaderReorderActive &&
+        pastColumns.length > 0 &&
+        !headerCellHoldActive &&
+        JSON.stringify(currentHeaders) !== JSON.stringify(pastColumns)
+      ) {
+        if (currentHeaders.length > 0) {
```
amal-k-joy, line 318: Hey, can you check if we can modify line 318 to 330 as below block ?
```
 onColDrag({
    headers: currentHeaders,
    data: activeCellContent.props.data,
  });
  if (currentHeaders.length === 0){
    setPastColumns([]);
  }
```

amal-k-joy (changes requested):

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
AustinGitHub: Sure, checking

AustinGitHub: This works, thanks, just pushed the code change

amal-k-joy (approved):
Nice work Austin

davidmenendez (approved):

PR #5639, opened 2024-07-02
ci: fix AVT tests not running

### LINKED ISSUE #5638: Fix AVT runners
@ariellalgilmore and I noticed while reviewing some AVT PRs that not all of our AVT tests run. Each AVT runner runs just short of 20 tests before stopping. Carbon Core’s tests can run about 90+.

After some investigation, we think it was due to our original Playwright setup where we allocated only 2 shared initially but split our tests into 4 shards (likely based on Core’s initial CI).

https://github.com/carbon-design-system/ibm-products/blob/main/.github/workflows/ci.yml#L151
https://github.com/carbon-design-system/ibm-products/blob/main/.github/workflows/ci.yml#L122
```[tasklist]
### Tasks
```

elycheea:
Closes #5638 

#### What did you change?
Increase our shards from 2 to 4

Should see more tests in our CI now.
~But also assuming we’re going to see some failed tests that were previously hidden.~
- Change `SearchBar` `labelText` to required since it’s required by underlying Carbon `Search` — tests updated accordingly
- Update `Toolbar` docs to current format — probably not necessary but ended up doing it while trying to figure out why the story wasn’t being captured at first 👀


davidmenendez:
interesting. well at least we can see failing tests now 🤔

matthewgallo (approved):

PR #5640, opened 2024-07-03
feat(ConditionBuilder): handle keyboard navigation for tree variant

amal-k-joy:
Contributes #4916

- Handle keyboard navigation for tree variant
- Handle focus for close icon on removal of conditions

Local storybook

### DIFF
--- a/packages/ibm-products-styles/src/components/ConditionBuilder/_condition-builder.scss
+++ b/packages/ibm-products-styles/src/components/ConditionBuilder/_condition-builder.scss
@@ -49,7 +49,7 @@ $block-class: #{c4p-settings.$pkg-prefix}--condition-builder;
 }
 .#{$block-class}__condition-block {
   display: flex;
-  width: max-content;
+  width: fit-content;
   flex-direction: row;
 }
 .#{$block-class}__content-container {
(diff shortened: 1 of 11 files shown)


davidmenendez (changes requested):
i'm noticing some strange interactions when i try to navigate with keyboard. i don't know if there are any specs for how the keyboard interfaces with condition builder, so if some of this is intentional please let me know.

the primary interaction being that i would expect to move left and right with the left and right arrow keys or with tab / shift + tab, not necessarily the up and down keys. also when `if` is focused and you use left or right keys the focus seems to go down another level within the item, which doesn't seem to go anywhere.

i also notice that the entire page moves slightly when using any of the arrow keys for navigation.

amal-k-joy:
> https://deploy-preview-5640--carbon-for-ibm-products.netlify.app

@davidmenendez 
Keyboard navigation is implemented cordinating with Jan Hassel (UX Engineer).

For sentence variant we have used the [layout grid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/layout-grids/#).

For tree variant, we have used [tree grid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/), which is the closest matching pattern. Here we have rowgroup ==> rows ==> cells
Arrow up/down : If currently row is focussed, it will take you though next rows. If any cell is focussed, it will focus the same type cell in the next row. 
Arrow Right/Left : Navigate through the cells within a row.

Extra scroll present is removed. Also scrolling happening on arrow key interaction is fixed using `preventDefault()`.

I am happy  to have more discussion on this.

davidmenendez (approved):
looks great! thanks for taking a second look

makafsal (approved):
Looks good

PR #5647, opened 2024-07-08
feat(conditionBuilder): accessibility and generic issue fixes

amal-k-joy:
Contributes #4916 , #5640 

- Implementation to show content as '**incomplete**' when user close the popover without selecting any values
- Edge case issue when user delete all plain condition keeping only subgroups in a group
- Mange tab index and focus
- Accessibility fixes

Local storybook

### DIFF
--- a/packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderItem/ConditionBuilderItem.js
+++ b/packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderItem/ConditionBuilderItem.js
@@ -43,7 +43,6 @@ export const ConditionBuilderItem = ({
       (rest['data-name'] === 'propertyField' && property === 'INVALID') ||
       (rest['data-name'] === 'operatorField' && operator === 'INVALID')
     ) {
-      label = 'INVALID';
       return {
         propertyLabel: invalidText,
         isInvalid: true,
--- a/packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderItem/ConditionBuilderItem.js
+++ b/packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderItem/ConditionBuilderItem.js
@@ -1,4 +1,4 @@
-import React, { useState, useRef, useEffect } from 'react';
+import React, { useState, useRef, useEffect, useContext } from 'react';
 
 import { Popover, PopoverContent, Layer } from '@carbon/react';
 import PropTypes from 'prop-types';
@@ -9,6 +9,8 @@ import {
 } from '../ConditionBuilderContext/DataConfigs';
 import { ConditionBuilderButton } from '../ConditionBuilderButton/ConditionBuilderButton';
 import { useTranslations } from '../utils/useTranslations';
+import { ConditionBuilderContext } from '../ConditionBuilderContext/ConditionBuilderProvider';
+import { handleKeyDownForPopover } from '../utils/handleKeyboardEvents';
 
 export const ConditionBuilderItem = ({
   children,
@@ -22,28 +24,28 @@ export const ConditionBuilderItem = ({
   popOverClassName,
   config,
   renderChildren,
+  onChange,
   ...rest
 }) => {
-  const contentRef = useRef(null);
   const popoverRef = useRef(null);
   const [open, setOpen] = useState(false);
 
-  const [invalidText, addConditionText, labelText, invalidDateText] =
-    useTranslations([
-      'invalidText',
-      'addConditionText',
-      label,
-      'invalidDateText',
-    ]);
+  const [invalidText, addConditionText, labelText] = useTranslations([
+    'invalidText',
+    'addConditionText',
+    label,
+  ]);
+  const { conditionBuilderRef } = useContext(ConditionBuilderContext);
   const getPropertyDetails = () => {
-    if (label === 'INVALID') {
+    const { property, operator } = condition || {};
+    if (
+      label === 'INVALID' ||
+      (rest['data-name'] === 'propertyField' && property === 'INVALID') ||
+      (rest['data-name'] === 'operatorField' && operator === 'INVALID')
+    ) {
+      label = 'INVALID';
       return {
         propertyLabel: invalidText,
-        isInvalid: true,
-      };
-    } else if (label === 'INVALID_DATE') {
-      return {
-        propertyLabel: invalidDateText,
         isInvalid: true,
       };
     }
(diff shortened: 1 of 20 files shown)


packages/ibm-products/src/components/ConditionBuilder/ConditionGroupBuilder/ConditionGroupBuilder.js
```diff
@@ -45,12 +46,46 @@ const ConditionGroupBuilder = ({
...
-        ),
-      });
+
+      if (group.conditions[1].conditions) {
+        //when we remove every plain conditions of a group without deleting  the subgroup, we need to restructure the group
+        //the inner group become outer group and same level subgroups become plain conditions
+
+        let newGroup = [],
+          allConditions = [];
+        group.conditions.map((eachItem) => {
```
davidmenendez, line 56: instead of using `map` you can use `reduce` similarly to the way you do below to simplify this
```
const allConditions = group.conditions.reduce((acc, item) => {
  if (item.conditions) {
    return acc.concat(item.conditions);
  }
  return acc;
}, []);
```

packages/ibm-products/src/components/ConditionBuilder/ConditionGroupBuilder/ConditionGroupBuilder.js
```diff
@@ -45,12 +46,46 @@ const ConditionGroupBuilder = ({
...
+          if (eachItem.conditions) {
+            allConditions = [...allConditions, ...eachItem.conditions];
+          } //spreading out the condition inside the subgroup
+        });
+        //we always have conditions first and then subgroups, so ordering accordingly
+        const groupedItems = allConditions.reduce((acc, item) => {
```
davidmenendez, line 62: we can eliminate some redundant code by modifying the checks
```
const groupedItems = allConditions.reduce((acc, item) => {
  if (item.conditions) {
    if (!acc.groups) {
      acc.groups = [];
    }
    acc.groups.push(item);
  } else {
    if (!acc.conditions) {
      acc.conditions = [];
    }
    acc.conditions.push(item);
  }
  return acc;
}, {});
```

packages/ibm-products/src/components/ConditionBuilder/ConditionGroupBuilder/ConditionGroupBuilder.js
```diff
@@ -45,12 +46,46 @@ const ConditionGroupBuilder = ({
...
-        conditions: group.conditions.filter(
-          (condition) => conditionId !== condition.id
-        ),
-      });
+
+      if (group.conditions[1].conditions) {
+        //when we remove every plain conditions of a group without deleting  the subgroup, we need to restructure the group
+        //the inner group become outer group and same level subgroups become plain conditions
+
+        let newGroup = [],
```
davidmenendez, line 54: since `newGroup` isn't actually used till line 73 i would just move it's instantiation to there. `const newGroup = [ ... ]`

packages/ibm-products/src/components/ConditionBuilder/ConditionGroupBuilder/ConditionGroupBuilder.js
```diff
@@ -45,12 +46,46 @@ const ConditionGroupBuilder = ({
...
+            acc['groups'] = acc['groups'] || [];
+            acc['groups'].push(item);
+          } else {
+            acc['conditions'] = acc['conditions'] || [];
+            acc['conditions'].push(item);
+          }
+          return acc;
+        }, {});
+
+        newGroup.conditions = [
```
davidmenendez, line 73: if i'm reading this correctly it looks like this is just `allConditions`. if that's the case then `groupedItems` really inst necessary at all since you can just do `newGroup.conditions = [].concat(allConditions)`. although i would like to point out that you're using property assignment on an array here, which should most likely be done on an object instead.

davidmenendez (changes requested):
few comments for ya

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderItem/ConditionBuilderItem.js
```diff
@@ -22,30 +24,30 @@ export const ConditionBuilderItem = ({
...
+  const { conditionBuilderRef } = useContext(ConditionBuilderContext);
   const getPropertyDetails = () => {
-    if (label === 'INVALID') {
+    const { property, operator } = condition || {};
+    if (
+      label === 'INVALID' ||
+      (rest['data-name'] === 'propertyField' && property === 'INVALID') ||
+      (rest['data-name'] === 'operatorField' && operator === 'INVALID')
+    ) {
+      label = 'INVALID';
```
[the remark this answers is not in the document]
amal-k-joy: thanks to notify this. Resolved.

packages/ibm-products/src/components/ConditionBuilder/ConditionGroupBuilder/ConditionGroupBuilder.js
amal-k-joy: modified as advised

packages/ibm-products/src/components/ConditionBuilder/ConditionGroupBuilder/ConditionGroupBuilder.js
amal-k-joy: Did some overall refactor in the logic.

packages/ibm-products/src/components/ConditionBuilder/ConditionGroupBuilder/ConditionGroupBuilder.js
amal-k-joy: Actually its just `allConditions` . But allConditions will contain normal conditions as well as any subgroups. So we need rearrange `allConditions` in such a way that conditions first and then any subgroups. This is an edge scenario when user delete all conditions in a group. 

https://github.com/carbon-design-system/ibm-products/assets/153802538/4e1e3043-e79d-499e-8c70-512ffd151f92

packages/ibm-products/src/components/ConditionBuilder/ConditionGroupBuilder/ConditionGroupBuilder.js
amal-k-joy: resolved

davidmenendez (approved):
awesome!

makafsal (approved):

PR #5654, opened 2024-07-09
fix(Datagrid): only remove single filter for correct table id

### LINKED ISSUE #4522: DataGrid Filter issues

- [ ] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive). Yes

## Detailed description

1.   Create DataGrid along with Filter.
2.   Add a filter of type Checkbox.
3.   Add checkbox filter details as shown in the notes section.
4.   Select the checkbox filter from the filter flyout  and click on Apply.
5.   Now try to remove the filter with label Glass using the X icon  , this fails. ( As per my observation this happens when using different value for id and value attribute of the checkbox filter)

DataGrid Filter

On click of X , it should remove the selected filter.

Chrome and safari

> package are you using?
"@carbon/ibm-products": "2.30.0",
 "@carbon/react": "1.52.0",

https://codesandbox.io/p/sandbox/brave-poincare-2ffcfr?file=%2Fpackage.json%3A56%2C2

- Notes
  {
    type: "checkbox",
    column: "lastName",
    props: {
      FormGroup: {
        legendText: "Last Name",
      },
      Checkbox: [
        {
          id: "glass1",
          labelText: "Glass",
          value: "glass",
        },
        {
          id: "regret",
          labelText: "Regret",
          value: "regret",
        },
        {
          id: "current",
          labelText: "Current",
          value: "current",
        },
      ],
    },
  },

matthewgallo:
Closes #4522 

This fixes a bug where removing a filter tag when there are multiple instances of a Datagrid component it gets removed for both Datagrids. With this change, the filter tag will only be removed from the appropriate Datagrid.

This is similar to the fix that @annawen1 made with the "Clear filters" button but for the removal of single filter tags.

#### What did you change?
```
packages/ibm-products/src/components/Datagrid/Datagrid/DatagridContent.tsx
packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/FilterProvider.js
```
Storybook, temporarily created a new story that used `Tabs` to render two separate Datagrid components.


packages/ibm-products/src/components/Datagrid/Extensions/Filtering/Panel.stories.jsx
```diff
@@ -24,7 +26,7 @@ export default {
   parameters: {
     styles,
     docs: { page: DocsPage },
-    layout: 'fullscreen',
+    layout: 'padded',
```
elycheea, line 29: Guessing we want to keep these `fullscreen`?

matthewgallo: Yep, that was leftover from the tab example. I'll revert to `fullscreen`.

elycheea (approved):
😆

AlexanderMelox (approved):
This looks good to me 👍🏻

PR #5657, opened 2024-07-09
chore(lint-license): add lint checker

### LINKED ISSUE #4792: [lint-check]: feat add check for Copyright/license header
## What will this achieve?

We are currrently manually checking if the Copyright header is on every file when as part of the release review process, but instead we can check this with a linter automatically like we do in Dotcom: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/main/tasks/check-license.js

When introducing new code to the repository the check-license linter will run and will fail and not allow user to create a commit if there is no Copyright header at the top of every file

ariellalgilmore:
Closes #4792 

#### What did you change?

added a lint checker to see if a license was in the file and update the year when necessary 

ran the lint checks locally


davidmenendez (approved):
wow this is awesome!

scripts/license-text.cjs
```diff
@@ -0,0 +1,27 @@
+/**
+ * Copyright IBM Corp. 2020
```
elycheea, line 2: I’m guessing this should also be 2024?

packages/ibm-products/src/components/AboutModal/AboutModal.stories.jsx
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corp. 2020, 2024
+ * Copyright IBM Corp. 2024
```
elycheea, line 2: I expected this one to retain the first year as it does in ActionBar 🤔

scripts/license-text.cjs
ariellalgilmore: updated!

packages/ibm-products/src/components/AboutModal/AboutModal.stories.jsx
ariellalgilmore: updated! was testing in aboutmodal

packages/ibm-products/src/components/AboutModal/AboutModal.stories.jsx
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corp. 2024
+ * Copyright IBM Corp. 2020
```
elycheea, line 2:
```suggestion
 * Copyright IBM Corp. 2020, 2024
```

matthewgallo (changes requested):

matthewgallo (dismissed):
👏

davidmenendez (approved):

davidmenendez (approved):
woot

matthewgallo (approved):

Issue #4380, opened 2024-02-21
CreateTearSheet: Add 4th button option


vnikiten01:
## What will this achieve?

- enhancement

The 4th button will allow users an additional path to not submit their created resource but navigate to a summary page where current configurations can be viewed alongside charts to better visualize intended changes and make further modifications before submission.

### Scenario

Create flows ( ~10) with multiple steps, 1-4 required and 1-4 optional steps.  the optional steps exist as a way to guide a user into making additional but unnecessary configurations in the context of what they are creating.

With the create flow the user has 2 paths to take

#### Path 1

The user can submit the form - this will create the resource and make an api call to persist data and kick something of.

#### Path 2

The user can Go to Summary Configuration - this does not create resource but allows the user to see the configurations they have made alongside some widgets to present other useful data.  Its a playground of sorts where the user can make additional changes that were not part of the intended workflow first presented.  Allowing a level of customization that advanced users might want/need while more basic users have the more guided flows with optional steps they might gravity to using. 

## How will success be measured?

Measured by the ability to add additional action support to create tearsheet that otherwise already exists in the standalone [tearsheet component](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-tearsheet-tearsheet--tearsheet&args=actions:9)

![image](https://github.com/carbon-design-system/ibm-products/assets/8129587/f753e24a-b37e-4ce0-8dd4-a42ea8162bed)

### Previous Request

[Previous request](https://github.com/carbon-design-system/ibm-products/issues/3835) 

### Other potential solutions

Toggle in influencer to turn off optional steps that would allow the user to fast submit

https://github.com/carbon-design-system/ibm-products/issues/4353

Sreejith-CS:
I would like to contribute this enhancement. Please assign this issue to me.

PR #5666, opened 2024-07-15
fix(CreateTearsheet): add custom button

Sreejith-CS:
Closes https://github.com/carbon-design-system/ibm-products/issues/4380

Add custom (4th) button to `CreateTearsheet`

#### What did you change?

- Added an optional prop for custom button text to `CreateTearsheet`
   - `customButtonText?: string;`
- Added optional props for custom button to `CreateTearsheetStep`
   -  `customButtonChangeText?: string;`
   -  `isCustomButtonDisabled?: boolean;`
   -  `isCustomButtonHide?: boolean;`
   -  `onCustomButtonClick?: () => void;`
- Created storybook for custom button 
- Added unit test

- Manual tests via storyboard
- Added unit test

https://github.com/user-attachments/assets/dbb4b401-1216-4c01-9325-271152498515

unit_test


Sreejith-CS:

davidmenendez (commented):
thank you for contributing. looks good to me!

i just want to double check with @sstrubberg on this. this was in `icebox` but didn't have any label for `needs design opinion`. just want to verify that this is fine on the design side before merging.

sstrubberg:
> thank you for contributing. looks good to me!
> 
> i just want to double check with @sstrubberg on this. this was in `icebox` but didn't have any label for `needs design opinion`. just want to verify that this is fine on the design side before merging.

Thanks for checking in! @oliviaflory @kingtraceyj @aubrey-oneal Can ya'll weigh-in here from a UX/Visual pov?

Sreejith-CS:
> > thank you for contributing. looks good to me!
> > i just want to double check with @sstrubberg on this. this was in `icebox` but didn't have any label for `needs design opinion`. just want to verify that this is fine on the design side before merging.
> 
> Thanks for checking in! @oliviaflory @kingtraceyj @aubrey-oneal Can ya'll weigh-in here from a UX/Visual pov?

It would be great if get a review and approval

davidmenendez:
@ljcarot can you weigh in?

oliviaflory:
@davidmenendez @Sreejith-CS I noticed that the `Skip` button doesn't actually take the user to the next step in the Create flow. 
I'm wondering if that is intentional? Is the 4th button just intended for demo purposes and an adopter would have to add the logic to go to the next step? From the original issue I am not 100% sure on the use case and UX intention

oliviaflory (dismissed):
@Sreejith-CS Thank you for your contribution!

I have a few changes that will need to be made in order to merge this into our library

1. Please remove the story `Create tearsheet with custom button`. It would be best to mirror the existing 4 button approach that is in the Tearsheet and showcase the 4th button in a control as seen here
 
Screenshot 2024-07-22 at 5 49 58 PM


2. The `Skip` button should be between the `Back` and `Replace` buttons (the same as the tear sheet examples)

Screenshot 2024-07-22 at 5 50 21 PM


3. I would like to see the new button be functional for the purposes of the demo. From the original issue it seemed that `Skip` should advance the user in the Create flow and move them onto the next step in the series of the navigation

Sreejith-CS:
@oliviaflory Thanks for the review. Consider the following points.

This change is to create a custom button on the `CreateTearsheet`. So let's add custom text to the button label. If there is a button labeled `Skip`, our requirement is to skip all the remaining steps and it will act as the submit button. If the custom button behaves like the next button, there's no use specifying it, because `Next` and `Skip` will do the same behaviour.

oliviaflory:
> @oliviaflory Thanks for the review. Consider the following points.
>
>This change is to create a custom button on the CreateTearsheet. So let's add custom text to the button label. If there is a button labeled Skip, our requirement is to skip all the remaining steps and it will act as the submit button. If the custom button behaves like the next button, there's no use specifying it, because Next and Skip will do the same behaviour.

@Sreejith-CS that is fair, I would advise your UX designer to consider whether skipping all remaining steps and submitting is really clear to the user when a button is labeled "Skip", but will not block the addition.

Revised request:

1. Please remove the story `Create tearsheet with custom button`. It would be best to mirror the existing 4 button approach that is in the Tearsheet and showcase the 4th button in a control as seen here

`actions` prop with options to select the buttons:
- Three buttons with one ghost
- Four buttons with one ghost
 
Screenshot 2024-07-22 at 5 49 58 PM


2. The `Custom action` button should be between the `Back` and `Replace` buttons (the same as the tear sheet examples)

Screenshot 2024-07-22 at 5 50 21 PM

Sreejith-CS:
@oliviaflory Please consider the following points.

> @Sreejith-CS that is fair, I would advise your UX designer to consider whether skipping all remaining steps and submitting is really clear to the user when a button is labeled "Skip", but will not block the addition.
> 

We can add label like `Skip all steps'.

> Revised request:
> 
> 1. Please remove the story `Create tearsheet with custom button`. It would be best to mirror the existing 4 button approach that is in the Tearsheet and showcase the 4th button in a control as seen here
> 
> `actions` prop with options to select the buttons:
> 
> * Three buttons with one ghost
> * Four buttons with one ghost
> 

We cannot display the custom (4th) button (as a dropdown option) because there are no `actions` props in the CreateTearsheet. So I think we should keep the story of `Create tearsheet with custom button`.  So that user can understand each of the props (`customButtonText , customButtonChangeText, isCustomButtonDisabled, isCustomButtonHide, onCustomButtonClick`). Controls have option `customButtonText` so we can show custom button.


https://github.com/user-attachments/assets/3582dbca-0239-4575-8b4c-404def1d6ada



> 2. The `Custom action` button should be between the `Back` and `Replace` buttons (the same as the tear sheet examples)
> 
This has been implemented.

packages/ibm-products/src/components/CreateTearsheet/CreateTearsheet.test.js
```diff
@@ -79,6 +82,10 @@ const renderCreateTearsheet = ({
   onPrevious = onPreviousStepFn,
   finalOnNextFn = finalStepOnNext,
   rejectOnSubmitNext = false,
+  isCustomButtonDisabled = false,
+  isCustomButtonHide = false,
```
davidmenendez, line 86: for these i think it's better to go with the opposite naming convention. so like `customButtonEnabled` and `customButtonVisible`

packages/ibm-products/src/components/CreateTearsheet/preview-components/MultiStepWithCustomButton.js
```diff
@@ -0,0 +1,103 @@
...
+import { Button } from '@carbon/react';
+import cx from 'classnames';
+import { pkg } from '../../../settings';
+import { CreateTearsheet } from '../CreateTearsheet';
+import { CreateTearsheetStep } from '../CreateTearsheetStep';
+import { SlugSample } from '../../../global/js/story-parts/slug';
+
+const blockClass = `${pkg.prefix}--tearsheet-create-multi-step`;
+
+export const MultiStepWithCustomButton = ({
```
davidmenendez, line 19: after some discussion we decided that we didn't want to highlight this functionality too much. instead of giving it it's own story could you move it to a control similar to the [4 button layout in regular tearsheet](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-tearsheet--tearsheet&args=actions:9&globals=viewport:basic)?

davidmenendez (changes requested):

packages/ibm-products/src/components/CreateTearsheet/CreateTearsheet.test.js
elycheea: I think this one probably should be `customButtonDisabled` since it’s about the disabled state. 🤔

matthewgallo:
Hey @Sreejith-CS! Thanks for all your hard work here. But I think we need to re-evaluate the approach taken here as it adds ~5 new props and is not a composable approach that we'd ideally like to see for the `CreateTearsheet` actions. Our team will add more detailed acceptance criteria to #4380 and move it from the "Icebox" when ready.

Sreejith-CS:
@matthewgallo I have updated the code as per discussion and review comments. Please review it

Sreejith-CS:
@matthewgallo @elycheea @davidmenendez It would be great if get a review and approval

matthewgallo:
Thanks for working on these changes @Sreejith-CS! We'll work on reviewing this again 🎉

matthewgallo (dismissed):
Thanks for including the changes we discussed @Sreejith-CS!

Sreejith-CS:
Thanks @matthewgallo 

@davidmenendez @oliviaflory Could you please review

elycheea (commented):
Hey @Sreejith-CS!

@davidmenendez @matthewgallo @oliviaflory and I had some additional discussion on the prop(s) naming and wanted to propose one more change to the naming —
instead of `experimentalAlternateSubmit` we want to suggest `experimentalSecondarySubmit`.

I know we originally discussed the former as a possible name, but on further review think it could also be interpreted as an alternative to the existing submit rather than additional. We believe the `experimentalSecondarySubmit` prefix would be a bit clearer than `experimentalAlternateSubmit`.

Lastly, we would still like to remove the story before we bring this change in. While it’s helpful for us to see and review it, I do think it’s not necessarily a feature we’re currently encouraging, especially as an experimental option.

Sreejith-CS:
> Hey @Sreejith-CS!
> 
> @davidmenendez @matthewgallo @oliviaflory and I had some additional discussion on the prop(s) naming and wanted to propose one more change to the naming — instead of `experimentalAlternateSubmit` we want to suggest `experimentalSecondarySubmit`.
> 
> I know we originally discussed the former as a possible name, but on further review think it could also be interpreted as an alternative to the existing submit rather than additional. We believe the `experimentalSecondarySubmit` prefix would be a bit clearer than `experimentalAlternateSubmit`.
> 
> Lastly, we would still like to remove the story before we bring this change in. While it’s helpful for us to see and review it, I do think it’s not necessarily a feature we’re currently encouraging, especially as an experimental option.

@davidmenendez @matthewgallo @oliviaflory @elycheea  Thanks for the review. I have updated according to all your review comments. Could you please review again.

elycheea (approved):

Issue #5726, opened 2024-07-25
[Feature Request]: Disable Modal for TagSet component


jeesonjohnson:
### The problem

Currently, we do not have the option to disable the TagSet overflow modal. This is currently an issue for our product, since we chose to disable this component, in favour of our own custom component with custom functionality. However, with how the code is set up currently, the inbuilt `TagSetModal` cannot be disabled.  See [here](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products/src/components/TagSet/TagSet.tsx#L402-L412). We have disabled it manually using CSS overrides. However, we would rather it be disabled by the use of a prop, so that the DOM does not become overly populated unnecessarily.
 

### The solution

Pass down a prop called, `showOverflowModal` (defaulted to `true`) that will disable the rendering of the component within the `TagSet` component. 

NOTE: I am happy to do this change, raise the appropriate PRs and do related testing. Just wanted to approval before attempting to make this change.

### Design link

N/A

### Links to other materials

Related code that needs to be changed: https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products/src/components/TagSet/TagSet.tsx#L402-L412

### Owner/main maintainer(s)

IBM Products

### Product/offering

IBM Cloud Pak For Integration

### Business priority

Low Priority = release date is not dependent on fix or not upcoming

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

oliviaflory:
@matthewgallo following up with Michael G

jeesonjohnson:
Hey @matthewgallo / @oliviaflory, has there been any update on this?

PR #5753, opened 2024-07-29
feat: adding the ability to disable modals for TagSet component

jeesonjohnson:
Closes https://github.com/carbon-design-system/ibm-products/issues/5726

Added the ability to disable the overflow modal for additional tags. This gives consumers the customisability to use their component to manage and view tags.

#### What did you change?
`packages/ibm-products/src/components/TagSet/TagSet.stories.jsx`
`packages/ibm-products/src/components/TagSet/TagSet.test.js`
`packages/ibm-products/src/components/TagSet/TagSet.tsx`
`packages/ibm-products/src/components/TagSet/TagSetOverflow.tsx`

Added additional unit tests

Visual testing:

https://github.com/user-attachments/assets/04d31bcf-538a-4268-a86d-35b0353f0648


jeesonjohnson:

jeesonjohnson:
recheck

packages/ibm-products/src/components/TagSet/TagSet.tsx
```diff
@@ -161,6 +165,7 @@ export let TagSet = React.forwardRef<HTMLDivElement, TagSetProps>(
       allTagsModalSearchLabel,
       allTagsModalSearchPlaceholderText,
       showAllTagsLabel,
+      disableOverflowPopup = false,
```
vladbalanescu, line 168: this can have an undefined value, since !undefined will be false anyway, in the render beneath.

packages/ibm-products/src/components/TagSet/TagSetOverflow.tsx
```diff
@@ -87,6 +91,7 @@ export const TagSetOverflow = React.forwardRef(
       // The component props, in alphabetical order (for consistency).
 
       allTagsModalSearchThreshold = defaults.allTagsModalSearchThreshold,
+      disablePopOver = false,
```
vladbalanescu, line 94: same goes here, but that's just a suggestion, not mandatory change!

vladbalanescu (approved):
LGTM

packages/ibm-products/src/components/TagSet/TagSet.tsx
jeesonjohnson: I think this is fine in this case, undefined would only be an issue if the user explicitly defines it as undefined. And since undefined is often associated with false, it wont be an issue in this case.

packages/ibm-products/src/components/TagSet/TagSetOverflow.tsx
jeesonjohnson: See above.

matthewgallo:
Hey @jeesonjohnson, someone will review this PR after the team has triaged the related issue (https://github.com/carbon-design-system/ibm-products/issues/5726). Thanks for your patience! 😄

jeesonjohnson:
Sounds good, thanks for the update @matthewgallo!

davidmenendez:
@jeesonjohnson can you please resolve the merge conflicts? thanks!

jeesonjohnson:
Hey @davidmenendez, gone ahead and resolved the merge conflicts.

devadula-nandan:
Hey @jeesonjohnson,

After these changes, I am wondering how you could set the onClick on the overflow tag to show your custom modal. 

You still need to access the dom directly, get the overflow tag, and attach an onClick event listener. which isn't a good approach in a react-handled application. also, it could be a mess targeting a tagset if there are multiple of them on the page.

instead of introducing `disablePopOver` we could introduce something like `onOverflowClick` like below
```
<TagSet tags={tags} onOverflowClick={customOverflowHandler} />
```
and in TagSetOverflow.jsx
```
...
  const handleOverflowClick = () => {
    if (onOverflowClick) {
      onOverflowClick(overflowTags);  // Invoke the custom handler
    } else {
     setPopoverOpen?.(!popoverOpen) // Opens default popover
    }
  };
...
   <Tag onClick={handleOverflowClick} className={cx(`${blockClass}__popover-trigger`)} >
...
```
Now if `onOverflowClick` is not passed, we can see the default behavior, and if passed, will invoke the custom handler.
which would give better control to the user.

cc: @davidmenendez any second thoughts/suggestions on this?

devadula-nandan:
@jeesonjohnson can you fix the spell checks?

jeesonjohnson:
@davidmenendez Done! :)

packages/ibm-products/src/components/TagSet/TagSet.tsx
```diff
@@ -499,6 +508,10 @@ TagSet.propTypes = {
    * display tags in multiple lines
    */
   multiline: PropTypes.bool,
+  /**
+   * Changes the behavior that occurs on selecting the overflow tag component
```
davidmenendez, line 512: i think we need to change this to be a little more specific because i don't think it clearly describes what happens when you use this prop. maybe something like `an optional click handler that replaces the default display all tags modal functionality`

packages/ibm-products/src/components/TagSet/TagSetOverflow.tsx
```diff
@@ -51,6 +51,10 @@ interface TagSetOverflowProps {
    * className
    */
   className?: string;
+  /**
+   * changes the behavior of clicking the overflow tags button
```
davidmenendez, line 55: same comment

packages/ibm-products/src/components/TagSet/TagSetOverflow.tsx
```diff
@@ -133,6 +138,39 @@ export const TagSetOverflow = React.forwardRef(
       }
     };
 
+    const handleOverflowClick = () => {
+      // If a custom overflow function is provided then trigger that function
+      // on clicking the overflow
+      if (onOverflowClick) {
+        onOverflowClick(overflowTags);
+      } else {
+        setPopoverOpen?.(!popoverOpen);
```
davidmenendez, line 147: i think this was an oversight on our end, but can you do me a favor and make `setPopoverOpen` to be a required prop? it shouldn't be optional.

packages/ibm-products/src/components/TagSet/TagSetOverflow.tsx
```diff
@@ -133,6 +138,39 @@ export const TagSetOverflow = React.forwardRef(
...
+          }
+          aria-hidden={overflowTags.length === 0}
+          className={cx(`${blockClass}`, {
+            [`${blockClass}--hidden`]: overflowTags.length === 0,
+          })}
+          ref={ref || localRef}
+        >
+          <OperationalTag
+            onClick={() => handleOverflowClick()}
+            className={cx(`${blockClass}__popover-trigger`)}
```
davidmenendez, line 166: usage of `cx` isn't necessary here

davidmenendez (changes requested):
thanks for all your hard work! just a few things i noticed

packages/ibm-products/src/components/TagSet/TagSetOverflow.tsx
jeesonjohnson: Sure!

packages/ibm-products/src/components/TagSet/TagSet.tsx
jeesonjohnson: Ive updated all the relevant descriptions accordingly. Thanks for the suggestion.

jeesonjohnson:
Thanks @davidmenendez. I've gone ahead and addressed the changes requested. Mind doing another review please.

jeesonjohnson:
Thanks for @davidmenendez . I have gone ahead and addressed all the comments. Mind having another look please?

davidmenendez (approved):
LGTM

devadula-nandan (approved):
LGTM

PR #5791, opened 2024-08-01
fix(PageHeader): correct conditional for non zero offsets

### LINKED ISSUE #4427: Pageheader is not reliably calculating top offset

- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

- The page header component is ignoring our UI shell product header as a result of a recent change. This is causing the header to collapse behind the UI shell header rather than under it.
- Last version I could reliably test that had a properly working header was v2.19.2. The next version I could upgrade to was v2.26.0.
- I believe the change https://github.com/carbon-design-system/ibm-products/pull/4068 in v2.24.0 was were the breaking change was introduced.

- Pageheader

> see changed?

- Expect the pageheader to accurately calculate the top offset based on any components above it.
- Currently the pageheader is ignoring a UI shell header component

- Chrome

> package are you using?

- `"@carbon/ibm-products": "2.26.0",`

> should be aware of?

CP4AIOps. Multiple squads are reporting similar behavior using the same Pageheader components.

>
- https://codesandbox.io/p/devbox/frosty-hoover-d2r9md?file=%2Fsrc%2FExample%2F_example.scss%3A12%2C27

## Additional information

- Screenshots or code
- Notes

richardpilot:
Closes #4427

`headerOffset` is used to ensure the PageHeader works correctly if the component is not at the top of the viewport. The conditional for calculating a header offset was incorrect.

#### What did you change?

Previously it checked whether the offset value was zero. This could only be true if the header was already at the top of the viewport or it could not obtain a reference to the `offsetTopMeasuringRef`. In either case there would be no need to calculate the offset anyway.

By inverting the conditional, headerOffset will now be calculated (i.e. set to non-zero) if the offset component used to identify the PageHeader's position within the viewport if actually offset from the top.

### DIFF
--- a/packages/ibm-products/src/components/PageHeader/PageHeaderUtils.js
+++ b/packages/ibm-products/src/components/PageHeader/PageHeaderUtils.js
@@ -86,7 +86,7 @@ export const utilCheckUpdateVerticalSpace = (
     // The header offset calculation is either going to work out at 0 if we have no gap between scrolling container
     // top and the measuring ref top, or the difference between.
     update.headerOffset =
-      offsetMeasuringTop === 0
+      offsetMeasuringTop !== 0
         ? offsetMeasuringTop - scrollableContainerTop
         : 0;
 


richardpilot:

recheck

richardpilot:
recheck

davidmenendez (approved):
thanks for contributing this! works for me 👍

davidmenendez:
recheck

matthewgallo (approved):

PR #5792, opened 2024-08-01
feat(RemoveModal): update body to use ReactNode

### LINKED ISSUE #2289: Enhancement: change RemoveModal prop type for body
# Name of component or pattern
[RemoveModal](https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-patterns-remove-removemodal--standard)
## Brief description
The idea is to make the `body` prop able to accept node as value instead of just "string" in order to allow adding more detailed texts to the component. It doesn't break if passing `ReactNode` component, however it causes warning messages about the prop type.

## Example:
This is an example of the a `RemoveModal` that receives as `body` prop the following HTML:
```jsx
  const Body = () => (
    <>
      <span>
        Removing bx1001 will permanently remove the configuration.
      </span>
      <span>
        <strong>Note: </strong> This action cannot be undone.
      </span>
    </>
  );
``` 
RemoveModal_Example

## Details
[Link to example or existing component - RemoveModal](https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-patterns-remove-removemodal--standard)

adamlsmith981:
Closes #2289 

Update the Remove Modal to use ReactNode

#### What did you change?
Changed the typing for the body from `string` to `ReactNode`. 

- Added a test to check that a `Link` react node exists in the modal when added. 
- Added a new story in storybook which includes a `Link` react node in the body of text, linking to the carbon design homepage. Note: the naming of the story might want some work, and for my use case, I am looking at using Links, but are there other potential components that should be tested here?

### DIFF
--- a/packages/ibm-products/src/components/RemoveModal/RemoveModal.stories.jsx
+++ b/packages/ibm-products/src/components/RemoveModal/RemoveModal.stories.jsx
@@ -6,7 +6,7 @@
  */
 
 import React, { useState } from 'react';
-import { Button } from '@carbon/react';
+import { Button, Link } from '@carbon/react';
 // import styles from './_storybook-styles.scss?inline'; // import index in case more files are added later.
 import { RemoveModal } from '.';
 import DocsPage from './RemoveModal.docs-page';
@@ -82,4 +82,18 @@ export const DeletePattern = Template.bind({});
 DeletePattern.args = {
   ...defaultProps,
   textConfirmation: true,
+};
+
+export const ComponentInBodyPattern = Template.bind({});
+ComponentInBodyPattern.args = {
+  ...defaultProps,
+  body: (
+    <React.Fragment>
+      {`Before removing ${resourceName}, you can find out more information on the `}
+      <Link href={'https://www.carbondesignsystem.com'}>
+        Carbon Design System
+      </Link>
+      {' website.'}
+    </React.Fragment>
+  ),
 };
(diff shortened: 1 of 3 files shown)


matthewgallo:
Hey @adamlsmith981, thanks for the contribution! It looks like the email associated with your commit does not match the one on your GitHub account. I think the DCO should work once you add your IBM address to your public GitHub account! (Although you might still need to comment recheck again once you add it.)

adamlsmith981:
recheck

devadula-nandan (approved):
Hi @adamlsmith981 ,
Thanks for the contribution 🎉
the code looks good to me.
just a thought
after referring stories that have links, i think we haven't given any external links in our stories.

devadula-nandan:
> @devadula-nandan Hey! Thanks for the observation, just to clarify, shall I remove the href attribute so that the link does not proceed to go to another url?
> 
> That being said, I think I took inspiration from another component that used the Carbon Design System home page as the target URL. For example: https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-patterns-empty-state-emptystate--with-link&globals=viewport:basic

i feel its fine, lets see it in the second review.

adamlsmith981:
recheck

davidmenendez (approved):

davidmenendez:
approved. thanks @adamlsmith981

Issue #5782, opened 2024-07-31
Investigate `ConditionBuilder` test causing timeout issues when running in actions


matthewgallo:
### Package

Carbon for IBM Products

### Description

There is a `ConditionBuilder` test that intermittently causes timeout issues when running in github actions. This causes release actions such as `rc` and `full minor` actions to fail.

https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.test.js#L151-L177

https://github.com/carbon-design-system/ibm-products/actions/runs/10185091190/job/28173861841

### Component(s) impacted

`ConditionBuilder`

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

Latest

### Suggested Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Product/offering

Our team

### CodeSandbox or Stackblitz example

N/A

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

amal-k-joy:
@matthewgallo , it seems the test is not failing anymore after trying multiple times. Also I couldnt any network interaction for any of the tests in condition builder.
 However there can be a possible network interaction by `accessibility-checker` package while running the test '`it('has no accessibility violations'`)'
 Also in local, below error is thrown sometimes , how ever yarn test is success even with this error. So not sure the intermittent  test failure is connected to this error.
` [ERROR] ValidPoliciesMissing: No valid policy has been provided. Valid policy ids for the specified archive are: EXTENSIONS,IBM_Accessibility,IBM_Accessibility_next,WCAG_2_2,WCAG_2_1,WCAG_2_0`

_Note: a funtion in ACHelper.ts inside accessibility-checker is receiving curPolicy as `Custom_Ruleset`(from config file achecker.js) , whereas it expects one of `EXTENSIONS,IBM_Accessibility,IBM_Accessibility_next,WCAG_2_2,WCAG_2_1,WCAG_2_0`_

amal-k-joy:
@matthewgallo , Should we keep this issue open or can be closed for now?

matthewgallo:
[...]

PR #5794, opened 2024-08-01
fix(Conditionbuilder): add await for accessibility test

amal-k-joy:
Contributes #5782

Skip the condition builder tests for now to pass CI checks

#### What did you change?
changed `describe` to  `xdescribe`
yarn test


matthewgallo (changes requested):
Can you use `it.skip()` so that we don't skip all of the tests (see [here](https://github.com/carbon-design-system/ibm-products/pull/5781/files#diff-c8d80f7bd755807c19bd99953c9566a305dce88b549a4d795f99593ab3660ec7R151))? It only ever seems to be one test that times out

matthewgallo (approved):

devadula-nandan (approved):

PR #5798, opened 2024-08-01
fix: datagrid hidden virtual body width

### LINKED ISSUE #4579: Error with DataGrid and virtual scrolling (table blank if hidden and then shown)
Problem when DataGrid is hidden (in a tab) and then shown.  When using normal pagination, this works fine. However, if I switch to using virtual scrolling, when the tab is visited, the table is blank. I see the tool bar, but no table contents.
The culprit is handleVirtualGridResize in DataGridVirtualBody.js. It is setting the width on the virtual body container to the clientWidth property, which happens to be 0, when the table is in the non-active tab. Once I switch tabs to the one with the table, since it has a width of 0px, clientWidth remains 0, and so its stuck. 

```
  var handleVirtualGridResize = function handleVirtualGridResize() {
    var gridRefElement = gridRef === null || gridRef === void 0 ? void 0 : gridRef.current;
    gridRefElement.style.width = gridRefElement === null || gridRefElement === void 0 ? void 0 : gridRefElement.clientWidth;
  };
```

- [ x] Carbon for IBM Products (or Carbon for IBM Cloud & Cognitive)

Tested in both FF and Chrome

Using IBM Products version  "2.32.0"

davidmenendez:
Closes #4579 

fixes a bug where a datagrid using infinite scrolling when hidden by something like a carbon `Tab` the width was getting set to 0. this fix simply checks to make sure the datagrid is on screen before applying a new width

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridVirtualBody.tsx
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridVirtualBody.tsx
@@ -47,7 +47,9 @@ const DatagridVirtualBody = (datagridState: DataGridState) => {
   /* istanbul ignore next */
   const handleVirtualGridResize = () => {
     const gridRefElement = gridRef?.current;
-    if (gridRefElement) {
+    // isHidden checks for an edge case where if the table is hidden by something like a Tab that the width can't be calculated to 0
+    const isHidden = gridRefElement?.offsetParent === null;
+    if (gridRefElement && !isHidden) {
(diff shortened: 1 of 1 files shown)


davidmenendez:
@matthewgallo since you're the DG guru i figure i'll tag you to take a look 😂

davidmenendez:
the documentation doesn't mention it specifically, but i've confirmed locally that `hidden` does return `null` when using `offsetParent` https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent

AlexanderMelox (approved):
LGTM

Issue #5799, opened 2024-08-01
DataSpreadsheets: Dragdrop with custom header shows weird callback 


AustinGitHub:
On our code we get this crash 

![Screenshot 2024-08-01 at 2 45 49 PM](https://github.com/user-attachments/assets/60b56756-9772-4e85-9a53-56f05a334ded)

But I replicated this on carbon products code and I don't see this crash, instead I see this 

![Screenshot 2024-08-01 at 2 47 59 PM](https://github.com/user-attachments/assets/ffe1480b-88fd-4426-b52a-4eb38e5e9f67)

So I am not sure why on our code we see it crash, possibly due to how it gets imported as a package. 

To resolve this, I think I need to check the type for the header and if it is an object maybe user can add specific key to where the header is at so something like 
`<span headerLocated={true}>COLUMN_1 </span>`

Then we take the inner text of this as the current column name. Any other alternatives, feel free to suggest but I will work on this change as drag/drop on our code currently crashes (but doesn't crash on storybook when replicated)

AustinGitHub:
Making note: We are looking at next 2 weeks to wrap up the project we are working on, so the sooner the better on this item would be helpful for us, thanks! I'll be active on any PR comments I get. Thank you!

fyi @matthewgallo

AustinGitHub:
Will validate once we get new release, thanks!

AustinGitHub:
Verified this is fixed, thanks!

PR #5800, opened 2024-08-01
fix: data spreadsheet object drag drop

AustinGitHub:
Closes # https://github.com/carbon-design-system/ibm-products/issues/5799

#### What did you change?

I added a check for if the type of header is an object and to check for specific key property set to true `headerKey` to pick up this text value as the header value otherwise in depth json.stringify does not work properly 

Storybook

https://github.com/user-attachments/assets/4c2f8abe-d3c2-48c4-878a-c4a9b1eb6523


packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
```diff
@@ -295,11 +295,40 @@ export let DataSpreadsheet = React.forwardRef(
       [cellEditorValue, onDataUpdate]
     );
 
+    const getHeaderKeyText = useCallback((element: any) => {
```
matthewgallo, line 298: If this function just tries to get the text of a given node, we have a utility for that which I think would make most sense to use here, `getNodeTextContent` (`packages/ibm-products/src/global/js/utils/getNodeTextContent`).

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
```diff
@@ -295,11 +295,40 @@ export let DataSpreadsheet = React.forwardRef(
...
+      if (!element || typeof element !== 'object') {
+        return null;
+      }
+
+      if (element.type === 'span' && element.props.headerKey === true) {
```
matthewgallo, line 303: How come you check explicitly for `span` here?

AustinGitHub:
Thanks, I will address the comments today

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
AustinGitHub: Hi Matthew, I took a look at this and this returns back the entire text content but in our case we have a new line like this where the header name is actually `COLUMN_1` and below it is sort of like a subheader we provide to it but in the context of wanting to know the header text it does not matter to us. 

Screenshot 2024-08-02 at 1 32 00 PM

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
AustinGitHub: I was following how we had our setup with, will change this out so it isn't specific, thanks!

AustinGitHub:
Hi @matthewgallo is there any update on this? Thanks!

packages/ibm-products/src/components/DataSpreadsheet/DataSpreadsheet.tsx
```diff
@@ -295,11 +295,40 @@ export let DataSpreadsheet = React.forwardRef(
       [cellEditorValue, onDataUpdate]
     );
 
+    const getHeaderKeyText = useCallback((element: any) => {
+      if (!element || typeof element !== 'object') {
+        return null;
+      }
+
+      if (element.props.headerKey === true) {
```
matthewgallo, line 303:
```suggestion
      if (element.props.headerKey) {
```

AustinGitHub: fixed, thanks!

AustinGitHub:
Screenshot 2024-08-14 at 9 37 10 AM

With provided feedback. Also noting `column_name` is not required. I don't see anywhere in code to really mention this as even things like `placement` or `sticky` on the columns properties aren't fully defined somewhere so I think implementation for this is ok.

I also added safety net so if user didn't define `column_name` but the header is an object, we will use the `getNodeTextContent` function.

AustinGitHub:
We have 1 more week for our deadline on our project, would love to be able to get these in and a version released with them if possible. That way we can implement them on our UI as soon as possible, thanks!

matthewgallo (approved):

AustinGitHub:
Hi @annawen1 can you review when you get the chance? Thanks!

davidmenendez (approved):

AustinGitHub:
@davidmenendez can you add this back to merge queue please? Thanks!

AustinGitHub:
Is there anything left for this? Thanks!

AustinGitHub:
Not sure why tests are failing, they passed before and nothing major changed :(

matthewgallo:
We're adding support for visual regression testing that isn't fully configured yet. Everything looks good here, we can merge this now. Thanks for you patience with this one Austin!

AustinGitHub:
Thanks!

Issue #5722, opened 2024-07-24
Rename all component `index.js` files to `index.ts`


matthewgallo:
For example:
```
packages/ibm-products/src/components/AboutModal/index.js
```
will become
```
packages/ibm-products/src/components/AboutModal/index.ts
```

PR #5807, opened 2024-08-05
feat(typescript): exports all types in main index

### LINKED ISSUE #5723: Once all components are exported as .ts, update the root `index.js` to `index.ts`, and ensure all type exports are made available from the root - e.g. `import { type AboutModalProps } from '@carbon/ibm-products';`

AlexanderMelox:
Closes #5723 

⚠️ Depends on #5722 to be merged

This PR exports all the components and types in typescript. I followed what core Carbon is doing with their `index.ts`.
https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/index.ts

#### What did you change?
`+ packages/ibm-products/src/components/index.ts`

Manually check if i can import into another file.


packages/ibm-products/src/components/index.ts
```diff
@@ -0,0 +1,86 @@
...
+export * from './InterstitialScreenViewModule';
+export * from './DelimitedList';
+export * from './Decorator';
+export * from './DecoratorLink';
+export * from './DecoratorSingleButton';
+export * from './DecoratorDualButton';
+export * from './DescriptionList';
+export * from './FullPageError';
+export * from './SearchBar';
+export * from './Nav';
```
matthewgallo, line 71: There are some [ci issues](https://github.com/carbon-design-system/ibm-products/actions/runs/10321685255/job/28575148442?pr=5807).

If we look at this export for example, this will export everything from `Nav/index.ts`. Previously we only exported `Nav` but now `NavItem` and `NavList` are also exported. This is what we want (and is actually a bug previously as these should be exported), but it also means those components need to be added `package-settings.js` and also need to include the `pkg.checkComponentEnabled` check in the component.

matthewgallo (changes requested):

packages/ibm-products/src/components/index.ts
AlexanderMelox: I see, so NavList and NavItem should be exported?

AlexanderMelox: I exported them in the latest changes

packages/ibm-products/src/components/Nav/NavItem.js
```diff
@@ -52,7 +52,7 @@ export let NavItem = ({
 }) => {
   const [hrefHasDifferentHost, setHrefHasDifferentHost] = useState(false);
   useEffect(() => {
-    if (href.indexOf(window.location.host) === -1) {
+    if (href?.indexOf(window.location.host) === -1) {
```
AlexanderMelox, line 55: This was failing tests, so added an optional chaining to it

annawen1 (approved):

matthewgallo (approved):

PR #5817, opened 2024-08-06
feat(typescript): update generate scripts to use typescript

### LINKED ISSUE #4627: Update generate component script to use TypeScript
## What will this achieve?

When we create new components, there is a generate component script that can be run to scaffold the correct setup of files, configuring package-settings, exporting the new component. This is an issue to think about when the right time is to have the actual component itself moved from `.js` to `.tsx` file.

I don't think we need to set a timeline on this for now, but wanted to make sure it's something we plan for later on.

AlexanderMelox:
Closes #4627 

⚠️ Depends on https://github.com/carbon-design-system/ibm-products/pull/5807 to be merged

This PR updates the generate script to output Typescript components when generating new components. I also refactor the generate script itself to add the typescript props exports to `index.ts`. `index.ts` does not exist on this PR but it is being added in https://github.com/carbon-design-system/ibm-products/pull/5807.

#### What did you change?
- packages/ibm-products/scripts/generate/index.js
- packages/ibm-products/scripts/generate/templates/DISPLAY_NAME.tsx
- packages/ibm-products/scripts/generate/templates/index.ts
- packages/ibm-products/src/components/index.ts

By running `yarn generate TestComponent`


packages/ibm-products/scripts/generate/templates/DISPLAY_NAME.tsx
```diff
@@ -52,7 +60,7 @@ export let DISPLAY_NAME = React.forwardRef(
 
       // Collect any other property values passed in.
       ...rest
-    },
+    }: DISPLAY_NAMEProps,
     ref
```
matthewgallo, line 64: Adding the following will get rid of a ts error from the forwarded ref

```
ref: ForwardedRef<HTMLDivElement>
```

image

matthewgallo (changes requested):

matthewgallo (approved):

davidmenendez (approved):

PR #5853, opened 2024-08-13
fix(Datagrid): sortable column focus issue

### LINKED ISSUE #5079: [a11y]: User must press TAB to focus before pressing RETURN to apply ordering on datagrid columns
### Package

@carbon/ibm-products

### Browser

Chrome, Firefox

### Operating System

MacOS

### Package version

2.36.0

### React version

18.2.0 

Manual

### Assistive technology

Macos

### Description

Datagrid sort loses focus after a sort,  forcing a user using keyboard navigation to {TAB} and press {ENTER} again to sort in the opposite direction instead of just pressing {ENTER} 
User should not have to TAB on the focus state so it should act where you TAB and can press ENTER to sort and press ENTER again to sort another direction or TAB to jump to the next column.

### WCAG 2.1 Violation

SC 2.4.3 Focus order

### Reproduction/example

https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid--sortable-columns

### Steps to reproduce

Steps to reproduce
1. Navigate to a data grid 
2. Use TAB to navigate to the column.
3. Press RETURN to sort column.
4. Press RETURN to sort column in opposite direction 

Expected 
sort in opposite direct

Actual

https://github.com/carbon-design-system/ibm-products/assets/156674431/ed12c466-75cf-496d-8b6f-a3dcade23d16

### Code of Conduct

- [X] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate problems

anamikaanu96:
Closes #5079 

Instead of losing the column header focus after sorting, now user will be able to sort in opposite direction without pressing the Tab again ( just needs to press Enter to sort in opposite direction).

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/useSortableColumns.tsx
+++ b/packages/ibm-products/src/components/Datagrid/useSortableColumns.tsx
@@ -126,6 +126,14 @@ const useSortableColumns = (hooks: Hooks) => {
         }
         return <ArrowsVertical {...iconProps} />;
       };
+
+      const handleKey = (e, columnId) => {
+        const { key } = e;
+        if (key === 'Enter') {
+          setTimeout(() => document.getElementById(columnId)?.focus(), 0);
+        }
+      };
+
       const Header = (headerProp) =>
         column.disableSortBy === true ||
         column.id === 'datagridSelection' ||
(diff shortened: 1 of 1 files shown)


devadula-nandan:
i'm just imagining a scenario in which if 2 datagrids are used on a page, and they have a column with the same name, there could be 2 same ids in the rendered html, potentially causing an error.

davidmenendez (approved):

makafsal:
@anamikaanu96 It is not working in Firefox.

makafsal (approved):

Issue #5837, opened 2024-08-08
[storybook]: broken usage guidelines links


davidmenendez:
### Package

Carbon for IBM Products

### Description

minor thing here i've noticed throughout our storybook. there appears to be numerous broken links throughout the storybook. probably because the website was updated at some point. [AboutModal](https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-patterns-about-modal-aboutmodal--docs) and [RemoveModal](https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-patterns-remove-removemodal--docs) being just two examples

### Component(s) impacted

i haven't audited every single component and pattern, but `AboutModal` and `RemoveModal` for starters

### Browser

Chrome

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

2.47

### Suggested Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Product/offering

carbon products

### CodeSandbox or Stackblitz example

NA

### Steps to reproduce the issue (if applicable)

1. Go to storybook
2. Go to docs tab of component or pattern
3. Verify that the required `usage guidelines` link is going to the appropriate place 

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues
```[tasklist]
### Components that have custom guidelines url
- [ ] ExpressiveCard
- [ ] ProductiveCard
- [ ] Datagrid
- [ ] Cascade
- [ ] APIKeyModal
- [ ] ImportModal
- [ ] RemoveModal
- [ ] ActionBar
- [ ] DelimitedList
```

matthewgallo:
I started compiling a list but after further investigation, I think most of the issues come from how the url is created [here](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products/src/global/js/utils/story-helper.js#L133).

One issue in particular is that for the "Components" section in our storybook the `kind` value will be `Components`, but the way the url is constructed now will cause it to be the following:

```
https://pages.github.ibm.com/cdai-design/pal/Componentss/${paramCase(...)}
```

the capital "C" and the extra `s`  in `Componentss` cause 404s.

[...]

PR #5855, opened 2024-08-13
fix: correct auto-generated guidelines links

janhassel:
Addresses _most_ cases from https://github.com/carbon-design-system/ibm-products/issues/5837

Currently, all auto-generated guideline links for components and patterns are leading to 404 pages as they…:
- use an uppercase first characters instead of lowercase
- add an unwanted "s" suffix

Before
```
https://pages.github.ibm.com/cdai-design/pal/Componentss/side-panel/usage
```

After
```
https://pages.github.ibm.com/cdai-design/pal/components/side-panel/usage
```

#### What did you change?

- Transform kind (`Components | Patterns`) to lowercase
- Remove unwanted "s" suffix

Local storybook

### DIFF
--- a/packages/ibm-products/src/global/js/utils/story-helper.js
+++ b/packages/ibm-products/src/global/js/utils/story-helper.js
@@ -130,7 +130,7 @@ export const storyDocsPageInfo = (csfFile) => {
 
     result.section = a;
 
-    result.guidelinesHref = `https://pages.github.ibm.com/cdai-design/pal/${kind}s/${paramCase(
+    result.guidelinesHref = `https://pages.github.ibm.com/cdai-design/pal/${kind.toLowerCase()}/${paramCase(
       result.section
     )}/usage`;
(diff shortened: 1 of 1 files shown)


matthewgallo (approved):
Thanks @janhassel! 🎉

davidmenendez (approved):
woot. ty

davidmenendez:
@matthewgallo i wanted to point out that this did fix some things, like `AboutModal` and `Tearsheet`, which were initially called out, but it looks like some of these urls are still broken. like for [cards](https://pages.github.ibm.com/cdai-design/pal/components/cards/usage) and [datagrid](https://pages.github.ibm.com/cdai-design/pal/components/datagrid/usage). these may be special circumstances based on the gatsby url structure. so while i think this is merge ready and does help alleviate some problems it's not fixed entirely. just wanted to point that out 👍

PR #5900, opened 2024-08-20
chore(stylelint): remove products specific config and use carbon package

annawen1:
Closes #

Running into issues with publishing the release - yarn is not able to find the `stylelint-config-ibm-products` local private package. This package can be replaced by the Carbon supported stylelint package instead (`stylelint-config-carbon`).

Disabling some stylelint rules for now. Those rules will be re-enabled once the percy testing work is introduced.

#### What did you change?

`package.json`
`yarn.lock`
removed all the `styleline-config-ibm-products` files

- Locally and on my testing repo


kennylam (approved):
LGTM!

matthewgallo (approved):

PR #5920, opened 2024-08-22
fix(Datagrid): enabled spacer column conditionally

### LINKED ISSUE #5646: [Datagrid]: last column always stretches to the full width
### Package

Carbon for IBM Products

### Description

When having only a few columns and using them with right-aligned content (e.g. for numbers), the last column can easily be overseen as it stretches the full remaining width of the Datagrid.

### Component(s) impacted

Datagrid

### Browser

_No response_

2.43.1

### Suggested Severity

### Product/offering

watsonx

https://stackblitz.com/edit/github-pdwkyw-ukdvo5?file=src%2FExample%2FExample.jsx

image

_No response_

### Code of Conduct

- [ ] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

anamikaanu96:
Closes #5646

Enabled spacer column conditionally, If the spacer column is enabled in case of few columns the last column won't stretches to the full remaining width of the Datagrid.

#### What did you change? 
packages/ibm-products/src/components/Datagrid/types/index.ts
packages/ibm-products/src/components/Datagrid/useFlexResize.js


davidmenendez (changes requested):
can you include a story and update the documentation to include this feature? thanks!

davidmenendez (changes requested):
last request: please include a unit test for this functionality. thanks!

packages/ibm-products/src/components/Datagrid/Datagrid/DatagridVirtualBody.tsx
```diff
@@ -85,6 +87,17 @@ const DatagridVirtualBody = (datagridState: DataGridState) => {
       if (headWrapEl) {
         headWrapEl.scrollLeft = virtualBody?.scrollLeft;
       }
+      if (enableSpacerColumn === true) {
```
matthewgallo, line 90: Just a bit more concise this way
```suggestion
      if (enableSpacerColumn) {
```

packages/ibm-products/src/components/Datagrid/useFlexResize.js
```diff
@@ -6,21 +6,43 @@
...
+    width: 0,
+    disableSortBy: true,
+    disableResizing: true,
+  };
+
   hooks.visibleColumns.push((columns) => {
     // always move actions to the end
     const actionsIdx = columns.findIndex((col) => col.isAction);
-    if (actionsIdx === -1) {
+    if (spacerColumn === false && actionsIdx === -1) {
```
matthewgallo, line 29:
```suggestion
    if (!spacerColumn && actionsIdx === -1) {
```

packages/ibm-products/src/components/Datagrid/useFlexResize.js
```diff
@@ -6,21 +6,43 @@
...
       const lastCol = columns.at(-1);
       lastCol.isFlexCol = true;
       return [...columns];
+    } else if ((spacerColumn === true) & (actionsIdx === -1)) {
```
matthewgallo, line 33:
```suggestion
    } else if ((spacerColumn) & (actionsIdx === -1)) {
```

packages/ibm-products/src/components/Datagrid/Datagrid/DatagridVirtualBody.tsx
```diff
@@ -85,6 +87,17 @@ const DatagridVirtualBody = (datagridState: DataGridState) => {
...
+        const spacerColumn: HTMLDivElement | null = document.querySelector(
```
matthewgallo, line 91: We might want to add an optional chain here to be sure to avoid any SSR issues, like above. Actually, you could use the element above `headWrapEl` and do `headWrapEl.querySelector('thead th:last-child')` since you already have that element.

matthewgallo (changes requested):
Just made some suggestions on a few small things.

davidmenendez (dismissed):

packages/ibm-products/src/components/Datagrid/Datagrid/DatagridVirtualBody.tsx
```diff
@@ -84,13 +84,10 @@ const DatagridVirtualBody = (datagridState: DataGridState) => {
...
-      if (headWrapEl) {
+      if (headWrapEl && enableSpacerColumn) {
         headWrapEl.scrollLeft = virtualBody?.scrollLeft;
-      }
-      if (enableSpacerColumn === true) {
-        const spacerColumn: HTMLDivElement | null = document.querySelector(
-          `#${tableId} .${blockClass}__head-wrap thead th:last-child`
-        );
+        const spacerColumn: HTMLDivElement | null =
+          headWrapEl.querySelector(`thead th:last-child`);
```
matthewgallo, line 90: I was just thinking about this, it's possible that the spacer column isn't _always_ the last `th` element, specifically if there are row actions then the row actions will be the last `th`.

packages/ibm-products/src/components/Datagrid/Datagrid/DatagridVirtualBody.tsx
```diff
@@ -84,13 +84,10 @@ const DatagridVirtualBody = (datagridState: DataGridState) => {
       const headWrapEl = document?.querySelector(
         `#${tableId} .${blockClass}__head-wrap`
       );
-      if (headWrapEl) {
+      if (headWrapEl && enableSpacerColumn) {
```
matthewgallo, line 87: I don't think you want to combine these conditions here. They should be separate like before, otherwise there is scrolling issues between the table body and table headers.

https://github.com/user-attachments/assets/e668afa4-bd95-4f56-b01c-ecde895a0b96

matthewgallo: You can also see there are multiple horizontal scroll bars

packages/ibm-products/src/components/Datagrid/Datagrid.test.js
```diff
@@ -913,6 +942,21 @@ describe(componentName, () => {
...
   });
 
+  it('renders a table with spacer column', () => {
+    render(<SpacerColumn />);
+    expect(screen.getByRole('table')).toHaveClass(
+      `${carbon.prefix}--data-table`
+    );
+
+    expect(
+      screen
```
matthewgallo, line 952: A better way to get all table headers could be to do something like this:
```
expect(screen.getAllByRole('columnheader').length).toEqual(3);
```

This way you don't have to chain multiple `getElementsByTagName`.

matthewgallo (approved):

makafsal (approved):

PR #5933, opened 2024-08-23
fix(Datagrid): select all checkbox to select current page

### LINKED ISSUE #5898: DataGrid select all doesn't work as expected
### Package

Carbon for IBM Products

### Description

- Create a DataGrid component , set  batchActions: true, 
- When select all button is clicked , it selects all the button 
- Uncheck Select All, only unselects the current page data

https://github.com/user-attachments/assets/b13e32da-23d1-42ca-b15e-9070537070a8

### Component(s) impacted

DataGrid batch action

### Browser

Chrome, Safari, Firefox, Microsoft Edge

"@carbon/ibm-products": "2.46.0",

### Suggested Severity

### Product/offering

NA

https://codesandbox.io/p/sandbox/datagrid-tooltip-hover-issue-26hqww?file=%2Fsrc%2FExample.js%3A203%2C25

_No response_

_No response_

### Code of Conduct

devadula-nandan:
Closes #5898 

this pr adds the intended behaviour to the component.

#### What did you change?
packages/ibm-products/src/components/Datagrid/Datagrid/DatagridSelectAll.tsx

storybook, checked locally combining `batch actions` and `pagination` story

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridSelectAll.tsx
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridSelectAll.tsx
@@ -79,12 +79,13 @@ const SelectAll = (datagridState: DataGridState) => {
         target: { checked: false },
       } as any);
     }
-    handleSelectAllRowData({
-      dispatch,
-      rows,
-      getRowId,
-      isChecked: event.target.checked,
-      indeterminate,
+    // deselect all rows first
+    (getToggleAllRowsSelectedProps as any)?.()?.onChange({
+      target: { checked: false },
+    });
+    // select all row on current page
+    (getToggleAllPageRowsSelectedProps as any)().onChange({
+      target: { checked: true },
     });
     onAllRowSelect?.(rows, event);
     return onChange?.(event);
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridSelectAll.tsx
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridSelectAll.tsx
@@ -9,7 +9,7 @@ import React, { useLayoutEffect, useState } from 'react';
 import { TableSelectAll } from '@carbon/react';
 import cx from 'classnames';
 import { pkg } from '../../../settings';
-import { handleSelectAllRowData } from './addons/stateReducer';
+import { handleOnPageSelectAllRowData } from './addons/stateReducer';
 import { DataGridState, DataGridToggleAllRowsProps } from '../types';
 
 const blockClass = `${pkg.prefix}--datagrid`;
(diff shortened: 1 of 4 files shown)


packages/ibm-products/src/components/Datagrid/Datagrid/DatagridSelectAll.tsx
```diff
@@ -79,12 +79,13 @@ const SelectAll = (datagridState: DataGridState) => {
...
-      getRowId,
-      isChecked: event.target.checked,
-      indeterminate,
+    // deselect all rows first
+    (getToggleAllRowsSelectedProps as any)?.()?.onChange({
+      target: { checked: false },
+    });
+    // select all row on current page
+    (getToggleAllPageRowsSelectedProps as any)().onChange({
+      target: { checked: true },
```
AlexanderMelox, line 88: I think this might be too extensive. A better way is to only select the row if the row isn't disabled. What do you think?

devadula-nandan: Hi @AlexanderMelox ,

I have noticed a few other inconsistencies, with this select-all feature
1) while we have selections on other pages, and if we click on the indeterminate state on any page, it deselects other pages as well, while the expected behavior is, that it should deselect the rows only on the current page. (will be fixed in this pr)
2) `Select all` button in the batch actions, selects the disabled rows, on other pages, maybe [this](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products/src/components/Datagrid/Datagrid/addons/stateReducer.js#L116) action needs a refactor (will be raising a separate issue for this)

I have converted this pr to a draft, to recreate a multi-use case on the pagination story, to verify the changes, and I'll be reverting the story once its ready.

elycheea (commented):
Looks good but we should probably add a test case for this so we’re less likely to see another regression. 😅

packages/ibm-products/src/components/Datagrid/Datagrid/addons/stateReducer.js
```diff
@@ -137,6 +149,53 @@ export const stateReducer = (newState, action) => {
...
+      const newSelectedRowIds = {};
+      if (rows) {
+        const newSelectedRowData = {};
+        rows.forEach((row) => {
+          const props = row.getRowProps?.();
+          if (props && props.disabled) {
+            return;
+          }
+          if (props) {
+            if (!indeterminate && !isChecked) {
```
matthewgallo, line 166:
```suggestion
            if (!indeterminate && isChecked) {
```

Did you mean this instead? Otherwise, the value of `datagridState?.state?.selectedRowData` is always `{}`.

devadula-nandan: right

matthewgallo (approved):

davidmenendez (approved):

PR #5943, opened 2024-08-26
fix(EditInPlace): removes focus when pressing esc or enter key

### LINKED ISSUE #5852: EditInPlace keeps focused state when invalid but field is not actually focused on
### Package

Carbon for IBM Products

### Description

Normally, if you focus on the `EditInPlace` component to give input, the state of the component gets focused by adding class `c4p--edit-in-place--focused`, and when you click away (remove focus or focus on another component), the focused state is removed.

But when the component has invalid input and you click away, the EditInPlace component stays focused even when it's not actually focused anymore. The focused state should be removed when component is no longer focused on, regardless of the invalid state. 

### Component(s) impacted

EditInPlace component

### Browser

Chrome

2.46.0

### Suggested Severity

### Product/offering

IBM Cloud Data Replication

https://codesandbox.io/p/devbox/serene-wave-nlgmk5

Reproduced here: https://codesandbox.io/p/devbox/serene-wave-nlgmk5 

## Normal scenario

Not focused
Screenshot 2024-08-12 at 4 22 43 PM

Focused
Screenshot 2024-08-12 at 4 24 06 PM
^Clicking on EditInPlace component to edit, component is focused and `c4p--edit-in-place--focused` class is added. 

Remove focus
Screenshot 2024-08-12 at 4 26 58 PM
Clicking away from EditInPlace component to remove focus, and `c4p--edit-in-place--focused` class is removed.

## Bug scenario

Not focused
Starting with same scenario above with just "abc" input.. not focused on EditInPlace component.
Screenshot 2024-08-12 at 4 28 49 PM

Focused and create invalid scenario (add numbers)
Screenshot 2024-08-12 at 4 30 28 PM
^Clickin on EditInPlace component to edit, which focuses on the component and adds `c4p--edit-in-place--focused` class.
Then added "1" to input to go into invalid state. This additionally adds `c4p--edit-in-place--invalid` class.. Both classes exist now.

Remove focus
Screenshot 2024-08-12 at 4 33 30 PM
First, hit the Enter key to submit/save the changes (keeping the invalid value).
At this time, technically you are no longer focused on the EditInPlace component.. but click somewhere else anyways.
The component keeps both invalid and focused state, but really it's not focused anymore.
This screenshot seems like i am still focused on the EditInPlace component, but i'm not.. If i had other components on the page, i could be clicking and typing in other fields and this would still look like this. 

_No response_

### Code of Conduct

AlexanderMelox:
Closes #5852 

This PR removes the focus of the EditInPlace after pressing Enter or the Escape key.

#### What did you change?
`packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx`

Modified storybook story to reflect the code sandbox props

https://github.com/user-attachments/assets/29621692-3efc-4c88-9293-6b49327c1fb5

### DIFF
--- a/packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx
+++ b/packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx
@@ -248,16 +248,21 @@ export let EditInPlace = forwardRef<HTMLDivElement, EditInplaceProps>(
       onCancelHandler();
     };
 
+    const removeFocus = () => {
+      inputRef.current?.blur();
+      setFocused(false);
+    };
+
     const onKeyHandler = (e) => {
       // to prevent blur handler from being called twice add additional state to check if escape is being used
       escaping.current = true;
       switch (e.key) {
         case 'Escape':
-          inputRef.current?.blur();
+          removeFocus();
           escapeHandler();
           break;
         case 'Enter':
-          inputRef.current?.blur();
+          removeFocus();
           returnHandler();
           break;
         default:
--- a/packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx
+++ b/packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx
@@ -207,13 +207,11 @@ export let EditInPlace = forwardRef<HTMLDivElement, EditInplaceProps>(
 
     const onSaveHandler = () => {
       setInitialValue(value);
-      setFocused(false);
       setDirtyInput(false);
       onSave();
     };
 
     const onCancelHandler = () => {
-      setFocused(false);
       setDirtyInput(false);
       onCancel(initialValue);
     };


packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx
```diff
@@ -248,16 +248,21 @@ export let EditInPlace = forwardRef<HTMLDivElement, EditInplaceProps>(
       onCancelHandler();
     };
 
+    const removeFocus = () => {
```
amal-k-joy, line 251: Hey @AlexanderMelox ,
Changes look fine. Just one doubt.
 I could see on `escape` press,  `inputRef.current?.blur();` and `setFocused(false);` was already happening previously. Inside `escapeHandler`, they are setting `false` to `setFocused`. I was just wondering why this issue existed earlier on `Escape` key.

amal-k-joy (changes requested):

packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx
AlexanderMelox: Okay I see, I realized that the focused state was only being triggered on enter when canSave is true. But for consistency, I'm going to only do it on the removeFocus function to keep the code DRY

AlexanderMelox: Thank you!

makafsal (approved):

amal-k-joy (approved):

Issue #5930, opened 2024-08-23
Datagrid: useSelectRows() should memoize rows and columns


wkeese:
### Package

Carbon for IBM Products

### Description

`useSelectRows()` has two issues that cause unnecessary redraws:

### `datagridState.rows` is continually reset

It's from this code that presumably just needs to be wrapped in `useMemo()`:

```js
const { rows } = instance;
const rowsWithSelect = rows.map((row) => ({ ...row, isSelectable: true }));
Object.assign(instance, { rows: rowsWithSelect });
```

### `datagridState.allColumns` is continually reset

It's from the code below, which continually resets `datagridSate.allColumns` even when the `columns` input hasn't changed:

```js
hooks.visibleColumns.push((columns) => {
  // Ensures that the first column is the row expander in the
  // case of selected rows and expandable rows being used together
  const newColOrder = [...columns];
  const expanderColumnIndex = newColOrder.findIndex(
    (col) => col.id === 'expander'
  );
  const expanderCol =
    expanderColumnIndex > -1
      ? newColOrder.splice(expanderColumnIndex, 1)
      : [];
  return [
    ...(expanderColumnIndex > -1 && expanderCol && expanderCol.length
      ? expanderCol
      : []),
    {
      id: selectionColumnId,
      Header: (gridState: DataGridState<any>) => <SelectAll {...gridState} />,
      Cell: (gridState) => <SelectRow {...gridState} />,
    },
    ...newColOrder,
  ] as ColumnInstance[];
});
```

I don't fully understand how the hook architecture works, and I'm not sure how to fix this.  Ideally you would use `useMemo()` here too, but I'm not sure if that's allowed because the callback doesn't look like a hook.  Maybe it needs to be fixed with something like lodash's `memoize()`, or a `WeakMap`, etc.

See also #5908, a similar bug about `useRowRenderer()`.

### Component(s) impacted

Datagrid

### Browser

Chrome

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

2.46.0

### Suggested Severity

None

### Product/offering

IKC

### CodeSandbox or Stackblitz example

const { rows } = instance; const rowsWithSelect = rows.map((row) => ({ ...row, isSelectable: true })); Object.assign(instance, { rows: rowsWithSelect });

### Code of Conduct

[...]

PR #5949, opened 2024-08-27
fix(datagrid): add useMemo() calls

### LINKED ISSUE #5908: Datagrid: useRowRenderer() makes grid constantly rerender
### Package

Carbon for IBM Products

### Description

I spent a long time tracing down a problem with Datagrid where it constantly rerenders.

AFAICT our root component is (for some reason) constantly rendering, which means that it constantly calls `useDatagrid()`, which constantly calls `useRowRenderer()`.

And then, `useRowRenderer()` keeps needlessly resetting the value of `rows[]`, even though the row data hasn't changed.

Not sure how to fix it, as the whole architecture (of adding a `RowRenderer` field to each row) seems wrong.  I guess the practical solution is to just modify `useRowRenderer()`, adding a `useMemo()`  around everything:

```js
const rowsWithRenderer = useMemo(() => {
  let newRows = null;
  if (state?.dynamicRowSkeleton) {
     ...
  }
  newRows = newRows ?? rows;
  
  const addRowRenderer = (row) =>
    Object.assign(row, {
      RowRenderer: DatagridRow,
      subRows: (row.subRows || []).map(addRowRenderer),
    });
  return newRows.map(addRowRenderer);
}, [rows, state])
```

### Component(s) impacted

Datagrid

### Browser

Chrome

2.46.0

### Suggested Severity

None

### Product/offering

IKC

https://stackblitz.com/edit/github-1szsco?file=src%2FExample%2FExample.jsx

See https://stackblitz.com/edit/github-1szsco?file=src%2FExample%2FExample.jsx and check the console to see how `TestComponent` is constantly rerendered, even though the data and columns haven't changed.

In our actual code, this makes Chrome completely hang, although it's possible that it's partly our fault.

But anyway, hopefully this example is enough to argue for adding a `useMemo()` call?

_No response_

### Code of Conduct

wkeese:
Closes #5908, refs #5930.

Add useMemo() wrappers to useRowRenderer() and useSelectRows() to avoid needless Datagrid rerenders.

#### What did you change?

Added useMemo() wrappers in useRowRenderer() and useSelectRows() to avoid resetting datagridState.rows unnecessarily.

Ran storybook examples and tested in my own app.

I suggest [diffing ignoring whitespace](https://github.com/carbon-design-system/ibm-products/pull/5949/files?diff=unified&w=1).


devadula-nandan:
Hi @wkeese,

I observe that these changes prevents the re-updation of rows[], but aren't fixing the re-rendering of the rows, which was the main objective of the issue if I am right.
pls find the recording below

https://github.com/user-attachments/assets/55420b91-6ba0-44bb-9b22-b40dd3cff694

devadula-nandan:
I suspect the root cause of this is the mentioned props, might not have memoization.

Screenshot 2024-08-29 at 9 15 25 PM

wkeese:
Hi @devadula-nandan.  Thanks for looking at this!  I think my PR is just one step on a long road to fixing the problem.

I didn't know about the problem with callbacks.  Can we fix it with `useCallback()`?  Or if that's impossible due to closure variables like `datagridState`, then [`useEventCallback()`](https://usehooks-ts.com/react-hook/use-event-callback)?

I thought there was another issue too though, with the `allColumns` variable, which is much harder to memoize.  Strangely it didn't show up in your trace.

I've never used that react profiling tab.  Is that what told you about the callback props changing?

wkeese:
Well, the code you're talking about was just to make a simple test case.  The actual culprit, for unknown reasons, is useColumnRightAlign().

AlexanderMelox:
@wkeese should we close this PR?

wkeese:
Sure, now that datagrid is deprecated I guess there's no point.

PR #5955, opened 2024-08-27
fix: update toHaveNoAxeViolations and remove skips

lee-chase:
Updates toHaveNoAxeViolations and fixes a small number of use cases.

#### What did you change?

The function was creating a new Promise which is not necessary, refactored.

Not entirely sure the async/await is doing much in the tests that were refactored, but with async tests the accessible checks need to be waited for when there are multiple uses of the accessible matchers.

Ran the tests.

### DIFF
--- a/config/jest-config-ibm-cloud-cognitive/setup/matchers/toHaveNoAxeViolations.js
+++ b/config/jest-config-ibm-cloud-cognitive/setup/matchers/toHaveNoAxeViolations.js
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corp. 2020, 2020
+ * Copyright IBM Corp. 2020, 2024
  *
  * This source code is licensed under the Apache-2.0 license found in the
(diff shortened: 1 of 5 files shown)


amal-k-joy (approved):
LGTM

lee-chase:
> LGTM
> 
> @elycheea , @lee-chase Looks like we have not added `await` in most of our accessibility test cases across components (eg: AboutModal.test, AddSelect.test ...) Do we need to create an new issue to add `await` ?
> 
> ```
>  it('has no accessibility violations', async () => {
>     const { container } = render(<AddSelect {...defaultProps} />);
>     expect(container).toBeAccessible(componentName);
>     expect(container).toHaveNoAxeViolations();
>   });
> ```

@amal-k-joy it only seems to be needed if you want to run multiple accessibility tests from within the same module.

elycheea (approved):
Thank you for cleaning up those `.skip()`s! 🙏

PR #5970, opened 2024-08-28
fix(datagrid): fixes multiselect for instant update bug

### LINKED ISSUE #5935: [Datagrid]: using MultiSelect filtering with instant update is broken
### Package

Carbon for IBM Products

### Description

i discovered this after discussing a separate issue from slack:

* in order to be more aligned with carbon functionality `MultiSelect` should be passing the entire `event` object to the consumer and not just `{ selectedItems }`

while investigating i also discovered these:

* using `MultiSelect` with batch filtering throws a console warning: `Warning: Cannot update a component ('FilterFlyout') while rendering a different component ('MultiSelect'). To locate the bad setState() call inside 'MultiSelect', follow the stack trace as described in...`
* using `MultiSelect` with instant filtering breaks the app `a.filter is not a function`

### Component(s) impacted

`Datagrid`

### Browser

Chrome

2.47

### Suggested Severity

### Product/offering

ibm products

https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-filtering-panel--panel-instant&globals=viewport:basic

for the broken instant update:

go to the [storybook](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-filtering-panel--panel-instant&globals=viewport:basic), open the filter, and select an option under one of the multi select drop downs.

for the batch update warning:

*this only appears locally*. run storybook locally and open the [batch update story](http://localhost:3000/?path=/story/ibm-products-components-datagrid-filtering-panel--panel-batch&globals=viewport:basic). select an option from one of the multi select drop downs.

_No response_

### Code of Conduct

AlexanderMelox:
Closes #5935 

This PR fixes the error `a.filter is not a function`. Replaced the dropdown filter with multiselect

#### What did you change?
`packages/ibm-products/src/components/Datagrid/Extensions/Filtering/Panel.stories.jsx`

Storybook

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Extensions/Filtering/Panel.stories.jsx
+++ b/packages/ibm-products/src/components/Datagrid/Extensions/Filtering/Panel.stories.jsx
@@ -294,15 +294,11 @@ PanelInstant.args = {
           {
             filterLabel: 'Status',
             filter: {
-              type: 'dropdown',
+              type: 'multiSelect',
               column: 'status',
               props: {
-                Dropdown: {
-                  id: 'marital-status-dropdown',
-                  ['aria-label']: 'Marital status dropdown',
-                  items: ['relationship', 'complicated', 'single'],
-                  label: 'Marital status',
-                  titleText: 'Marital status',
+                MultiSelect: {
+                  ...multiSelectProps,
                 },
               },
             },
--- a/packages/ibm-products/src/components/Datagrid/Extensions/Filtering/Panel.stories.jsx
+++ b/packages/ibm-products/src/components/Datagrid/Extensions/Filtering/Panel.stories.jsx
@@ -485,15 +485,11 @@ PanelWithInitialFilters.args = {
           {
             filterLabel: 'Status',
             filter: {
-              type: 'dropdown',
+              type: 'multiSelect',
               column: 'status',
               props: {
-                Dropdown: {
-                  id: 'marital-status-dropdown',
-                  ['aria-label']: 'Marital status dropdown',
-                  items: ['relationship', 'complicated', 'single'],
-                  label: 'Marital status',
-                  titleText: 'Marital status',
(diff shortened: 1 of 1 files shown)


davidmenendez (changes requested):
i think the underlying issue of using a dropdown needs to be addressed here. there are other examples in storybook of a dropdown being used in the datagrid filter and it's breaking the same way. for example https://deploy-preview-5970--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid-filtering-panel--panel-many-checkboxes&globals=viewport:basic

AlexanderMelox:
Thanks David, yeah I've been working on that too. Just wanted to get the stories consistent

AlexanderMelox:
@davidmenendez after further inspection, the reason that was happening was because in the header the filter was multiSelect but the filter in the panel story was dropdown. So nothing to fix, just had to make sure they match. I did double check when they both are dropdown everything works well 

**Status column is multiselect:** https://github.com/carbon-design-system/ibm-products/blob/0c592d2a20f015cf3167344aec8e51ac3f7ea2f9/packages/ibm-products/src/components/Datagrid/utils/FilteringUsage.js#L63

**Doesn't match here:** https://github.com/carbon-design-system/ibm-products/blob/0c592d2a20f015cf3167344aec8e51ac3f7ea2f9/packages/ibm-products/src/components/Datagrid/Extensions/Filtering/Panel.stories.jsx#L297

AlexanderMelox:
Working with dropdown
https://github.com/user-attachments/assets/2f398c5a-e67d-429f-906f-35eb18eb7920

AlexanderMelox:
Also talked to @matthewgallo and agreed to just change them all to multi select

davidmenendez (approved):
TY!

sangeethababu9223 (approved):

PR #5999, opened 2024-09-04
fix(datagrid): use same empty array every time

### LINKED ISSUE #5998: Datagrid: FilterPanel endless redraws: fix empty array issues
### Package

Carbon for IBM Products

### Description

`FilterPanel` isn't working in our app, due to endless redraws.  Part of the problem seems to be bugs in the filter code causing array attributes to get reset every time.

See the `reactTableFiltersState`  attribute passed to `FilterPanel`:

```js
const FilterPanel = ({
  …
  reactTableFiltersState = [],
  …
}) => {
```

Assuming that the caller does **not** specify `reactTableFiltersState`, it gets a new value every time the `FilterPanel()` function is called.

And that becomes a disaster when the variable is used in the dependencies array for `useEffect()` etc., for example in `useFilters.js`:

```js
// Re-applies filters if the Datagrid goes into a fetching state while panel is open
// and has had filters changed without applying
useEffect(() => {
  …
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, [
  …
  reactTableFiltersState,
  …
]);
```

useFilters.js also has the same problem with its `filters` attribute:

```js
const useFilters = ({
  …
  filters = [],
  …
}) => {
```

The solution is trivial: to have an empty array variable, ex:

```js
// Use same empty array every time, for benefit of useEffect() etc. dependency checking.
const emptyArray = [];

const FilterPanel = ({
  …
  reactTableFiltersState = emptyArray,
  …
}) => {
```

### Component(s) impacted

Datagrid

### Browser

Chrome

2.48

### Suggested Severity

None

### Product/offering

IKC

Couldn't create one yet.

Possibly the problem is triggered by table height (resize observers) or scrolling observers triggering the initial redraw. 
 The grid has a lot of bizarre code doing height calculation in Javascript.  That should all be removed too, but that's for another PR.

_No response_

_No response_

### Code of Conduct

wkeese:
Closes #5998

Since [] !== [], need to be careful when setting [] as a default value and then using it in the dependencies of useEffect() etc.

#### What did you change?

Created constant variables so that filters, reactTableFiltersState, etc. variables get the same default value every time.

Storybook

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/FilterFlyout.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/addons/Filtering/FilterFlyout.js
@@ -36,11 +36,14 @@ const defaults = {
   align: 'bottom',
 };
 
+// Use same empty array every time, for benefit of useEffect() etc. dependency checking.
+const emptyArray = [];
+
 const FilterFlyout = ({
   updateMethod,
   flyoutIconDescription = defaults.flyoutIconDescription,
   align = defaults.align,
-  filters = [],
+  filters = emptyArray,
   title = defaults.title,
   primaryActionLabel = defaults.primaryActionLabel,
   onFlyoutOpen = () => {},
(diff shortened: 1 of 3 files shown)


devadula-nandan (approved):
LGTM

matthewgallo (approved):
Thanks for contributing this @wkeese! We just saw something similar in a different component that was moving from `defaultProps` to default function params.

PR #6007, opened 2024-09-05
feat(ConditionBuilder): option to default enable with initial state

### LINKED ISSUE #5874: [Feature Request]: Condition builder with initial state
### The problem

The conditional builder populates with the initial state after the user clicks on the Add condition. As the rootState is not initialized with the `initialState` props, the `getConditionState` function calls with an empty state on the component load.

### The solution

Initialize the `rootState` with `initialState`
```
useEffect(() => {
    if(initialState) {
      setRootState(initialState)
    }
  }, [initialState]);
```

### Design link

_No response_

### Links to other materials

_No response_

### Owner/main maintainer(s)

NA

_No response_

### Product/offering

_No response_

### Business priority

None

### Code of Conduct

amal-k-joy:
Closes #5874

This will add an option to start the condition builder by default when initial state is provided. Currently initial state populates when _Add Condition_ button is clicked. 
`initialState` prop will now accept a initial state as well as boolean that decides whether to default populate the state.

Storybook

### DIFF
--- a/packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.tsx
+++ b/packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.tsx
@@ -28,13 +28,14 @@ import {
   ConditionBuilderContextProps,
   ConditionBuilderState,
   ConditionGroup,
+  InitialState,
 } from '../ConditionBuilder.types';
 import { blockClass, HIERARCHICAL_VARIANT } from '../utils/util';
 interface ConditionBuilderContentProps {
   startConditionLabel: string;
   getConditionState: (state: ConditionBuilderState) => void;
   getActionsState?: (state: Action[]) => void;
-  initialState?: ConditionBuilderState;
+  initialState?: InitialState;
   actions?: Action[];
 }
 const ConditionBuilderContent = ({
@@ -46,6 +47,10 @@ const ConditionBuilderContent = ({
 }: ConditionBuilderContentProps) => {
   const { rootState, setRootState, variant, actionState } =
     useContext<ConditionBuilderContextProps>(ConditionBuilderContext);
+
+  const [initialConditionState, setInitialConditionState] = useState(
+    initialState?.state
+  );
   const [isConditionBuilderActive, setIsConditionBuilderActive] =
     useState(false);
   const [showConditionGroupPreview, setShowConditionGroupPreview] =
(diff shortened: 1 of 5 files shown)


packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.tsx
```diff
@@ -274,35 +293,38 @@ ConditionBuilderContent.propTypes = {
...
+                ]),
+              }),
+              PropTypes.object,
+            ])
+          ),
+        })
+      ),
+      operator: PropTypes.string,
+    }),
+    enabledDefault: PropTypes.bool,
```
szinta, line 327: Just a thought: do we really need an extra prop to determine whether to load initalState by default or not? Can we do it like, if the user passes the initialState, it is loaded by default, and if the initialState is not supplied or it is an empty object, then 'Add Conditions' button is displayed?

szinta (dismissed):
Looks good to me except just one clarification regarding the additional prop _enabledDefault_

davidmenendez (changes requested):
i'm not sure i see the benefit here of adding `initialConditionalState`. using props as state is viewed as an anti-pattern and i think it's something we should take a look at here. if we can derive what we need directly from props then adding additional state only complicates things.

take a look at these and let me know what you think

https://react.dev/learn/sharing-state-between-components#a-single-source-of-truth-for-each-state

https://react.dev/learn/choosing-the-state-structure#avoid-redundant-state

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.tsx
amal-k-joy: Hey @szinta ,
Here the scenario is, when the initial state is passed, we need to :
Populate the passed initial state initially itself with out required to click on _Add Condition_ Button (enabledDefault:true)
Populate the passed initial state only after clicking on _Add Condition_ Button (enabledDefault : false). 
Also we have not exposed an extra prop, instead managed with `initialState. enabledDefault`

amal-k-joy:
> i'm not sure i see the benefit here of adding `initialConditionalState`. using props as state is viewed as an anti-pattern and i think it's something we should take a look at here. if we can derive what we need directly from props then adding additional state only complicates things.
> 
> take a look at these and let me know what you think
> 
> https://react.dev/learn/sharing-state-between-components#a-single-source-of-truth-for-each-state
> 
> https://react.dev/learn/choosing-the-state-structure#avoid-redundant-state


Hi @davidmenendez ,
Thanks for pointing this out. I have managed this with ref.

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.tsx
```diff
@@ -46,6 +53,10 @@ const ConditionBuilderContent = ({
 }: ConditionBuilderContentProps) => {
   const { rootState, setRootState, variant, actionState } =
     useContext<ConditionBuilderContextProps>(ConditionBuilderContext);
+
+  const initialConditionState = useRef(
```
davidmenendez, line 57: i don't think this is still quite right. if `initialState` is being passed as a prop then there's no need to store the value again anywhere. not in state or as a ref. if this is simply needed to establish the base state in `rootState` then i think it just needs to be set once in the effect on line 100.

davidmenendez (changes requested):

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.tsx
amal-k-joy: I got your point. This is to fix this issue. Consider we pass an  initial state prop and it will populate that in the condition builder. And now if you remove all conditions till no condition , we show _Add Condition_ button., and again when i click on _add condition_ ,this time  the builder should start from an empty state , rather than from the initial state. so we cannot use initial state prop directly in all scenarios, and since prop is immutable I had to either use a state/ref.
Current behaviour: 

https://github.com/user-attachments/assets/551b98d2-66f2-4527-b95d-bc01bf1d7857

Updated behaviour:

https://github.com/user-attachments/assets/7c968708-8bd1-4105-8dfb-244eb2bea95e

davidmenendez: ok i see. makes sense

davidmenendez (approved):

anamikaanu96 (approved):

PR #6084, opened 2024-09-20
fix(update-example-gallery): resolve create-pr failure issue

### LINKED ISSUE #6067: Fix example gallery action
@sangeethababu9223 noticed that our Stackblitz and CodeSandbox examples are out of date. We do have an action to update these examples[^1], but the action has never been working correctly.

[^1]: https://github.com/carbon-design-system/ibm-products/actions/workflows/update-example-gallery.yml
```[tasklist]
### Tasks
- [x] add commit changes from update gallery
- [x] reduce frequency of example gallery to once a month ?
```

makafsal:
Closes #6067 

#### What did you change?
`- cron: '0 0 1 * *'`
`if: **failure()** && steps.up_to_date.outcome == 'failure'`
`GITHUB_TOKEN `

Github UI


matthewgallo (approved):

elycheea (commented):
I believe we still need to add a commit step as is included in the release workflows[^1].

Also wonder if we want to set example gallery updates to middle of the month rather than start of the month. It probably doesn’t matter in the long run but until we update the `update.yml` action too, we’ll be getting 3 automated PRs on the 1st of each month. 😓 


[^1]: https://github.com/carbon-design-system/ibm-products/blob/main/.github/workflows/release-base.yml#L60

makafsal:
Raising, another PR due to I messed up this PR 😄

PR #6091, opened 2024-09-23
fix(dataGrid): header unscrollable in infinite scroll

### LINKED ISSUE #5948: Datagrid: can't scroll headers
### Package

Carbon for IBM Products

### Description

QA filed a ticket against our product that you can't scroll a table horizontally from the header... i.e. there's no effect when you move the cursor over the header columns, and then move two fingers on the trackpad (or similar action on mobile devices).

From a user's standpoint, I can understand the complaint.

From a development standpoint, I can see neither `useFloatingScroll()` nore `DatagridVirtualBody` have any code to support this.  They only have code to:

```js
// Sync the scrollLeft position of the virtual body to the table header
```

I.E. Code to monitor the `scrollLeft` position of the `<tbody>` and copy it to the `<thead>`. There's no code to allow manually scrolling the `<thead>` nor to copy scroll position from the `<thead>` to the `<tbody>`.

I think to support it you would need to set `overflow-x:auto` on the `<thead>`, but then use the tricks listed in https://stackoverflow.com/questions/16670931/hide-scroll-bar-but-while-still-being-able-to-scroll to hide the scrollbar.

### Component(s) impacted

Datagrid

### Browser

Chrome

2.46.0

### Suggested Severity

None

### Product/offering

IKC

https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-datagrid--infinite-scroll&globals=viewport:basic

It's hard to give a test case for this problem due to all the other bugs.

The [test case listed above](https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-datagrid--infinite-scroll&globals=viewport:basic) uses the `withVirtualScroll` flag, but the horizontal scrolling is broken in other ways: notice how you can't see the vertical scrollbar until you scroll all the way to the right.

I also made a [test case for useFloatingScroll()](https://stackblitz.com/edit/github-qcefwb-kgcczw?file=README.md), but it has multiple other problems that need to be fixed first before you can see the problem listed in this ticket.

_No response_

### Code of Conduct

devadula-nandan:
Closes #5948 
WIP

noticed the last column is not resizable and doesn't expand in the infinite scroll story

#### What did you change?
packages/ibm-products-styles/src/components/Datagrid/styles/_datagrid.scss

storybook

### DIFF
--- a/packages/ibm-products-styles/src/__tests__/__snapshots__/styles.test.js.snap
+++ b/packages/ibm-products-styles/src/__tests__/__snapshots__/styles.test.js.snap
@@ -3978,6 +3978,12 @@ p.c4p--about-modal__copyright-text:first-child {
 .c4p--datagrid .c4p--datagrid__head-wrap {
   overflow: hidden;
   background-color: var(--cds-layer-accent);
+  scrollbar-width: none; /* Firefox */
+  -ms-overflow-style: none; /* IE and Edge */
+}
+
+.c4p--datagrid .c4p--datagrid__head-wrap::-webkit-scrollbar {
+  display: none;
 }
 
 .c4p--datagrid .cds--action-list .cds--btn.c4p--button-menu {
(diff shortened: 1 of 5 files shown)


devadula-nandan:
closing in favour of a better approach in another pr

PR #6112, opened 2024-09-24
feat(tag): updates tags to use DismissibleTag

### LINKED ISSUE #4556: Tags with filter need to be replaced with `DismissibleTag`
`@carbon/react@1.53.0` introduces `DismissibleTag` which should replace use of Tag with filters.

```[tasklist]
### Acceptance criteria
- [ ] Investigate the Tag updates from Carbon (is DismissibleTag stable??)
- [ ] Replace `<Tag filter>` with `<DismissibleTag>`
- [ ] Update tests
- [ ] Reinstate tests from #4552
- [ ] Search for any other cases not documented
```

To see the full list of tests that were modified as a result of this change, please see commits from #4552[^1].

[^1]: https://github.com/carbon-design-system/ibm-products/pull/4552/files/ed643916b2474f300f4d329d06746f1c769e2dee..19d36540c5673c25feae7ecd141d210c82319a1c

AlexanderMelox:
Closes #4556 

This Pr replaces the deprecated filter prop on tags and replaces them with DismissibleTag.

#### What did you change?
All places that utilize filter prop in the Tag components

Storybook


davidmenendez (commented):
everything looks fine. my only comment is still on the usage of the `filter` prop. based on the implementation here i think it's fine to ignore / deprecate the usage of that prop and base the usage of `DismissibleTag` on if an `onClose` handler has been passed. since `filter` only every worked with `onClose` i don't think there should be any gaps by just using that as a check `isFilterable = typeof onClose === 'function'`

if there's some other edge case i'm missing please let me know!!

davidmenendez:
oh and just a reminder that you can get rid of the `has no exported member` errors by including the new imports here https://github.com/carbon-design-system/ibm-products/blob/6dd626ac445a255ca9e7c64eb851c11dbb0117f7/packages/ibm-products/src/custom-typings/index.d.ts

anamikaanu96 (approved):
LGTM

AlexanderMelox:
@matthewgallo After further investigation why this might have happened, it looks like Carbon automatically adds the full label when the text overflows

Screenshot 2024-10-22 at 10 14 52 AM
Screenshot 2024-10-22 at 10 15 40 AM

matthewgallo (approved):

PR #6114, opened 2024-09-25
fix(Tearsheet): add missing declaration for headerActions prop

wkeese:
Refs #546.

`headerActions` was defined for `TearsheetShell` but not for `Tearsheet`.

This caused Typescript errors, and you were also missing `PropTypes` validation for any plain Javascript code using `headerActions`.

Note that ideally, Tearsheet.tsx and TearsheetNarrow.tsx would not repeat the type declarations from TearsheetShell.tsx.  But that's no addressed in this PR.

#### What did you change?

Added Tearsheet `headerActions` Typescript declaration, and also added it to the `PropTypes`.

Built locally.

### DIFF
--- a/packages/ibm-products/src/components/Tearsheet/Tearsheet.tsx
+++ b/packages/ibm-products/src/components/Tearsheet/Tearsheet.tsx
@@ -88,6 +88,14 @@ export interface TearsheetProps extends PropsWithChildren {
   hasCloseIcon?: boolean;
 
   /**
+   * The content for the header actions area, displayed alongside the title in
+   * the header area of the tearsheet. This is typically a drop-down, or a set
+   * of small buttons, or similar. NB the headerActions is only applicable for
+   * wide tearsheets.
+   */
+  headerActions?: ReactNode;
+
+  /**
    * The content for the influencer section of the tearsheet, displayed
    * alongside the main content. This is typically a menu, or filter, or
    * progress indicator, or similar.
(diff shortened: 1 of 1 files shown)


packages/ibm-products/src/components/Tearsheet/Tearsheet.tsx
```diff
@@ -87,6 +87,14 @@ export interface TearsheetProps extends PropsWithChildren {
...
   hasCloseIcon?: boolean;
 
+  /**
+   * The content for the header actions area, displayed alongside the title in
+   * the header area of the tearsheet. This is typically a drop-down, or a set
+   * of small buttons, or similar. NB the headerActions is only applicable for
+   * wide tearsheets.
+   */
+  headerActions?: ReactNode;
+
```
amal-k-joy, line 97: Hey @wkeese ,
Thanks for the changes. 
I was trying to see this prop type warning in the story `Tearsheet with all header items and influencer`  which passes `headerActions` prop and couldn't find any. Could you help me to reproduce this issue ?

image

amal-k-joy (changes requested):

packages/ibm-products/src/components/Tearsheet/Tearsheet.tsx
wkeese: @amal-k-joy - I guess `PropTypes` doesn't warn about having **extra** parameters.   I updated the ticket description to just say that we don't have proper `PropTypes` validation, i.e. something like `headerActions={567}` would **not** currently be flagged.

Note that this PR is mainly for fixing the Typescript declaration, or at least that's my main goal.

amal-k-joy:
@makafsal , As we discussed , could you check the feasibility of lifting the type declaration from `TearsheetShell`

amal-k-joy (approved):
LGTM

elycheea (approved):

Issue #6115, opened 2024-09-25
Datagrid: useDatagrid() return type incompatible with Datagrid.datagridState


wkeese:
### Package

Carbon for IBM Products

### Description

I’m getting Typescript errors using Datagrid because `useDatagrid()` returns  type `TableInstance`, yet `Datagrid`’s `datagridState` parameter is supposed to be type `DataGridState`.

So this fails Typescript checking, even though it’s basically what Storybook and the unit tests are doing:

```js
const datagridState = useDatagrid(…);
return <Datagrid datagridState={datagridState}/>;
```

The Typescript errors are numerous.  Here's a partial screenshot:

![image](https://github.com/user-attachments/assets/7753c571-e6f9-4651-9e6c-6c00948cb142)

`useDatagrid()` basically extends react-table’s `useTable()`, and both returning a `TableInstance`.   And `DatagridState` extends `Omit<TableInstance<T>, 'state' | 'headers' | 'rows' | 'columns'>` adding a bunch more fields like:

* DatagridPagination
* DatagridActions
* DatagridBatchActions
* rows
* columns
* rowSize
* HeaderRow
* withStickyColumn
* 
So the storybook examples seem wrong, for example this one includes `DatagridPagination`, `DatagridActions`, and `DatagridBatchActions` into the `useDatagrid()` call:

```js
const datagridState = useDatagrid(
  {
    columns,
    data,
    initialState: {
      pageSize: 10,
      pageSizes: [5,10,25,50]
    },
    selectAllToggle: {
      labels: {
        allRows: "Select all randomly generated rows",
      },
      onSelectAllRows: (areAllSelected) => console.log('are all selected: ', areAllSelected),
    },
    DatagridPagination,
    DatagridActions,
    DatagridBatchActions,
  },
  useSelectRows,
  useSelectAllToggle,
);
```

But according to Typescript all of those fields should be sent directly to the `Datagrid`, i.e.:

```js
const tableInstance = useDatagrid(
  {
    columns,
    data,
    initialState: {
      pageSize: 10,
      pageSizes: [5,10,25,50]
    },
    selectAllToggle: {
      labels: {
        allRows: "Select all randomly generated rows",
      },
      onSelectAllRows: (areAllSelected) => console.log('are all selected: ', areAllSelected),
    },
  },
  useSelectRows,
  useSelectAllToggle,
);

return <Datagrid datagridState={{
     ...tableInstance,
    DatagridPagination,
    DatagridActions,
[...]

PR #5257, opened 2024-05-17
refactor(DataGrid): implement types for DataGrid component

makafsal:
Closes #5175 

#### What did you change?
1. `packages/ibm-products/src/components/Datagrid/Datagrid/Datagrid.tsx`
2. `packages/ibm-products/src/components/Datagrid/Datagrid/DatagridContent.tsx`
3. `packages/ibm-products/src/components/FilterSummary/FilterSummary.tsx`
4. `packages/ibm-products/src/components/Datagrid/Datagrid/DatagridBody.tsx`
5. `packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHead.tsx`
6. `packages/ibm-products/src/components/Datagrid/Datagrid/DatagridEmptyBody.tsx`
7. `packages/ibm-products/src/components/Datagrid/Datagrid/DatagridVirtualBody.tsx`
8. `packages/ibm-products/src/components/Datagrid/Datagrid/DatagridRefBody.tsx`
9. `packages/ibm-products/src/components/Datagrid/Datagrid/DatagridSimpleBody.tsx`
10. `packages/ibm-products/src/components/Datagrid/Datagrid/DatagridExpandedRow.tsx`
11. `packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHeaderRow.tsx`
12. `packages/ibm-products/src/components/Datagrid/Datagrid/addons/Slug/DatagridSlug.tsx`
13. `packages/ibm-products/src/components/Datagrid/Datagrid/DatagridRow.tsx`
14. `packages/ibm-products/src/components/Datagrid/Datagrid/DatagridSelectAll.tsx`
15. `packages/ibm-products/src/components/Datagrid/Datagrid/DatagridSelectAllWithToggle.tsx`
16. `packages/ibm-products/src/components/Datagrid/Datagrid/DatagridToolbar.tsx`
17. `packages/ibm-products/src/components/Datagrid/Datagrid/DraggableElement.tsx`

#### How did you test and verify your work?
Storybook
`yarn test`
Manual test

PR #6122, opened 2024-09-25
fix: datagrid typescript declarations

wkeese:
Refs #5257, #6115.

This fixes some (but not all) of the Typescript issues with Datagrid.

Specifically:

1. Export DataGridState type.  Maybe other types should be exported too. Note that in other places the G is not capitalized, but it is for the type.

2. onAllRowSelect() is (and should be) optional.

3. Presumably you aren't required to have a button on your empty state screens, so "emptyStateAction" should be optional.

4. "allRowsLabel" should be optional because you only need it if your table has a "Select all" checkbox.  And also because it's optional in PropTypes.

5. Properties like "page" are only necessary if your table has pagination, so make UsePaginationInstanceProps props optional.

6. UseFiltersInstanceProps defines a bunch of weird and required properties like "preFilteredRows", "preFilteredFlatRows", "preFilteredRowsById".  AFAICT none of those are valid for DataGridState except for setFilter() and setAllFilters(), and even those should be optional.

7. The original DataGridState class extended both UseRowSelectInstanceProps and Pick<UseRowSelectInstanceProps<T>, 'toggleAllRowsSelected'>.  Obviously that doesn't make sense because of the redundancy.  But I'm not sure what does make sense.  AFAIK users don't directly defined functions like toggleRowSelected() or toggleAllRowsSelected()... but there is code that calls them.

This PR doesn't address the main issues with the types, where useDatagrid() returns a TableInstance but Datagrid.datagridState is a DataGridState.

Further, I think maybe DataGridState needs to be split into two types. Consider that:

1. The object passed to Datagrid.datagridState does not contain "row".
2. DataGridRow's props DO include "row".
3. Both Datagrid.datagridState and DataGridRow's props are defined as DataGridState.

#### What did you change?

Typescript declarations for Datagrid.

Tested locally with my own app.

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/index.ts
+++ b/packages/ibm-products/src/components/Datagrid/index.ts
@@ -27,3 +27,7 @@ export { default as useEditableCell } from './useEditableCell';
 export { default as useFiltering } from './useFiltering';
 export { getAutoSizedColumnWidth } from './utils/getAutoSizedColumnWidth';
 export { useFilterContext } from './Datagrid/addons/Filtering/hooks';
+
+// types/index.ts defines a lot of types.  Unclear which ones should be exported,
+// but presumably not all of them because some have generic names like "Size".
+export type { DataGridState } from './types';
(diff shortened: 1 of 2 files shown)


makafsal (approved):
LGTM

matthewgallo (approved):

elycheea:
@all-contributors add @wkeese for code

PR #6135, opened 2024-09-30
fix(dataGrid): header scroll issues

### LINKED ISSUE #5948: Datagrid: can't scroll headers

devadula-nandan:
Closes #5948

- The width is now dynamically set on window resize. Previously, useIsomorphicEffect used to set the width only during the initial load. the effect didn't used to run on gridRef changes, now we are getting the width from it and passing it as dependency.
- Added a scroll event to the header to sync with the body. Previously, only scrolling the body would sync the header, but now scrolling the header also syncs the body.
- Fixed the last column using the `flex: 1 0 auto` property.

https://github.com/user-attachments/assets/9f9916ab-78c7-41b2-b737-b2d83f10ab38

- Found an unexpected console warning already existing before this pr, and raised #6144 although very minor, raising so we don't lose it.

#### What did you change?
packages/ibm-products-styles/src/components/Datagrid/styles/_datagrid.scss
packages/ibm-products/src/components/Datagrid/Datagrid/DatagridRow.tsx

storybook, dev tools touch simulation.

### DIFF
--- a/config/jest-config-ibm-cloud-cognitive/CHANGELOG.md
+++ b/config/jest-config-ibm-cloud-cognitive/CHANGELOG.md
@@ -3,6 +3,18 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [1.10.0-rc.0](https://github.com/carbon-design-system/ibm-products/compare/jest-config-ibm-cloud-cognitive@1.7.0-rc.0...jest-config-ibm-cloud-cognitive@1.10.0-rc.0) (2024-09-30)
+
+
+### Bug Fixes
+
+* adds typescript to coverage report ([#6140](https://github.com/carbon-design-system/ibm-products/issues/6140)) ([2466c5b](https://github.com/carbon-design-system/ibm-products/commit/2466c5bb001f18ad065eb669f380cd32db0d0bfc))
+* update toHaveNoAxeViolations and remove skips ([#5955](https://github.com/carbon-design-system/ibm-products/issues/5955)) ([694128b](https://github.com/carbon-design-system/ibm-products/commit/694128bb707a2400d67e6618af3fea3f3ffba86f))
+
+
+
+
+
 # [1.9.0](https://github.com/carbon-design-system/ibm-products/compare/jest-config-ibm-cloud-cognitive@1.9.0-rc.0...jest-config-ibm-cloud-cognitive@1.9.0) (2024-09-25)
 
 **Note:** Version bump only for package jest-config-ibm-cloud-cognitive
--- a/config/jest-config-ibm-cloud-cognitive/index.js
+++ b/config/jest-config-ibm-cloud-cognitive/index.js
@@ -8,43 +8,7 @@
 'use strict';
 
 module.exports = {
-  collectCoverageFrom: ['src/**/*.js', '!**/*.stories.js'],
-  coverageReporters: [
-    [
-      'html',
-      {
-        skipEmpty: true,
-        // this next part doesn't actually seem to work
-        // see https://github.com/facebook/jest/issues/9734
-        watermark: {
-          statements: [80, 100],
-          lines: [80, 100],
-          functions: [80, 100],
-          branches: [80, 100],
-        },
-      },
-    ],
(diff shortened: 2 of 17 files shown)


amal-k-joy (approved):
Good work

AlexanderMelox (approved):
LGTM!

wkeese:
Thanks @devadula-nandan, I confirm it works on https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid--infinite-scroll&globals=viewport:basic and https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid--with-virtualized-data&globals=viewport:basic.

Do you have an opinion about @carbon/react's scrolling strategy, which scrolls on the table wrapper via the [stickyHeader](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/DataTable/Table.tsx#L43)... vs. @carbon/ibm-products strategy of scrolling on thead/tbody?

devadula-nandan:
Hi @wkeese, here is my [playground](https://devadula-nandan.github.io/TD-TABLE/?path=/story/example-tdtable--basic&args=stickyHeader:!true;tableHeight:600px;tableWidth:800px&globals=dataTheme:light) while i was experimenting on sticky header with horizontal and vertical scroll, on a constrained table. its not a carbon example but
the div that contains table element needs to have overflow auto and the thead to have sticky and top-0, which i consider an ideal implementation
Screenshot 2024-11-26 at 6 57 01 PM


however carbon's implementation does have sticky and top 0 on thead, but columns doesnt have defined widths like tanstack does, hence we cant get horizontal scrolling to work effectively afaik
Screenshot 2024-11-26 at 7 05 25 PM

and coming to datagrid there is this dependency that got built up on certain implementations within datagrid, to an extent that we need to re architect and could also potentially bring regressions, hence we moved to tanstack-carbon way, in which u can use carbon's datatable features, and u can also bring in your own custom changes.
but i do agree overflow on tbody and thead and syncing them is an overkill

hope this helps

wkeese:
Thanks @devadula-nandan, that all makes sense, although it will personally take me a while to get used to a vertical scrollbar that extends up to the top of the headers.

I wonder why ibm-products Datagrid decided to break ranks with @carbon's DataTable and put separate scrolling on the table body.

But iIn any case, I agree that scrolling on the table itself (or a wrapper around the table) makes sense.  Your playground seems to work perfectly.

PR #6148, opened 2024-10-01
feat(`ibm-products-web-components`): setup new web component package and storybook

matthewgallo:
Addresses part of #6130
Addresses most of #6129

This is the initial setup for our web component package, `@carbon/ibm-products-web-components`. This PR includes a new storybook for the web components package as well as the Side panel from `@carbon/web-components`.

The testing setup is new/different compared to what we have in `packages/ibm-products` since these components need to be tested in a real browser. I've setup `vitest`, configuring it to use `playwright` and added the tests to be included in our ci check.

Follow up tasks after this PR is reviewed/merged:
- Setup new netlify site/deploy previews
- Confirm publishing is working as expected by doing a dry run release action
- Add tearsheet web component

#### What did you change?
`packages/ibm-products-web-components`
#### How did you test and verify your work?
I tested by verifying that:
- the web component storybook runs locally
- the web component storybook builds as expected
- running the build script for the new package
- confirming new testing setup runs and `side-panel.test.ts` passes as part of PR ci checks

PR #6150, opened 2024-10-02
fix: adds html output for coverage reporting

davidmenendez:
fixes small mistake i made when cleaning up the jest config. also includes skipping `.story` files, which apparently exist...

### DIFF
--- a/config/jest-config-ibm-cloud-cognitive/index.js
+++ b/config/jest-config-ibm-cloud-cognitive/index.js
@@ -8,7 +8,12 @@
 'use strict';
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


amal-k-joy (approved):

matthewgallo:
We have some AVT failures (I _think_ because of a new rule). You can look at this [commit](https://github.com/carbon-design-system/ibm-products/pull/6148/commits/048fca13ab5b9389c4686b9b0d6cca3759c128f0) and possibly bring in those changes to get it passing again until we can connect with someone regarding the `target_spacing_sufficient` rule.

matthewgallo (approved):

PR #6167, opened 2024-10-04
feat(cards): renames slug to aiLabel

### LINKED ISSUE #6094: ExpressiveCard / ProductiveCard: `slug` --> `aiLabel` renaming

AlexanderMelox:
Closes #6094 

This PR changes the prop `slug` to `aiLabel`.

#### What did you change?
Changed the prop and any variables that has the name slug, updated documentation and proptypes

Storybook


davidmenendez (dismissed):

packages/ibm-products/src/components/Card/Card.tsx
```diff
@@ -252,7 +258,7 @@ export const Card = forwardRef(
             [`${blockClass}__productive`]: productive,
             [`${blockClass}__clickable`]: clickable,
             [`${blockClass}__media-left`]: mediaPosition === 'left',
-            [`${blockClass}--has-slug`]: !!slug,
+            [`${blockClass}--has-slug`]: !!aiLabel || !!slug,
```
elycheea, line 261: I think we discussed replacing the classnames too since these are exposed to the consumers.

> Double check if _any styles reference the word slug_ and replace with aiLabel

```suggestion
            [`${blockClass}--has-aiLabel`]: !!aiLabel || !!slug,
```

packages/ibm-products/src/components/Card/CardHeader.tsx
```diff
@@ -118,7 +132,7 @@ export const CardHeader = ({
       <div
         className={cx([
           `${headerClass}-container`,
-          { [`${headerClass}-container--has-slug`]: !!slug },
+          { [`${headerClass}-container--has-slug`]: !!aiLabel || !!slug },
```
elycheea, line 135:
```suggestion
          { [`${headerClass}-container--has-aiLabel`]: !!aiLabel || !!slug },
```

elycheea (changes requested):
I believe we discussed in a previous call that we should rename the class names as well since these are exposed to the user.

I only called out the references I spotted, but we’d need to update the SCSS files too, of course.

(Third task in #6094.)

packages/ibm-products/src/components/Card/CardHeader.tsx
davidmenendez: yes you're absolutely correct. good call

packages/ibm-products/src/components/Card/Card.tsx
AlexanderMelox: okay got it

packages/ibm-products-styles/src/components/Card/_card.scss
elycheea, line 1: For duplicating the styles, I think it’d be better for us to group the selectors rather than maintaining them separately.

For example:

```scss
.#{$block-class}__header-container--has-slug,
.#{$block-class}__header-container--has-ai-label {
  width: 100%;
  padding-right: $spacing-07;
}
```

Reduces our duplication so that if there is a bug later, we will only need to change these once to reflect them for both `slug` or `aiLabel`.

AlexanderMelox: That makes a lot of sense....I'll go ahead and do that. Thanks 😅

elycheea (approved):

matthewgallo (approved):

PR #6226, opened 2024-10-16
fix(Datagrid): useflexresize infinite redraw

wkeese:
Refs #5920, #5646.

Fix regression from 9bd307e4836d986383a0464e799834712f3904e0 where Datagrid keeps redrawing.  On our unit tests we get the exception:

> Too many re-renders. React limits the number of renders to prevent an infinite loop.

#### What did you change?

Added correct dependencies to useEffect().

Tested locally.


makafsal (approved):
LGTM

anamikaanu96 (approved):
Thank you for your contribution! Approved 🎉

Issue #6189, opened 2024-10-08
Address CSP in Datagrid subcomponents


elycheea:
DatagridActions.js is negotiable. It’s meant to be an example, but we have heard instances where consumers will try to import this to use anyway.

- [ ] Include the respective component name in the forbid configuration inside package.json
- [ ] Run yarn lint and fix the errors
- [ ] Replace `style={{}}` usage with classNames where appropriate
- [ ] Set [dynamic styles](https://github.com/carbon-design-system/carbon/blob/main/docs/style.md#authoring-dynamicinline-styles) using `useIsomorphicEffect()`

```[tasklist]
### Tasks
- [x] DatagridExpandedRow.tsx
- [ ] DatagridRow.tsx (need forwardRef on Carbon’s `TableCell`)
- [ ] DatagridVirtualBody.tsx (styles breaking currently)
- [ ] DraggableElement.tsx
- [x] DatagridActions.js
```

anamikaanu96:
DatagridRow.tsx  -> For implementing forwardRef in table cell I have raised a [PR](https://github.com/carbon-design-system/carbon/pull/17760) in carbon for [issue](https://github.com/carbon-design-system/carbon/issues/17759)

anamikaanu96:
Two tasks(DatagridExpandedRow.tsx and DatagridActions.js) in this ticket is already merged with [PR](https://github.com/carbon-design-system/ibm-products/pull/6229).

As discussed for below tasks I have opened separate tickets and closing this issue.

- [ ] DatagridRow - [issue-6269](https://github.com/carbon-design-system/ibm-products/issues/6269)
- [ ] DraggableElement and DatagridVirtualBody - [issue-6270](https://github.com/carbon-design-system/ibm-products/issues/6270)

PR #6229, opened 2024-10-16
fix(Datagrid): resolve CSP in subcomponents

anamikaanu96:
Addressing 2 files from [6189](https://github.com/carbon-design-system/ibm-products/issues/6189)

Replaced inline style in DatagridExpandedRow.tsx
Removed styles from DatagridActions.js since its not getting applied anywhere

#### What did you change? 
packages/ibm-products/src/components/Datagrid/utils/DatagridActions.js
packages/ibm-products/src/components/Datagrid/Datagrid/DatagridExpandedRow.tsx


packages/ibm-products/src/components/Datagrid/utils/DatagridActions.js
```diff
@@ -101,7 +96,7 @@ export const DatagridActions = (datagridState) => {
         {!mobileToolbar ? (
           <>
             {renderFilterPanelButton()}
-            <div style={style}>
+            <div>
```
elycheea, line 99: Do we still need these empty `div`s? 🤔 Or are they safe to remove if they were only used for that `style` attribute.

elycheea (approved):

sangeethababu9223 (approved):

PR #6293, opened 2024-10-24
fix(UserAvatar): release review fixes

### LINKED ISSUE #6168: User avatar follow up fixes from release review
### Summary from #5544 
- Noticed that in Safari that the ring is missing. Should check with design if we intend for this to appear on click or just focus[^1].
- Theming works when wrapped in a <Theme> component but this isn’t apparent in Storybook. May need to add documentation on different ways to handle theming using UserAvatar[^2].
- User avatar SCSS files need to add imports for `Tooltip` and possibly `TooltipTrigger` in the `_carbon-imports.scss` and `_user-avatar.scss` files respectively.
- Code coverage look GOOD 🔥

### Tasks
- [ ] Check with design and address focus ring in Safari as needed
- [ ] Update stories to demonstrate theme handling
  - [ ] Add documentation in Storybook docs
- [ ] Add missing SCSS imports to `carbon-imports.scss` and `user-avatar.scss`
- [ ] Address console error around `renderIcon` prop type[^4]
- [ ] Add extra comments to clarify theme mixins/handling (for anyone coming in a few years later :))

[^1]: In Safari, focus state only appears on tab but not on click, which is inconsistent with other browsers. ([#5544 comment](https://github.com/carbon-design-system/ibm-products/issues/5544#issuecomment-2383529408))
[^2]: Suggestion on how we show `Theme` in Storybook based on Carbon React Storybook. ([#5544 comment](https://github.com/carbon-design-system/ibm-products/issues/5544#issuecomment-2384028989))
[^4]: https://github.com/carbon-design-system/ibm-products/issues/5544#issuecomment-2383583111

anamikaanu96:
Closes #6168

Addressed released review fixes.
Also regarding "focus ring" I discussed with Marion Hekeler(@marion-bruells) she mentioned it should work only on tabbing.

#### What did you change? 
packages/ibm-products/src/components/UserAvatar/UserAvatar.tsx
packages/ibm-products-styles/src/components/UserAvatar/_user-avatar.scss
packages/ibm-products/src/components/UserAvatar/UserAvatar.mdx
packages/ibm-products-styles/src/components/UserAvatar/_carbon-imports.scss

### DIFF
--- a/packages/ibm-products-styles/src/components/UserAvatar/_user-avatar.scss
+++ b/packages/ibm-products-styles/src/components/UserAvatar/_user-avatar.scss
@@ -65,6 +65,10 @@ $sizes: (
   }
 }
 
+.#{$block-class}__tooltip button {
+  pointer-events: none;
+}
+
 @mixin photo-styling {
   border-radius: 100%;
   object-fit: fill;
--- a/packages/ibm-products-styles/src/components/UserAvatar/_user-avatar.scss
+++ b/packages/ibm-products-styles/src/components/UserAvatar/_user-avatar.scss
@@ -12,6 +12,7 @@
 @use '@carbon/styles/scss/theme' as *;
 @use '@carbon/styles/scss/spacing' as *;
 @use '@carbon/styles/scss/type';
+@use '@carbon/react/scss/components/tooltip';
 
 // Other Carbon settings if needed
 // TODO: @use '@carbon/styles/scss/grid';
@@ -74,6 +75,7 @@ $sizes: (
   background-color: $color;
 }
 
+//this mixin allows you to set background color based on the lighter themes such as g10 and white
 @mixin setLightBg($order, $color) {
   :root .#{$block-class}--#{$order},
   .#{$carbon-prefix}--g10 .#{$block-class}--#{$order},
@@ -84,6 +86,7 @@ $sizes: (
   }
 }
 
+//this mixin allows you to set background color based on the darker themes such as g90 and g100
 @mixin setDarkBg($order, $color) {
   .#{$carbon-prefix}--g90 .#{$block-class}--#{$order},
   .#{$carbon-prefix}--g100 .#{$block-class}--#{$order},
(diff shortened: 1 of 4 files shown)


packages/ibm-products-styles/src/components/UserAvatar/_user-avatar.scss
```diff
@@ -64,6 +65,10 @@ $sizes: (
   }
 }
 
+.#{$block-class}__tooltip button {
+  pointer-events: none;
+}
+
```
elycheea, line 71: This actually prevents the tooltip from showing up all on hover.

I think it’s better to leave this as is than to add this. You can compare this to similar components like the [read-only `Tag`](https://react.carbondesignsystem.com/?path=/story/components-tag--read-only) from Carbon. These may be more specific to the browser defaults, which is okay.

packages/ibm-products-styles/src/components/UserAvatar/_user-avatar.scss
elycheea, line 1: I missed this before but in lines 61-64, I think we should add one more line —

```scss
outline-offset: 1px;
```
This will give us the same outline as the `OperationalTag`, which will give us better consistency and better proximity to the avatar itself.

image
image


```scss
.#{$block-class}__tooltip {
  &:focus-within .#{$block-class} {
    // stylelint-disable-next-line carbon/theme-token-use
    outline: 2px solid $focus;
    outline-offset: 1px;
  }
}
```

elycheea (changes requested):
A few small comments since `pointer-events` actually prevents the tooltip from working. I also found precedent in the read-only and interactive tags — I think it’s better to leave the browser default behavior in case. 🙂

But also think we can adjust the focus ring just a little bit after looking at those examples.

Lastly, I had some thoughts on showing theme this way as some of the theme stories in the `@carbon/react` Storybook do. But I also still have this stashed on a branch, so I’m happy to help open that as a followup PR to this one.

image

elycheea (approved):

amal-k-joy (approved):
LGTM

PR #6204, opened 2024-10-11
feat(Tearsheet): Move Tearsheet into @carbon/ibm-products-web-components

sangeethababu9223:
Closes #6183

Move Tearsheet from `@carbon/web-components` to `@carbon/ibm-products-web-components`

#### What did you change?

- Copied the `Tearsheet` folder from core carbon to `ibm-products-web-components`.
- Updates styles & Styles
- Add test cases for Tearsheet

#### How did you test and verify your work?
Storybook

PR #6294, opened 2024-10-24
fix: netlify check

davidmenendez:


davidmenendez:
issue was identified- the `netlify.toml` in the web-components package was causing the issue and simply needed to be removed. should be resolved in this PR https://github.com/carbon-design-system/ibm-products/pull/6204

closing this out.

PR #6296, opened 2024-10-25
fix: dynamically importing lottie web package

### LINKED ISSUE #6243: Test dynamic import of `media` prop
```[tasklist]
### Tasks
- [ ] test the dynamic import of lottie web package `SteppedAnimatedMedia`
- [ ] can we dynamically import `SteppedAnimatedMedia` to coachmarkOverlayElements
- [ ] create CRA template and test the difference in sizing (not an issue with vite, but is in webpack)
```

amal-k-joy:
Closes #6243 

lottie-web package will be dynamically imported at the time of usage only. That is only when media prop is passed with animation paths.

#### What did you change?
used `import().then()` syntax to dynamically import the package
local

### DIFF
--- a/packages/ibm-products/src/components/SteppedAnimatedMedia/SteppedAnimatedMedia.tsx
+++ b/packages/ibm-products/src/components/SteppedAnimatedMedia/SteppedAnimatedMedia.tsx
@@ -13,7 +13,7 @@ import React, {
   useState,
   MutableRefObject,
 } from 'react';
-import lottie, { type AnimationItem } from 'lottie-web';
+import { type LottiePlayer, type AnimationItem } from 'lottie-web';
 import clamp from 'lodash/clamp';
 // Other standard imports.
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


packages/ibm-products/src/components/SteppedAnimatedMedia/SteppedAnimatedMedia.tsx
```diff
@@ -84,25 +84,27 @@ export const SteppedAnimatedMedia = React.forwardRef(
...
-        });
-        prefersReducedMotion
-          ? animRef.current?.goToAndStop(0)
-          : animRef.current?.goToAndPlay(0);
-      }
+      import('lottie-web').then((lottie) => {
+        const prefersReducedMotion = window?.matchMedia
+          ? window.matchMedia('(prefers-reduced-motion: reduce)').matches
+          : true;
+        if (localRefValue) {
```
davidmenendez, line 91: i think this should be moved out before the import. the lottie code isn't even run unless `localRefValue` is true, so as an extra level of optimization i would wrap the import with this if statement.

davidmenendez (changes requested):
looks good just one comment.

also as part of the verification process could you test and shows the results of the bundle size to show that the bundle is smaller? maybe using something like [webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer) if it's helpful

packages/ibm-products/src/components/SteppedAnimatedMedia/SteppedAnimatedMedia.tsx
amal-k-joy: Hi David, 
Thats make sense, however `localRefValue` will never be null with the implementation here.
```
 const backupRef = useRef<HTMLDivElement>(null);
    const localRef = ref ?? backupRef;
```

amal-k-joy:
> looks good just one comment.
> 
> also as part of the verification process could you test and shows the results of the bundle size to show that the bundle is smaller? maybe using something like [webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer) if it's helpful

Hi David,
I was not able to replicate the original issue the user was referring. I couldnt see lottie-web being bundled when none of the related components (Coachmark, inLineTip) is used. I used a package `source-map-explorer ` to explore the bundles.
### With lottie-web
image
**without lottie-web**
image

When `import.then()` is used, lottie-web will be bundled separately and get downloaded at the time of usage

https://github.com/user-attachments/assets/a5a70c7c-91fb-4ecf-9558-21e7c304a9d4

davidmenendez:
@matthewgallo i believe we agreed that instead of trying to implement the dynamic importing that we would just cut lottie in favor of allowing the consumer to choose to to implement the animation, right?

matthewgallo:
Yep, that's what I also remember after we last reviewed this. And also because `SteppedAnimatedMedia` and the other components that use it are experimental/preview we have a little more flexibility to just remove it (`lottie-web`).

davidmenendez:
@amal-k-joy can you open up a new issue for removing the lottie dependency? we can keep this PR up for now unless anyone else wants to close it

amal-k-joy:
> #6243

@matthewgallo , @davidmenendez ,
I have opened [this issue](https://github.com/carbon-design-system/ibm-products/issues/6366) to decouple lottie-web from our component. Currently we have exposed a `media` prop that expects either a render callback(media.render) that returns a component( ``) or filepaths(media.filepaths) with animation data.

The new approach I am thinking is to deprecate the `media` prop and introduce an new callback prop called `renderMedia` which can return any react component. So if user needs lottie animation, they can import the package from their end and pass down the animated component as it is.

Here in this example , `<SteppedAnimatedMedia>` that has lottie-web will be imported only in stories . User can either use `<SteppedAnimatedMedia>` or their own custom component. 

```
  <InlineTip {...args}  action={selectedAction} renderMedia={() => <SteppedAnimatedMedia
          filePaths={[InlineTipAnimation]}
          playStep={1}
        />} />

<InlineTip {...args}  action={selectedAction} renderMedia={() => <img alt="" src={InlineTipImage} />} />

<InlineTip {...args}   action={selectedAction}  />  
```

Please share your thoughts.

elycheea:
Thanks @amal-k-joy for opening the follow up and documenting it nicely! We can move further discussion to the issue. 😀

PR #6389, opened 2024-11-07
feat(MultiAddSelect): add floating ui to sort and filter buttons

### LINKED ISSUE #6328: Add floating ui to Sort section and filterByOpts in MultiAddSelect component
the requirement for this issue is to migrate `overflowMenu` usage from `@carbon/react` in our `MultiAddSelect` component from v11 to v12 via feature flag `enableV12Overflowmenu`.

current v11 usage screenshots
<table>
  <tr>
    <td>Image 1</td>
    <td>Image 2</td>
  </tr>
</table>

### Why is this necessary?  
To stay ahead as early adopters, identify potential conflicts with the upcoming major version, and ensure we receive the latest and ongoing support for both Floating UI and the new version.

### Why is this blocking us?  
The technical challenge lies in the shared `MenuItem` component between the `Menu` and the v12 `OverflowMenu`. The behavior of this shared child differs when used with the `OverflowMenu` parent, specifically preventing the `renderIcon` prop from working unless the parent is set to `mode="basic"`, a configuration only supported by the `Menu` component. 

Additionally, the `label` prop and `shortcut` prop for `MenuItem` in v12's `OverflowMenu` is restricted to a string type, which limits our ability to render icons alongside the label. In contrast, the previous v11 `OverflowMenu` used its own child component, `OverflowMenuItem`, which allowed more flexibility by supporting a React node for the `itemText` prop. This enabled us to render custom components, such as icons, within the menu.

this is a hard block unless the v12 overflowMenu provides a way to get `renderIcon` to work, and checkbox on `menuItem` to work

### Acceptance criteria
- [ ] Update sort to use v12 OverflowMenu with text only option
   - [ ] Remove `ArrowUp` and `ArrowDown` from `AddSelectSort`
- [ ] Refactor `AddSelectColumn` to use [tab tip popover](https://carbondesignsystem.com/components/popover/usage/#tab-tip) with checkboxes

devadula-nandan:
Closes #6328 

This pr adds following upgrades to `MultiAddSelect`
- Refactors `filter` to use `popover` as per design recommendation (discussed in the issue), and enables floating ui styles by enabling flag `enableV12DynamicFloatingStyles`
- Enables `enableV12Overflowmenu` on `sort` overflow menu, refactors to new api changes.
- Fixes `IconButton` label tooltip clipping situation

https://github.com/user-attachments/assets/85315c8d-d0f8-4a49-be2f-f92de28a2c77

related
https://github.com/carbon-design-system/carbon/issues/18872


packages/ibm-products/src/components/AddSelect/AddSelectColumn.js
```diff
@@ -139,33 +151,53 @@ export let AddSelectColumn = ({
...
+                  <IconButton
+                    size="md"
+                    id={filterBtnId}
+                    hasIconOnly
+                    aria-expanded={filterOpen}
+                    badgeCount={filters.length > 0 ? 0 : undefined} // setting to 0 renders an empty dot
+                    kind="ghost"
+                    label={filterByLabel || 'Filter'}
+                    onClick={() => {
+                      setFilterOpen(!filterOpen);
```
matthewgallo:
```suggestion
                      setFilterOpen(prev => !prev);
```

packages/ibm-products/src/components/AddSelect/AddSelectColumn.js
```diff
@@ -139,33 +151,53 @@ export let AddSelectColumn = ({
...
+                  }}
+                  onRequestClose={() => setFilterOpen(false)}
+                  open={filterOpen}
+                >
+                  <IconButton
+                    size="md"
+                    id={filterBtnId}
+                    hasIconOnly
+                    aria-expanded={filterOpen}
+                    badgeCount={filters.length > 0 ? 0 : undefined} // setting to 0 renders an empty dot
```
elycheea, line 179: Ooh, this is a nice addition! Is _that_ why you were looking at the badge/icon indicators? 😉

matthewgallo (dismissed):

matthewgallo:
Looks like there's a formatting [issue](https://github.com/carbon-design-system/ibm-products/actions/runs/15446201800/job/43476464865?pr=6389)

packages/ibm-products/src/components/AddSelect/AddSelectSort.tsx
```diff
@@ -33,23 +37,13 @@ export const AddSelectSort = ({
...
           {
             id: `${cur}-asc`,
-            itemText: (
-              <>
-                <ArrowUp size={16} />
-                {cur}
-              </>
-            ),
+            label: cur,
             direction: 'asc',
```
elycheea, line 41: I think this is what threw me off when I originally looked at this issue. It looks like the `direction` labels are hardcoded.

Since the original implementation used arrows, we probably didn’t flag this for g11n/i18n initially.

matthewgallo (approved):

packages/ibm-products/src/components/AddSelect/types/index.ts
```diff
@@ -56,5 +56,6 @@ export interface SortOption {
   id?: string;
   direction?: string;
   attribute?: string;
-  itemText?: ReactNode;
+  label?: string;
+  shortcut?: ReactNode;
```
elycheea: Are we using `shortcut` anywhere or is this left from an earlier implementation?

packages/ibm-products/src/components/AddSelect/AddSelectSort.tsx
elycheea: We could alternatively open a followup issue for this improvement since in my previous discussions with @davidmenendez, we thought it would likely affect the API a little bit to provide a `directionLabel` prop or something.

packages/ibm-products/src/components/AddSelect/types/index.ts
devadula-nandan: cleaned up

packages/ibm-products/src/components/AddSelect/AddSelectSort.tsx
devadula-nandan: we have 2 options here, 
1) we could hardcode the direction labels to standard unicode characters for up and down arrow
↑↓, so they still comply with i18n for being language-neutral
here is how it looks when added before or after the label
Screenshot 2025-06-08 at 3 02 13 PM

2) user could pass the direction (translation or language-neutral unicode) inside label so we don't have to hardcode direction

shall i add the changes following the cleaner option 1?

amal-k-joy (approved):
LGTM

PR #6401, opened 2024-11-11
chore: deprecate edit and update

devadula-nandan:
Closes #6321 

deprecates below components

edit side panel
edit fullpage
edit tearsheet
edit tearsheetnarrow
edit update cards

#### What did you change?
added deprecation annotation, deprecated the component, and added a notice in story docs, and expected consol warning in test.

packages/ibm-products/src/components/CreateFullPage/_storybook-styles.scss
packages/ibm-products/src/components/EditFullPage/EditFullPage.stories.jsx
packages/ibm-products/src/components/EditFullPage/EditFullPage.test.js
packages/ibm-products/src/components/EditFullPage/EditFullPage.tsx
packages/ibm-products/src/components/EditSidePanel/EditSidePanel.stories.jsx
packages/ibm-products/src/components/EditSidePanel/EditSidePanel.test.js
packages/ibm-products/src/components/EditSidePanel/EditSidePanel.tsx
packages/ibm-products/src/components/EditTearsheet/EditTearsheet.test.js
packages/ibm-products/src/components/EditTearsheet/EditTearsheet.tsx
packages/ibm-products/src/components/EditTearsheet/preview-components/MultiFormEditTearsheet.js
packages/ibm-products/src/components/EditTearsheetNarrow/EditTearsheetNarrow.stories.jsx
packages/ibm-products/src/components/EditTearsheetNarrow/EditTearsheetNarrow.test.js
packages/ibm-products/src/components/EditTearsheetNarrow/EditTearsheetNarrow.tsx
packages/ibm-products/src/components/EditUpdateCards/EditUpdateCards.stories.jsx
packages/ibm-products/src/components/EditUpdateCards/EditUpdateCards.test.js
packages/ibm-products/src/components/EditUpdateCards/EditUpdateCards.tsx
and respective avt files

#### How did you test and verify your work?
storybook, yarn test

PR #6419, opened 2024-11-12
chore: datagrid deprecation notice

### LINKED ISSUE #6174: Update Storybook to show deprecation (Datagrid)
A utility component has deprecation, we can follow:

- Add migration notice in Datagrid.mdx
- Add deprecation notice in docs-page ([see example](packages/ibm-products/src/components/UserProfileImage/UserProfileImage.docs-page.js))
- Add `.deprecated` to Datagrid
- Verify tests are passing as adding the deprecation notice will typically lead to failures from console logs / warnings

davidmenendez:
Closes #6174

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid.docs-page.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid.docs-page.js
@@ -13,7 +13,7 @@ import toolbarScreenshot from './storybook-assets/datagrid-actions-example.png';
 
 const DocsPage = () => (
   <StoryDocsPage
-    deprecationNotice={`\`Datagrid\` will be deprecated in the next major version and moving to an example based approach. For more information please refer to the [migration docs](https://github.com/carbon-design-system/tanstack-carbon).`}
+    deprecationNotice={`\`Datagrid\` is deprecated and will be moving to an example-based approach. For more information, please refer to the [migration docs](https://github.com/carbon-design-system/tanstack-carbon).`}
     blocks={[
       {
         title: 'Getting started',
--- a/packages/ibm-products/src/components/Datagrid/Datagrid.docs-page.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid.docs-page.js
@@ -16,11 +16,6 @@ const DocsPage = () => (
     deprecationNotice={`\`Datagrid\` will be deprecated in the next major version and moving to an example based approach. For more information please refer to the [migration docs](https://github.com/carbon-design-system/tanstack-carbon).`}
     blocks={[
       {
-        title: 'Deprecation notice',
-        description:
-          '`Datagrid` will be deprecated in the next major version and moving to an example based approach. For more information please refer to the [migration docs](https://github.com/carbon-design-system/tanstack-carbon).',
-      },
-      {
         title: 'Getting started',
         description: `You can use the \`Datagrid\` component to create various kinds of Carbon data tables. Simply import the \`Datagrid\` along with any accompanying hooks (listed below) depending on what extensions you'd like to be included.`,
         source: {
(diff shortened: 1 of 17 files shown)


matthewgallo:
Nice! I left some [comments](https://github.com/carbon-design-system/ibm-products/pull/6401) on @devadula-nandan's deprecation PR which I think will be applicable here too. Can you also drop in the `Annotation` component which will visually indicate in storybook that it is deprecated?

You can add a `decorators` property in the story default export so it wraps every story.

packages/ibm-products/src/components/Datagrid/Datagrid.docs-page.js
```diff
@@ -13,6 +13,7 @@ import toolbarScreenshot from './storybook-assets/datagrid-actions-example.png';
 
 const DocsPage = () => (
   <StoryDocsPage
+    deprecationNotice={`\`Datagrid\` will be deprecated in the next major version and moving to an example based approach. For more information please refer to the [migration docs](https://github.com/carbon-design-system/tanstack-carbon).`}
```
matthewgallo, line 16: Lets change this verbiage (`in the next major`) because we are announcing deprecation beginning Nov 20. And maybe add that support will remain into the next major (with sev 1 and security vul support only in v3).

elycheea:
```suggestion
    deprecationNotice={`\`Datagrid\` is deprecated and will be moving to an example-based approach. For more information, please refer to the [migration docs](https://github.com/carbon-design-system/tanstack-carbon).`}
```

Details on major versions and sev/security should be captured in the `tanstack-carbon` README.

davidmenendez:
@elycheea @matthewgallo wording updated 👍

packages/ibm-products/src/components/Datagrid/Datagrid.stories.jsx
```diff
@@ -56,6 +57,27 @@ export default {
...
   },
   excludeStories: ['getBatchActions'],
+  decorators: [
+    (story) => (
+      <div>
+        <Annotation
+          type="deprecation-notice"
+          text={
+            <div>
+              This component is deprecated and will be removed in the next major
```
matthewgallo, line 67: Found one more spot to use the updating wording from @elycheea

matthewgallo: Or maybe just a slight update, since we'll be keeping it in the next major

davidmenendez: updated it to match the other wording

elycheea: Haha, I removed the “major” language. I expect it to still be around in v3 but remain deprecated, but if we do see usage drop in telemetry, that would change things. ;)

matthewgallo: Yep, that aligns with the changes David made [here](https://github.com/carbon-design-system/ibm-products/pull/6419/commits/f37e48a11d07a6f34cd65b6fa7c1d0409a2dfe62)

elycheea (approved):
I don’t love the nested Datagrid folder structure, but better than losing the docs tab. 🥲

Not sure if we need to add the deprecation notices to the extension stories too, but including them in the main Datagrid stories is definitely priority.

davidmenendez:
> I don’t love the nested Datagrid folder structure

nor do i. the problem is that the helper function assumes a certain story structure `{ibm products | experimental} / {components | patterns | onboarding} / {name of thing}`

since the deprecation section isn't organizing everything into components or patterns it causes the helper to break when there's only two folder items 🫠

`{deprecation} / {name of thing}` ❌ 

`{deprecation} / {name of thing} / {name of thing}` ✅

matthewgallo:
Lets take a look at refactoring `story-helper` once we get this out, whether we decide to categorize `Deprecated` with subdirs `Components` and `Patterns` or we just update `story-helper` to handle the different structure.

matthewgallo (approved):

PR #6446, opened 2024-11-15
fix: percy temporary fix

devadula-nandan:
fixes percy temporarily

#### What did you change?
ci.yml

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -201,7 +201,7 @@ jobs:
       - name: Run VRT
         working-directory: packages/core
         env:
-          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN_IBM_PRODUCTS }}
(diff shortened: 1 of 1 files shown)


kennylam (approved):
This is fine for now. We can explore a better solution moving forward.

matthewgallo (approved):

davidmenendez:
@kennylam @devadula-nandan can we capture this in an issue so we don't lose track of it? i'd hate to see us get ding'ed in a security check for having a hardcoded token that we forgot about! 😱

kennylam:
> @kennylam @devadula-nandan can we capture this in an issue so we don't lose track of it? i'd hate to see us get ding'ed in a security check for having a hardcoded token that we forgot about! 😱 


Yep, we're going to meet next week to discuss an alternate solution I've used for this. I've already mentioned it to @matthewgallo, I'll add you as well.

PR #6463, opened 2024-11-18
fix: remove tooltip from tagoverflow

### LINKED ISSUE #5736: Tagoverflow: dom nesting and tooltip accessibility
### Package

Carbon for IBM Products

### Description

i noted that the [TagOverflow](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-tag-overflow-tagoverflow--tags-with-truncation&globals=viewport:basic) story is displaying a invalid dom error. `button cannot be a descendent of button`. this notification appears to only pop up locally.

the issue stems from `Tooltip` being wrapped around the `Tag` components. because tooltip is a button and tag's have an `onClick` mechanism they are rendered as buttons.

in addition to this in and of itself being an accessibility issue, this is also tied to our ongoing efforts to remove unnecessary usage of tooltips and `title` attributes. my suggestion is to remove the tooltip entirely. since carbon tags already utilize `text-overflow` i think what we should do is allow the ellipsis to show as usual and then adjust our popover detection mechanism to check for truncated values and display them in full in the modal. i believe that because this offers a way for users who need screen readers a way to access the full value that this fix would be accessible.

### Component(s) impacted

[TagOverflow](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-tag-overflow-tagoverflow--tags-with-truncation&globals=viewport:basic)

### Browser

Chrome

latest

### Suggested Severity

None

### Product/offering

c4p

NA

_No response_

_No response_

### Code of Conduct

davidmenendez:
Closes #5736 

fixes an issue where the `Tag` component was being wrapped by a `button` from `Tooltip` which was causing an `invalid dom nesting` error. since carbon tags include `title` they should already be screen reader accessible when they are being truncated.


makafsal (changes requested):
According to Codecov, I think we need to write some tests.

amal-k-joy (changes requested):
Looks good, but found one glitch.
With this change we are unable to tab navigate through the tags. Since you removed Tooltip wrapper , removed button elements as well  that made it not  focusable i guess.

davidmenendez:
@amal-k-joy this appears to be working as intended as they are [read only](https://react.carbondesignsystem.com/?path=/story/components-tag--read-only)

davidmenendez:
@makafsal coverage should be acceptable now 👍

amal-k-joy (approved):
Looks good

makafsal (approved):

PR #6464, opened 2024-11-18
chore: remove lint configs from package and move into own files

### LINKED ISSUE #6303: Remove `eslint` and `commitlint` config from root `package.json`
We should move our `eslint` and `commitlint` configs into their own config files and remove from root `package.json`. Specifying an `eslint` config in the root `package.json` is no longer supported in v9 either so it will help us in our migration to eslint v9 which has been coming up in our upgrade dependencies PR.

davidmenendez:
Closes #6303 

removes the following configs from `package.json` and moves them into their own separate files in preparation of updating to eslint v9

also cleaned up some old unused config from `.stylelintrc.js`


matthewgallo (approved):
Pulled changes down locally, everything looks good! Noticed that core is using `eslint.config.mjs` (esm) for their eslint config, I'm guessing we should be able to do the same when we update to eslint 9.

davidmenendez:
> `eslint.config.mjs` (esm)

yep! next step is to update that file and config as part of the upgrade

elycheea (approved):

elycheea:
@davidmenendez I had some of the eslint 9 changes in [this branch](https://github.com/elycheea/ibm-products/tree/eslint-v9).

davidmenendez:
@elycheea ok awesome! these changes should be compatible with that branch and shouldn't introduce too many conflicts based on what i'm seeing. i guess i just did the not entirely necessary step of decoupling the configs before updating where you just did both in that branch 😂

PR #6477, opened 2024-11-20
feat: Decouple lottie-web dependency from our repo

### LINKED ISSUE #6366: Decouple lottie-web dependency from our repo
Currently we have exposed a media prop that expects either a render callback(`media.render`) that returns a component( ) or filepaths(`media.filepaths`) with animation data.

The new approach  is to deprecate the `media` prop and introduce an new render prop called `renderMedia` which is a callback that can return any react component. So if user needs lottie animationed component, they can import the package from their end and can return the animated component as it is.

We have the component` <SteppedAnimatedMedia>` that has implemented lottie-web. User can either import and use that wrapper component or define their own custom component.
#### Usage at adopter
```
  <InlineTip  renderMedia={(params) => <SteppedAnimatedMedia
          filePaths={[InlineTipAnimation]}
          playStep={1}
        />} />

<InlineTip  renderMedia={(params) => <img alt="" src={InlineTipImage} />} />

<InlineTip    action={selectedAction}  />  
```
#### Sample implementation in component
`{
            renderMedia && <>{renderMedia({className:`${blockClass}__media`})}</>
          }`

Here `<SteppedAnimatedMedia>` will still be part of our repo , but will not be imported in any of our components , there by should be excluded from bundling. 

- [x] Deprecate `media` prop with lots of warnings
- [ ] Replace `SteppedAnimatedMedia` with `renderMedia` prop

#### Followup later
- [ ] #6434

amal-k-joy:
Contributes #6366 

Introduced new prop renderMedia to replace media prop
Deprecated media prop

local


davidmenendez (changes requested):
saw the code coverage dropped some. can you implement some tests for this? thanks!

Hey , added more coverage.

devadula-nandan:
hi @amal-k-joy,
i noticed the animations were running twice, could you check that?
rest looks good to me.

https://github.com/user-attachments/assets/b5629ef8-d1dd-4d87-87e7-d26b4b86eb38

amal-k-joy:
> hi @amal-k-joy, i noticed the animations were running twice, could you check that? rest looks good to me.
> 
>  Screen.Recording.2024-11-28.at.4.14.33.PM.mov

Thanks for finding that out. Fixed that.

devadula-nandan (approved):
looks good

davidmenendez (approved):

Issue #6248, opened 2024-10-18
[a11y]: API Key Modal, Screen reader not identified the Read only state


anamikaanu96:
### Package

@carbon/ibm-products

### Browser

Firefox

### Operating System

Windows

### Package version

v2.52.0-rc.0

### Automated testing tool and ruleset

Manual testing

### Assistive technology

JAWS

### Description

The Api key input field is non-editable/read-only. Same state info is not conveyed by the Screen reader. If selected all content in input field and given backspace, Screen reader announces as "Selection Deleted" whereas visually it is not.

Expected behaviour : Screen reader expected to announce the "Read only state" on focus(Tab/Arrow), ensure the same.

### Reproduction/example

https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-generating-an-api-key-apikeymodal--generate&globals=viewport:basic

### Steps to reproduce

This was reported in all variants of API Key Modal.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate problems

PR #6481, opened 2024-11-20
fix(APIKeyModal): improve screen reader announcement

### LINKED ISSUE #6253: [a11y]: Improve screen reader annoucement for API Key Modal
### Package

@carbon/ibm-products

### Browser

Firefox

### Operating System

Windows

### Package version

v2.52.0-rc.0

### React version

_No response_

Manual Testing

### Assistive technology

JAWS

### Description

Issues

1. Screen reader not identified the visually presented content 

   Issue :
   Once the Password  field is Masked :  the API Key shows as "******" for visual users , however on Tab  focus - Screen reader announces as- 
   Blank ,Screen reader not informing  what is presented  in the field.

   Expected behaviour : Screen reader expected to convey the Visual Info from the page, the mentioned content must be Navigable/Readable 
    with standard methods.

2. The users assistance Text , given along with  the Password field is not announced on Tab focus.

     Issue :
     The information "This is your unique API key…" is not announced when the API key input field receives focus.

     Expected behaviour : Recommend to associate the  mentioned instruction  to the Password Edit field , screen reader  expected to 
     understand this info focusing on the Same field.

3. Error message  not announced instantly

     Issue : "Failed to create API key" error message not announced immediately by the screen reader when the "Generate Key" button is 
     activated.
    
     Expected behaviour : Make sure the screen reader announce the Error information instantly.

4. Screen reader announces incorrect label
     
      Issue : If we activate "Previous" button on "Choose which resources the API will have access to" (second)dialog, focus falls on "Close" 
      button of the previous dialog. Screen reader still announces as "Previous" button instead of "Close button"
    
     Expected behaviour : Ensure that screen reader announces appropriate label to convey correct information to the screen reader user.

5. Heading change not announced instantly

      Issue : After editing & saving the key, "API Key Successfully saved" heading replaces "Save an API key" heading. This change in heading is 
      not conveyed by SR.
    
     Expected behaviour : Make sure the screen reader announce the changed information  instantly.

6. Heading change & success message  not announced instantly .

      Issue : 
      1. After editing & saving the key, "API Key Sucessfully saved" heading replaces "Save an API key" haeding.
      2. Edited Successfully is the success message appears on screen which is not announced as well
          This change in heading/success message is not conveyed by SR.
    
     Expected behaviour : Ensure that either of the success message is announced by the SR immedialtely as it appears on screen.

7. Show/Hide label announced inconsistently

      Issue : The Show/Hide Key label is announced inconsistently,
      1. Password input circle bullet circle bullet
      2. Show key announces as Hide Key or vice versa.
    
     Expected behaviour : Ensure that the Show/Hide key button is announced appropriately

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://ibm-products.carbondesignsystem.com/?path=/docs/ibm-products-components-generating-an-api-key-apikeymodal--docs

### Steps to reproduce

1. Screen reader not identified the visually presented content 
   Reported in all variants of API Key Modal 

2. The users assistance Text , given along with  the Password field is not announced on Tab focus.
     Reported in API Key Modal /Instant Generate 

3. Error message not announced instantly
    Reported in API Key Modal -  Generate with Error

4. Screen reader announces incorrect label
    Reported in API Key Modal -  Custom Generate

5. Heading change not announced instantly
    Reported in API Key Modal - Edit

6. Heading change & success message not announced instantly 
    Reported in API Key Modal - Custom Edit

7. Show/Hide label announced inconsistently
    Reported in all variants of API Key Modal 

### Code of Conduct

anamikaanu96:
Closes #6253 #6248

Improved the screen reader announcement.
Issue 1 : This issue is occur in PasswordInput component from Carbon, attaching the recording of PasswordInput.
Issue 2 : The information "This is your unique API key…" was changed from body of APIKeyDownloader to helperText of PasswordInput. Had a discussion with @davidmenendez on this.
Issue 3: Working as expected
Issue 4 to 6 : The success message is currently showing in heading, in that case it won't read success message immediately if we give aria-live or alert ( Also if we follow reading heading approach, in Custom Edit variant we have both Heading change & success message at a time so it will ignore reading heading since giving more importance to success message)
Had a discussion with @davidmenendez and  Fix done : Added a successMessage below
Issue 7 :  Divya Vani Bollayyagari from Raghu's team confirmed that it is working as expected now

Required design input on issue 2 and 4-6
Carbon's PasswordInput https://github.com/user-attachments/assets/163f4d12-e6d5-478b-a254-a7df3f6f907a

#### What did you change? 
packages/ibm-products/src/components/APIKeyModal/APIKeyDownloader.js
packages/ibm-products/src/components/APIKeyModal/APIKeyModal.stories.jsx
packages/ibm-products/src/components/APIKeyModal/APIKeyModal.test.js
packages/ibm-products/src/components/APIKeyModal/APIKeyModal.tsx
packages/ibm-products/src/components/APIKeyModal/APIKeyModal.types.ts


sangeethababu9223:
@matthewgallo ,
Just noticed the `needs design opinion`. Not approving for now.

@oliviaflory,
Could you please look into this?

elycheea (commented):
| Before | After |
---------|--------
image | image

So I see the design changes and understand why we want a design opinion on them. Can you and @davidmenendez provide more clarity on what you tried before changing the design (1. move to helper text, 2. repeated success message)?

I would have expected us to use a combination of `aria-live` and `aria-relevant`[^1]. I do think it’d be nice if we’re able to prevent redundancy in the pattern rather than repeating the success message twice.

[^1]: [`aria-live`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live) and [`aria-relevant`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-relevant)

davidmenendez (changes requested):
waiting for design feedback & approval.

davidmenendez:
lets bring this up with michael gower at the next carbon access guild 👍

anamikaanu96:
Hey all, thanks for reviewing!

Initially I tried to make the title readable by screen reader/jaws, in the current implementation, title of the modal header is wrapped inside a <div>. This means that applying attributes like aria-live or aria-relevant will only affect the <div> itself and will not trigger the title announcement.

`  <div className={headerClass} {...rest} ref={ref}>
        {label && <h2 className={labelClass}>{label}</h2>}
        {title && <h2 className={titleClass}>{title}</h2>}
        {children}`
        
To address this, I tried passing the title as a child of the ModalHeader and applied the necessary accessibility attributes. While in this approach, the successTitle is announced but it does not happen immediately. I believe the successMessage should be announced promptly as soon as it appears on the screen.

Also its valid that we are duplicating the successMessage by displaying it both in the title and at the bottom of the modal.
At the same time in current behaviour of API Key Modal, there are some inconsistency where only error message and editSuccess message is displayed at bottom.

So for now as I mentioned in the Engineering huddle, the alternative solution I proposed is to avoid changing the title for a success. Instead, we will update only the success message displayed below.

Regarding the helper text, it was previously being passed in the body of APIKeyDownloader, but it functions more like a helper text. Additionally, Raghu requested that the text needs to be announced when the TextInput gains focus. For this reason, I moved it as a helper text.

As we discussed in call, lets bring this up with Michael Gower for more clarity!

@oliviaflory The "more content menu" is included as part of the VoiceOver narration, I tried in Carbon TextInput too.

Proposed Solution: Please check deployment preview

davidmenendez:
based on the carbon accessibility meeting today, this solution is the preferred method. Michael Gower agrees that having the header update isn't the most intuitive and that utilizing the existing area that we are using for status updates (failure, loading) is a better place for that information.

that being said, it sounds like there's an interest in having a larger conversation regarding how status updates are handled in modals. even across the modals in C4P (export, import, create, delete) there's an inconsistency between how each handles success.

export:

Screenshot 2024-12-02 at 3 06 55 PM

also should be noted that the error state for export doesn't appear to be working

apikey:

Screenshot 2024-12-02 at 3 07 17 PM

create & delete: neither have a success, failure, loading, or error state, unlike the other modals which have specific props for displaying and handling those states.

might be something we want to look at 👍 🤔  

as far as this specific issue and PR are concerned the solution Anamika has submitted is Gower approved.

anamikaanu96:
Thank you @davidmenendez for discussing this with Michael Gower!

I updated the style of status message with a `CheckmarkFilled` icon as suggested by @oliviaflory. Now its looking similar to Export Modal success message. Additionally I have deprecated two props `editSuccessTitle` and `generateSuccessTitle` instead introduced `editSuccessMessage` and `generateSuccessMessage`.

davidmenendez:
tagging @RichKummer for any other design insight. he mentioned to me removing the info icon next to `download` and using the check mark, but this new update adds it next to the success message. thoughts?

Screenshot 2024-12-03 at 1 36 23 PM

RichKummer (dismissed):
Tried out the latest and I think this works well visually until we've had that larger conversation about status treatment that @davidmenendez mentioned in his last comment. It also looks like VoiceOver calls out the success status after clicking the "Generate API key" button – it should hopefully be working the same in JAWS.

Thanks for the fixes @anamikaanu96 @davidmenendez

RichKummer (approved):

davidmenendez (approved):
🎉

PR #6508, opened 2024-11-26
fix(EmptyState): empty states throws hydration errors in next js due to dynamic id

### LINKED ISSUE #6315: [Hydration Error with EmptyState]: Nested NoDataEmptyState throws hydration errors on Next JS 15
### Package

Carbon for IBM Products

### Description

When using NoDataEmptyState as a nested component, a hydration error is thrown
 Image

```
              {contentSwitcherIndex == 2 ? (
                <>
                  {array.length > 0 ? (
                    <Grid>
                      {array.map((s) => (
                        <Column
                          key={"Column" + s.s_name}
                          xlg={5}
                          lg={5}
                          md={4}
                          sm={4}
                        >
                          <ServiceTile
                            key={s.s_name}
                            service={s}
                          />
                        </Column>
                      ))}
                    </Grid>
                  ) : (
                    <NoDataEmptyState
                      style={{ padding: "2rem" }}
                      illustrationDescription="No data available"
                      subtitle="No data available at this time. If you think this is a mistake, please contact admin."
                      title="No data available"
                    />
                  )}
                </>
              ) : null}
```

### Component(s) impacted

[NoDataEmptyState](https://ibm-products.carbondesignsystem.com/?path=/docs/ibm-products-patterns-empty-state-nodataemptystate--docs)
Image

Both CodeSandbox and Stackblitz examples are not working
Image

### Browser

Chrome

V2.38.0

### Suggested Severity

### Product/offering

IBM Research

Link is not working

Create a Next js react project
Use NoDataEmptyState in a conditional rendering function or directly in the page body 

_No response_

### Code of Conduct

amal-k-joy:
Closes #6315 

Hydration error will happen in next js when any dynamic ids is used in the dom. Here we were using unique dynamic ids to the svg elements generated using uuid4() which was added as part of another [issue fix](https://github.com/carbon-design-system/ibm-products/pull/1290). 
Currently , i have exposed a prop, `svgId` , where user can optionally pass any static id, which will solve dynamic id issues in empty states.

next js application

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
       - name: Install
         run: yarn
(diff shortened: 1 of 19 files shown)


davidmenendez (changes requested):
this actually seems like a good use case for [useId](https://react.dev/reference/react/useId). i think we should consider using this instead of `uuidv4`. this should alleviate the problem without need for introducing a new prop. the docs do mention a pitfall with SSR though:

> With [server rendering](https://react.dev/reference/react-dom/server), useId requires an identical component tree on the server and the client. If the trees you render on the server and the client don’t match exactly, the generated IDs won’t match.

i'm not very familiar with next and SSR's though, so i'm not sure if this will solve this specific issue. if not then a static prop might be the way to go. but i'm confident that `useId` will be fine here.

amal-k-joy:
> this actually seems like a good use case for [useId](https://react.dev/reference/react/useId). i think we should consider using this instead of `uuidv4`. this should alleviate the problem without need for introducing a new prop. the docs do mention a pitfall with SSR though:
> 
> > With [server rendering](https://react.dev/reference/react-dom/server), useId requires an identical component tree on the server and the client. If the trees you render on the server and the client don’t match exactly, the generated IDs won’t match.
> 
> i'm not very familiar with next and SSR's though, so i'm not sure if this will solve this specific issue. if not then a static prop might be the way to go. but i'm confident that `useId` will be fine here.

Hey @davidmenendez ,
Thanks for the good catch. `useId` can be used here, since its consistent across client and server side rendering.

> The useId hook in React 18 and later is a powerful tool for generating unique IDs that are consistent across client-side and server-side rendering.

davidmenendez (approved):

makafsal (approved):

PR #6509, opened 2024-11-26
chore(cards): rename aiLabel to decorator

### LINKED ISSUE #6452: Rename aiLabel prop to decorator in ExpressiveCard / ProductiveCard

sangeethababu9223:
Closes #6452

Rename aiLabel prop to decorator in ExpressiveCard / ProductiveCard

#### What did you change?
Updated the prop name `aiLabel` in Expressive Card,  Productive Card and Card files to `decorator`.
Updated the card style to only show gradient styling when `AILabel` component is passed.

Storybook


packages/ibm-products-styles/src/components/Card/_card.scss
```diff
@@ -176,16 +188,20 @@ $block-class: #{c4p-settings.$pkg-prefix}--card;
 }
 
 .#{$block-class}--has-slug,
-.#{$block-class}--has-ai-label {
+.#{$block-class}--has-decorator {
```
elycheea, line 191: I would probably expect us to move these styles to only if the component contains an `AILabel`. So more like how @ariellalgilmore  implemented it in Tiles[^1].

[^1]:https://github.com/carbon-design-system/carbon/pull/18109/files#diff-14ba41396ff7353279cbcbab448a45d6a4bc22fbdf70cda85eefb0ee823fe145R447-R449

packages/ibm-products/src/components/Card/Card.tsx
```diff
@@ -262,7 +262,10 @@ export const Card = forwardRef(
             [`${blockClass}__clickable`]: clickable,
             [`${blockClass}__media-left`]: mediaPosition === 'left',
             [`${blockClass}--has-slug`]: !!slug,
-            [`${blockClass}--has-ai-label`]: !!aiLabel,
+            [`${blockClass}--has-decorator-non-ai-label`]:
+              !!decorator && decorator['type']?.displayName !== 'AILabel',
+            [`${blockClass}--has-decorator`]:
+              !!decorator && decorator['type']?.displayName === 'AILabel',
```
elycheea, line 268: Can we simplify this to just use `[`${blockClass}--has-decorator`]` and rely on the presence of `cds--slug`/`cds--ai-label` classes to add the gradients?

elycheea, line 268: Can we simplify this to just use ```[`${blockClass}--has-decorator`]``` and rely on the presence of `cds--slug`/`cds--ai-label` classes to add the gradients?

packages/ibm-products-styles/src/components/Card/_card.scss
```diff
@@ -150,19 +150,31 @@ $block-class: #{c4p-settings.$pkg-prefix}--card;
   right: $spacing-05;
 }
 
+.#{$block-class}__header__inner-wrapper--decorator {
+  position: absolute;
+  top: $spacing-05;
+  right: $spacing-05;
+
+  .#{c4p-settings.$carbon-prefix}--slug {
```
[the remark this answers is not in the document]
elycheea: Or if this is to target the current `AILabel`, then maybe this should include both `--slug` and `--ai-label`?

elycheea (changes requested):
Made some suggestions to have the default `decorator` styles be set without the AI styles and only add them when there’s an `--ai-label` or `--slug` present, similar to the implementation we see in Core.

packages/ibm-products-styles/src/components/Card/_card.scss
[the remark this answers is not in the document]
sangeethababu9223: Thanks @elycheea.
I've updated the section to include `--ai-label` too. Initially we only had  `--slug` selector for the style we are overwriting here. But I've updated both sections now.

packages/ibm-products-styles/src/components/Card/_card.scss
sangeethababu9223: Thank you for pointing this out. It does make more sense to do this. 
This is updated.

packages/ibm-products/src/components/Card/Card.tsx
sangeethababu9223: Thanks @elycheea.
This is updated.

amal-k-joy (changes requested):
Can we add some tests for decorator?

sangeethababu9223:
> Can we add some tests for decorator?

I've added a basic test case for the decorator.

amal-k-joy (approved):
LGTM

packages/ibm-products-styles/src/components/Card/_card.scss
```diff
@@ -185,7 +199,8 @@ $block-class: #{c4p-settings.$pkg-prefix}--card;
 }
 
 .#{$block-class}__clickable.#{$block-class}--has-slug::before,
-.#{$block-class}__clickable.#{$block-class}--has-ai-label::before {
+.#{$block-class}__clickable.#{$block-class}--has-ai-label::before,
+.#{$block-class}__clickable.#{$block-class}--has-decorator::before {
```
elycheea, line 203:
```suggestion
.#{$block-class}__clickable.#{$block-class}--has-decorator::before {
```

I think for `has--decorator`, we don’t want to tie this to the popover styles since this could be separate.

sangeethababu9223: Thanks @elycheea,
This is updated.

elycheea (approved):

amal-k-joy (approved):

PR #6510, opened 2024-11-26
chore: update stylelint from v15 to v16

### LINKED ISSUE #6424: Upgrade stylelint to v16
Stylelint v16 does include some breaking changes with it going ESM only.

https://stylelint.io/migration-guide/to-16/

REF https://github.com/carbon-design-system/carbon/pull/17770

- [ ] Review the migration docs
- [ ] Review prior art from Core
- [ ] Update stylelint to 16 and resolve and breaking changes

davidmenendez:
Closes #6424 

upgrades `stylelint` from `v15.11` to `v16.10` and `stylelint-config-carbon` from `v1.19.1` to `v1.20.0`

this PR also does some additional cleanup by removing unnecessary `stylelint-disable` comments

### DIFF
--- a/examples/carbon-for-ibm-products/APIKeyModal/src/ThemeSelector/_theme-dropdown.scss
+++ b/examples/carbon-for-ibm-products/APIKeyModal/src/ThemeSelector/_theme-dropdown.scss
@@ -58,8 +58,8 @@
 
 .carbon-theme-dropdown {
   position: fixed;
-  bottom: 1rem; // stylelint-disable-line carbon/layout-token-use
-  left: 1rem; // stylelint-disable-line carbon/layout-token-use
+  bottom: 1rem;
+  left: 1rem;
   min-width: 11.5rem;
 }
 
--- a/examples/carbon-for-ibm-products/AboutModal/src/Example/Example.jsx
+++ b/examples/carbon-for-ibm-products/AboutModal/src/Example/Example.jsx
@@ -82,7 +82,7 @@ export const Example = () => {
             IBM{' '}
             <span
               style={
-                // stylelint-disable-next-line carbon/type-token-use
+                
                 { fontWeight: '600' }
               }
             >
--- a/examples/carbon-for-ibm-products/AboutModal/src/ThemeSelector/_theme-dropdown.scss
+++ b/examples/carbon-for-ibm-products/AboutModal/src/ThemeSelector/_theme-dropdown.scss
@@ -58,8 +58,8 @@
 
 .carbon-theme-dropdown {
   position: fixed;
-  bottom: 1rem; // stylelint-disable-line carbon/layout-token-use
-  left: 1rem; // stylelint-disable-line carbon/layout-token-use
+  bottom: 1rem;
+  left: 1rem;
   min-width: 11.5rem;
 }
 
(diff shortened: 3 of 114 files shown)


devadula-nandan (approved):

examples/carbon-for-ibm-products/AboutModal/src/Example/Example.jsx
```diff
@@ -82,7 +82,7 @@ export const Example = () => {
             IBM{' '}
             <span
               style={
-                // stylelint-disable-next-line carbon/type-token-use
+
```
elycheea, line 85:
```suggestion
```
Probably don’t mean to keep this blank line here?

examples/carbon-for-ibm-products/APIKeyModal/src/ThemeSelector/_theme-dropdown.scss
elycheea, line 1: I would expect we’d still want to flag non-Carbon token usage? I actually thought that comes from `stylelint-plugin-carbon-tokens` though.

https://www.npmjs.com/package/stylelint-plugin-carbon-tokens

davidmenendez: you are absolutely correct. good catch! i assumed that was included in `stylelint-config-carbon` but it is not. i'll add that in and make the appropriate changes.

examples/carbon-for-ibm-products/AboutModal/src/Example/Example.jsx
davidmenendez: yes unfortunately this added a bunch of new lines i'll have to manually remove because `prettier` won't haha

examples/carbon-for-ibm-products/APIKeyModal/src/ThemeSelector/_theme-dropdown.scss
elycheea: Haha, it helps that I opened the issue for Lee back when we first started downgrading it in our updates PRs. 😜

.stylelintrc.js
```diff
@@ -1,11 +1,18 @@
...
     'csstools/use-logical': null,
     'scss/load-no-partial-leading-underscore': null,
     'scss/double-slash-comment-inline': null,
     'no-duplicate-selectors': null,
+
+    'carbon/layout-use': null,
+    'carbon/theme-use': null,
+    'carbon/type-use': null,
+    'carbon/motion-duration-use': null,
+    'carbon/motion-easing-use': null,
```
elycheea, line 16: Doesn’t this defeat the purpose of using `stylelint-plugin-carbon-tokens`? 😹

elycheea: I think we would prefer to err on the side of caution and disable lines in our examples/story-specific SCSS files than to enable our team to bypass the tokens. This helps us be a bit more intentional about disabling the token usage (e.g. onboarding components that use colors that are not currently available as theme tokens).

davidmenendez: oh yeah sorry i'm doing this in stages because i saw a large amount of errors so i'm just starting with updating the rules haha. next step is enabling these and fixing them 🫡 

> I think we would prefer to err on the side of caution and disable lines in our examples/story-specific SCSS files than to enable our team to bypass the tokens. This helps us be a bit more intentional about disabling the token usage (e.g. onboarding components that use colors that are not currently available as theme tokens).

agreed. removing the line specific ones in favor of consolidating the rules to the top of the files is definitely a short-term fix since i think long term i'd we'd like to see those removed entirely (if possible).

davidmenendez:
closing this one because i got carried away trying to remove a bunch of what i thought where now useless `stylelint-disable` comments, which i learned was not the case, and it would be easier to just start fresh.

PR #6395, opened 2024-11-08
feat(productive card): floating ui for overflow menu

sangeethababu9223:
Closes #6329

Add floating ui to productive card with overflow

#### What did you change?

Wrapped the `card` component inside `FeatureFlags` with `enableV12Overflowmenu` prop

#### How did you test and verify your work?
Storybook

PR #5803, opened 2024-08-05
fix(ProductiveCard): with overflow tooltip issue

sangeethababu9223:
Closes #5399

No option to edit tooltip text on ProductiveCard with Overflow menu

#### What did you change?

Added `iconDescription` attribute to ProductiveCard and passes the same as prop to CardHeader.

#### How did you test and verify your work?
Storybook

PR #6518, opened 2024-11-29
test(ProductiveCard): add avt test coverage

### LINKED ISSUE #6472: Evaluate ProductiveCard for WCAG 2.2 readiness
### Parent issue summary

> Product teams can confidently build WCAG 2.2 AA accessible products using Carbon for IBM Products.

- Increase transparency on what accessibility standards are currently met and tested.
- Increase accessibility coverage by introducing more automated tests.
- Evaluate, update, and add testing to ensure that we meet WCAG 2.2 (included in latest `accessibility-checker` rulesets).

### Acceptance criteria
- [ ] Test latest `accessibility-checker` rulesets on various states of productive card
- [ ] Add tests for disabled states
- [ ] Add tests for open and close states in overflow menus
- [ ] Review any additional states we may need to capture or add tests for
- [ ] Add keyboard navigation tests to AVT

devadula-nandan:
Closes #6472

adds accessibility tests for productive card

Since we are in v12 OverflowMenu. i noticed that passing aria-label prop to v12 OverflowMenu spreads it to the overflow menu container which is a div and is reported as violation
v12 OverflowMenu uses single `Label` prop to set tooltip text on menu button and aria-label for list container (ul)
which means we might need to deprecate either `iconDescription` or `overflowAriaLabel` as `Label` prop handles both with v12 OverflowMenu.

useful references
https://github.com/carbon-design-system/ibm-products/pull/5803
https://github.com/carbon-design-system/ibm-products/pull/6395

i have passed label as below, and awaiting for any review suggestions
```
label={iconDescription || overflowAriaLabel}
```

#### What did you change?
e2e/components/ProductiveCard/ProductiveCard-test.avt.e2e.js
packages/ibm-products/src/components/Card/Card.tsx

yarn avt


e2e/components/ProductiveCard/ProductiveCard-test.avt.e2e.js
```diff
@@ -23,4 +24,226 @@ test.describe('ProductiveCard @avt', () => {
...
+    await visitStory(page, {
+      component: 'ProductiveCard',
+      id: 'ibm-products-components-cards-productivecard--with-action-ghost-button',
+    });
+
+    await expect(page).toHaveNoACViolations('ProductiveCard @avt-disabled');
+
+    const buttons = page.locator(`button.${carbon.prefix}--btn`);
+    const disabledButton = buttons.nth(2);
+    expect(await disabledButton.getAttribute('disabled')).not.toBeNull();
```
elycheea, line 50: I think we typically try to query by role or text over a specific query selector.

In this case, we could probably do 

```js
const disabledButton = page.getByRole('button', { name: 'Read more' });
```

Or some similar variation of that.

e2e/components/ProductiveCard/ProductiveCard-test.avt.e2e.js
```diff
@@ -23,4 +24,226 @@ test.describe('ProductiveCard @avt', () => {
...
+  test('@avt-overflow-menu: validates overflow menu interactions', async ({
+    page,
+  }) => {
+    await visitStory(page, {
+      component: 'ProductiveCard',
+      id: 'ibm-products-components-cards-productivecard--with-overflow',
+    });
+
+    const menuButton = page.locator('button[aria-haspopup="true"]');
+    const menu = page.locator('ul[role="menu"]');
```
elycheea, line 78: We can probably use the roles here as well. :)

https://playwright.dev/docs/locators#locate-by-role

elycheea (commented):
Looks good overall, but I think there are a few places where we could also use role/text locators rather than by query selectors. :)

devadula-nandan:
@elycheea fixed it

elycheea (approved):
Also agree it might be good to deprecate `iconDescription` or `overflowAriaLabel` if they are now redundant. If `iconDescription` overrides `overflowAriaLabel`, that may be preferable unless the latter is more widely used. 🤔

devadula-nandan:
> Also agree it might be good to deprecate `iconDescription` or `overflowAriaLabel` if they are now redundant. If `iconDescription` overrides `overflowAriaLabel`, that may be preferable unless the latter is more widely used. 🤔

im pretty sure we can deprecate `iconDescription` as it has been introduced recently compared to `overflowAriaLabel`.
also changed the override order to prioritise `overflowAriaLabel`. and updated the pr accordingly.

although a more generic name `label` could be more convenient (just like we have in core. a common prop to pass to button tooltip and aria-label for v12 OverflowMenu) when compared to having `overflowAriaLabel` that also sets tooltip text.

any additional thoughts or suggestions?

elycheea (approved):

anamikaanu96 (changes requested):
Hi @devadula-nandan can you add test for hover state.

elycheea (approved):

devadula-nandan:
@anamikaanu96 done

anamikaanu96 (approved):

Issue #5596, opened 2024-06-25
[Feature Request]: Open CoachmarkOverlayElements for extension


gcattan:
### The problem

We want to change the solution of the CoachmarkOverlayElements at each step. As a walk me.

As discussed in slack with @elycheea and @jlongshore , the CoachmarkOverlayElements is not designed for this behavior, but rather to draw attention at a static element (not in a flow).

### The solution

We don't change the design, but we allow developers to implement custom actions on click next or back, and to set the current progression of the component.

### Owner/main maintainer(s)

gcattan

### Product/offering

lineage-ui

### Business priority

Low Priority = release date is not dependent on fix or not upcoming

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

elycheea:
Adding context from Slack 

@jlongshore: 
> the beacon is rendered along side the element you are wanting to call attention to ( It’s not in the flow so it doesn’t take up space - potentially moving your layout )… and it can then have its position tuned by the `positionTune` props to get the beacon in just the right spot… Jumping off from one element and having it follow almost like a walkme like flow isn’t something that particular component is designed to do at the moment, happy to take feature requests - **maybe it can be modified to do that, or maybe it’s a different pattern all together**.

@gcattan: 
>  May be we can find something in the middle, that does not require too much changes. Now, I am thinking that we can add a callback on the next and back button, and allow to pass the current progression as a props too.

gcattan:
@elycheea probably a bit early, but do you have maybe a rough estimate for this? 
Let us know also if some help is needed from our part :)

gcattan:
@elycheea @jlongshore We would like to prioritize this feature for December release.
Are you able to provide some estimate for this task?  I see this issue is in Icebox right now.

PR #6548, opened 2024-12-05
feat(CoachmarkOverlayElements): Add next/back callbacks and currentStep properties

gcattan:
Closes https://github.com/carbon-design-system/ibm-products/issues/5596

This PR:
- Adds callbacks when clicking on Next and Previous buttons.
- Allows to set the current progression.

#### What did you change?

Add three optional properties to CoachmarkOverlayElements:
- onNextCallback
- onBackCallback
- currentStep

I modified the story book as follows:

image

Then manually checked that:
- next and back callbacks correctly logged into the console
- the progression was correctly set.

### DIFF
--- a/packages/ibm-products/src/components/CoachmarkOverlayElements/CoachmarkOverlayElements.test.js
+++ b/packages/ibm-products/src/components/CoachmarkOverlayElements/CoachmarkOverlayElements.test.js
@@ -22,12 +22,18 @@ const children = `hello, world (${uuidv4()})`;
 const dataTestId = uuidv4();
 const className = `class-${uuidv4()}`;
 
-const childrenContent = (
+const childrenContent = [
   <CoachmarkOverlayElement
+    key="element1"
     title="Hello World"
     description="this is a description test"
-  />
-);
+  />,
+  <CoachmarkOverlayElement
+    key="element2"
+    title="Hello World"
+    description="this is a another description test"
+  />,
+];
 
 const renderCoachmarkWithOverlayElements = (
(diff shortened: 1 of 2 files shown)


gcattan:
@AlexanderMelox CI looks good. But I still cannot access the logs of `percy`. Can you help with this issue?

AlexanderMelox:
> @AlexanderMelox CI looks good. But I still cannot access the logs of `percy`. Can you help with this issue?

I just got access to Percy so i can approve now

gcattan:
@AlexanderMelox  Following our suggestion and summarizing here ou exchange on slack: 
- I renamed the callbacks to `onNext` and `onBack`
- Added a test for the `onNext`.

However, testing the `onBack` is a bit more complicated as it is only showing when `scrollPercent > 0`, but this value never changes in testing (not sure why though).

Anyway, this helped me fix a bug with the scroll position not being correct (the position of the carrousel was not correctly updated).

AlexanderMelox:
> @AlexanderMelox Following our suggestion and summarizing here ou exchange on slack:
> 
> * I renamed the callbacks to `onNext` and `onBack`
> * Added a test for the `onNext`.
> 
> However, testing the `onBack` is a bit more complicated as it is only showing when `scrollPercent > 0`, but this value never changes in testing (not sure why though).
> 
> Anyway, this helped me fix a bug with the scroll position not being correct (the position of the carrousel was not correctly updated).

Jest doesn't really update css styles in testing so thats why it can be a little tricky to test scrolling and such, but we can handle it from now on

PR #6549, opened 2024-12-05
fix(actions): use correct reviewing team name

matthewgallo:
Another follow up PR to get the PR review label action working. I've confirmed that the auth issues are fixed now but then noticed a 404 for the teams api because I forgot to update the `team_slug` with the name of our reviewing team.

One more small change included here is to just specify `APP_INSTALLATION_ID` as a valid input to be passed in.

#### What did you change?
```
actions/add-review-labels/action.yml
actions/add-review-labels/index.js
```

### DIFF
--- a/actions/add-review-labels/action.yml
+++ b/actions/add-review-labels/action.yml
@@ -7,6 +7,9 @@ inputs:
   APP_PRIVATE_KEY:
     description: GitHub app private key
     required: true
+  APP_INSTALLATION_ID:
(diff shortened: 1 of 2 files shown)


davidmenendez (approved):

elycheea (approved):
🙈

matthewgallo:
😂 I even left a comment in the code for my future self to update the name and still forgot!

Issue #6274, opened 2024-10-22
Create overflow utilities


elycheea:
`ActionBar`, `DelimitedList`, `StringFormatter`, and `TagOverflow` are all similar in that they help handle overflow of items, using truncation or an other overflow component (e.g. interactive tag, overflow menu, tooltip) to handle additional display.

We should do a deep dive to learn where we might be able to consolidate some of these or reuse overflow logic for these very similar use cases.

### Explore the following features
- [x] Line length or container width
- [x] Number of lines or multiline truncation
- [x] Single line truncation with overflow content
- [x] Max visible items
- [x] Overflow rendering (tooltip, overflow menu, popover)
- [x] **Can this new utility deprecate these 4 separate components**
- [x] Floating UI support

Another thought on delimited list — could this be used _alongside_ “string formatter”? Current version does not show overflow content at all — only truncation, but most likely an accessibility violation if not.

### Acceptance criteria
- [ ] Work with design to determine which features need to be included
- [ ] These should be written agnostically to support both packages (`@carbon/utilities` and `@carbon/utilities-react`)

davidmenendez:
for reference: i touched on this briefly in [#5341](https://github.com/carbon-design-system/ibm-products/issues/5341#issuecomment-2445318757)

elycheea:
Moved the `DelimitedList` and `StringFormatter` evaluation to the umbrella issue so we can close out this epic. :)

PR #6551, opened 2024-12-05
fix: pageheader subtitle truncation visibility

### LINKED ISSUE #6534: [Feature Request]: Add title to the PageHeader subtitle
### The problem

picture/source: https://ibm-cloud.slack.com/archives/C013ZTX0N6B/p1733257027336759

The PageHeader component adds an ellipsis when the subtitle gets too large to display but you can't hover over it to see what the missing text is.

### The solution

Please add an HTML title or equivalent to the string.

### Design link

_No response_

### Links to other materials

_No response_

### Owner/main maintainer(s)

David Menendez

_No response_

### Product/offering

_No response_

### Business priority

None

### Code of Conduct

davidmenendez:
Closes #6534 

two issues i'd like to point out

1. `StringFormatter` needs to be updated to be TypeScript
2. `StringFormatter` doesn't dynamically hide the overflow text. I'm not sure if we have an alternative to `StringFormatter` that can display the text without making the popover always visible. would like some thoughts on this.

Screenshot 2024-12-05 at 1 22 40 PM

Screenshot 2024-12-05 at 1 28 02 PM


davidmenendez:
updated the subtitle to use `DefinitionTooltip` similar to title and added a new hook that re-uses the overflow logic for the title to be used in subtitle

makafsal (approved):
LGTM

anamikaanu96 (dismissed):
Hey @davidmenendez, It looks like the styles and tooltip are breaking when we update the subtitle using the controls.
Screenshot 2024-12-09 at 2 01 12 PM

amal-k-joy (changes requested):
Change looks good. However found this behaviour , but not sure is really applicable to adopter. However , could not see this issue for `title`

https://github.com/user-attachments/assets/06c9b6ca-3ebc-4982-955e-68b530fb3cf5

davidmenendez:
@anamikaanu96 @makafsal @amal-k-joy thank you for catching this. i realized that my problem was that i wasn't accounting for calculating the overflow by height. i refactored my code to take that into consideration for subtitle.

davidmenendez:
i know this needs to include testing, but i wanted to get feedback on the implementation of this solution. specifically with regards to `replaceWithOverflow.ts`.

would love to get y'all's opinion on this!

cc: @elycheea @matthewgallo

amal-k-joy:
> i know this needs to include testing, but i wanted to get feedback on the implementation of this solution. specifically with regards to `replaceWithOverflow.ts`.
> 
> would love to get y'all's opinion on this!
> 
> cc: @elycheea @matthewgallo

Solution looks good to me , since we we clamping vertically based on number of lines, its sensible to check for height overflow in case of title.
Also a thought, regarding the util naming, the functionality looks more like checking for overflow. (checkForOverflow, checkForOverflowWidth, checkForOverflowHeight )

packages/ibm-products-styles/src/components/PageHeader/_page-header.scss
```diff
@@ -559,6 +553,23 @@ $right-section-alt-width: 100% - $left-section-alt-width;
     @include type.type-style('body-01');
   }
 
+  .#{$block-class}__subtitle-tooltip .cds--definition-term {
```
matthewgallo, line 556: Don't forget to use `carbon.prefix` here

packages/ibm-products/src/components/PageHeader/PageHeader.tsx
```diff
@@ -1035,13 +1044,32 @@ export let PageHeader = React.forwardRef(
...
+                      <DefinitionTooltip
+                        definition={subtitle}
+                        className={`${blockClass}__subtitle-tooltip`}
+                      >
+                        <span
+                          ref={overflowSubtitleRef}
+                          className={`${blockClass}__subtitle-text`}
+                        >
+                          {subtitle}
+                        </span>
```
matthewgallo, line 1060: You could create a function to return this jsx and use it below as well, just to cut down on some repitition

davidmenendez:
> Solution looks good to me , since we we clamping vertically based on number of lines, its sensible to check for height overflow in case of title. Also a thought, regarding the util naming, the functionality looks more like checking for overflow. (checkForOverflow, checkForOverflowWidth, checkForOverflowHeight )

to be honest i struggled a bit with naming convention here and i think it's because i got too caught up in maintaining the "normal" vs "overflow" refs, but now that i think about it those could (should) be done in the component instead of the utility.

packages/ibm-products/src/global/js/utils/replaceWithOverflow.ts
```diff
@@ -0,0 +1,47 @@
...
+   * the ref for the element not overflowing
+   */
+  ref: RefObject<HTMLElement>;
+  /**
+   * the ref for the element that is overflowing
+   */
+  overflowRef: RefObject<HTMLElement>;
+};
+
+export const replaceOverflowWidth = ({ ref, overflowRef }: Props): boolean => {
```
matthewgallo, line 27: Stilling thinking this one through, but interested in your thoughts around combining these two utilities and having it default to check `width`, ie something like:
```ts
type Props = {
  ref: RefObject<HTMLElement>;
  overflowRef: RefObject<HTMLElement>;
  mode: 'width' | 'height'
};

export const replaceOverflow = ({ ref, overflowRef, mode = 'width' }) => {
...
}
```

Not necessarily saying `mode` is the best naming, but it would provide a way to merge these utilities so we just had one. I also agree with @amal-k-joy that it might make sense to change the naming, `shouldOverflow`, `willOverflow`, `checkOverflow`?

matthewgallo: This really starts to hit on some of the issues detailed in https://github.com/carbon-design-system/ibm-products/issues/6274 🎉

matthewgallo: Ok, yeah that makes sense, a bit easier to use "right out of the box". Lets stick with this approach.

davidmenendez:
tests have been added 👍

matthewgallo (approved):

amal-k-joy (approved):
Awesome work .

davidmenendez:
also for #6636

PR #6555, opened 2024-12-09
fix(Condition builder): initial state getting reset on re-rendering

### LINKED ISSUE #6457: Condition Builder: initial state getting reset on re-rendering
This issue was highlighted in our [support channel](https://ibm-studios.slack.com/archives/C013ZTX0N6B/p1729773104440639).

Currently initial state is respected during first render only, need to support re-renders as well.

https://github.com/user-attachments/assets/c07e8946-2a50-4308-83b8-baf06c9bdded

amal-k-joy:
Closes #6457 

Initial state issue fix.

#### What did you change?
Set the initial state to root state and clear initial state on manually deleting all conditions

local

### DIFF
--- a/packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.tsx
+++ b/packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.tsx
@@ -99,8 +99,7 @@ const ConditionBuilderContent = ({
   }, [actionState]);
   useEffect(() => {
     if (initialState?.enabledDefault) {
-      setRootState?.(initialConditionState.current as ConditionBuilderState);
-      initialConditionState.current = null;
+      setRootState?.(initialState.state);
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [initialState]);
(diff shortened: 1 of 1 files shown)


davidmenendez (approved):

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.tsx
```diff
@@ -120,12 +119,17 @@ const ConditionBuilderContent = ({
 
   const onRemove = useCallback(
     (groupId) => {
+      const groups = rootState?.groups?.filter(
```
davidmenendez, line 122: question: is this only removing a single entry from the `groups` array? because if so it would be more efficient to use a combination of `findIndex` and `toSpliced` instead. operations like `filter` don't stop after the first match: it will continue to search the array for additional matches. if we know that `onRemove` only removes one group at a time then i would refactor it to use the other methods. it's a small performance consideration, but it's also just good practice to think about operations in this manner.

if multiple groups are being removed though then i would ignore this comment 😄

amal-k-joy: The result are multiple groups except the removed group.

sangeethababu9223 (approved):

PR #6658, opened 2025-01-03
fix(conditionbuilder): property text in each condition should be bold

### LINKED ISSUE #6436: Condition Builder : Property text in each conditions should be bold
As per design , property text should be bold
current
Image

expected
Image

amal-k-joy:
Closes #6436 

property text in each condition should be bold

#### What did you change?
style correction
local


anamikaanu96 (approved):
LGTM.

Issue #6672, opened 2025-01-08
Unable to build 2.57.0 due to framer-motion bump


tom-youd:
### Package

Carbon for IBM Products

### Description

When using IBM products to build say a webpack build or build a storybook you get the following error -

```
<i> [webpack-dev-middleware] wait until bundle finished
WARN Module not found: Error: Can't resolve '@emotion/is-prop-valid' in '/Users/tomyoud/cp4i/cp4i-canvas/node_modules/framer-motion/dist/es/render/dom/utils'
[
  "ModuleNotFoundError: Module not found: Error: Can't resolve '@emotion/is-prop-valid' in '/Users/tomyoud/cp4i/cp4i-canvas/node_modules/framer-motion/dist/es/render/dom/utils'\n" +
    '    at /Users/tomyoud/cp4i/cp4i-canvas/node_modules/webpack/lib/Compilation.js:2123:28\n' +
    '    at /Users/tomyoud/cp4i/cp4i-canvas/node_modules/webpack/lib/NormalModuleFactory.js:923:13\n' +
    '    at eval (eval at create (/Users/tomyoud/cp4i/cp4i-canvas/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:10:1)\n' +
    '    at /Users/tomyoud/cp4i/cp4i-canvas/node_modules/webpack/lib/NormalModuleFactory.js:339:22\n' +
    '    at eval (eval at create (/Users/tomyoud/cp4i/cp4i-canvas/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:9:1)\n' +
    '    at /Users/tomyoud/cp4i/cp4i-canvas/node_modules/webpack/lib/NormalModuleFactory.js:520:22\n' +
    '    at /Users/tomyoud/cp4i/cp4i-canvas/node_modules/webpack/lib/NormalModuleFactory.js:151:10\n' +
    '    at /Users/tomyoud/cp4i/cp4i-canvas/node_modules/webpack/lib/NormalModuleFactory.js:783:25\n' +
    '    at /Users/tomyoud/cp4i/cp4i-canvas/node_modules/webpack/lib/NormalModuleFactory.js:1008:8\n' +
    '    at /Users/tomyoud/cp4i/cp4i-canvas/node_modules/webpack/lib/NormalModuleFactory.js:1137:5\n' +
[...]

PR #6673, opened 2025-01-08
fix: add @emotion/is-prop-valid to dependency list to fix builds

tom-youd:
Closes https://github.com/carbon-design-system/ibm-products/issues/6672

Added @emotion/is-prop-valid as a dependency to fix builds

#### What did you change?

As per above - added dependency to package.json (for Ibm-products)

Building storybook locally


matthewgallo:
Thanks for helping out with all of the investigation you did here @tom-youd! This shouldn't be an issue now that we've rolled `framer-motion` back to `6.5.1` in the patch we released yesterday (`2.57.1`).

PR #6734, opened 2025-01-20
fix(BigNumbers): Convert to TypeScript

### LINKED ISSUE #6689: Migrate `BigNumbers` to TypeScript
As we add Typescript support, want to call out a few helpful tips from the Carbon TypeScript guide that is largely relevant to our library as well.

### Changing the extension via git

When changing the file extension — **do not simply rename the file**. **Instead, change the extension via `git`.**

For example,

```
git mv packages/ibm-products/src/components/ComponentName/ComponentName.js packages/ibm-products/src/components/ComponentName/ComponentName.tsx
```

### Do not add types to internal components or functions
- Leave internals as `.js`
- Add a [jsdoc type annotation](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html#providing-type-hints-in-js-via-jsdoc) of `/** @type any */`
- [Enable errors](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html#ts-check) by adding `// @ts-check` to the first line in the file.

### Testing changes
A few different options for testing changes —

- At the bottom of a file, write a dummy component that uses the component that you’re converting to ensure that you can still pass all appropriate props.
- Take a storybook example for a component, copy and paste it into the bottom of the `.tsx` file and validate whether or not it accepts the props as you’ve defined them.

```[tasklist]
### Tasks
- [x] Change extension to `.tsx` **via git** (**do not simply rename**, see above)
- [x] Copy the PropTypes definitions above the component definition
- [x] Retool the PropTypes to be a TypeScript interface
- [x] Fix errors as they appear
- [ ] Do NOT add types to internal components or functions that are not exported as part of the public API
- [ ] Test your changes (see above)
```

### Resources
Carbon’s guide on [adding component types](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/adding-component-types.md)

paul-balchin-ibm:
Closes #6689 

- Converted to TypeScript and updated code.
- Rewrote prop descriptions.
- Converted CSS margins and padding from standard to logical.
- Pushed tests/coverage to 100%.
- Updated Storybook Controls.

#### What did you change?

```
BigNumbers.mdx
BigNumbers.stories.jsx
BigNumbers.test.js
BigNumbers.tsx
BigNumbersSkeleton.tsx
_storybook-styles.scss
constants.js
index.ts
```

- Storybook
- Tests/coverage


packages/ibm-products-styles/src/components/BigNumbers/_big-numbers.scss
```diff
@@ -43,16 +43,16 @@ $skeleton-block-class: #{c4p-settings.$pkg-prefix}--big-numbers-skeleton;
...
+  margin-block-start: auto;
 }
 
 .#{$block-class}__info {
   vertical-align: top;
 }
 
 .#{$block-class}__trend {
-  margin-top: $spacing-03;
+  margin-block-start: $spacing-03;
```
paul-balchin-ibm, line 55: Add a little space so that the button isn't touching the number.

packages/ibm-products/src/components/BigNumbers/BigNumbers.stories.jsx
```diff
@@ -32,44 +34,37 @@ const numericOptions = {
   'undefined ': undefined,
 };
 
+const iconButtonOptions = {
```
paul-balchin-ibm, line 37: Added the ability to show/hide the Edit button via a Storybook dropdown. This means we could remove the `withEditButton` page.

packages/ibm-products/src/components/BigNumbers/BigNumbers.stories.jsx
```diff
@@ -125,28 +143,26 @@ export default {
...
+  locale: 'en-US',
   percentage: false,
   size: BigNumbersSize.Default,
-  forceShowTotal: false,
+  tooltipDescription: '',
+  total: 13,
   trending: false,
   truncate: true,
-  locale: 'en-US',
+  value: 5,
```
paul-balchin-ibm, line 158: Fixed `value` so that it's pointing to an item in the list be default.

packages/ibm-products/src/components/BigNumbers/BigNumbers.stories.jsx
```diff
@@ -32,44 +34,37 @@ const numericOptions = {
...
+const iconButtonOptions = {
+  'undefined ': null,
+  'Example <Button> ': (
+    <Button
+      renderIcon={Edit}
+      iconDescription="Icon Description"
+      kind="ghost"
+      size={'sm'}
+      hasIconOnly
+      onClick={action('Button.onClick()')}
```
paul-balchin-ibm, line 46: Added an `action` so that clicking the Edit button shows an entry in the *Actions* tab.

packages/ibm-products/src/components/BigNumbers/BigNumbers.stories.jsx
```diff
@@ -157,20 +173,3 @@ export const bigNumbers = Template.bind({});
 bigNumbers.args = {
   ...defaultProps,
 };
-
-export const withEditButton = Template.bind({});
```
paul-balchin-ibm, line 161: With the added ability to show/hide the Edit button, this template is no longer req'd.

packages/ibm-products/src/components/BigNumbers/BigNumbers.test.js
```diff
@@ -12,36 +12,25 @@ import { Edit } from '@carbon/react/icons';
...
-    <BigNumbers
-      label="Label"
-      value={12345}
-      total={55555}
-      percentage={false}
-      size={BigNumbersSize.Default}
-      forceShowTotal={false}
-      trending={false}
-      truncate={true}
-      locale="en-US"
```
paul-balchin-ibm, line 37: Removing these "default" values means each test below only has to pass the props specific to the test, instead of having to consider "add/change/remove" every time.

packages/ibm-products/src/components/BigNumbers/BigNumbers.tsx
```diff
@@ -184,126 +180,64 @@ BigNumbers.displayName = componentName;
 // in alphabetical order (for consistency).
 // See https://www.npmjs.com/package/prop-types#usage.
 BigNumbers.propTypes = {
-  /**
-   * Optional class name.
-   * @type number
```
paul-balchin-ibm, line 189: Removed all `propTypes > @type ...` since it doesn't seem to be used elsewhere.

packages/ibm-products/src/components/BigNumbers/constants.js
```diff
@@ -64,7 +64,13 @@ export const getIconSize = (size) => {
  *
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#locales
  */
-export const getSupportedLocale = (locale) =>
-  Intl.NumberFormat.supportedLocalesOf(locale).length > 0
-    ? locale
-    : DefaultLocale;
+export const getSupportedLocale = (locale) => {
```
paul-balchin-ibm, line 67: Updated to `try/catch` because
```
Intl.NumberFormat.supportedLocalesOf(locale);
```
will throw an error if an unsupported locale string is used.

packages/ibm-products/src/components/BigNumbers/BigNumbers.tsx
```diff
@@ -63,27 +60,27 @@ export let BigNumbers = React.forwardRef(
...
+      trending = false,
+      truncate = true,
+      value,
       // Collect any other property values passed in.
       ...rest
-    },
-    ref
+    }: BigNumbersProps,
+    ref: React.ForwardedRef<HTMLDivElement>
   ) => {
```
davidmenendez, line 80: minor stylistic thing here, but i think it's cleaner to adopt this signature
```
const Component = forwardRef<HTMLDivElement, Props>((props, ref,) => {
  const { prop1 = 'default value', prop2, prop3, ...rest } = props;
  ...
});
```
this way you get more stuff out of the signature and into the body

packages/ibm-products/src/components/BigNumbers/BigNumbersSkeleton.tsx
```diff
@@ -0,0 +1,76 @@
...
+ * SkeletonBigNumbers is used to display a skeleton version while
+ * content is loading (handled by the BigNumbers prop `loading').
+ *
+ * Note: This component is only used within BigNumbers.
+ */
+
+// Use the same properties and values as parent BigNumbersProps
+type BigNumbersSkeletonProps = Pick<BigNumbersProps, 'className' | 'size'>;
+
+export const BigNumbersSkeleton = React.forwardRef(
```
davidmenendez, line 38: same comment about signature structure

packages/ibm-products/src/components/BigNumbers/BigNumbersSkeleton.tsx
```diff
@@ -0,0 +1,76 @@
...
+    ref: React.ForwardedRef<HTMLDivElement>
+  ) => {
+    const BigNumbersSkeletonClasses = classnames(className, {
+      [`${blockClass}--lg`]: size === BigNumbersSize.Large,
+      [`${blockClass}--xl`]: size === BigNumbersSize.XLarge,
+    });
+    return (
+      <div
+        {...rest}
+        className={cx(className, blockClass, BigNumbersSkeletonClasses)}
```
davidmenendez, line 50: you can just include `className` and `blockClass` in the `BigNumbersSkeletonClasses` definition and simplify it to `className={BigNumbersSkeletonClasses}`

packages/ibm-products/src/components/BigNumbers/BigNumbersSkeleton.tsx
```diff
@@ -0,0 +1,76 @@
...
+// Import portions of React that are needed.
+import React from 'react';
+
+// Other standard imports.
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+
+import { getDevtoolsProps } from '../../global/js/utils/devtools';
+import { pkg } from '../../settings';
+import classnames from 'classnames';
```
davidmenendez, line 18: `cx` is already being imported above

packages/ibm-products/src/components/BigNumbers/constants.js
davidmenendez: this looks like a handy utility. i would move this into the global utilities folder!

davidmenendez (changes requested):

packages/ibm-products/src/components/BigNumbers/BigNumbers.tsx
paul-balchin-ibm: Normally I would not have an opinion about what is cleaner code. But I found that moving the props into the body breaks the "Default" column in Storybook.

image

packages/ibm-products/src/components/BigNumbers/constants.js
paul-balchin-ibm: Done.

packages/ibm-products/src/components/BigNumbers/BigNumbers.tsx
paul-balchin-ibm: Changed as requested, with the exception that the spread `props` were put back as a function argument.

packages/ibm-products/src/components/BigNumbers/BigNumbersSkeleton.tsx
paul-balchin-ibm: Done.

packages/ibm-products/src/components/BigNumbers/BigNumbersSkeleton.tsx
paul-balchin-ibm: Done.

packages/ibm-products/src/components/BigNumbers/BigNumbersSkeleton.tsx
paul-balchin-ibm: Fixed.

packages/ibm-products/src/components/BigNumbers/BigNumbers.mdx
```diff
@@ -33,12 +33,6 @@ of a button as well as tool tip functionality. The default locale is English
   <Story of={BigNumbersStories.bigNumbers} />
 </Canvas>
 
-### Big numbers with edit button
```
paul-balchin-ibm, line 36: Removed story; this feature is now available through the default story **Controls**.

packages/ibm-products/src/components/BigNumbers/BigNumbers.tsx
```diff
@@ -1,92 +1,89 @@
...
+      value,
       // Collect any other property values passed in.
       ...rest
     },
     ref
   ) => {
-    const BigNumbersClasses = classnames(className, {
-      [`${blockClass}--lg`]: size === BigNumbersSize.Large,
-      [`${blockClass}--xl`]: size === BigNumbersSize.XLarge,
+    const BigNumbersClasses = cx(className, {
```
davidmenendez, line 81: another small stylistic observation: i think as a general react convention that uppercase variables like this are reserved for components. could you just set this to lowercase?

packages/ibm-products/src/components/BigNumbers/BigNumbersSkeleton.tsx
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+// Use the same properties and values as parent BigNumbersProps
+type BigNumbersSkeletonProps = Pick<BigNumbersProps, 'className' | 'size'>;
+
+export const BigNumbersSkeleton = forwardRef<
+  HTMLDivElement,
+  BigNumbersSkeletonProps
+>(({ className, size, ...rest }, ref) => {
+  const BigNumbersSkeletonClasses = cx(className, blockClass, {
```
davidmenendez, line 41: same comment about the uppercase variables

davidmenendez (changes requested):
one minor style issue i wanted to point out.

something i just noticed, `constants.js` remains as a javascript file. can we convert that to typescript also? same with `getSupportedLocale`?

packages/ibm-products/src/components/BigNumbers/BigNumbers.tsx
paul-balchin-ibm: Done

packages/ibm-products/src/components/BigNumbers/BigNumbersSkeleton.tsx
paul-balchin-ibm: Done

davidmenendez (approved):

packages/ibm-products/src/components/BigNumbers/BigNumbers.stories.jsx
anamikaanu96: Can we change the iconButton labels to 'No icon' and 'Edit Icon' instead of 'undefined' and 'Example <Button>' ?

anamikaanu96 (changes requested):

packages/ibm-products/src/components/BigNumbers/BigNumbers.stories.jsx
paul-balchin-ibm: Hi @anamikaanu96,

In Storybook, each **Control** should reflect as much as possible what the developer is expected to use as a _prop value_. The more technically accurate we can be, the sooner the developer can get on with their job.

If the prop is opinionated, then we try to be "prescriptive" by showing exactly what they must use: 
- the `(False|True)` toggle for any boolean, and nothing else, 
- the `'default'` `'lg'` `'xl'` strings for **size**, and nothing else,
- etc.

If the prop is unopinionated, then we try to be more "descriptive" since we leave it up to the developer to use their best judgement. **iconButton** does not force an "Edit icon". If we specify 'No icon' and 'Edit Icon', we are forcing the developer to look at the Storybook code or component source code to interpret what we mean by that.

The next best fallback documentation we can give the developer is the prop's **Description**, which includes what _type_ of property is expected. In this case "Displays an icon button next to value" and "node".

Just by looking at the control's details, the developer should know that the prop is expecting an "icon button" in the form of `node` and that the default value is "undefined". By choosing "Example <Button>" in the drop down, the developer can instantly see what changed in the live preview.

image

I admit that `Example <Button>` is technically correct, if a bit generic, but we also don't have any standards for how to do this. If you have something that is more informative, I'd appreciate your input. Maybe `<Button> example` (the more important word "Button" comes first), or `with <Button>`? Thanks :)

anamikaanu96: Thats sounds reasonable.

anamikaanu96 (approved):

PR #6746, opened 2025-01-22
refactor(sizing): implements css logical props and adds stylelint rule 

### LINKED ISSUE #3556: Implement CSS logical properties
Discovered through #3495 — it appears that [`stylelint-config-carbon@1.15.0`](https://github.com/carbon-design-system/carbon/releases/tag/v11.38.0) introduces CSS logical properties, part of the RTL improvements they’ve been working through.

Probably a larger initiative for us to consider as well, but don’t believe these changes seem too widespread for us to update in one of our routine dependency updates.

### Tasks
- [x] https://github.com/carbon-design-system/ibm-products/issues/5236
- [x] Remove local stylelint in favor of `stylelint-config-carbon`
- [ ] Replace CSS properties with logical counterparts
- [ ] `yarn lint:styles`

AlexanderMelox:
Closes #3556 

This PR changes the following to use CSS Logical Properties:

Screenshot 2025-01-24 at 2 32 44 PM

Note: This PR has a lot of file changes because it was done automatically using `stylelint --fix`. 

Similar PR in core -> https://github.com/carbon-design-system/carbon/pull/14531

**New**

- `stylelint-use-logical` was added a plugin to stylelint config

### ⚠️ ATTENTION REVIEWER ⚠️
**Things to look out for while reviewing:** Since I mostly did a find and replace, some lines might be changed that shouldn't that includes the following:

-  Css custom properties for example `--nav-height:` shouldn't be changed
- media queries: should be kept the same
- Sass variables


AlexanderMelox:
Maybe both reviewers can split the files ?

davidmenendez:
looks like linter is failing 🤔

AlexanderMelox:
yeah i'm checking, i'll move to draft until ready

.stylelintrc.js
```diff
@@ -14,5 +14,8 @@ module.exports = {
     'carbon/motion-easing-use': true,
     'carbon/theme-use': true,
     'carbon/type-use': true,
+
+    // CSS Logical properties
+    'csstools/use-logical': 'ignore',
```
AlexanderMelox, line 19: We added the rule but for now, we are ignoring so that we can roll them out in separate PR's, the last step would be to enable it at the end

AlexanderMelox:
we added the stylelint rules in this PR
https://github.com/carbon-design-system/ibm-products/pull/6767

matthewgallo (approved):

.stylelintrc.js
matthewgallo: Just noticed we already have this listed in `.stylelintrc.js`, don't need to add it again

davidmenendez (approved):

PR #6767, opened 2025-01-27
feat: adds stylelint-use-logical rule

AlexanderMelox:
This PR adds the use logical properties stylelint rule. This PR is necessary to go ahead and change all physical properties to logical i.e. `width` -> `inline-size`

https://www.npmjs.com/package/stylelint-use-logical

#### What did you change?

- Installed `stylelint-use-logical`
- Added  `stylelint-use-logical` `to stylelintrc.js`.

**Note:** we are ignoring the rules for now until all other PR's change the logical properties. Then there will be a follow up PR to enforce the lint

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,6 +1,6 @@
 module.exports = {
   extends: ['stylelint-config-carbon'],
-  plugins: ['stylelint-plugin-carbon-tokens'],
+  plugins: ['stylelint-plugin-carbon-tokens', 'stylelint-use-logical'],
   rules: {
     'max-nesting-depth': null,
     'scss/no-global-function-names': null,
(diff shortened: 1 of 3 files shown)


.stylelintrc.js
```diff
@@ -14,5 +14,8 @@ module.exports = {
     'carbon/motion-easing-use': true,
     'carbon/theme-use': true,
     'carbon/type-use': true,
+
+    // CSS Logical properties
+    'csstools/use-logical': 'ignore',
```
AlexanderMelox, line 19: Again, ignoring this for now, until we successfully change all physical properties to logical, then we will add a follow-up PR to enable this. 

The reason is so that new created PR's branching off of main do not have hundreds of stylelint errors

matthewgallo (approved):

davidmenendez (approved):

PR #6775, opened 2025-01-28
refactor(ActionBar): implement useOverflowItems hook in ActionBar

### LINKED ISSUE #6720: Refactor ActionBar to use useOverflowItems
`ActionBar` has similar overflow functionality as `TagOverflow` so it should be refactored to use the new utility function.

makafsal:
Closes #6720 

#### What did you change?
1. Implemented `useOverflowItems` hook in `ActionBar`
2. Fixed `OverflowMenu` issue
3. Fixed `Actiobar` width issue `PageHeader`

Storybook
yarn test


packages/ibm-products/src/components/ActionBar/ActionBar.tsx
```diff
@@ -119,178 +118,52 @@ export let ActionBar = React.forwardRef(
...
-                data-original-id={id}
-                key={`hidden-item-${key}`}
-                className={`${blockClass}__hidden-sizing-item`}
-                tabIndex={-1}
-              />
-            ))}
-          </span>
-        </div>
+    // Calculate the displayed items
+    const displayedItems = visibleItems.map(({ key, id, ...rest }) => (
```
davidmenendez, line 137: we should be able to move this into the render function. i would take a look at the [example implementation code](https://github.com/carbon-design-system/ibm-products/pull/6807/files#diff-b3bc1c13c1dde53425d434e6bf787214e52de32d4302f2d95339c0043eb11ca0R127) to see how you could move this and the hidden items into the render body.

packages/ibm-products/src/global/js/hooks/useOverflowItems.ts
```diff
@@ -17,15 +17,21 @@ export const useOverflowItems = <T extends Item>(
...
   const itemsRef = useRef<Map<string, number> | null>(null);
   const [maxWidth, setMaxWidth] = useState(0);
   const visibleItemCount = useRef<number>(0);
+  const minWidthRef = useRef<number>();
+  const requiredWidthRef = useRef<number>();
+  const offset = offsetRef?.current?.offsetWidth || 0;
 
   const handleResize = () => {
     if (containerRef.current) {
-      const offset = offsetRef?.current?.offsetWidth || 0;
```
davidmenendez, line 28: the offset should always be included otherwise when it switches to the overflow menu it will go beyond the container. consider these screenshots and look at the position of the overflow button within the container

PR
Screenshot 2025-02-03 at 3 30 56 PM

main
Screenshot 2025-02-03 at 3 33 32 PM

davidmenendez (changes requested):
an additional comment regarding the offset. i would take a second look at the changes you added and make sure that they are working correctly.

packages/ibm-products/src/components/ActionBar/ActionBar.tsx
makafsal: Done.

packages/ibm-products/src/global/js/hooks/useOverflowItems.ts
makafsal: Resolved.

packages/ibm-products/src/components/ActionBar/ActionBar.stories.jsx
```diff
@@ -52,10 +52,15 @@ export default {
...
 const Template = (argsIn) => {
-  const { containerWidth, ...args } = { ...argsIn };
+  const [containerWidth, setContainerWidth] = useState();
+
+  useEffect(() => {
+    setContainerWidth(argsIn?.containerWidth);
+  }, [argsIn]);
+
   return (
     <div style={{ width: containerWidth }}>
```
davidmenendez, line 62: minor issue i noticed in storybook: console is throwing an error because `containerWidth` is being spread to the underlying div and it's an unrecognized attribute.

Screenshot 2025-02-04 at 8 35 25 AM

packages/ibm-products/src/components/ActionBar/ActionBar.tsx
```diff
@@ -119,186 +118,72 @@ export let ActionBar = React.forwardRef(
...
+
+    const overflowMenuItems = hiddenItems?.map(({ id: key, ...rest }) => (
+      <ActionBarItem {...rest} key={key} />
+    ));
 
-    // create hidden sizing items
-    useEffect(() => {
-      // Hidden action bar and items used to calculate sizes
-      setHiddenSizingItems(
+    const overflowHiddenMenu =
```
davidmenendez, line 140: this shouldn't be required now since the calculations are being done in the hook. this was originally used as a way to calculate the values without the user seeing, but now it's unnecessary.

packages/ibm-products/src/global/js/hooks/useOverflowItems.ts
```diff
@@ -85,10 +102,24 @@ export const useOverflowItems = <T extends Item>(
 
   const visibleItems = getVisibleItems();
   const hiddenItems = items.slice(visibleItems.length);
+  const firstItemKey: string = getMap()?.keys()?.next()?.value || '';
+  const firstItemWidth = getMap()?.get(firstItemKey) || 0;
```
davidmenendez, line 106: i'm slightly confused by what these variables are trying to accomplish. can you explain their purpose?

packages/ibm-products/src/global/js/hooks/useOverflowItems.ts
```diff
@@ -85,10 +102,24 @@ export const useOverflowItems = <T extends Item>(
...
+    requiredWidth !== requiredWidthRef.current
+  ) {
     visibleItemCount.current = visibleItems.length;
-    onChange?.(hiddenItems);
+    minWidthRef.current = maxWidth;
+    requiredWidthRef.current = requiredWidth;
+
+    onChange?.({
+      hiddenItems,
+      minWidth: maxWidth,
```
davidmenendez, line 120: i also feel slightly confused by the naming here. why is `minWidth` being set to `maxWidth` and why is `maxWidth` being set to something that's beyond the `maxWidth`? could we make this slightly more clear / readable?

davidmenendez (changes requested):
looking great! got a few comments 👍 🚀

packages/ibm-products/src/components/ActionBar/ActionBar.stories.jsx
makafsal: Done.

packages/ibm-products/src/components/ActionBar/ActionBar.tsx
makafsal: Without a dummy `ActionBarOverflowItems` button, we'll get the `offsetRef?.current?.offsetWidth` as `undefined`. This is causing issues in the calculation because initially the offsetWidth is unknown for the hook and the ref pointing to nowhere.
With a dummy element:

https://github.com/user-attachments/assets/c41cd3b2-2767-46b6-8b4c-d480226358e5

Without a dummy element: 

https://github.com/user-attachments/assets/76fa848c-1bdf-418a-9f90-7a4db1b4c446

Without a dummy element, you can observe that in the initial resize, the hook cannot calculate the offset, but after the appearance of the offset element once then, it is calculated properly.

packages/ibm-products/src/global/js/hooks/useOverflowItems.ts
makafsal: Renamed.

davidmenendez (commented):
great! looks like a test is failing. you just need to update the tests are this should be good to go.

davidmenendez (commented):
i figured out the solution to dealing with the offset ref not being defined. instead of passing that ref to `ActionBarOverflowItems` just wrap that component with a div that has the ref.
```
<div className="hidden-items" ref={offsetRef}>
  {hiddenItems && hiddenItems?.length > 0 && (
    ...
  )}
</div>
```
this should ensure the offset it set correctly and will make the hidden item sizing unnecessary.


https://github.com/user-attachments/assets/e777af54-27b1-4921-bef4-ec0e83aec212

makafsal:
> `<div className="hidden-items" ref={offsetRef}>`


https://github.com/user-attachments/assets/88cdab88-73ea-4ded-b3d8-89a06a2f6701

I've tried this approach, but it is not always giving the actual offset width and it causes problem sometimes, not often.

makafsal:
> i figured out the solution to dealing with the offset ref not being defined. instead of passing that ref to `ActionBarOverflowItems` just wrap that component with a div that has the ref.
> 
> ```
> <div className="hidden-items" ref={offsetRef}>
>   {hiddenItems && hiddenItems?.length > 0 && (
>     ...
>   )}
> </div>
> ```
> 
> this should ensure the offset it set correctly and will make the hidden item sizing unnecessary.
> 
>  Screen.Recording.2025-02-06.at.4.02.43.PM.mov

I've tested this approach, but it is not working as expected in most of the cases.


https://github.com/user-attachments/assets/2b9e3cde-a339-4d88-824f-2b51ae6f5a0f

davidmenendez (changes requested):
i think a bit more than what was needed was added in the last commit. i don't see the added benefit to the additional effects or refs. i think the only thing that was needed to fix the issue with the offset not being initially calculated when not present. just the inclusion of that example code i sent on friday

```
  const offsetRefHolder = useRef<number | null>(null);

  const handleResize = () => {
    if (containerRef.current) {
      const offset = offsetRef?.current?.offsetWidth || 0;
      if (offset && !offsetRefHolder.current) {
        offsetRefHolder.current = offset;
      }
      const newMax =
        containerRef.current.offsetWidth - (offsetRefHolder.current || 0);
      setMaxWidth(newMax);
    }
  };
```

if i'm missing something with the new implementation please let me know.

packages/ibm-products/src/global/js/hooks/useOverflowItems.ts
```diff
@@ -80,20 +122,54 @@ export const useOverflowItems = <T extends Item>(
...
+
+      onChange?.({
+        hiddenItems: getVisibleItems()?.hiddenItems,
+        minWidth: remainingWidth,
+        maxWidth: requiredWidth + firstItemWidth,
+      });
+    }
+  }, [getVisibleItems, onChange, remainingWidth, requiredWidth]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
makafsal: You're right, I've given a second look and refactored that.

packages/ibm-products/src/global/js/hooks/useOverflowItems.ts
```diff
@@ -80,20 +122,54 @@ export const useOverflowItems = <T extends Item>(
...
+    getVisibleItems();
+    onChangeVisibleItems();
+  }, [getVisibleItems, onChangeVisibleItems]);
 
   return {
-    visibleItems,
+    visibleItems: getVisibleItems()?.visibleItems,
```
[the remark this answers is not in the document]
makafsal: That makes sense 😇, so refactored to return only the `visibleItems`.

packages/ibm-products/src/global/js/hooks/useOverflowItems.ts
```diff
@@ -14,23 +14,44 @@ type Item = {
...
-      const newMax = containerRef.current.offsetWidth - offset;
-      setMaxWidth(newMax);
+      offsetWidthRef.current = offsetRef?.current?.offsetWidth || 0;
+      const usableWidth =
+        containerRef.current.offsetWidth - offsetWidthRef.current;
+      setRemainingWidth(usableWidth);
     }
   };
 
+  const offsetRefHandler = (node) => {
```
makafsal, line 44: This is the new callback I've added to use instead of the hidden `overflowMenuItem`. This handle will give the instant value of the offset element the moment it is mounted in the DOM.

davidmenendez (approved):
ok looks good! i think we're ready to move on

devadula-nandan (approved):
LGTM

PR #6777, opened 2025-01-28
refactor(positions):  implements css logical props

### LINKED ISSUE #3556: Implement CSS logical properties

AlexanderMelox:
Closes #3556 

This PR changes the following to use CSS Logical Properties:
Screenshot 2025-02-03 at 1 02 09 PM

Similar PR in core -> https://github.com/carbon-design-system/carbon/pull/14531

**Changed**

- Change all border related physical properties to logical -> https://css-tricks.com/css-logical-properties-and-values/#aa-borders

### ⚠️ ATTENTION REVIEWER ⚠️
**Things to look out for while reviewing:** Since I mostly did a find and replace, some lines might be changed that shouldn't that includes the following:

-  Css custom properties for example `--nav-height:` shouldn't be changed
- media queries: should be kept the same
- Sass variables


packages/ibm-products-styles/src/components/Coachmark/_coachmark-overlay.scss
```diff
@@ -91,7 +91,7 @@ $draghandle-btn-class: #{$block-class}__handle;
 
   &--bottom {
     $horizontal-push: calc($distance-offset - $caret-center);
-    $top: calc(-1 * $caret-height);
+    $inset-block-start: calc(-1 * $caret-height);
```
matthewgallo, line 94: Not sure if this was intended but it doesn't look like this is being used either

matthewgallo (approved):

packages/ibm-products-styles/src/components/Coachmark/_coachmark-overlay.scss
AlexanderMelox: yes good catch, thanks Matt

AlexanderMelox, line 94:
```suggestion
    $top: calc(-1 * $caret-height);
```

devadula-nandan (approved):

PR #6810, opened 2025-01-29
refactor: remove immutability helper 

### LINKED ISSUE #6806: Refactor to remove `immutability-helper` as a dependency
We do not need to be including `immutability-helper` as a dependency, rather we should be relying on functionality already native to js in order to decrease our footprint as a package. This is only used in one component (Datagrid, `Columns.js`).

matthewgallo:
Closes #6806

This PR removes `immutability-helper` from our package, it is only used in one place (column customization in Datagrid) and  we were able to achieve the same thing without a library.

#### What did you change?
```
packages/ibm-products/package.json
yarn.lock
```
Verified in Column customization Datagrid story

### DIFF
--- a/packages/ibm-products/package.json
+++ b/packages/ibm-products/package.json
@@ -104,7 +104,6 @@
     "@dnd-kit/utilities": "^3.2.2",
     "@ibm/telemetry-js": "^1.5.0",
     "framer-motion": "^6.5.1 < 7",
-    "immutability-helper": "^3.1.1",
     "lodash": "^4.17.21",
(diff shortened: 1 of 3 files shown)


devadula-nandan (approved):
looks good to me, just wanted to know, if we ignore datagrid in codecov?

matthewgallo:
No, we don't ignore it yet. @AlexanderMelox was looking into configuring codecov to separate deprecated components but I think there's still some steps left for that.

davidmenendez (approved):

Issue #6815, opened 2025-01-30
[a11y]: CreateFullPage headings are not semantically correct


shocquu:
### Package

@carbon/ibm-products

### Browser

Chrome, Safari, Firefox, Edge

### Package version

v2.57.1

### React version

v18.0.0

### Automated testing tool and ruleset

Google Lighthouse, IBM Equal Access Accessibility Checker, Storybook

### Description

The headings in **CreateFullPage** do not follow the proper heading sequence; they skip from `h1` directly to `h5`. Ideally, we would prefer to manage the starting order, especially when Success Criterion 2.4.6: Headings and Labels is integrated into a modal, where the headings should commence from `h2`.

### WCAG 2.1 Violation

Success Criterion 2.4.6: Headings and Labels

### Reproduction/example

https://ibm-products.carbondesignsystem.com/iframe.html?globals=viewport%3Abasic&args=&id=ibm-products-patterns-create-flows-createfullpage--create-full-page-with-global-header&viewMode=story

### Steps to reproduce

1. Import \<CreateFullPage/> component,
2. Run the accessibility checker (Lighthouse, IBM Equal Access Accessibility Checker, etc.)
3. Get the `Heading levels should only increase by one` violation

![Image](https://github.com/user-attachments/assets/8538608a-c21d-4ab0-8e3e-5cbd91e6dd35)

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate problems

PR #6816, opened 2025-01-30
refactor: use Carbon's Section component for headings in some components

shocquu:
Closes https://github.com/carbon-design-system/ibm-products/issues/6815

This PR addresses accessibility issues with the **CreateFullPage** component.

#### What did you change?
Replaced headings in **CreateFullPage** and **CreateFullPageStep** components with Carbon's Heading components. 

Manually tested the changes locally and in Storybook, and ran `yarn test` to ensure no new bugs.


shocquu:
recheck

shocquu:
recheck

packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
```diff
@@ -481,18 +490,22 @@ export const TearsheetShell = React.forwardRef(
...
-                      element="h3"
-                      className={cx(
-                        `${bcModalHeader}__heading`,
-                        `${bc}__heading`
-                      )}
-                    >
-                      {title}
-                    </Wrap>
+                    <Section level={headingLevel}>
+                      <Wrap element="h2" className={`${bcModalHeader}__label`}>
```
wkeese, line 494: Did you mean to leave this "h2" here?

shocquu: updated

packages/ibm-products/src/components/CreateFullPage/CreateFullPageStep.tsx
```diff
@@ -157,7 +157,7 @@ export let CreateFullPageStep = forwardRef(
     const [shouldIncludeStep, setShouldIncludeStep] = useState<boolean>();
     const previousState = usePreviousValue({
       currentStep: stepsContext?.currentStep,
-    });
+    }) as unknown as { currentStep: number | undefined };
```
makafsal, line 160: Do we need to assert to type `unknown` here?

makafsal (changes requested):

packages/ibm-products/src/components/CreateFullPage/CreateFullPageStep.tsx
shocquu: The optional return "undefined" was missing, but it has now been added and it is no longer complaining

packages/ibm-products/src/components/CreateFullPage/CreateFullPage.tsx
```diff
@@ -248,6 +257,7 @@ export let CreateFullPage = React.forwardRef(
       title,
       secondaryTitle,
       breadcrumbOverflowTooltipAlign = 'right',
+      headingLevel = 1,
```
wkeese, line 260: I looked at this PR some more.

I'm OK with having an optional `headingLevel` property, although I'm not sure it's necessary.

Regardless, when `headingLevel` is undefined, why default to `1`?    `Heading` will figure out its own level based on how many ancestor `<Section>` components there are.   You are breaking that.

shocquu: I removed the default value. Do you have any suggestions for a better implementation? This solution seemed the simplest and worked without requiring many changes when I tested it locally.

wkeese:
> Do you have any suggestions for a better implementation? This solution seemed the simplest and worked without requiring many changes when I tested it locally.

I guess my question is why you think a `headingLevel` parameter is useful?   Can you give some examples when the application will need to override the inherited level for CreateFullPage, OptionsTile?   I agree that for tearsheets (attached to document.body) it makes sense for the headings to start at H2.

packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
```diff
@@ -481,18 +490,25 @@ export const TearsheetShell = React.forwardRef(
...
+                          element={Heading}
+                          className={cx(
+                            `${bcModalHeader}__heading`,
+                            `${bc}__heading`
+                          )}
+                        >
+                          {title}
+                        </Wrap>
+                      </Section>
+                    </Section>
```
wkeese, line 511: I think there are a couple problems with these changes:

1. `Section` should wrap the heading and its associated content, not just the heading.
2. Whenever you add a property to `TearsheetShellProps`, you need to add it to `propTypes` too.  (At least until Carbon stops using PropTypes.)
3. Whenever you add a property to TearsheetShell.tsx, you need to repeat the same changes in Tearsheet.tsx and TearsheetNarrow.tsx.

shocquu:
> I guess my question is why you think a `headingLevel` parameter is useful? Can you give some examples when the application will need to override the inherited level for CreateFullPage, OptionsTile? I agree that for tearsheets it makes sense for the headings to start at H2, unless the tearsheet is anchored somewhere other than `document.body`, in which case it might need to start at a higher level.

**CreateFullPage** does not inherit the **Section's** level if the Section is not defined within the same app. This is likely due to the [HeadingContext](https://github.com/carbon-design-system/carbon/blob/b67877bdd28e41838033296ce14174aa98b8b672/packages/react/src/components/Heading/index.tsx#L14). The `headingLevel` is intended to reset the Section's level to the specified value.

CleanShot 2025-02-21 at 17 16 54

https://stackblitz.com/edit/github-xqrpayxv?file=src%2FExample%2FExample.jsx

wkeese:
> CreateFullPage does not inherit the Section's level if the Section is not defined within the same app. 

I don't get it.  There's no concept of "app" in HTML or React, except for iframes.

The current `CreateFullPageStep` doesn't inherit the ancestor `Section`'s level because it is hardcoded to H5 and H6.

If you remove the hardcoded H5 and H6, and use `Section` and `Heading` instead, then the inheritance should work.

The only reason I can think of that it wouldn't work is that you've ended up with two copies of @carbon/ibm-products, either because you're using iframes, or something more subtle like this:

* myApplication/
    * MyPage.tsx <-- `<Section level={3}> <MyLibraryComponent> </Section>`
    * node_modules/
        * @carbon/
             * ibm-products/ <-- Defines `HeadingContext` used by application
        * MyLibrary/
             * MyLibraryComponent.tsx
             * node_modules/
                 * @carbon/
                     * ibm-products/ <-- Defines `HeadingContext` used by MyLibrary

shocquu:
> The current `CreateFullPageStep` doesn't inherit the ancestor `Section`'s level because it is hardcoded to `H5` and `H6`

What I'm trying to say is, even when the headings are not hardcoded and instead Section and Heading are applied, the heading level does not update or inherit properly, remaining set at level 1. I will post a screenshot once I manage to get the carbon linked locally.

shocquu:
CreateFullPage inside Section:

https://github.com/user-attachments/assets/87aa418e-c9fd-4559-8c2f-ca143215fd22

With `headingLevel`:

https://github.com/user-attachments/assets/7e7e0854-eb32-4176-9dd0-0ae701b2d4b1

wkeese:
I understand what you're saying, but the thing in your video shouldn't be happening.  What's the output when you run `npm list @carbon/react`?

shocquu:
> I understand what you're saying, but the thing in your video shouldn't be happening. What's the output when you run `npm list @carbon/react`?

CleanShot 2025-02-27 at 14 06 50

packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
shocquu: done

wkeese:
OK, well it looks like you do have two copies of @carbon/ibm-products, and also two copies of @carbon/react.  Not sure if that's what's causing this weird behavior, but I can't think of anything else that would cause it.

shocquu:
Would you recommend removing the `headingLevel` prop to see if that resolves the issue? If it doesn't, should we introduce it in a separate PR?

wkeese:
> Would you recommend removing the headingLevel prop 

Yes.

> ... to see if that resolves the issue? If it doesn't, should we introduce it in a separate PR?

The way to resolve the "issue" is by deduping, you should only have one copy of @carbon/react.  That's just a requirement of how Section and Heading work.

packages/ibm-products/src/components/Tearsheet/Tearsheet.tsx
```diff
@@ -385,5 +391,11 @@ Tearsheet.propTypes = {
    */
   title: PropTypes.node,
 
+  /**
+   * Sets the heading level for the title element (1 = \<h1>, 2 = \<h2>, ..., 6 = \<h6>).
+   * Helps control the semantic structure of the page.
+   */
+  // headingLevel: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
+
```
makafsal, line 399: Can you please remove the unwanted code rather than commenting?

wkeese:
This has gotten kind of big, I'm splitting out some of the changes into smaller PRs, starting with this one since a prereq for everything else:

- https://github.com/carbon-design-system/ibm-products/pull/7161

AlexanderMelox:
@wkeese I'm going to go ahead and close this PR since it has been fixed with the other PR's. Thank you for opening these, it helps the team a lot. Happy Coding! 🚀

PR #6825, opened 2025-01-31
build(`eslint`): upgrade to v9

### LINKED ISSUE #6187: reconfigure eslint config for v9
our latest dev dependencies update updates eslint from[ v8 to v9](https://github.com/carbon-design-system/ibm-products/pull/6185/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R88), which is causing breaking changes

Image

- [ ] Move to v9 requires `eslint.config.mjs`

matthewgallo:
Closes #6187 

These changes upgrade `eslint` to v9 and moves our config from `.eslintrc.js` to `eslint.config.mjs`, using the new flat config. I picked up where @elycheea left off with this one, so shout out to Elysia 🎉 

#### What did you change?
eslint config related files
Removed some disabling of rules
Manually ran `npx eslint` from project root


elycheea (commented):
I only did the easy part of the update to find the initial blockers 🙈. Thanks for finishing it out!

We also have good timing since System released yesterday so we can update `eslint-config-carbon` to the full release instead of the RC. 🎉

davidmenendez (approved):
🚀

PR #6828, opened 2025-01-31
Fix percy notification panel timestamps

### LINKED ISSUE #6808: NotificationsPanel: remove any dynamic values from storybook implementation
Percy continues to flag `NotificationsPanel` as having visual changes, but this is only because the story implementations contains dynamic values for the timestamps. In order to stop the percy failures we need to just refactor the story to remove dynamic values.

AlexanderMelox:
Closes #6808 

This PR removes date timestamps so that it doesn't show up on percy builds and fail them.

#### What did you change?
- Timestamps in notifications panel
- Timestamps in page header

storybook


paul-balchin-ibm (approved):
LGTM!

PR #6829, opened 2025-01-31
fix(NotificationsPanel): remove focus trap

### LINKED ISSUE #6432: Focus trap in NotificationPanel prevents user from navigating to next item in UIShell header
I am trying to work only with the keyboard strokes, and on click of Tab I go to the notifications panel, click Enter to open it, and Tab again to bring focus to the different Tabbable elements in the panel. But it takes us into a loop and not able to exit from the panel after we reach the end of the panel. While we do have Esc key to exit and close the panel, we want to have other options to exit the panel so as to bring consistency among our other header action components fhttps://github.com/user-attachments/assets/a923b755-ea36-4d07-976d-4a39da52a4b7
https://github.com/user-attachments/assets/cd8f696d-22a0-4be0-8592-b941ad0e2615
or the ease of the user. Attached is one screen recording of the same behavior.

paul-balchin-ibm:
Closes #6432

For accessibility reasons: remove the built-in focus trap.

After detailed discussions (see also the [refinement comment](https://github.com/carbon-design-system/ibm-products/issues/6432#issuecomment-2536400005)), the following changes were approved:
1. Remove the focus trap entirely.
2. Solve why the Panel "flickers" when it's open _and_ a new notification is added.
3. Keep the "on click outside → automatically close the Panel" feature.


packages/ibm-products-styles/src/components/NotificationsPanel/_notifications-panel.scss
```diff
@@ -179,8 +179,9 @@ $block-class: #{c4p-settings.$pkg-prefix}--notifications-panel;
       .#{$block-class}__notification-read-less-button {
         @include type.type-style('label-01');
 
+        display: flex;
+        align-items: center;
```
paul-balchin-ibm, line 183: Fixes styling for "Read more"

**Before**
image

**After**
image

packages/ibm-products-styles/src/components/NotificationsPanel/_notifications-panel.scss
```diff
@@ -270,6 +271,10 @@ $block-class: #{c4p-settings.$pkg-prefix}--notifications-panel;
     inset-block-end: 0;
     min-block-size: 2.5rem;
     .#{$block-class}__view-all-button {
+      display: flex;
+      width: 100%;
+      max-width: calc(100% - 2.5rem);
+      align-items: center;
```
paul-balchin-ibm, line 277: Fixes styling for "View all"

**Before**
image

**After**
image

packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.stories.jsx
```diff
@@ -45,149 +56,206 @@ export default {
...
+
+  const addNewNotification = () => {
+    const notification = newNotification();
+    setNotificationsData((data) => [...data, { ...notification }]);
+  };
+
+  return (
+    <div className={`${storyBlockClass}--full-height`}>
+      {/* <HeaderContainer
+        render={() => ( */}
```
paul-balchin-ibm, line 129: Temp fix for draft review.

Removing the `HeaderContainer`:
1. Solves the problem of re-rendering all children every time there's a state change. 
2. Solves the problem of tabbing order by allowing the Panel to be rendered inline with the Shell's notification icon button.

Keeping the `HeaderContainer`:
1. Requires the Panel to be rendered outside the Shell, but it will lose the expected tabbing order.

packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel_data.js
```diff
@@ -39,6 +39,7 @@ const data = [
     title: 'System alert',
     description: 'Email classification was exported successfully.',
     timestamp: new Date(currentDate.getTime() - 11 * msInOneMinute),
+    unread: false,
```
paul-balchin-ibm, line 42: Added back missing `unread:false` key/value. Expected keys should always be included even if their value is false or null.

packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.stories.jsx
```diff
@@ -45,149 +56,206 @@ export default {
...
+  const notification = { ...data[random] };
+  // Update dynamic data.
+  notification.id = uuidv4();
+  notification.timestamp = new Date();
+  notification.unread = true;
+
+  return notification;
+};
+
+const Template = (args) => {
```
paul-balchin-ibm, line 80: `stories.jsx` has been refactored to more modern standards.

The **data** control now shows a dropdown with "Sample data" and "Empty data". We no longer need an "Empty" story.

matthewgallo (commented):
Just a few questions for you, I noticed that the exit animation of the panel is gone. Is that something you can check out? Also, `triggerButtonRef` isn't used at all anymore. Thoughts on this? This is used across modals, panels in Carbon, but since we removed focus trap, suppose we might be able to deprecate it?

paul-balchin-ibm:
> ...the exit animation of the panel and the `triggerButtonRef` are gone. 

Oops. That was probably part of my aggressive "delete one thing at a time until I see a change in behavior" tactic; and then forgot to put them back in. (All for nothing, it was the UI Shell that was the source of the issues.)

Thanks, @matthewgallo.

matthewgallo:
The actual exit animation of the panel is missing now. This is what is currently looks like with the latest changes from `main`.

https://github.com/user-attachments/assets/415a2990-df22-4690-985b-7c8bcecfa9a2

paul-balchin-ibm:
> The actual exit animation of the panel is missing now

Fixed with latest commit.

e2e/components/NotificationsPanel/NotificationsPanel-test.avt.e2e.js
```diff
@@ -39,21 +39,6 @@ test.describe('NotificationsPanel @avt', () => {
     );
     await expect(notificationPanel).toBeVisible();
 
-    const firstElement = page.locator(
-      `button.${pkg.prefix}--notifications-panel__dismiss-button`
-    );
```
paul-balchin-ibm, line 44: Removed testing related to focus trapping.

matthewgallo (approved):

devadula-nandan:
hi @paul-balchin-ibm,
i remember fixing the notification panel flicker issue
https://github.com/carbon-design-system/ibm-products/pull/4771

but looks like it got resurrected from the grave after moving them to useIsomorphicEffect as part of fixing csp violations by removing inline css.
image

PR #6834, opened 2025-02-03
feat(interstitial screen): deprecate media prop and introduce new prop renderMedia

### LINKED ISSUE #6833: Deprecate media prop and lottie-web from InterstitialScreen.
Media prop is used in InterstitialScreen
- [ ] need to deprecate existing media prop
- [ ] replace media prop with renderMedia prop 
- [ ] Animations not working in storybook

amal-k-joy:
Closes #6833 
 This component was somehow missed out while we [implement this in other components](https://github.com/carbon-design-system/ibm-products/pull/6477)

Deprecate `media` prop from interstitial screen
Introduce new prop `renderMedia`
Introduce new prop `breakpointsWithMedia` to replace existing `media.breakpoints`

storybook


davidmenendez (changes requested):
this looks good, my only concern is that we can't just cut off support right away. i would add the deprecation notices, but i would also leave in the support for the previous props until we have decided the appropriate time to deprecate.

davidmenendez (commented):
this looks good, my only concern is that we can't just cut off support right away. i would add the deprecation notices, but i would also leave in the support for the previous props until we have decided the appropriate time to deprecate.

if the lottie removal is ready to move forward then let me know so i can approve. otherwise we need to maintain backwards compatibility for a time.

amal-k-joy:
> this looks good, my only concern is that we can't just cut off support right away. i would add the deprecation notices, but i would also leave in the support for the previous props until we have decided the appropriate time to deprecate.
> 
> if the lottie removal is ready to move forward then let me know so i can approve. otherwise we need to maintain backwards compatibility for a time.

Hey David,
We are not moving ahead to completely remove the existing prop as of now. This component should support old media prop as well. I have pushed a change right now to fix the support for `media.breakpoint` which got removed in this PR. Apart from that, do you see any other backward incompatibility ?

davidmenendez (dismissed):
just wanted to make sure the backward compatibility was taken into account. looks good!

makafsal (approved):
LGTM

makafsal (approved):

devadula-nandan (approved):

PR #6873, opened 2025-02-10
fix: glitches on open

### LINKED ISSUE #6872: flickering issues on some panel animations
### Package

Carbon for IBM Products

### Description

with the csp violation fixes, some previously seen flickering issues returned, but now at on open

these will be hard to notice, if the component is loaded in isolation in storybook, but in a practical scenario, this would become noticeable, when used in any application. so in order to notice them in storybook its recommended to check them in mid-tier or low-end mobile setting for throttling in dev tools. as the ultimate aim is to render these in any condition.

more details and reasoning why this has resurfaced

Before https://github.com/carbon-design-system/ibm-products/issues/4651
1. Panel enters the dom WITH STYLE (fade in animation), no flickers observed
2. Panel leaves the dom with fade out animation. Animation resets its position in the end before leaving the dom. Flickering observed on close, as animation is resetting in the end. 

After the fix from [pr](https://github.com/carbon-design-system/ibm-products/pull/4771)
1. Panel enters the dom WITH STYLE (fade in animation), no flickers observed
2. Panel leaves the dom with fade out animation and [forwards](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-fill-mode) fill mode. Animation doesn’t reset its position in the end before leaving the dom. Flickering not observed. 

After csp refactors
1. Panel enters the dom WITHOUT STYLE and the style (fade in animation) is applied later with useIsomorphicEffect. Flickering can be observed on open
2. Panel leaves the dom with fade out animation and [forwards](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-fill-mode) fill mode. Animation doesn’t reset its position in the end before leaving the dom. Flickering not observed. 

### key differences 

1. before the csp refactors, when the styles were inline. they were applied to the component before being rendered. so there were no issues on `on open` animations, and the flickering which was observed with `on close` was rectified with forwards fill mode as mentioned 
2. with the useIsomorphicEffect (useLayoutEffect or useEffect under the hood) hook that was added as part of csp refactors. the styles gets applied after the element renders in the dom.

### Solution/s
1) create `styleTune` and pass it as dependency in useIsomorphicEffect as seen here https://github.com/carbon-design-system/ibm-products/pull/6228/files#diff-759c2c32b74cf5c7a1e81a69a85326a03ce89430c4aea2ccc31b8699aecb7fb0R131 with necessary dependencies in memo. this way the styles will be calculated and applied to the element before render

2) ditch any kind of effect usage and simply move the styles to the sass file and apply 2 classnames conditionally for fade in and fade out

### Component(s) impacted

- [ ] notifications panel
- [ ] web terminal

### Browser

Chrome

2.59.0

### Suggested Severity

### Product/offering

c4ip

noticed in storybook

in storybook, observe closely on open animation, easily noticeable if throttling is set to mid-tier mobile in dev tools

-

### Code of Conduct

devadula-nandan:
Closes #6872 

moves the styles to scss and removed any effects applying styles.
also used motion tokens as per carbon motion standards

#### What did you change?
notifications panel, web terminal

storybook

### DIFF
--- a/packages/ibm-products-styles/src/components/NotificationsPanel/_notifications-panel.scss
+++ b/packages/ibm-products-styles/src/components/NotificationsPanel/_notifications-panel.scss
@@ -45,6 +45,29 @@ $block-class: #{c4p-settings.$pkg-prefix}--notifications-panel;
     transform: translateY(-38.5rem); // the height of the notification panel
   }
 }
+.#{$block-class} {
+  @media screen and (prefers-reduced-motion: reduce) {
+    &__entrance {
+      animation: none;
+    }
+  }
+
+  &__entrance {
+    animation: fade-in $duration-moderate-02 motion(standard);
+  }
+  @media screen and (prefers-reduced-motion: reduce) {
+    &__exit {
+      animation: none;
+    }
+  }
+
+  &__exit {
+    animation: fade-out forwards $duration-moderate-02 motion(standard);
+  }
+  @media (prefers-reduced-motion) {
+    animation: none;
+  }
+}
 
 .#{$block-class}__container {
   @include theme($g100);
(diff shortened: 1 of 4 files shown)


paul-balchin-ibm (dismissed):
LGTM. Thanks, @devadula-nandan

paul-balchin-ibm (approved):

sangeethababu9223 (approved):

PR #6169, opened 2024-10-07
feat(Tearsheet): Slug to aiLabel

sangeethababu9223:
Closes #6096

Rename the prop `slug` to `aiLabel`.

#### What did you change?

- Added Prop `aiLabel`
- Deprecated `slug` prop
- Updated the style with aiLabel classes as per the [ Core carbon standard](https://github.com/carbon-design-system/carbon/issues/17714)

#### How did you test and verify your work?
Storybook

PR #6904, opened 2025-02-13
fix(NotificationsPanel): new date-time formatting function

### LINKED ISSUE #6761: NotificationsPanel: "relative" time stamp is bespoke code and not localized
### Package

Carbon for IBM Products

### Description

The `NotificationsPanel` component uses a one-off `timeAgo()` function to create a timestamp like "6 hours ago".

**Issues**

The developer is required to provide _all_ the translations for relative permutations such as "`value` minute ago" and "`value` months ago", singular and plural, for seconds, minutes, hours, days, months, and years. There are 12-15 time-specific props to be supplied for each supported language. English is supplied as the default fallback language.

The `timeAgo()` function is not used for any other component in **Products**.

**Solution**

Replace the bespoke code with Carbon's new `dateTimeFormat` utility function found [here](https://github.com/carbon-design-system/carbon/blob/main/packages/utilities/src/dateTimeFormat/README.md). It follows Carbon's date and time formatting guidelines, provides for various time-stamp styles or formats, and is localized.

### Component(s) impacted

NotificationsPanel

### Browser

N/A

v2.58.0

### Suggested Severity

### Product/offering

Internal to Carbon for IBM Products 

N/A

_No response_

_No response_

### Code of Conduct

---

### Acceptance criteria
- [ ] Deprecate text callback props (ie `minutesAgoText`, `secondsAgoText`, etc.), will allow us to remove internal `timeAgo` function
- [ ] Add `timeAgoText` prop (`string`), allows consumer to utilize the new `dateTimeFormat` from `@carbon/utilities` with _relative_ time format, consumer passes in locale
- [ ] Include Docs for using the `dateTimeFormat` utility.

paul-balchin-ibm:
Closes #6761

[Direct link to **NotificationsPanel** Story](https://deploy-preview-6904--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-notifications-panel-notificationspanel--default&globals=viewport:basic)

Introduced new [Carbon utilities dateTimeFormat()](https://github.com/carbon-design-system/carbon/blob/main/packages/utilities/src/dateTimeFormat/README.md) function, which follows [official date and time and language guidance](https://pages.github.ibm.com/carbon/ibm-products/guidelines/content/date-and-time/).

`dateTimeFormat()` and 2 new props replaces:
- 13 existing date/time props, which must be filled out by the app developer, for each supported language,
- bespoke code to support those 13 props.

_* These older props are now marked as deprecated and remain functional._

#### What did you change?

```
packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.tsx
packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.stories.jsx
```
Added `@carbon/utilities`.
```
packages/core/package.json
```

- [x] Storybook


packages/core/package.json
```diff
@@ -68,6 +68,7 @@
     "vite": "^6.0.7"
   },
   "dependencies": {
+    "@carbon/utilities": "^0.4.0",
```
elycheea, line 71: Since this is being used by the actual `NotificationsPanel` component, we would need to move this into the `packages/ibm-products/package.json`.

However, re-reading the original issue, it looks like we actually intended for the user to use `dateTimeFormat`, which means we would probably want to use the utility in the data or story rather than in the component itself.

packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.tsx
elycheea, line 1: I think our intention was to introduce a new `timeAgoText` prop and deprecating but not removing the existing props. Since we’re in a major version, we don’t want to remove functionality that could break for adopters when they update to the latest version. Deprecation first is the softer approach and allows us to direct adopters to the new recommended implementation.

You can see an example of how we’ve previously deprecated props [here](https://github.com/carbon-design-system/ibm-products/pull/6169/files#diff-767f1b7d29ac95b328f742a9bdc243f38168b363f007ba01f52a58f32b805132).

elycheea (changes requested):

paul-balchin-ibm:
**Put into draft mode.** 

After further discussion, there are still open questions on how to deprecate the old props gracefully.

We can keep the old props, mark them as deprecated, and remove the `timeAgo()` function. The new, internal `dateTimeFormat()` will take over gracefully.

But, if the app developer was providing translations using the old props, these will now only render the default English, since the old functionality has been removed. For the app developer to specify other languages, they will need to know to use the new `dateTimeLocale` prop (i.e. `'en-US'`, `'fr_CA'`, etc).

packages/core/package.json
paul-balchin-ibm: Hi @elycheea, based on recent meetings, is this still the case?

packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.tsx
paul-balchin-ibm: Right. Deprecated props pushed.

packages/core/package.json
elycheea: Yep. Move it to our `ibm-products/package.json`!

paul-balchin-ibm: Done.

packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.tsx
```diff
@@ -273,12 +316,35 @@ export interface NotificationsPanelProps {
...
+ * [date and time guidance](https://pages.github.ibm.com/carbon/ibm-products/guidelines/content/date-and-time/).
+ *
+ * **Step 1:** ignore the following _deprecated_ props: `daysAgoText`,
+ * `hourAgoText`, `hoursAgoText`, `minuteAgoText`, `minutesAgoText`,
+ * `monthAgoText`, `monthsAgoText`, `nowText`, `secondsAgoText`,
+ * `yearAgoText`, `yearsAgoText`, `yesterdayAtText`.
+ *
+ * **Step 2:** provide the `dateTimeLocale` prop with a valid locale—such as "de" or "fr-CA"—and it will do the work.
+ * You must provide a value, otherwise the deprecated props above will be used instead.
+ */
```
elycheea, line 339: I think we’d typically just include this in the actual documentation (StoryDocs or mdx) rather than in the component file.

I would consider reversing the order for adopting the new props though. Rather than “ignore” as a task, I think it makes more sense to 1) adopt, 2) remove.

> To adopt the new localization:
>
> **Step 1:** Provide a valid locale to the `dateTimeLocale` prop, such as `de` or `fr-CA`.
> **Step 2:** Remove the now _deprecated_ props: `daysAgoText`, `hourAgoText`, `hoursAgoText`, `minuteAgoText`, `minutesAgoText`, `monthAgoText`, `monthsAgoText`, `nowText`, `secondsAgoText`, `yearAgoText`, `yearsAgoText`, `yesterdayAtText`.
>
> If you do not provide a locale, the deprecated props will be applied instead.

paul-balchin-ibm: `<StoryDocsPage />` is picking up this comment and dropping into the Overview section on the **Docs** page. I can't find an example of telling `<StoryDocsPage />` to fill in the Overview section directly.

elycheea: I actually imagine we’ll eventually move back to MDX through the docs epic 👀, but for now, I’d just focus on revising the docs you added rather than unpacking the `StoryDocsPage`. I want to say that was a workaround when we ran into issues with MDX when we first moved to Storybook 7 or 8? 🤔

paul-balchin-ibm: Works for me 👍 I'll stop digging into `StoryDocsPage` if we're planning to move back to `mdx`.

paul-balchin-ibm:
I noticed the doc change while resolving the conflict. 

Both are completed.

elycheea (approved):

davidmenendez (approved):

PR #6918, opened 2025-02-17
feat(userAvatar): backgroundcolor change based on theme

### LINKED ISSUE #6790: Add support for background colors based on theme

anamikaanu96:
Closes #6790 

Now the user avatar background color will dynamically change according to the storybook theme.

#### What did you change? packages/ibm-products-web-components/src/components/user-avatar/user-avatar.scss
packages/ibm-products-web-components/src/components/user-avatar/user-avatar.ts

### DIFF
--- a/packages/ibm-products-web-components/src/components/user-avatar/user-avatar.scss
+++ b/packages/ibm-products-web-components/src/components/user-avatar/user-avatar.scss
@@ -38,9 +38,38 @@ $block-class: #{$prefix}--user-avatar;
 
 //this mixin allows you to set background color based on the lighter themes such as g10 and white
 @mixin setLightBg($order, $color) {
-  :host .#{$block-class}--#{$order} {
+  :host([data-carbon-theme='g10']) .#{$block-class}--#{$order},
+  :host([data-carbon-theme='white']) .#{$block-class}--#{$order} {
     @include setBgColor($color);
   }
 }
-
+@mixin setDarkBg($order, $color) {
+  :host([data-carbon-theme='g90']) .#{$block-class}--#{$order},
+  :host([data-carbon-theme='g100']) .#{$block-class}--#{$order} {
+    @include setBgColor($color);
+  }
+}
 @include setLightBg(order-1-cyan, $cyan-60);
+@include setLightBg(order-2-gray, $gray-60);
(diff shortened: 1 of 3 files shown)


packages/ibm-products-web-components/src/components/user-avatar/user-avatar.scss
```diff
@@ -38,9 +38,38 @@ $block-class: #{$prefix}--user-avatar;
 
 //this mixin allows you to set background color based on the lighter themes such as g10 and white
 @mixin setLightBg($order, $color) {
-  :host .#{$block-class}--#{$order} {
+  :host([data-carbon-theme='g10']) .#{$block-class}--#{$order},
+  :host([data-carbon-theme='white']) .#{$block-class}--#{$order} {
+    @include setBgColor($color);
```
elycheea, line 43: I think `data-carbon-theme` comes specifically from our Storybook implementation so I don’t think we’d want to hardcode this value.

Might need to discuss our options a bit more here, but wondering if we are going to have to use a theme prop for this one after all.

paul-balchin-ibm:
Something's not quite right. If each avatar is wrapped in its own `<Theme>`, then changing the Storybook theme should not change the avatar background colors.

|Storybook theme "white" (correct) |Storybook theme "g100"|
|-|-|
|image|image|

anamikaanu96:
> Something's not quite right. If each avatar is wrapped in its own `<Theme>`, then changing the Storybook theme should not change the avatar background colors.
> 
> Storybook theme "white" (correct)	Storybook theme "g100"
> image	image

Yes, I did notice this and will make the change. There's no need to use `data-carbon-theme`, as @elycheea suggested. 
I'll eliminate the `data-carbon-theme` related CSS to resolve this problem. Thank you @paul-balchin-ibm and @elycheea for reviewing!

Additionally in this PR, I added a theme prop as we discussed in the refinement call.

elycheea (approved):
Ahh, this does bring back the old awkwardness of needing to separately manage the `theme` prop and story 😅.

But looks good and at least gives the adopter some control over it in the meantime. We’ll see what comes out of the discussion with System.

paul-balchin-ibm (approved):
LGTM!

Issue #6529, opened 2024-12-03
Evaluated OptionsTile for WCAG 2.2 readiness


elycheea:
### Parent issue summary

> Product teams can confidently build WCAG 2.2 AA accessible products using Carbon for IBM Products.

- Increase transparency on what accessibility standards are currently met and tested.
- Increase accessibility coverage by introducing more automated tests.
- Evaluate, update, and add testing to ensure that we meet WCAG 2.2 (included in latest `accessibility-checker` rulesets).

### Acceptance criteria
- [x] Test latest `accessibility-checker` rulesets on various states of options tile
- [ ] Add tests for disabled states
- [ ] Add tests or error states
- [x] Add tests for open and close states
- [x] Review any additional states we may need to capture or add tests for (locked, enabled, etc)
- [x] Add keyboard navigation tests to AVT

davidmenendez:
regarding acceptance criteria 2 & 3

> Add tests for disabled states
> Add tests or error states

currently there are tests for unit tests these states. there's not exactly a lot going on here though. the states for locked, warning, and error are mostly user implementation states. beyond showing the message those states don't have any other direct impact on the behavior of the component. even the locked state doesn't prevent the tile from opening 🤔 

i'm not sure if there's other considerations for these two points that i'm not aware of, but otherwise they are covered.

as of right now the only thing left is to just get the AVT test in.

elycheea:
@davidmenendez I actually noticed that too when I was reviewing the PR. There _are_ states, but iirc, they were message-based only and didn’t change the behavior of the options tile itself. 👍

PR #6520, opened 2024-12-02
chore: improve percy snapshot coverage

devadula-nandan:
Closes #6512 #6035 

Opens components' default state so the Percy snapshots are meaningful and also align with core.

#### What did you change?
The default state of the components has been updated to 'open' for the relevant stories.
Race conditions have been added in certain stories to address reported flakiness.
Focus-related issues have been resolved by @makafsal for some initially open stories.

#### How did you test and verify your work?
percy tests, storybook

PR #6920, opened 2025-02-17
fix: optionstile tests and UseControllable State update

davidmenendez:
contributes to #6529  

adds some additional tests to `OptionTile` to increase coverage to 100% & updates the `useControlledState` hook to typescript

note to self: a build error occurs when using generics with arrow syntax

```
// breaks
const useHook = <T>(value: T) =>
...

// works
function useHook<T>(value: T) {
...
```

although this works in `useOverflowItems` because of extends use. `T extends SomeType`


devadula-nandan (commented):
Hi @davidmenendez ,

although these changes are good to have which also increases our test coverage.

i think we need to add accessibility tests, which would verify all the states mentioned in the issue. like disabled state, collapsed state, invalid / error state etc. as per wcag 2.2 guidelines. what do you think?

davidmenendez:
@devadula-nandan yes this is still ongoing. i'm just breaking it up into several smaller PR's instead of a single large one. that's why i used `contributes to` and not `closes` at the top 😉

devadula-nandan (approved):
LGTM

devadula-nandan (approved):
Hi @davidmenendez,

Previously, I noticed a similar issue with the open state. We can observe it in the main if we set the open prop to true and then try to close it by clicking on the options tile.
https://github.com/carbon-design-system/ibm-products/pull/6520#issuecomment-2511887322
It looks like this issue has been fixed now. However, I'm not sure if it was related to syncing the internal open state with the open prop.

The only concern I have is whether it will be reactive to the open prop. From the story controls, toggling the open prop doesn’t seem to update it, but a refresh does. 🤔

looks like this https://github.com/carbon-design-system/ibm-products/issues/6936 issue addresses it?

davidmenendez:
@devadula-nandan yes the new issue i created should address that

packages/ibm-products/src/global/js/hooks/useControllableState.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { useState, useCallback } from 'react';
+
+type Callback<T> = (value: T) => void;
+
+export function useControllableState<T>(
```
paul-balchin-ibm, line 12: We are missing the function's descriptive comment above.

davidmenendez:
@paul-balchin-ibm resolved 👍

paul-balchin-ibm (approved):
LGTM!

devadula-nandan (approved):

PR #6951, opened 2025-02-21
docs: init hooks section of storybook with useOverflowItems

### LINKED ISSUE #6838: Export `useOverflowItems` and include a hooks section in storybook demonstrating how to use it
As we start to lean into patterns and utilities more, we should demonstrate how to use these things. The `useOverflowItems` hook is a perfect test case for this.

There should be two steps to completing this work:

- [x] This hook should be exported from our library so that it can be consumed externally
- [x] We should add a hooks section under the utils section in storybook - present the same level of guidance as [`useContextMenu`](https://react.carbondesignsystem.com/?path=/docs/hooks-usecontextmenu--overview)
- [x] Put together an example story or two demonstrating how to use this hook in a practical setting. One way to do this is to construct the same design as `TagOverflow` or `TagSet` but with using the `useOverflowItems` hook, so one can understand the value in the utility and see it in practice.

davidmenendez:
Closes #6838 

initiates the hooks section of storybook. docs section still required. looking for feedback on implementation & sidebar IA before proceeding.


matthewgallo (changes requested):
Much improved experience since your demo in engineering huddle earlier this week. The only small suggestion I have is to just rename `_useOverflowItems.scss` to `_storybook-styles.scss` since those are specific to just the story.

Other than that, looks great!

davidmenendez:
@matthewgallo @elycheea `useOverflowItems` has been placed in its own folder with it's specific assets

matthewgallo (approved):

elycheea (approved):
Woohoo! Thanks for making the updates and sharing it all this week.

Issue #6698, opened 2025-01-14
Read-only TextInput has incorrect background color due to Tearsheet overriding style


richardpilot:
### Package

Carbon for IBM Products

### Description

The scss for the `Tearsheet` overrides the styling for TextInput, however when a TextInput component is in read-only mode, it's background-color is designed to match the background colour of the container.

https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products-styles/src/components/Tearsheet/_tearsheet.scss
Search for the override `background-color: $field;`

**Current behaviour:**
![Image](https://github.com/user-attachments/assets/fc4e11ba-32d0-4b6d-8f57-d7ced8280564)

**Expected behaviour:**
![Image](https://github.com/user-attachments/assets/ed37f2a0-5522-4b22-88d8-6bbee530c30f)

### Component(s) impacted

`TextInput` inside of any `Tearsheet` variant

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

2.47.0

### Suggested Severity

None

### Product/offering

MQ

### CodeSandbox or Stackblitz example

https://codesandbox.io/p/devbox/wizardly-ellis-mf45yv

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

makafsal:
![Image](https://github.com/user-attachments/assets/b2df1caf-068f-4cdd-b258-c379028a183a)




Observed similar issues with components `text-area`, `search-input`, `select-input`, `dropdown`, `dropdown-list`, and `date-picker`.

Fix: Avoid this style override for components in `readOnly` state [Line 408](https://github.com/carbon-design-system/ibm-products/blob/e4c667d9e057f589291079fc1ea8f42e980c26ef/packages/ibm-products-styles/src/components/Tearsheet/_tearsheet.scss#L408)

Test: Make sure working with all themes

paul-balchin-ibm:
After doing a little digging, I found that there are styling issues with the `disabled` prop as well. This prop will have to be considered as well.

PR #6961, opened 2025-02-24
fix(Tearsheet): input field carbon styling

paul-balchin-ibm:
# Draft mode: for technical review only.

This is not for approval, only for review.

Refers to issue #6698

[Direct link to Tearsheet with input field tests.](https://deploy-preview-6961--carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-tearsheet--tearsheet-with-input-fields&globals=viewport:basic)

**Issue (As per the meeting of February 24)**

Issue #6698 requests a style fix for `TextInput.readOnly` inside `Tearsheet`. 

I found related style fixes for some other input fields, but there is no **story** to validate any of the input fields (number, text, date, etc.) and their possible states (enabled, disabled, readOnly) per Carbon guidelines.

_How do we validate that the input field styles have been fixed for all form elements and their states?_

**Partial solution**

We create a `Tearsheet` **story** that includes controls to switch between states for all listed input fields.

But, we also don't want to show this to the casual user or developer. This is only a technical, internal tool to validate that the developer is getting the expected look and feel.

Question(s) for next meeting:
- How can we hide the story in the UI, but still make it available through a URL only?
- It may be possible to apply a unique **tag** (i.e. `tags: ['autodocs']`) that is accessible through the Storybook UI → Search bar → filter dropdown. [Storybook docs, custom tags](https://storybook.js.org/docs/writing-stories/tags#filtering-by-custom-tags)

### DIFF
--- a/packages/ibm-products-styles/src/components/Tearsheet/_tearsheet.scss
+++ b/packages/ibm-products-styles/src/components/Tearsheet/_tearsheet.scss
@@ -401,18 +401,21 @@ $motion-duration: $duration-moderate-02;
   }
 
   &.#{$block-class}--wide .#{$block-class}__content {
-    // Revert background color overridden by Carbon's modal - https://github.com/carbon-design-system/carbon/blob/main/packages/styles/scss/components/modal/_modal.scss#L54
-    .#{$carbon-prefix}--pagination,
-    .#{$carbon-prefix}--pagination__control-buttons,
-    .#{$carbon-prefix}--text-input,
-    .#{$carbon-prefix}--text-area,
-    .#{$carbon-prefix}--search-input,
-    .#{$carbon-prefix}--select-input,
-    .#{$carbon-prefix}--dropdown,
-    .#{$carbon-prefix}--dropdown-list,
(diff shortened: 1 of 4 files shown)


davidmenendez:
i like the presentation of the story. ultimately i think this kind of approach is a great way to utilize our E2E testing and i'm a fan.

i've given it a lot of thought and i think that even though this kind of kitchen sink style of story might not represent a real scenario / implementation that it's still useful for showcasing these kinds of small distinctions in context. since special consideration has been made for certain elements in the context of a tearsheet (inputs in this case) i think it makes sense to showcase those distinctions directly. anything that we do with purpose should be highlighted, especially when it's an override to a core component. this way we are more closely aligned with the usage guidelines.

this feels a little contradictory to my thoughts on simplifying stories to be more dumb, but again: because inputs behave differently in a tearsheet i think it warrants highlighting that in this manner.

elycheea:
Discussed with Paul and closing this for now since the actual fix was implemented in a separate PR (#6975). However, this is still a good exploration on how we can include hidden stories specifically for testing that we may not want to elevate in our actual Storybook examples.

PR #7029, opened 2025-03-04
chore: init utilities section of WC storybook

### LINKED ISSUE #6716: [Tag set] in Web Components

davidmenendez:
contributes to #6716 

this PR only establishes the skeletal structure for the new utilities section. a follow up will be done to add actual documentation content and flesh out the storybook content.


davidmenendez:
thoughts & considerations;

Internally, how do we separate recipe / example code from exportable components?

* there's an `examples` folder in the root, but that's designated for stackblitz
* this PR adds a `utilities` folder under `src` but it doesn't really convey that the repo is just hosting example code
* maybe instead of `Experimental/Utilities/[utility]`in the sidebar we do something like `Experimental/Recipes/[recipe ie Overflowing Items]`
* we need to make sure we're making the distnction clear internally for devs and externally for consumers

sangeethababu9223:
Hey @davidmenendez ,
Is this in progress now?

devadula-nandan:
> thoughts & considerations;
> 
> Internally, how do we separate recipe / example code from exportable components?
> 
> * there's an `examples` folder in the root, but that's designated for stackblitz
> * this PR adds a `utilities` folder under `src` but it doesn't really convey that the repo is just hosting example code
> * maybe instead of `Experimental/Utilities/[utility]`in the sidebar we do something like `Experimental/Recipes/[recipe ie Overflowing Items]`
> * we need to make sure we're making the distnction clear internally for devs and externally for consumers

some more thoughts & considerations

looking at what fluentui did with presentation of their version of overflow utility
https://react.fluentui.dev/?path=/docs/components-overflow--docs
isn't it simpler, to show all examples that are using this utility (TagSet, ActionBar, UserAvatars(overflowing), Buttons, etc) in a single doc page? which could be placed in the path you have mentioned `Experimental/Recipes/[recipe ie Overflowing Items]`

AlexanderMelox (changes requested):

AlexanderMelox:
Screenshot 2025-04-16 at 2 38 48 PM

I would say I prefer if utlities is a header like Experimental, unless the hook is also experimental. I feel like we had a talk about this and I forgot what the consensus was

packages/ibm-products-web-components/src/utilities/overflow-handler/overflow-handler.mdx
```diff
@@ -0,0 +1,12 @@
+import { ArgTypes, Markdown, Meta } from '@storybook/blocks';
```
sangeethababu9223, line 1: We could remove the `ArgTypes ` here, since this story is not part of any specific component and we don't have to list any attributes/props.

davidmenendez:
> Screenshot 2025-04-16 at 2 38 48 PM
> 
> I would say I prefer if utlities is a header like Experimental, unless the hook is also experimental. I feel like we had a talk about this and I forgot what the consensus was

yeah i think we'll want to sort this out after this gets merged

sangeethababu9223 (approved):

AlexanderMelox (approved):

PR #7031, opened 2025-03-04
feat(custom-element): new es build module

ariellalgilmore:
Closes https://github.com/carbon-design-system/carbon/issues/18761

creates a new `es-custom` folder which will generate our components with `cds-custom-` prefix instead of `cds-`

#### Changelog

**New**

- bump `@carbon/web-components` to latest version (v2.25.0) that includes the `es-custom` build folder
- updated build to add a post build which copies the generates `es` files and then does a find and replace to swap out all instances of `cds` with `cds-custom`
- add `es-custom` to gitignore

#### Testing / Reviewing

run yarn build and see if `es-custom` folder is generated in web components and if the components have the correct prefix

**note**
- User's will need to import two different style files. One with
```
@use '@carbon/styles' with (
  $prefix: 'cds-custom' // Default prefix
);
```
and the other will be just 
```
@use '@carbon/styles' 
```
- Also, they need to set in their tsconfig files if using an angular version before angular 19 to
```
"skipLibCheck": true
```


packages/ibm-products-web-components/package.json
```diff
@@ -51,7 +53,7 @@
   "dependencies": {
     "@carbon/ibm-products-styles": "^2.57.0-rc.0",
     "@carbon/styles": "1.75.0",
-    "@carbon/web-components": "2.23.0",
+    "@carbon/web-components": "file:./package.tgz",
```
annawen1, line 56: oh might want to change it back here 😅

annawen1 (approved):

Issue #7019, opened 2025-03-03
[a11y]: [Tearsheet]: launcherButtonRef broken with stacked tearsheets


wkeese:
### Package

@carbon/ibm-products

### Browser

Chrome

### Operating System

MacOS

### Package version

2.59.0

### React version

20.0

### Automated testing tool and ruleset

manual test

### Description

Problem: `launcherButtonRef` doesn't work on stacked tearsheets.

Example markup:

```jsx
      <Tearsheet title="First tearsheet" selectorPrimaryFocus="input" open={isOpen1}>
          <TextInput .../>
          <Button ref={button2Ref} onClick={() => setIsOpen2(true)}>
            Open stacked tearsheet
          </Button>
      </Tearsheet>

      <TearsheetNarrow
        launcherButtonRef={button2Ref}
        actions={[
          {
            key: 1,
            kind: 'primary',
            label: 'Close',
            onClick: () => {
              setIsOpen2(false);
            },
          },
        ]}
        title="Tearsheet 2"
        open={isOpen2}
      >
```

Situation: User opens Tearsheet1, then opens Tearsheet2, then closes Tearsheet2.

Expected result: When Tearsheet 2 is closed, TearsheetShell should focus button2Ref (the button specified by `launcherButtonRef`).

Actual result: When the user closes the stacked tearsheet, TearsheetShell.tsx does focus `launcherButtonRef`.  However, it also focuses the element specified by `selectorPrimaryFocus`, or if `selectorPrimaryFocus` is not specified, then the first element.  Thus a race condition.

Presumably the `selectorPrimaryFocus` related code should only run when a tearsheet is opened.    So I'd expect code like this:

```js
    useEffect(() => {
      if (open) {
        // Focusing the first element or selectorPrimaryFocus element
        claimFocus(firstElement, modalRef, selectorPrimaryFocus);
      }
    }, [open]
```

Instead you have this code that calls `claimFocus()` when Tearsheet 2 is closed, because `depth` has changed:

```js
    useEffect(() => {
      if (open && position === depth) {
        // Focusing the first element or selectorPrimaryFocus element
        claimFocus(firstElement, modalRef, selectorPrimaryFocus);
      }
    }, [
      currentStep,
      depth,
      firstElement,
      modalRef,
      modalRefValue,
      open,
      position,
      selectorPrimaryFocus,
      hasError,
    ]);
```

### Reproduction/example

[...]

PR #7063, opened 2025-03-11
fix(useFocus): resolve focus missing in state change

### LINKED ISSUE #6946: [Tearsheet]: The focus moves to the first element when the action button's disabled property is toggled.
### Package

Carbon for IBM Products

### Description

The focus moves to the first element when the action button's disabled property is toggled.
UseCase:- We have a form with multiple fields, and the requirement is to enable the "Continue" button only after all the required fields are filled.the focus is moving away from the currently editing field once this condition is met.

### Component(s) impacted

Tearsheet

### Browser

Chrome, Safari, Firefox

2.59.0

### Suggested Severity

### Product/offering

OpenPages

https://stackblitz.com/edit/github-cqjvhd-k4woiiav?file=src%2FExample%2FExample.jsx,src%2FApp.jsx,package.json

1) Enter the data in fields.
2) While entering data is second field, focus is immediately moving to first field.

_No response_

### Code of Conduct

makafsal:
Closes #6946 #7019 

#### What did you change?
Moved `claimFocus` into the hook

`yarn test`
`yarn avt`


davidmenendez (changes requested):
looks good but code coverage dropped. can you take a look at that? thanks!

makafsal:
> looks good but code coverage dropped. can you take a look at that? thanks!

@davidmenendez This is used for the `Createsheet` component, which is challenging to cover in the unit test (Covered in AVT).

davidmenendez (dismissed):

packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
```diff
@@ -318,29 +317,40 @@ export const TearsheetShell = React.forwardRef(
...
+        if (
+          open &&
+          depth === position &&
+          !modalRef?.current?.contains(document.activeElement)
+        ) {
+          claimFocus();
+        }
+      });
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [depth, position]);
```
jlongshore, line 346: Missing `open` and potentially `modalRef` in the dep array... potentially `claimFocus` (see comment below)

packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
```diff
@@ -318,29 +317,40 @@ export const TearsheetShell = React.forwardRef(
...
-      if (prevOpen && !open && launcherButtonRef) {
+      if (prevOpen && !open && launcherButtonRef?.current) {
         setTimeout(() => {
-          launcherButtonRef.current.focus();
+          launcherButtonRef?.current.focus();
         }, 0);
       }
-    }, [launcherButtonRef, open, prevOpen]);
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [open, prevOpen]);
```
jlongshore, line 333: We could add `launcherButtonRef` to the dep array and remove the `eslint-disable-next-line...` rule above. No harm in adding a ref to the array when it's kind of unknown to the linter what that really is ( [reference](https://www.epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array#custom-hooks-and-the-ref-object-itself) )

packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
```diff
@@ -318,29 +317,40 @@ export const TearsheetShell = React.forwardRef(
...
+        ) {
+          claimFocus();
+        }
+      });
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [depth, position]);
+
+    useEffect(() => {
+      if (hasError && !modalRef?.current?.contains(document.activeElement)) {
+        claimFocus();
```
jlongshore, line 350: This may go beyond the scope of this review... but should claimFocus be wrapped in a useCallback? That would allow us to add that in the dep array of these useEffects and not worry about runaway re-renders... and we could lift some of these disable next line rules.

packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
```diff
@@ -318,29 +317,40 @@ export const TearsheetShell = React.forwardRef(
...
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [hasError]);
```
jlongshore, line 353: `modalRef` in the dep array... potentially `claimFocus` (see comment below)

packages/ibm-products/src/global/js/hooks/useFocus.js
```diff
@@ -85,40 +85,27 @@ export const useFocus = (modalRef, selectorPrimaryFocus) => {
     }
   };
 
+  const claimFocus = () => {
```
jlongshore, line 88: Should this be wrapped in a useCallback so we can add it to the dep arrays in our useCallbacks above? If we do that we could potentially lift all of the `eslint-disable-next-line react-hooks/exhaustive-deps` rules

jlongshore (commented):
Just a few observations... they could be outside the scope of what this PR is intending to do also, just looking for discussion.

packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
```diff
@@ -318,29 +317,40 @@ export const TearsheetShell = React.forwardRef(
...
-      firstElement,
-      modalRef,
-      modalRefValue,
-      open,
-      position,
-      selectorPrimaryFocus,
-      hasError,
-    ]);
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [open, currentStep, effectiveHasCloseIcon]);
```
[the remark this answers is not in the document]
makafsal: 1. `currentStep` - to handle the focus when step change in `CreateTearsheet`

2. `effectiveHasCloseIcon` - Currently, we don't have a direct way to hide the close button in ComposedModal, so we're hiding it using styles, so this will cause to appear (not visibly) the close icon then become hidden (after applying style). This causes a focus issue if the focus is on the close button initially, so we've to refocus.

packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
makafsal: 👍

jlongshore (approved):
LGTM

PR #7125, opened 2025-03-19
chore: update build job to include multiple major react versions

### LINKED ISSUE #7044: [Infrastructure]: update GH workflow to test different major versions of react
reopened on 3/26/25

after a lot of trial and error I've decided to reopen this issue. after working through https://github.com/carbon-design-system/ibm-products/issues/7137 i've come to the conclusion that in order to properly ensure backwards compatibility with major react versions we would have to update our tests to be able to run in all the major versions we support.

we discussed the idea of being able to test for this in the build, but that's not sufficient. i went into detail [here](https://github.com/carbon-design-system/ibm-products/issues/7137#issuecomment-2755500176) but basically the problem here is that we can't test for runtime errors during the build process. these can only be checked for in unit tests.

so while there's validity to ensuring builds compile with different major react versions we would still have to fix all of our tests to be backwards compatible. the good news is that a sizable amount of our tests run successfully in 17 and 18. a sizeable amount of tests can easily be fixed by switching the `act` import from `react` to `@testing-react`. the remainder seem to revolve around asynchronous actions we're testing with `act` which will take some time to refactor to work with 17 and 18.

our options seem to be the following

* continue the effort and refactor the tests to be backwards compatible
* ignore versions 16 and 17 of react and focus this effort moving forward with 18 and 19
* see if we can implement testing that only occurs for specific versions of react

if we can determine the version of react we are testing at test time we should be able to use specific selector or tag like `v17` to run tests specific to that version. this would alleviate having to try make complex test cases backwards compatible.

Image

davidmenendez:
Closes #7044 & #7137

adds a new job for testing other major versions of React. since we are currently on v18 i didn't think it made sense to include `18` in the matrix, but if we wanted to be extra safe we could also include it `[17, 18]`


.github/workflows/ci.yml
```diff
@@ -119,6 +119,32 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
       - name: CI snapshot tests for c4p
         run: yarn ci-check:test:c4p:snapshot
+  test-c4p-major-version:
```
matthewgallo, line 122: Small thing but `test-react-major-version` or `test-c4p-react-major-version` might make more sense

.github/workflows/ci.yml
```diff
@@ -119,6 +119,32 @@ jobs:
...
+        uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 #4.2.0
+        with:
+          path: '**/node_modules'
+          key:
+            test-c4p-major-versions-modules-${{ runner.os }}-${{
+            hashFiles('yarn.lock') }}
+      - name: Install
+      - run: yarn
+      - name: Install React version ${{ matrix.react }}
+      - run: yarn add react@${{ matrix.react }} react-dom@${{ matrix.react }}
```
matthewgallo, line 145: You might also need `--frozen-lockfile` to prevent changes to `yarn.lock` in this step

.github/workflows/ci.yml
```diff
@@ -119,6 +119,34 @@ jobs:
...
+          cache: yarn
+      - name: Cache dependencies
+        uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 #4.2.0
+        with:
+          path: '**/node_modules'
+          key:
+            test-c4p-react-major-versions-modules-${{ runner.os }}-${{
+            hashFiles('yarn.lock') }}
+      - name: Install
+      - run: yarn
```
matthewgallo, line 143:
```suggestion
        run: yarn
```

I saw when the action ran in your fork there was some yml issues, `run` shouldn't need the dash since it's supposed to be within `- name: Install`

.github/workflows/ci.yml
```diff
@@ -119,6 +119,34 @@ jobs:
...
+      - name: Install React version ${{ matrix.react }}
+      - run:
```
matthewgallo, line 145: Same here about the yml formatting

.github/workflows/ci.yml
```diff
@@ -119,6 +119,34 @@ jobs:
...
+          yarn add react@${{ matrix.react }} react-dom@${{ matrix.react }}
+          --frozen-lockfile
+      - name: CI tests for c4p with React version ${{ matrix.react }}
+      - run: yarn ci-check:test:c4p
```
matthewgallo, line 149: Same here about the yml formatting

.github/workflows/ci.yml
davidmenendez: oops!

davidmenendez:
update: the action appears to be working as intended, but unfortunately the tests are failing. i'm currently investigating getting the tests to pass in v17

davidmenendez:
this should be ready to review.

there's a lot of file changes, but the bulk of it is just to test files to add `async` and fix the `act` import.

i added `.skip` to some tests because they were causing failures in v17. i think this should be fine for now and i'm going to open up a follow up issue to address these skipped tests. basically the majority of the problems i faced dealing with this issue is the asynchronous implementation in tests. it changed either in react or the testing library between 17 and 18 and the tests were failing because async wasn't included. anything that's being skipped as part of this PR i think is just indicative of our flaky testing, which we need to look at anyways.

anamikaanu96:
Hey @davidmenendez The test is failing in createFullpage and some merge conflicts. Could you take a look?

davidmenendez:
code coverage dropped, but it's necessary at the moment until we can fix a lot of these flaky tests.

anamikaanu96 (approved):
Code looks good to me 🎉 
Please open a separate ticket for code coverage.

matthewgallo (commented):
There are a lot of `.skip`s that have been added. Do we still need them?

davidmenendez:
@matthewgallo i'm going to add a followup issue to address this. but basically there's some issue with async functionality between the major test versions. rather then try to hunt down the problematic code on this PR i decided to just skip it for now.

packages/ibm-products/src/components/FilterPanel/FilterPanelCheckboxWithOverflow/FilterPanelCheckboxWithOverflow.test.js
```diff
@@ -33,7 +33,8 @@ const renderComponent = ({ ...rest } = {}) =>
       labelText={labelText}
       {...{ ...rest }}
     >
-      <OverflowMenuItem itemText="Option 1" />
+      {/* <OverflowMenuItem itemText="Option 1" /> */}
+      <OverflowMenuItem itemText="FART" />
```
matthewgallo, line 37: 😂 Lets stick with "Option 1"

davidmenendez: oh gosh 🫢

packages/ibm-products/src/components/CoachmarkStack/CoachmarkStackHome.tsx
```diff
@@ -96,183 +95,170 @@ const componentName = 'CoachmarkStackHome';
...
-                setLinkFocusIndex(0);
-                onClose();
-              }}
-            />
-            <div className={`${overlayClass}__body`}>
-              <div className={`${overlayClass}-element`}>
-                {!renderMedia && (
-                  <Idea size={20} className={`${blockClass}__icon-idea`} />
-                )}
+  return createPortal(
```
matthewgallo, line 195: The [Next.js portal example](https://github.com/vercel/next.js/blob/canary/examples/with-portals/components/ClientOnlyPortal.js) closer aligns to what we currently have where a portal is conditionally rendered if the `portalNode` ref (or some value tracking if the component has mounted) has been set.

packages/ibm-products/src/components/CoachmarkStack/CoachmarkStackHome.tsx
```diff
@@ -96,183 +95,170 @@ const componentName = 'CoachmarkStackHome';
...
-          document?.querySelector('body'))
-        : document?.querySelector('body');
-    }, [portalTarget]);
+  const portalNode = useRef<Element>(null);
 
-    if (!navLinkLabels) {
-      return pconsole.warn(
-        `${componentName} is an Onboarding internal component and is not intended for general use.`
-      );
+  useEffect(() => {
```
matthewgallo, line 128: Will this cause SSR issues? It was a while back, but didn't we move to using `useIsomorphicEffect` because of that?

.github/workflows/ci.yml
```diff
@@ -121,6 +121,43 @@ jobs:
...
+          path: '**/node_modules'
+          key:
+            test-c4p-react-major-versions-modules-${{ runner.os }}-${{
+            hashFiles('yarn.lock') }}
+      - name: Install
+        run: yarn --immutable
+      - name: Add React version ${{ matrix.react }}
+        run: |
+          yarn add react@${{ matrix.react }} react-dom@${{ matrix.react }}
+          yarn workspace @carbon/ibm-cloud-cognitive-core add react@${{ matrix.react }} react-dom@${{ matrix.react }} -D
```
matthewgallo, line 151:
```suggestion
          yarn workspace @carbon/ibm-products add react@${{ matrix.react }} react-dom@${{ matrix.react }} -D
```

.github/workflows/ci.yml
```diff
@@ -121,6 +121,43 @@ jobs:
...
+      - name: Add legacy dependencies
+        if: ${{ startsWith(matrix.react, '^17.') }}
+        run: |
+          yarn add @testing-library/react@^12.0.0 -D
+          yarn workspace @carbon/ibm-cloud-cognitive-core add @carbon/react@^1.28.0 -D
+          yarn workspace @carbon/ibm-products add @carbon/react@^1.28.0 -P
```
matthewgallo, line 157: Do you need both of these lines? Probably just the second, right? `packages/core` isn't really used for anything now.

Also, how come you're installing `1.28.0`?

packages/ibm-products/src/components/CoachmarkStack/CoachmarkStackHome.tsx
```diff
@@ -96,183 +95,170 @@ const componentName = 'CoachmarkStackHome';
...
+    renderMedia,
+    navLinkLabels,
+    onClickNavItem,
+    onClose,
+    portalTarget,
+    closeButtonLabel,
+    title,
+    tooltipAlign,
+    ...rest
+  } = props;
```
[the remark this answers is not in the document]
davidmenendez: i find this pattern to be useful because you can easily log all the props before separating them. it also declutters the function signature a bit.

.github/workflows/ci.yml
davidmenendez: yeah good call now that core has been decoupled.

`1.28` if i remember correctly i think this version didn't include a bug that was added in a subsequent `1.x` version. this PR has been sitting around for a while so i wish i could remember the specifics of why up top, but i remember there was a specific reason

packages/ibm-products/src/components/CoachmarkStack/CoachmarkStackHome.tsx
davidmenendez: ok let me take another look at this 👍

packages/ibm-products/src/components/CoachmarkStack/CoachmarkStackHome.tsx
davidmenendez: oh interesting. let me put that back

davidmenendez:
its been 4 months since this PR was opened and now that we're on the cusp of react 19 support i don't think it makes sense to keep this going. i will open up a follow up issue to investigate CI testing for 18 + 19. based on what i've seen with the latest major version of jest it may prove to be just as tricky to implement, but we will see!

PR #7132, opened 2025-03-20
feat: setup renovate, add batch upgrade for all carbon packages

### LINKED ISSUE #7118: Add renovate for dependency upgrades

matthewgallo:
Closes #7118 

Adds renovate config, this will allow us to upgrade packages more easily since renovate does upgrades for single packages. However, you can batch packages together which I've configured in this PR so that all `@carbon/*` packages are upgraded together. I used the same time interval that we've been using (1st and 15th of every month).

These changes should also allow us to delete the workflow files we have for upgrading packages via `npm-check-updates`, want to make sure this setup works before deleting those.

#### What did you change?
```
.github/renovate.json5
```
I've been able to test that there are no renovate configuration errors via the following:
```
npx --yes --package renovate -- renovate-config-validator
```
image

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,0 +1,33 @@
+{
+  "extends": [
+    // https://docs.renovatebot.com/presets-config/#configjs-lib
+    "config:js-lib",
+
+    // https://docs.renovatebot.com/presets-default/#maintainlockfilesweekly
+    ":maintainLockFilesWeekly",
+
+    // https://docs.renovatebot.com/presets-default/#preservesemverranges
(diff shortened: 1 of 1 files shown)


elycheea (commented):
I don’t see the Renovate App added to our repo. Do we not need to have that set up first?

From the docs, it looks like we should install the GitHub app first and then add the config?
https://docs.renovatebot.com/getting-started/installing-onboarding/

matthewgallo:
Going to close this PR in favor of https://github.com/carbon-design-system/ibm-products/pull/7134 that was automatically opened by the renovate gh app.

PR #7161, opened 2025-03-25
fix(custom typings): include section

wkeese:
Refs #4336, #4225 

Quick fix for custom-typings/index.d.ts to include the `Section` component. 

#### What did you change?

Added `Section` to the list of components exported from @carbon/react.

Also removed all the duplicate exports.

Of course, ideally we would just fix the missing exports in @carbon/react, and then get rid of this whole declaration workaround.

cc @matthewgallo.

Built locally


elycheea (approved):
Thanks @wkeese for the fix and cleaning up the duplicates!

amal-k-joy (approved):

Issue #6981, opened 2025-02-26
[a11y]: Focus escapes from OptionsTile to the first focusable element


JKob207:
### Package

@carbon/ibm-products

### Browser

Chrome

### Operating System

MacOS

### Package version

v2.56.1

### React version

v17.0.2

### Automated testing tool and ruleset

IBM Accessibility Checker

### Description

There is an issue when using `Tearsheet` component with `OptionsTile`. When in `OptionsTile` I have enabled the switch button and try to toggle that, focus goes to the first focusable element within the tearsheet. It happens only during the first click of the button. Focus should remain on `OptionsTile` switch button.

### Reproduction/example

https://stackblitz.com/edit/github-q1hum2y6-cz9b5naw?file=src%2FExample%2FExample.jsx

### Steps to reproduce

1. Using mouse or keyboard navigation toggle the switch button.
2. Focus escapes to the first focusable element within tearsheet.
3. Another clicking on the switch button doesn't cause the problem.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate problems

wkeese:
I believe both these issues were already fixed via #7063.  (It doesn't use `firstElement` any more.)  But @timfuhrmann said he would confirm.

timfuhrmann:
I can confirm that both issues have been fixed with https://github.com/carbon-design-system/ibm-products/pull/7063, see forked stackblitz: https://stackblitz.com/edit/github-q1hum2y6-w2ygg34a?file=package.json

PR #7183, opened 2025-03-27
fix(Tearsheet): remove nodes from dependencies

timfuhrmann:
Closes https://github.com/carbon-design-system/ibm-products/issues/6981

When rendering conditionally instead of passing `open` to the tearsheet, the focus jumps back to the primary focus on re-render: https://stackblitz.com/edit/github-vqpu3jc2?file=src%2FExample%2FExample.jsx

#### What did you change?
- Removed `firstElement` from dependency array, calling getFocusable from within instead
- Removed unused `modalRefValue` from dependency array 

- Storybook `selectorPrimaryFocus` is still working as expected 
- Storybook with new story for rendering conditionally and case from issue linked above (not included) is working as expected now

### DIFF
--- a/packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
+++ b/packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
@@ -281,11 +281,10 @@ export const TearsheetShell = React.forwardRef(
     const modalRef = (ref || localRef) as MutableRefObject<HTMLDivElement>;
     const { width } = useResizeObserver(resizer);
     const prevOpen = usePreviousValue(open);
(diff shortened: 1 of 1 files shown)


timfuhrmann:

timfuhrmann:
recheck

devadula-nandan:
@makafsal looks like the issue got assigned to you, but @timfuhrmann has opened this pr which closes the issue. could you check it to avoid any duplicity of efforts if you are already working on it? 

@timfuhrmann
also there is a test failing, could you pls look into it?

timfuhrmann:
Fixed already with https://github.com/carbon-design-system/ibm-products/pull/7063

PR #7205, opened 2025-04-01
feat: options tile webcomponent init

### LINKED ISSUE #7140: Create options tile component
The options tile Web Component can be added directly to the web-components directory in this monorepo. It should be based on the existing [React implementation](https://ibm-products.carbondesignsystem.com/?path=/docs/ibm-products-components-options-tile).

> [!NOTE]
> The current React implementation of the OptionsTile includes `locked`, `invalid`, and `warn` as separate if statements, but in reality, the options tile component can only reflect one of the statuses at any given moment. This is partially as designed[^1] from the hierarchy section of the design guidance but options tile API is quite opaque with this implementation.

This refinement also includes improvements suggested from #6936 as these are breaking changes in React but  adopting these could result in a cleaner implementation in Web Components.

### Acceptance criteria
- [x] Add support for `title`
  - [x] Include `titleId` as optional
- [x] Add support for `open` state
- [x] Add support for `lg` and `xl` sizes
- [x] Add support for `summary`
- [x] Add default story — can see if we can use select with options to show single story with/without toggle

#### Questions for discussion
- [x] Should we use an `enabled` prop[^2] or pass in `Toggle` via slot? Feels like slot makes more sense for WC implementation, because if we refactored the React version, we might consider `renderToggle`.
- [x] Validation states
  - Can we change the current `locked`, `invalid`, and `warn` into a single validation object or should we use a slot for handling these?
  - How do we want to handle the naming for these?
  - does this include `summary` as a single `statusMessage`?
- [x] Can we replace `onChange` with `onOpen` and `onClose` instead?

[^1]: [Options tile validation](https://pages.github.ibm.com/carbon/ibm-products/components/options-tile/usage/#validation)
[^2]: `enabled` is used for the child toggle component’s state _but also_ affects its rendering

davidmenendez:
Closes #7140


packages/ibm-products-web-components/src/components/options-tile/options-tile.ts
```diff
@@ -0,0 +1,137 @@
...
+   */
+  @property({ type: Function })
+  onClose?: (evt: Event) => void;
+
+  /**
+   * Callback fired when the component requests to be opened
+   */
+  @property({ type: Function })
+  onOpen?: (evt: Event) => void;
+
```
devadula-nandan, line 46: we define custom events in a different way
reference
https://github.com/carbon-design-system/carbon/blob/1651ed286434813b8ee6a24b81600b7448b0e8b5/packages/web-components/src/components/dropdown/dropdown.ts#L294

devadula-nandan: Screenshot 2025-04-09 at 11 25 40 AM
it adds a separate section like above.

devadula-nandan:
https://github.com/carbon-design-system/carbon/pull/19071
i have opened a pr for toggle focus ring issue for the web component

packages/ibm-products-web-components/src/components/options-tile/options-tile.ts
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * @element c4p-options-tile
+ * */
+
+@customElement(`${prefix}-options-tile`)
+class CDSOptionsTile extends HostListenerMixin(LitElement) {
+  /**
+   * If `true` the body of the component is shown
+   */
+  @property({ type: Boolean })
```
AlexanderMelox, line 32:
```suggestion
  @property({ type: Boolean, reflect: true })
```

I think we've been adding reflect true to properties

packages/ibm-products-web-components/src/components/options-tile/options-tile.ts
```diff
@@ -0,0 +1,137 @@
...
+  /**
+   * Callback fired when the component requests to be opened
+   */
+  @property({ type: Function })
+  onOpen?: (evt: Event) => void;
+
+  /**
+   * Determines the size of the header
+   */
+  @property({ type: String })
```
AlexanderMelox, line 50:
```suggestion
  @property({ type: String, reflect: true })
```
```suggestion
  @property({ type: String })
```

packages/ibm-products-web-components/src/components/options-tile/options-tile.ts
```diff
@@ -0,0 +1,137 @@
...
+  size?: 'lg' | 'xl' = 'lg';
+
+  /**
+   * Text that is displayed under the title
+   */
+  @property({ type: String })
```
AlexanderMelox, line 56:
```suggestion
  @property({ type: String, reflect: true })
```

packages/ibm-products-web-components/src/components/options-tile/options-tile.ts
```diff
@@ -0,0 +1,137 @@
...
+  summary?: string;
+
+  /**
+   * Text for the title
+   */
+  @property({ type: String })
```
AlexanderMelox, line 62:
```suggestion
  @property({ type: String, reflect: true })
```

packages/ibm-products-web-components/src/components/options-tile/options-tile.ts
```diff
@@ -0,0 +1,137 @@
...
+      </div>
+    `;
+  }
+
+  private _toggle() {
+    if (this.open) {
+      this._onClose();
+    } else {
+      this._onOpen();
+    }
```
AlexanderMelox, line 115: Prob nice to save some line space, just personal preference though
```suggestion
   this.open ? this._onClose() : this._onOpen();
```

packages/ibm-products-web-components/src/components/options-tile/options-tile.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  private _onOpen() {
```
AlexanderMelox, line 118: I thought I was under the impression the "on" function prefix is when users pass it to the components and "handle" is internal. I might be wrong but prob good to bring up in next dev engineering huddle

AlexanderMelox (changes requested):
Looking good! I would suggest also adding some basic testing for this component or opening a follow up issue to add tests

packages/ibm-products-web-components/package.json
```diff
@@ -89,7 +89,7 @@
     "playwright": "^1.51.1",
     "postcss": "^8.4.47",
     "remark-gfm": "^4.0.0",
-    "rimraf": "^5.0.5",
```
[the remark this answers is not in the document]
davidmenendez: yeah it fixes an issue where `yarn clean` wasn't was failing to run

packages/ibm-products-web-components/src/components/options-tile/options-tile.ts
```diff
@@ -0,0 +1,137 @@
...
+import HostListenerMixin from '@carbon/web-components/es/globals/mixins/host-listener.js';
+import styles from './options-tile.scss?lit';
+import { carbonElement as customElement } from '@carbon/web-components/es/globals/decorators/carbon-element.js';
+import ChevronDown20 from '@carbon/web-components/es/icons/chevron--down/20';
+
+const blockClass = `${prefix}--options-tile`;
+
+/**
+ * OptionsTile.
+ *
```
sangeethababu9223, line 23: Hey @davidmenendez,
Could we document our custom events here? As done in tearsheet?
like this: 
<pre> ```/**
 * Tearsheet.
 *
 * @element c4p-tearsheet
 * @csspart dialog The dialog.
 * @fires c4p-tearsheet-beingclosed
 *   The custom event fired before this tearsheet is being closed upon a user gesture.
 *   Cancellation of this event stops the user-initiated action of closing this tearsheet.
 * @fires c4p-tearsheet-closed - The custom event fired after this tearsheet is closed upon a user gesture.
 */
@customElement(`${prefix}-tearsheet`)
class CDSTearsheet extends HostListenerMixin(LitElement) {}``` </pre>

packages/ibm-products-web-components/src/components/options-tile/options-tile.ts
```diff
@@ -0,0 +1,137 @@
...
+    this.dispatchEvent(new CustomEvent('onOpen', options));
+  }
+
+  private _onClose() {
+    const options = {
+      bubbles: true,
+      composed: true,
+    };
+    this.dispatchEvent(new CustomEvent('onClose', options));
+  }
```
sangeethababu9223, line 132: Here dispatch event could be modified to add types as done [with Tearsheet.](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products-web-components/src/components/tearsheet/tearsheet.ts#L291C14-L291C27)

packages/ibm-products-web-components/src/components/options-tile/options-tile.stories.ts
```diff
@@ -0,0 +1,95 @@
...
+  return html`
+    <c4p-options-tile
+      id="my-tile"
+      ?open=${open}
+      size=${size}
+      summary=${summary}
+      title=${title}
+      titleId=${titleId}
+      @onOpen=${handleOpen}
+      @onClose=${handleClose}
```
sangeethababu9223, line 65: If we are to rename the custom events like `c4p-option-tile-on-close` and  `c4p-option-tile-on-open` then this should also be modified to `@c4p-option-tile-on-close=${handleClose}` and `@c4p-option-tile-on-open=${handleOpen}`

davidmenendez:
@AlexanderMelox @sangeethababu9223 updated per feedback. please re-review. thanks!

davidmenendez:
i think everything is working and looking correctly. the only thing i can't seem to figure out is positioning the toggle

Screenshot 2025-04-23 at 2 32 07 PM

i just need to apply `display: flex` to the label `.cds--toggle__label` but because it's a child of `<cds-toggle>` i don't think i can target it? i've tried using `::slotted` but it doesn't appear to work even with a catch all `::slotted(*)`. i _think_ the toggle component needs to use [exportparts](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/exportparts) to make that child targetable through styles. but i'm still kind of new to this so if i'm just missing something super obvious please let me know!

AlexanderMelox (approved):

AlexanderMelox:
Might be a good idea while we build our web components to have codecov not fail builds? What do you guys think? @matthewgallo @elycheea

sangeethababu9223 (approved):
Looks great! 🎉

Issue #7067, opened 2025-03-11
[a11y]: Datagrid sort button invalid ARIA attribute


Jonathan-Maciel:
### Package

@carbon/ibm-products

### Browser

Chrome

### Package version

v2.54.0

### React version

v16.10.0

### Automated testing tool and ruleset

IBM Equal Access Accessibility Checker

### Description

The sort button has the `aria-sort` attribute, but I believe this should be on the parent `<th>`.

![Image](https://github.com/user-attachments/assets/c6f8f095-98a1-4967-bb5d-16f37f1b6679)
![Image](https://github.com/user-attachments/assets/a0378509-8ab0-4332-8226-53f24a001834)

### Reproduction/example

https://carbon-for-ibm-products.netlify.app/?path=/story/deprecated-datagrid-datagrid--sortable-columns&globals=viewport:basic

### Steps to reproduce

Inspecting the linked storybook example, you can see that the `aria-sort` label is applied to the button and not the column head. 

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate problems

wkeese:
Note that it works better in vanilla Carbon Datagrid, or at least aria-sort is not on the button, see https://react.carbondesignsystem.com/?path=/story/components-datatable-sorting--default.

```html
<th aria-sort="ascending" class="cds--table-sort__header" scope="col">
  <div class="cds--table-sort__description" id="table-sort-:r39:">
    Click to sort rows by Name header in descending order
  </div>
  <button
    type="button"
    aria-describedby="table-sort-:r39:"
    class="cds--table-sort cds--table-sort--active"
  >
    <span class="cds--table-sort__flex"
      ><div class="cds--table-header-label">Name</div>
      <svg ... > ... </svg>
      <svg ... > ... </svg>
      <div class="cds--table-header-label--decorator-inner"></div
    ></span>
  </button>
</th>

{{{

Jonathan-Maciel:
[...]

PR #7206, opened 2025-04-01
fix: https://github.com/carbon-design-system/ibm-products/issues/7067

Jonathan-Maciel:
Closes https://github.com/carbon-design-system/ibm-products/issues/7067

move aria-sort to th on datagrid

#### What did you change?
I moved the aria-sort label from the `Button` component and into the `th` element of the header row. A carbon styles file is applying `padding 0` to the headers so that had to be overwritten as well.

I ran `yarn storybook` and verified that the sortable datagrid component styling was consistent with the existing code, and I verified that the page can now pass the accessibility checker.

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHeaderRow.tsx
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHeaderRow.tsx
@@ -114,6 +114,27 @@ const ResizeHeader = ({
   );
 };
 
+const getAriaSortValue = (col, datagridState) => {
+  const {
+    ascendingSortableLabelText,
+    descendingSortableLabelText,
+    defaultSortableLabelText,
+  } = datagridState;
+  if (!col) {
+    return;
+  }
+  const { isSorted, isSortedDesc } = col;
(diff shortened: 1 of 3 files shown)


Jonathan-Maciel:

packages/ibm-products/src/components/Datagrid/Datagrid/DatagridHeaderRow.tsx
```diff
@@ -256,6 +277,11 @@ const HeaderRow = (
               )}
               key={header.id}
               aria-hidden={header.id === 'spacer' && 'true'}
+              aria-sort={
```
amal-k-joy, line 280: Looks good, but in cases where a column is not sortable(`disableSortBy:true`), we shouldn't set aria-sort on it—even if the table supports sorting. That could confuse screen readers.

amal-k-joy (changes requested):
Thanks for your contribution. Though added a minor comment

amal-k-joy (approved):
Looks good.

devadula-nandan (approved):
LGTM

PR #7218, opened 2025-04-02
chore: add vrt script back into ci.yml

### LINKED ISSUE #7116: Move `visual-snapshot` script back to ci workflow
We currently run our visual snapshot script as part of our netlify build command, this was so that we didn't have to expose our percy secret from `ci.yml`. But in order to reduce our total snapshot counts, we should only run `visual-snapshot` when a PR is "Ready for Review" which means it needs to be removed from the netlify build command and added by to our `ci.yml`.

Reference core's PR:
https://github.com/carbon-design-system/carbon/pull/18826

### Acceptance criteria
- [ ] Move `visual-snapshot` from netlify build command back into `ci.yml` workflow (so we can prevent running percy on draft PRs). This will require access to netlify, in order to update the build command.
- [ ] When we initially did this there were issues with using the percy token as a repo secret, this will also need to be explored so that we don't expose it directly in `ci.yml`

davidmenendez:
Closes #7116 

need to confirm this works before disabling in netlify

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,6 +206,7 @@ jobs:
       - name: Run VRT
         env:
           PERCY_TOKEN: ${{ secrets.PERCY_TOKEN_IBM_PRODUCTS }}
+          TEST_TOKEN: 1234
         if:
           ${{ github.repository == 'carbon-design-system/ibm-products' &&
           github.event.pull_request.draft == false }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,8 +205,8 @@ jobs:
         run: yarn build
       - name: Run VRT
         if:
-          github.repository == 'carbon-design-system/ibm-products &&
-          github.event.pull_request.draft == false'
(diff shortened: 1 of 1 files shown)


davidmenendez:
the job is failing here, but according to the [secrets and workflows docs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow)

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

so i _think_ this has to be merged to verify that the token is set properly.

devadula-nandan (dismissed):

elycheea (dismissed):

davidmenendez:
will need to merge without requirements to bypass failure. will wait till tomorrow morning as a precaution to ensure this is working properly and not breaking builds for others.

davidmenendez:
@elycheea @devadula-nandan please re-review. i added an additional check just to ensure that if this doesn't work that the builds won't fail

devadula-nandan (approved):

elycheea (approved):

Issue #7110, opened 2025-03-17
[a11y]: The <li> element has the id (id) that is already in use


parul-26:
### Package

@carbon/ibm-products

### Browser

Chrome

### Operating System

MacOS

### Package version

v2.18.1

### Automated testing tool and ruleset

IBM Equal Access Accessibility Checker- Latest Deployment

### Description

We are encountering an issue where the <li> element has duplicate id attributes, which is causing the following error:

`The <li> element has the id "someone8" that is already in use.`

Root Cause:

The error occurs because the id values for the table headers are being reused within the Tearsheet's customized columns. This leads to the same id being assigned to both the table header and the corresponding list elements in the customized columns.

<img width="1726" alt="Image" src="https://github.com/user-attachments/assets/500c9495-e28a-40c4-8477-5719609af95e" />

### Reproduction/example

https://ibm-products.carbondesignsystem.com/iframe.html?globals=viewport%3Abasic&viewMode=docs&id=deprecated-datagrid-datagrid-columncustomization--docs&args=#component-api

### Steps to reproduce

1. Go to the shared documentation ([link provided](https://ibm-products.carbondesignsystem.com/iframe.html?globals=viewport%3Abasic&viewMode=docs&id=deprecated-datagrid-datagrid-columncustomization--docs&args=#component-api)).
2. Click on Customize Columns in the table or Tearsheet section.
3. Scan the page for the table headers and customized columns.
4. Observe the issue where the same id is being used for both table headers and the <li> elements in the customized columns.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate problems

elycheea:
@parul-26 Thanks for reporting this! The Datagrid is currently deprecated in favor of [`TanStack/Table`](https://github.com/carbon-design-system/tanstack-carbon), which unfortunately means support for bugs and enhancements is limited. If you’d like to [contribute a fix](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CONTRIBUTING.md), I’m happy to reopen this issue for you.

parul-26:
@elycheea can you please reopen this issue as i want to contribute to the fix. Thankyou!

elycheea:
[...]

PR #7221, opened 2025-04-03
fix(customizeColumns): update column ids for accessibility

parul-26:
Closes #https://github.com/carbon-design-system/ibm-products/issues/7110

{{short description}}
The data-grid table headers and the customised coloumns are sharing the same id's causing `The <li> element has the id (id) that is already in use` 

#### What did you change?
updated the id's of the customised coloumns to solve the accessibility voilations. 

image


amal-k-joy (approved):
LGTM

elycheea (dismissed):

amal-k-joy (changes requested):
Hey @parul-26 ,
Everything looks good, but one test is failing.

elycheea:
@amal-k-joy ahh, good catch. We had to manually approve CI to run. @parul-26 looks like we’ll need to update the customize columns test.

amal-k-joy (approved):

Issue #7121, opened 2025-03-19
Composable Interstitial: implement `InterstitialBody`


matthewgallo:
```
<InterstitialScreen.Body  contentRenderer={({all internal states})=>(
                    //Here user can render any custom content.
                    //They can return a array of children for multistep or a single child.
                    //Based on the children array , we manage carousal and progress indicator
                   // They can render any thing and even a progres indicator or media here since we are providing them with all required internal states and is completely composable.
              
                  )}
/>
```

### To do
- [ ] check with Susana on the patterns/designs to be given as example codes.

### Acceptance criteria
- [ ] Dev implementation with keyboard navigation
- [ ] Unit testing
- [ ] AVT testing
- [ ] Example patterns creation

Issue #6442, opened 2024-11-15
[Feature Request]: Add Distinct Callbacks for Buttons in InterstitialScreen


Yashrajsingh2001:
### The problem

The `InterstitialScreen` component currently allows developers to set only the labels for its buttons using the following props:

- `nextButtonLabel="Next"`
- `previousButtonLabel="Back"`
- `skipButtonLabel="Skip"`
- `startButtonLabel="Save"`

However, developers cannot assign separate functions to these buttons. Regardless of which button is clicked ("Skip," "Save," or "X" for Close), the `onClose` function is called. This creates a limitation where all actions are handled by the same function, making it impossible to differentiate the logic based on the button clicked.

### The solution

Introduce new callback props for individual button actions:
- **`onSkip`**: A function triggered when the "Skip" button is clicked.
- **`onSave`**: A function triggered when the "Save" button is clicked.
- **`onClose`**: A function specifically triggered when the "X" (Close) button is clicked.

This change will provide developers the flexibility to define separate logic for each action, improving the component's usability and extensibility.

### Owner/main maintainer(s)

NA

### Business priority

None

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

amal-k-joy:
Should we expose different callbacks for each actions or a unified callback like below for all actions passing the action type. Thus , we may need to introduce only 1 new callback prop
`handleActionButtonClick(actionType,e,...)`.

matthewgallo:
### Acceptance criteria

- [ ] Single async callback, returning the action type (potential naming, `onAction`, `onRequestAction`)
- [ ] Review with Carbon for any naming convenitions
- [ ] Include loading indicator 
- [ ] Include test for new callback prop

*Note, there will be some composability changes to this component in the future (1H 2025) impacting how actions are provided in this component

paul-balchin-ibm:
I would vote for individual callbacks:
- This would follow existing standards `onClick`, `onBlur`, `onToggle`, etc. i.e. one type of event calls one callback function.
- Naming conventions such as `onSkip`, `onSave`, and `onClose`, etc. are self documenting and don't require further explanation.
[...]

PR #7229, opened 2025-04-04
feat: composable interstitial screen implementation

### LINKED ISSUE #7120: Composable Interstitial: implement `InterstitialHeader`
**Proposed** 

```
<InterstitialScreen isFullScreen={true/false} hideProgressIndicator isOpen>  

           //Based on the disccussion with Susana and Tracy, they need a new prop in header for subtitle(headerSubTitle)
            <Interstitial.Header headerTitle={'headerTitle'} headerSubTitle ={'headerSubTitle'} hideProgressIndicator  >
        
                   //This need to be passed only in full screen mode. User can chose not to pass FullScreenScreenHeader and reuse their 
                   //existing header.
                  <Interstitial.FullScreenScreenHeader domainName productName>
                    children
                  </Interstitial.FullScreenScreenHeader>
                  
                  children
              
                  //when isFullScreen is false, we use carbon compose modal header and we can accept any prop supported by carbon 
                  </Interstitial.Header>
              
                  <Interstitial.Body  contentRenderer={({all internal states})=>(
                    //Here user can render any custom content.
                    //They can return a array of children for multistep or a single child.
                    //Based on the children array , we manage carousal and progress indicator
                   // They can render any thing and even a progres indicator or media here since we are providing them with all required internal states and is completely composable.
              
                  )}>
              
                   </Interstitial.Body>
        
        <Interstitial.Footer actionButtonConfig={} actionButtonRenderer={} />

</InterstitialScreen> 

```
```
Interstitial.HeaderTemplate={
 FullScreenScreenHeader

 headerTitle
 headerSubTitle
 ProgressIndicator

children

};

FullScreenScreenHeaderTemplate={
domainName | productName
children

}
```

### Acceptance criteria
- [ ] Dev implementation with keyboard navigation, remove full screen global header support (`domainName` and `productName`)
- [ ] Unit testing
- [ ] AVT testing
- [ ] Example patterns creation

amal-k-joy:
Closes #7120, #7121, #7122, #6442

Implementation of composable interstitial screen as a compound component pattern.
InterstitialScreen acting as a container and **InterstitialScreen.Header, InterstitialScreen.Body, InterstitialScreen.Footer** 
are static properties of the InterstitialScreen component, effectively acting as sub-components.

Removed `InterstitialScreenView` and `InterstitialScreenViewModule` as components and moved as storybook assets.

storybook


packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenBody.tsx
```diff
@@ -0,0 +1,158 @@
...
+    );
+    if (
+      isValidElement(_bodyContent) &&
+      Array.isArray((_bodyContent.props as { children: ReactNode }).children)
+    ) {
+      setIsMultiStep(
+        !!(_bodyContent.props as { children: ReactNode[] }).children?.length
+      );
+      setStepCount?.(
+        (_bodyContent.props as { children: ReactNode[] }).children?.length
```
davidmenendez, line 88: this seems redundant `(_bodyContent.props as { children: ReactNode[] }).children?.length` can we cast this as a variable earlier to just reference?

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenBody.tsx
```diff
@@ -0,0 +1,158 @@
...
+      ...disableButtonConfig,
+      ...config,
+    });
+  };
+
+  const onScrollHandler = (scrollPercent) => setScrollPercent(scrollPercent);
+
+  const renderBody = () => (
+    <div className={`${blockClass}--body ${className}`} ref={bodyScrollRef}>
+      <div className={cx(`${blockClass}--content`)}>
```
davidmenendez, line 121: don't need to use `cx` here

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenBody.tsx
```diff
@@ -0,0 +1,158 @@
...
+            <Carousel
+              disableArrowScroll
+              ref={scrollRef}
+              onScroll={onScrollHandler}
+            >
+              {bodyChildrenData}
+            </Carousel>
+          </div>
+        ) : (
+          <div>{bodyChildrenData}</div>
```
davidmenendez, line 133: div should have a class

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenBody.tsx
```diff
@@ -0,0 +1,158 @@
...
+  /**
+   * This is a required callback that has to return the content to render in the body section.
+   * It can be a single child or an array of children depending on your need
+   */
+  contentRenderer: (config: contentRendererArgs) => ReactNode;
+}
+
+const InterstitialScreenBody = ({
+  className,
+  contentRenderer,
```
davidmenendez, line 46: missing `...rest` and also moving forward i think we should go with this signature for components
```
const Component = (props) => {
  const { propA, propB, propC, ...rest } = props;
}
```
this can make debugging easier when you're able to reference the entire `props` variable

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenFooter.tsx
```diff
@@ -0,0 +1,205 @@
...
+  actionButtonRenderer?: (config: actionButtonRendererArgs) => React.ReactNode;
+}
+const InterstitialScreenFooter = ({
+  className,
+  skipButtonLabel = 'Skip',
+  previousButtonLabel = 'Back',
+  nextButtonLabel = 'Next',
+  startButtonLabel = 'Get Started',
+  actionButtonRenderer,
+}: InterstitialScreenFooterProps) => {
```
davidmenendez, line 61: same comment about component signature

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenFooter.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  const handleClickPrev = () => {
+    const targetStep = clamp(progStep - 1, progStepFloor, progStepCeil);
+    handleGotoStep?.(targetStep as number);
+  };
+
+  const handleClickNext = () => {
+    const targetStep = clamp(progStep + 1, progStepFloor, progStepCeil);
+    handleGotoStep?.(targetStep as number);
+  };
```
davidmenendez, line 86: i think we can merge this into a single click handler and pass the `+/-1` as a parameter

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenFooter.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  const getFooterContent = () => (
```
davidmenendez, line 88: a lot of what's going on in here is very redundant. i think it would be cleaner and more efficient to create a function that sets up any common props (like `size`), sets the remaining props conditionally, and returns it so you end up with something that looks more like
```
const btnProps = getBtnProps();
...
<Button {...props} />
```
this helps to eliminate a lot of the clutter with doing the conditionals inline

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenHeader.tsx
```diff
@@ -0,0 +1,149 @@
...
+};
+
+const InterstitialScreenHeader = ({
+  className,
+  headerTitle,
+  headerSubTitle,
+  closeIconDescription,
+  hideProgressIndicator,
+  children,
+}: InterstitialScreenHeaderProps) => {
```
davidmenendez, line 59: same comment about component signature

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenHeader.tsx
```diff
@@ -0,0 +1,149 @@
...
+        {!hideProgressIndicator &&
+          bodyChildrenData &&
+          Array.isArray(bodyChildrenData) && (
+            <div className={`${blockClass}--progress`}>
+              <ProgressIndicator vertical={false} currentIndex={progStep}>
+                {bodyChildrenData?.map((child: React.ReactNode, idx) => {
+                  if (React.isValidElement<EnrichedChildren>(child)) {
+                    return (
+                      <ProgressStep
+                        key={idx}
```
davidmenendez, line 84: it's considered an anti pattern to use an index as a key. can we use an id attribute or something similar?

packages/ibm-products/src/components/InterstitialScreen/_story-assets/InterstitialScreenViewModule/InterstitialScreenViewModule.tsx
```diff
@@ -59,11 +68,28 @@ export let InterstitialScreenViewModule = React.forwardRef<
...
   ) => {
+    useEffect(() => {
+      disableActionButton?.({
+        start: true,
+      });
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
+
+    const handleOnChange = (selected) => {
+      if (selected) {
```
davidmenendez, line 85: you can just shorten this to `disableActionButton?.({ start: !selected })`

davidmenendez (changes requested):
looks like a big improvement! just a few comments

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenFooter.tsx
amal-k-joy: `handleGotoStep`  is kept as reusable that is passed down to `contentRenderer` and `actionRenderer` callbacks and its expects the target step.
To merge to a single handler I may need to define one more method, which is redundant , since current way is  already readable.

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenHeader.tsx
amal-k-joy: we dont have any id attribute, since we are building steps from the body children. How ever will use `stepTitle` which is a mandatory prop to build an id

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenFooter.tsx
amal-k-joy: Yes, start button had some redundant props , so merged to a single component. Lets keep other buttons as it is, since there isn't much redundant attributes.

devadula-nandan (approved):

davidmenendez (approved):

PR #7256, opened 2025-04-08
feat(open in stackblitz): Options Tile, Saving and Full page error

### LINKED ISSUE #7176: Implement Open Stackblitz in Option Tile and Saving
- [ ] Option Tile
- [ ] Saving - create mdx file

sangeethababu9223:
Closes #7176

Implement Open Stackblitz in Option Tile

#### What did you change?

- Refactored `packages/ibm-products/previewer/codePreviewer.tsx` to filter out storybook specific components.
- Filtered out `StoryWrapper` from code in previwer.
- Updated the `Full page Error` story to remove `StoryWrapper` component.
- Added `.mdx` files to `Saving` and `Option Tile` components and added `Open in stackblitz` option in both the stories.

yarn storybook


makafsal (approved):

makafsal (changes requested):
@sangeethababu9223  Can you check the security issues?

sangeethababu9223:
> @sangeethababu9223 Can you check the security issues?

CodeQL Scanning issue is there in Core carbon set up also. This is because we're using regular expressions to sanitize the code. I had talked to Gui and he mentioned they arte dismissing it, so we could do the same.

makafsal (approved):

matthewgallo (approved):
Just one small thing I noticed in the FullPageError generated stackblitz, the props passed to the component aren't used from the generated `args`. I don't see a problem with that, but then there's no need to include the `args` object since it's not used at all.

For example, the `title` prop is passed `"[Error title]"` rather than `args.title`.

Screenshot 2025-04-23 at 12 01 34 PM

sangeethababu9223:
> Just one small thing I noticed in the FullPageError generated stackblitz, the props passed to the component aren't used from the generated `args`. I don't see a problem with that, but then there's no need to include the `args` object since it's not used at all.
> 
> For example, the `title` prop is passed `"[Error title]"` rather than `args.title`.
> 
> Screenshot 2025-04-23 at 12 01 34 PM

Will update this in the next PR for stackblitz

PR #7267, opened 2025-04-09
fix(sidepanel): aria-label for back button

### LINKED ISSUE #7174: [a11y]: SidePanel should provide accessible label to back button
### Package

@carbon/ibm-products

### Browser

Chrome

### Operating System

MacOS

### Package version

v2.63.0-rc.0

### React version

_No response_

IBM Accessibility Checker - Latest Deployment 

### Assistive technology

_No response_

### Description

When navigating to the next step in a `SidePanel` component, the back button that appears in the top-left corner has an accessibility issue:
`Form control element <button> has no associated label`

Image

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-side-panel-sidepanel--panel-with-second-step&globals=viewport:basic

### Steps to reproduce

N/A

### Code of Conduct

anamikaanu96:
Closes #7174

Added aria-label for back button.


wkeese:
I guess this fixes it, but it would be better to use [IconButton](https://react.carbondesignsystem.com/?path=/story/components-iconbutton--default).

The deprecated way to have icon-only buttons is with `<Button hasIconOnly>`, but the SidePanel code is even worse because it doesn't even use that deprecated `hasIconOnly` flag.

davidmenendez (approved):

elycheea (approved):

PR #7284, opened 2025-04-11
fix(card): use appropriate heading level

### LINKED ISSUE #7168: [a11y]: Cards: use proper headings
### Package

@carbon/ibm-products

### Browser

Chrome

### Operating System

MacOS

### Package version

2.62

### React version

18

accessibility-checker

### Assistive technology

_No response_

### Description

As explained in https://www.w3.org/WAI/tutorials/page-structure/headings/, you can't just randomly use an `<h6>` heading in the Card components.   The heading should be one level above the heading of the parent section.

> Skipping heading ranks can be confusing and should be avoided where possible: Make sure that a `<h2>` is not followed directly by an `<h4>`, for example.

Worse yet, [label only cards](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-cards-expressivecard--label-only&globals=viewport:basic) don't have any heading at all.  The "label" is really just a `<div>`.

Tearsheets also has this label/title duality, and they ended up making them H2 and H3 respectively.   So that's one option for Cards, although it would need to be less brittle because you don't know which heading you should start on.   So it would need to be like:

```
<Section>
    <Heading>label</Heading>
    <Section>
          <Heading>{title}</Heading>
           ...
     </Section>
</Section>
```

### WCAG 2.1 Violation

https://www.w3.org/TR/WAI-WEBCONTENT/wai-pageauth.html#tech-logical-headings

### Reproduction/example

https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-components-cards-expressivecard--docs&globals=viewport:basic

### Steps to reproduce

Just use Chrome developer tools to check heading levels.

### Code of Conduct

wkeese:
Closes #7168.

Use appropriate heading level for a card, i.e. one level below the parent heading.

Doesn't address label-only cards though, not sure if it's OK that they don't have any heading.

#### What did you change?

Used `Section` and `Heading` rather than hardcoding to `H6`.

Storybook.


wkeese:
Mend errors are unrelated to this PR, and Percy is a spurious failure.

elycheea (approved):

anamikaanu96 (approved):

PR #7286, opened 2025-04-11
fix(checklist): use appropriate heading levels

### LINKED ISSUE #7168: [a11y]: Cards: use proper headings

wkeese:
Closes #7168.

Use appropriate heading levels for the Checklist component, rather than hardcoding to a particular level.

Doesn't address label-only cards though, not sure if it's OK that they don't have any heading.

For the card content, I followed the example at
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/section, having a separate `<section>` for each heading,
rather than one `<section>` wrapping all the headings.

Note: I'm not sure why the outer node of Checklist is an `<aside>` rather than `<section>` etc., but I didn't change it in this PR.

#### What did you change?

Use `<Section>` and `<Heading>` rather than hardcoded headings.

Storybook and Jest tests.

### DIFF
--- a/packages/ibm-products/src/components/Checklist/Checklist.tsx
+++ b/packages/ibm-products/src/components/Checklist/Checklist.tsx
@@ -13,7 +13,7 @@
  */
 
 // Carbon and package components we use.
-import { Button, IconButton } from '@carbon/react';
+import { Button, Heading, IconButton, Section } from '@carbon/react';
 import { Kind, Kinds, Theme, Themes } from './Checklist.types';
 // Import portions of React that are needed.
 import React, {
@@ -205,12 +205,13 @@ export let Checklist = React.forwardRef(
     }, [open]);
 
     return (
-      <aside
+      <Section
         {
           // Pass through any other property values as HTML attributes.
           ...rest
         }
         aria-label={checklistAriaLabel}
+        as="aside"
         className={cx(
           blockClass, // Apply the block class to the main HTML element
           className, // Apply any supplied class names to the main HTML element.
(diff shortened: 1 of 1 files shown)


wkeese:
Mend error is unrelated to this PR.

davidmenendez (approved):

matthewgallo:
The Mend issue isn't related to @wkeese's changes here, so I'm going to put this in the merge queue. Thanks Bill!

PR #7315, opened 2025-04-16
chore: update to use commit hash for node in vrt runner

amal-k-joy:
Closes #7218

update to use commit hash for node in vrt runner to avoid node set up failure


sangeethababu9223 (approved):

elycheea:
Haha, I made the same fix in #7239 yesterday. I think this should already be resolved after pulling from main. 🙏

davidmenendez (approved):

PR #7316, opened 2025-04-16
feat(conditionBuilder): add pre add callback before adding

### LINKED ISSUE #7130: Condition builder: Introduce pre-add callback before adding condition, group, or sub-group
### The problem

If there is requirement to put a limit on the number of conditions, groups or subgroups, then there has to be a way to disable those add buttons, by passing some props.

### The solution

Image

### Design link

_No response_

### Links to other materials

_No response_

### Owner/main maintainer(s)

NA

_No response_

### Product/offering

_No response_

### Business priority

Medium Priority = upcoming release but is not pressing

### Code of Conduct

### Acceptance criteria
- [x] Add a new callback prop `onAddItem` which will be triggered before adding a condition or sub-group or group. User will have the option to stop adding by returning `{preventAdd: true}` . User will receive addItemType , rootState, currentGroup as arguments that helps them to perform any validations

amal-k-joy:
Closes #7130 

This add support for a new callback prop 'onAddItem' that will be called before adding any condition , subgroup or group. 
User can prevent add if they return `{preventAdd : true}`

#### What did you change?
Added a new callback prop
```
onAddItem?: (
    addItemType: 'add-condition' | 'add-subgroup' | 'add-group',
    rootState: ConditionBuilderState,
    group?: ConditionGroup
  ) => { preventAdd: boolean };
```

### DIFF
--- a/packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.types.ts
+++ b/packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.types.ts
@@ -158,7 +158,12 @@ export type statementConfig = Item & {
   connector: 'and' | 'or';
   secondaryLabel?: string;
 };
-type AddItemType = 'add-condition' | 'add-subgroup' | 'add-group';
+type AddItemType = 'condition' | 'subgroup' | 'group';
+type AddItemConfig = {
+  type: AddItemType;
+  state: ConditionBuilderState;
+  group?: ConditionGroup;
+};
 
 export type ConditionBuilderProps = {
   inputConfig: inputConfig;
@@ -176,11 +181,7 @@ export type ConditionBuilderProps = {
   variant?: 'Non-Hierarchical' | 'Hierarchical';
   translateWithId: (id: string) => string;
   statementConfigCustom: statementConfig[];
-  onAddItem?: (
-    addItemType: AddItemType,
-    rootState: ConditionBuilderState,
-    group?: ConditionGroup
-  ) => { preventAdd: boolean };
+  onAddItem?: (config: AddItemConfig) => { preventAdd: boolean };
 };
 
 export type InitialState = {
@@ -208,9 +209,5 @@ export type ConditionBuilderContextProps = {
   setRootState?: Dispatch<SetStateAction<ConditionBuilderState>>;
   actionState?: Action[];
   setActionState?: Dispatch<SetStateAction<Action[]>>;
-  onAddItem?: (
-    addItemType: AddItemType,
-    rootState: ConditionBuilderState,
-    group?: ConditionGroup
-  ) => { preventAdd: boolean };
+  onAddItem?: (config: AddItemConfig) => { preventAdd: boolean };
 } & ConditionBuilderContextInputProps;
--- a/packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.types.ts
+++ b/packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.types.ts
@@ -158,6 +158,7 @@ export type statementConfig = Item & {
   connector: 'and' | 'or';
   secondaryLabel?: string;
 };
+type AddItemType = 'add-condition' | 'add-subgroup' | 'add-group';
 
 export type ConditionBuilderProps = {
   inputConfig: inputConfig;
@@ -175,6 +176,11 @@ export type ConditionBuilderProps = {
   variant?: 'Non-Hierarchical' | 'Hierarchical';
   translateWithId: (id: string) => string;
   statementConfigCustom: statementConfig[];
+  onAddItem?: (
+    addItemType: AddItemType,
+    rootState: ConditionBuilderState,
+    group?: ConditionGroup
+  ) => { preventAdd: boolean };
 };
 
 export type InitialState = {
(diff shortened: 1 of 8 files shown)


packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.types.ts
```diff
@@ -158,6 +158,7 @@ export type statementConfig = Item & {
   connector: 'and' | 'or';
   secondaryLabel?: string;
 };
+type AddItemType = 'add-condition' | 'add-subgroup' | 'add-group';
```
davidmenendez, line 161: minor, but i think we can remove the `add-` and just keep `group`, etc

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.types.ts
```diff
@@ -175,6 +176,11 @@ export type ConditionBuilderProps = {
   variant?: 'Non-Hierarchical' | 'Hierarchical';
   translateWithId: (id: string) => string;
   statementConfigCustom: statementConfig[];
+  onAddItem?: (
```
davidmenendez, line 179: this is something i've been thinking about recently and i think that we should try sticking to a convention for functions where if there's more than one parameter we use an object instead of multiple individual arguments. this can help ensure simple mistakes aren't made due to argument position. so `({ type, state, group  }) =>` instead of `(type, state, group)`

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.types.ts
```diff
@@ -202,4 +208,9 @@ export type ConditionBuilderContextProps = {
   setRootState?: Dispatch<SetStateAction<ConditionBuilderState>>;
   actionState?: Action[];
   setActionState?: Dispatch<SetStateAction<Action[]>>;
+  onAddItem?: (
```
davidmenendez, line 211: same as above

davidmenendez (changes requested):
looks good! a few minor comments and could you please include testing to validate this new functionality? thanks!

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderAdd/ConditionBuilderAdd.tsx
```diff
@@ -65,6 +79,15 @@ const ConditionBuilderAdd = ({
     onBlur: hideConditionSubGroupPreviewHandler,
   });
 
+  const handleAddSubGroup = () => {
+    const { preventAdd } =
+      onAddItem?.('add-subgroup', rootState as ConditionBuilderState, group) ??
+      {};
```
jlongshore, line 85: To me, the nullish coalescing here to an empty object seems weird, I assume it's to pacify typescript because the onAddItem callback is optional... I wonder if it would be clearer if we checked for onAddItem first:

```
if (onAddItem) {
  const { preventAdd } = onAddItem('add-condition', rootState as ConditionBuilderState, group);
  if (!preventAdd) {
    hideConditionPreviewHandler?.();
    onClick();
  }
}
```
Totally your call - I believe preventAdd would just be undefined if it were trying to destructure out of an empty object and it wouldn't error (I think at least) - but it may be clearer to another dev down the road :D

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderContent/ConditionBuilderContent.tsx
```diff
@@ -153,27 +153,31 @@ const ConditionBuilderContent = ({
...
-    };
-    setRootState?.({
-      ...rootState,
-      groups:
-        rootState && rootState.groups
-          ? [...rootState.groups, newGroup]
-          : [newGroup],
-    });
+    const { preventAdd } =
+      onAddItem?.('add-group', rootState as ConditionBuilderState) ?? {};
```
jlongshore, line 157: Same as above, your call - its mainly a preference thing...

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.types.ts
jlongshore: ^ I agree with this too...

jlongshore (changes requested):
Just a few things... overall its good work!

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.types.ts
amal-k-joy: Yes , thats sounds right to me as well.

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.types.ts
```diff
@@ -202,4 +208,9 @@ export type ConditionBuilderContextProps = {
...
+    addItemType: AddItemType,
+    rootState: ConditionBuilderState,
+    group?: ConditionGroup
+  ) => { preventAdd: boolean };
```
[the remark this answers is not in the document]
amal-k-joy: I also thought of returning a boolean initially. But this approach has more clarity on intent and scalable as well. Also I feel returning boolean can be preferred when the function  name itself conveys its intent, incases with method names like `checkIsValid` , `checkIsDisabled`, `canProceed` , `preventAdd`

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderAdd/ConditionBuilderAdd.tsx
amal-k-joy: Here  I need to call hideConditionPreviewHandler?.() and onClick(),  even if onAddItem itself is undefined or onAddItem return undefined.

davidmenendez (approved):
awesome!

makafsal (approved):

jlongshore (approved):
LGTM, nice work!

PR #6686, opened 2025-01-13
chore(tearsheet): added definition tooltip if subtitle has ellipsis

devadula-nandan:
Closes #4230 

implements definition tooltip on the subtitle if it has ellipsis.
took reference from https://github.com/carbon-design-system/ibm-products/pull/6551, also feel like string formatter needs to be updated, to be used in these common use cases

#### What did you change?
packages/ibm-products-styles/src/components/Tearsheet/_tearsheet.scss
packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx

#### How did you test and verify your work?
yarn test, yarn avt, storybook

PR #7345, opened 2025-04-22
feat: string formatter height overflow, re-use in tearsheet

### LINKED ISSUE #6939: Tearsheet: redesign when it checks for definition overflow
### Package

Carbon for IBM Products

### Description

@devadula-nandan added code in #6686 (for #4230) to display a tooltip when the tearsheet description overflows.  I'm seeing some issues:

1. I've seen TearsheetShell re-rendering in an infinite loop.  Not sure why but perhaps this is related.
2. It measures overflow every time TearsheetShell() is called, even when nothing relevant has changed.  This happens even (a) when the tearsheet is closed and/or (b) before the description is even rendered.
3. It needs to run when the browser width changes... and it does... but just out of luck from some unrelated code.

What I would suggest is to replace the current code of:

```js
const isOverflowing = checkHeightOverflow(descriptionRef.current);
```

with:

```js
const [isOverflowing, setIsOverflowing] = useState(false);
useEffect(() => {
  setIsOverflowing(checkHeightOverflow(descriptionRef.current));
}, [description, width]);
```

Note that normally useState() and useEffect() can be replaced by a single useMemo(), but in this case I want to wait for the TearsheetShell to (re)render before measuring.

### Component(s) impacted

Tearsheet, maybe TearsheetNarrow too.

### Browser

Chrome

2.59.0

### Suggested Severity

### Product/offering

IKC

https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-tearsheet--tearsheet&globals=viewport:basic

_No response_

_No response_

### Code of Conduct

devadula-nandan:
Closes #6939 

this pr
1) reverts https://github.com/carbon-design-system/ibm-products/pull/6686
2) adds feature to stringFormatter to calculate height overflow based on lines prop, and show definition tooltip only if it is truely overflowing
3) re-uses the string formatter in tearsheet (string formatter can be used in description prop of tearsheet)
4) gives the adopter more composability as they have complete control over the stringFormatter re-usability and all its api. which can be passed through description prop
5) improves maintainability of tearsheet by removing the clutter

#### What did you change?
string formatter
tearsheet (clean up and minor refactor)

manually in storybook, mozilla, chrome.
failed tests fixed


davidmenendez (dismissed):

davidmenendez:
looks good!

devadula-nandan:
@davidmenendez Thanks for the review. can i get a re-review?
i just pushed some changes on storybook, to be more clear of this usage, also added a control to play with it in the storybook.

an interesting observation. if the text truncated in avt. it creates an additional tab stop, failing the avt. so i changed the default description for narrow tearsheet to use plain string. however these are unrelated to the component, and the user could try every option in the storybook

packages/ibm-products/src/components/StringFormatter/StringFormatter.js
```diff
@@ -49,16 +49,48 @@ export let StringFormatter = React.forwardRef(
...
+        ref.current = node;
+      }
+    };
 
     useIsomorphicEffect(() => {
-      contentRef.current.style.maxWidth = width;
-      contentRef.current.style.WebkitLineClamp = lines;
-    }, [lines, width]);
+      const checkTruncation = () => {
+        const element = outerRef.current?.querySelector(
```
davidmenendez, line 66: on second look i'm not a fan of this approach. i dont think it's a good idea to ever hard code selectors like this because it's not flexible. can we take a different approach by adding a parameter the consumer can send instead?

davidmenendez (changes requested):
one small comment that i noticed on second look. let me know what you think

packages/ibm-products/src/components/StringFormatter/StringFormatter.js
devadula-nandan: thanks for the review. i have changed to using refs, and we are already having an outer ref forwarded, so we don't need to expose any internal refs, as any internal element can be accessed through outer ref

davidmenendez (approved):

davidmenendez (approved):

makafsal (approved):

Issue #7162, opened 2025-03-25
[a11y]: OptionsTile heading level


wkeese:
### Package

@carbon/ibm-products

### Browser

Chrome

### Operating System

MacOS

### Package version

2.62.0

### React version

18

### Automated testing tool and ruleset

none

### Description

The heading rank in `OptionsTile` must match with the rest of the document, so it shouldn't be hardcoded to `<h6>`.   

As written in https://www.w3.org/WAI/tutorials/page-structure/headings/:

> Skipping heading ranks can be confusing and should be avoided where possible: Make sure that a `<h2>` is **not** followed directly by an `<h4>`, for example.

### WCAG 2.1 Violation

https://www.w3.org/TR/WAI-WEBCONTENT/wai-pageauth.html#tech-logical-headings

### Reproduction/example

https://carbon-for-ibm-products.netlify.app/?path=/docs/ibm-products-components-options-tile-optionstile--docs

### Steps to reproduce

Just look at the output in Chrome developer tools.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate problems

PR #7368, opened 2025-04-23
fix(optionstile): section needs to wrap around children

wkeese:
Refs #7162

It's important for the `OptionsTile` children to be inside the` <Section>`, in case they have their own subheadings.

#### What did you change?

Moved the Section to be wrapped around everything instead of just wrapped around the title.

Storybook


sangeethababu9223 (approved):

davidmenendez (approved):

wkeese:
As usual, I don't hae permission to look at the visual changes diff, but I assume it's a spurious error.  Please check and let me know.

PR #7380, opened 2025-04-24
feat(options-tile): allow props.tile to be a react node if it doesn't contain interactive children

### LINKED ISSUE #7282: Resolve warnings by allowing fragments in OptionsTile to be passed in string props
### Package

Carbon for IBM Products

### Description

This is a reopen of this issue: https://github.com/carbon-design-system/ibm-products/issues/2719

It looks like the previous issue had a fix that solved the Productive Card issues, but did not fix the issues with OptionsTile.

The title and summary props in OptionsTile only accept string props. I am trying to pass in an HTML fragment with a DismissibleTag.

### Component(s) impacted

OptionsTile

Image

### Browser

_No response_

2.63.0

### Suggested Severity

None

### Product/offering

IBM Verify

https://stackblitz.com/edit/github-b3eosmhj?file=src%2FExample%2FExample.jsx

_No response_

_No response_

### Code of Conduct

janhassel:
Closes #7282

Widens the allowed prop type for `props.title` in `OptionsTile` to also include `React.ReactNode`. Since the title is rendered within an interactive element though (`<summary>`), it is validated that there are no interactive children present.

#### What did you change?

- Updated OptionsTile prop type for `props.title`
- Added `@carbon/utilities-react` dependency for their `useNoInteractiveChildren` hook
- Add test to verify interactive children will cause the component to throw (in dev environment)

- Manual test with storybook
  - Use `<button>Test</button>` as `props.title` in OptionsTile story
- Unit test


packages/ibm-products/src/components/OptionsTile/OptionsTile.tsx
```diff
@@ -94,9 +95,9 @@ export interface OptionsTileProps {
   summary?: string;
 
   /**
-   * Provide the title for this OptionsTile.
+   * Provide the title for this OptionsTile. Must not contain any interactive elements.
    */
-  title: string;
+  title: string | React.ReactNode;
```
makafsal, line 100: We can remove `string`, since the type `string` is valid for type `ReactNode`.

```
type ReactNode =
        | ReactElement
        | string
        | number
        | bigint
        | Iterable<ReactNode>
        | ReactPortal
        | boolean
        | null
        | undefined
        | DO_NOT_USE_OR_YOU_WILL_BE_FIRED_EXPERIMENTAL_REACT_NODES[
            keyof DO_NOT_USE_OR_YOU_WILL_BE_FIRED_EXPERIMENTAL_REACT_NODES
        ]
        | Promise<AwaitedReactNode>;
        ```

**Review by makafsal — CHANGES_REQUESTED** [review]

File: packages/ibm-products/src/components/OptionsTile/OptionsTile.tsx (hunk 1 above)
Comment (janhassel) [reply packages/ibm-products/src/components/OptionsTile/OptionsTile.tsx]: Good point!

File: packages/core/package.json [hunk 2]
```diff
@@ -35,6 +35,7 @@
     "@carbon/react": "^1.81.0",
     "@carbon/themes": "^11.50.0",
     "@carbon/type": "^11.38.0",
+    "@carbon/utilities-react": "0.7.0",
```
matthewgallo, line 38: We've refactored our storybook to now be located within `packages/ibm-products`, so you can remove this change.

packages/ibm-products/package.json
```diff
@@ -114,6 +114,7 @@
     "@carbon/react": "^1.81.0",
     "@carbon/themes": "^11.50.0",
     "@carbon/type": "^11.38.0",
+    "@carbon/utilities-react": "0.7.0",
```
matthewgallo, line 117: Similar to `@carbon/utilities` I think this should be moved to `dependencies`.

matthewgallo (changes requested):
Thanks for these enhancements @janhassel! Just had a few comments around where to include the dependency. Otherwise, this looks great and I really like the inclusion of `useNoInteractiveChildren`.

We likely just missed this since there isn't an issue linked that is included in our current sprint, can you add one? If you don't have one, just link to the issue that was included in @makafsal's PR ([here](https://github.com/carbon-design-system/ibm-products/issues/7282)).

janhassel:
@matthewgallo I updated the dependencies as you suggested. fyi, this PR is in response to the following report on Slack: https://ibm-analytics.slack.com/archives/C013ZTX0N6B/p1677856103477479 (yes, it's a bit older; it took some time to coordinate and scaffold the `@carbon/utilities` packages in order to properly re-use the `useNoInteractiveChildren` hook as a sensible compromise of warning the user about the accessibility implications 😄)

PR #6073, opened 2024-09-18
fix(dataGrid): focus loss on radio select

devadula-nandan:
Closes #5683 

returns the radio select focus, if it's lost.

#### What did you change?
packages/ibm-products/src/components/Datagrid/useSelectRows.tsx

#### How did you test and verify your work?
storybook, screen reader for mac, jaws for win 11

Issue #5683, opened 2024-07-17
[a11y]: Datagrid, radio select: selection of radio buttons causes focus jump


elycheea:
### Package

@carbon/ibm-products

### Browser

Firefox

### Package version

v2.44.0-rc.1

### Automated testing tool and ruleset

manual

### Assistive technology

JAWS

### Description

Observation on row selection radio buttons selection:
- Selecting a radio button within the table row jumps focus to the top and announces:
"storybook-preview-frame frame"
- Screen reader doesn’t announce the radio button selection status immediately 

**Note**
Screen reader users experience difficulty navigating the table and confirming their selections.

**Recommended fix**
Ensure the selection status of the radio button elements is conveyed instantly, also avoid the focus jump caused by selection.

**User impact**
When the state of an element is not conveyed appropriately, it prevents screen reader users to interact smoothly with the page element.

### WCAG 2.1 Violation

4.1.2. Name Role Value

### Reproduction/example

https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-datagrid--radio-select&globals=viewport:basic

### Steps to reproduce

To reproduce this defect:

1. Use a screen reader on the row selection story.
2. Try navigating through radio buttons.
3. Try selecting a radio button.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate problems

PR #7385, opened 2025-04-24
fix(datagrid): clicking select all checkbox focus loss

### LINKED ISSUE #7348: [a11y]: Keyboard focus is lost after checking "Select all rows in the table" checkbox of Datagrid
### Package

@carbon/ibm-products

### Browser

Chrome

### Operating System

MacOS

### Package version

v2.65.0-rc.0

### React version

_No response_

manual

### Assistive technology

MacOS Voiceover

### Description

When checking the "Select all rows in the table" checkbox on toolbar of datagrid, keyboard focus is lost.
Losing focus can disrupt screen reader behavior.

Image

Image

**Expected behavior**
keyboard focus should maintain after checking the checkbox

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://ibm-products.carbondesignsystem.com/iframe.html?args=&globals=viewport%3Abasic&id=deprecated-datagrid-datagrid--selectable-row&viewMode=story

### Steps to reproduce

1. Navigate to the "Select all rows in the table" checkbox on the toolbar with keyboard
2. Check the checkbox

### Code of Conduct

wkeese:
Closes #7348, and refs #5638.

Clicking the select-all checkbox triggered React to remove the parent `TH` and reinsert it into the document... causing the focus to be lost.

Avoided that by fixing spurious recreation of functions.

#### What did you change?

The problem was that React's VDOM diffing wasn't working well.  But to solve it, it wasn't just about adding `useCallback()` for the event handler callbacks.  I also needed to prevent spurious recreation of the `Heading` and `Cell` functions.

##### DatagridSelectAll.tsx

* Add `useCallback()` for `handleOnPageSelectAllChange()` and `handleSelectAllChange()`.

* Subsequently, move around variable declarations to obey rules of hooks, specifically that all the `useCallback()` need to happen before any `return` statement.

##### useSelectRows.tsx

* Create static `renderSelectAll()` and `renderSelectCell()` functions so that `Header` and `Cell` functions aren't recreated every render. I think this was the root issue.  See https://stackoverflow.com/questions/58507656/why-does-react-discard-the-entire-dom-subtree-and-recreate-it-from-scratch and https://legacy.reactjs.org/docs/reconciliation.html#elements-of-different-types.  But it's hard to tell because Datagrid is confused about the difference between components (that should start with a capital letter) and mere callbacks (that should start with a lowercase letter).

* Add `useCallback()` for `onSelectHandler()`.

* Move `onSelectHandler()` so that `onChange()` is not referenced before defined.

* Revert workaround from #6073 to refocus radio buttons after click, because this fixes the root cause of #5683.

* Use `aria-label` rather than deprecated `ariaLabel`.

Storybook


wkeese:
CC @devadula-nandan.

devadula-nandan (dismissed):
@wkeese looks great

davidmenendez (changes requested):
thanks for submitting these changes. can you include or update the tests to account for this? seeing a few drops in code coverage. thanks!

wkeese:
> thanks for submitting these changes. can you include or update the tests to account for this? seeing a few drops in code coverage. thanks!

@davidmenendez - I'm pretty sure that's not true, and that it's just pointing out code paths that were _never_ covered.  It's just confused because I moved around the code.  Nonetheless, in the interest of merging this, I added more tests.

However, I'm not sure what's happening with [the codecov report](https://app.codecov.io/gh/carbon-design-system/ibm-products/pull/7385?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=carbon-design-system):

1. It says that in useSelectRows.ts, onRadioSelect() isn't being called but I tested in the debugger (running the 'Radio Select' test) and it is called.
2. It shows that I greatly improved testing in DatagridSelectAll.tsx.  The lines marked with ⚠ are all from the old code.
3. So why does this page claim that coverage went down?

anamikaanu96 (approved):
LGTM

PR #7397, opened 2025-04-28
chore: string formatter initial commit

devadula-nandan:
Closes #

{{short description}}

### DIFF
--- a/packages/ibm-products-web-components/src/components/string-formatter/string-formatter.ts
+++ b/packages/ibm-products-web-components/src/components/string-formatter/string-formatter.ts
@@ -1,0 +1,103 @@
+/**
+ * @license
+ *
+ * Copyright IBM Corp. 2025, 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import { html, LitElement } from 'lit';
+import { property, query } from 'lit/decorators.js';
+import { carbonElement as customElement } from '@carbon/web-components/es/globals/decorators/carbon-element.js';
+import { prefix } from '../../globals/settings';
+// todo: replace with definition tooltip to get tab stop
+import '@carbon/web-components/es/components/tooltip/index.js';
+
+import styles from './string-formatter.scss?lit';
+
+const componentName = 'string-formatter';
+export const blockClass = `${prefix}--${componentName}`;
+const elementName = `${prefix}-${componentName}`; // c4p-string-formatter
+
+/**
+ * StringFormatter.
+ *
+ * @element c4p-string-formatter
+ */
+@customElement(elementName)
+export class CDSStringFormatter extends LitElement {
+  @property({ type: Number }) lines = 0;
+  @property({ type: String }) value = '';
+
+  @query(`.${blockClass}_content`)
+  private textElement!: HTMLElement;
+
+  private isOverflowing = false;
+  private resizeObserver?: ResizeObserver;
+
+  static styles = styles;
+
+  protected firstUpdated() {
+    this.updateOverflowStatus();
+    this.setupResizeObserver();
+  }
+
+  protected updated(changedProperties: Map<string, unknown>) {
+    super.updated(changedProperties);
+    if (changedProperties.has('lines') || changedProperties.has('value')) {
+      this.updateOverflowStatus();
+    }
+  }
+
+  private setupResizeObserver() {
+    if (this.textElement) {
+      this.resizeObserver = new ResizeObserver(() => {
+        this.updateOverflowStatus();
+      });
+      this.resizeObserver.observe(this);
+    }
+  }
+
+  private updateOverflowStatus() {
+    if (this.textElement) {
+      this.textElement.style[`--line-clamp`] = this.lines;
+      const { scrollHeight, clientHeight } = this.textElement;
+      const buffer = clientHeight / (2 * this.lines);
+      // add a buffer of at least half of line height/clientHeight. to get a stable outcome.
+      this.isOverflowing = scrollHeight > clientHeight + buffer;
+      this.requestUpdate();
+    }
+  }
+
+  disconnectedCallback(): void {
+    super.disconnectedCallback();
+    this.resizeObserver?.disconnect();
+  }
+
+  render() {
+    const content = html`
+      <div class="${blockClass}_content" style="--line-clamp: ${this.lines};">
+        ${this.value}
+      </div>
+    `;
+
+    return this.isOverflowing
+      ? html`
+          <cds-tooltip align="bottom" enter-delay-ms="0" leave-delay-ms="0">
+            ${content}
+            <cds-tooltip-content id="content"
+              >${this.value}</cds-tooltip-content
+            >
+          </cds-tooltip>
+        `
+      : content;
+  }
+}
+
+declare global {
+  interface HTMLElementTagNameMap {
+    [elementName]: CDSStringFormatter;
+  }
+}
+
+export default CDSStringFormatter;
--- a/packages/ibm-products-web-components/src/components/string-formatter/index.ts
+++ b/packages/ibm-products-web-components/src/components/string-formatter/index.ts
@@ -1,0 +1,10 @@
+/**
+ * @license
+ *
+ * Copyright IBM Corp. 2025, 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import './string-formatter';
--- a/packages/ibm-products-web-components/src/components/string-formatter/story-styles.scss
+++ b/packages/ibm-products-web-components/src/components/string-formatter/story-styles.scss
@@ -1,0 +1,10 @@
+/*
+* Copyright IBM Corp. 2025, 2025
+*
+* This source code is licensed under the Apache-2.0 license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+@use '@carbon/styles/scss/theme' as *;
+
+$story-prefix: 'string-formatter-stories__';
--- a/packages/ibm-products-web-components/src/components/string-formatter/string-formatter.mdx
+++ b/packages/ibm-products-web-components/src/components/string-formatter/string-formatter.mdx
@@ -1,0 +1,47 @@
+import { ArgTypes, Markdown, Meta } from '@storybook/blocks';
+import { cdnJs, cdnCss } from '../../globals/internal/storybook-cdn';
+import * as StringFormatterStories from './string-formatter.stories';
+
+<Meta of={StringFormatterStories} />
+
+# FullPageError
+
+Display a full-page error when the requested page is unavailable to the user.
+This is typically caused by issues with the requested URL or access permissions.
+Errors caused by server connectivity issues are not covered in this guideline.
+
+> 💡 Check our
+> [Stackblitz](https://stackblitz.com/github/carbon-design-system/ibm-products/tree/main/packages/ibm-products-web-components/examples/string-formatter)
+> example implementation.
+> [![Edit carbon-web-components](https://developer.stackblitz.com/img/open_in_stackblitz.svg)](https://stackblitz.com/github/carbon-design-system/ibm-products/tree/main/packages/ibm-products-web-components/examples/string-formatter)
+
+## Getting started
+
+Here's a quick example to get you started.
+
+### JS (via import)
+
+```javascript
+import '@carbon/ibm-products-web-components/es/components/string-formatter/index.js';
+```
+
+### HTML
+
+```html
+<c4p-string-formatter
+  label="Error 404"
+  class="my-custom-class"
+  title="Page not found"
+  description="The page you requested has moved or is unavailable, or the specified URL is not valid. Please check the URL or search the site for the requested content."
+  kind="404"
+>
+  <!-- Pass any content you want to display to the user here -->
+  <a class="cds--link" href="#">– Forwarding link 1</a>
+  <br />
+  <a class="cds--link" href="#">– Forwarding link 2</a>
+</c4p-string-formatter>
+```
+
+## `<c4p-string-formatter>` attributes, properties and events
+
+<ArgTypes of="c4p-string-formatter" />
--- a/packages/ibm-products-web-components/src/components/string-formatter/string-formatter.scss
+++ b/packages/ibm-products-web-components/src/components/string-formatter/string-formatter.scss
@@ -1,0 +1,34 @@
+/*
+* Copyright IBM Corp. 2025, 2025
+*
+* This source code is licensed under the Apache-2.0 license found in the
+* LICENSE file in the root directory of this source tree.
+*/
+
+$css--plex: true !default;
+@use 'sass:map';
+
+@use '@carbon/styles/scss/reset';
+@use '@carbon/themes' as *;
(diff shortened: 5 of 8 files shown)


packages/ibm-products-web-components/src/components/string-formatter/string-formatter.ts
```diff
@@ -0,0 +1,103 @@
...
+ * Copyright IBM Corp. 2025, 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import { html, LitElement } from 'lit';
+import { property, query } from 'lit/decorators.js';
+import { carbonElement as customElement } from '@carbon/web-components/es/globals/decorators/carbon-element.js';
+import { prefix } from '../../globals/settings';
+// todo: replace with definition tooltip to get tab stop
```
devadula-nandan, line 13: definition tooltip is not ready in carbon/web-components yet.

we could leave this as it is as well.
IMO we don't need a tab stop on string formatter, as it could potentially cause inconsistencies in tab order in some tests. however we could still access it with mouse or touch

elycheea (commented):
Not a full review, but I know @davidmenendez was considering something like `useOverflowString`, but maybe that’s more for utility format. 

I do think this should be classified as a utility component. But I’d love for us to consider the naming if there are better ways to describe its purpose.

(Maybe @RichKummer has thoughts?)

Also we should label this a `feat` not a `chore` :D

devadula-nandan:
> Not a full review, but I know @davidmenendez was considering something like `useOverflowString`, but maybe that’s more for utility format.
> 
> I do think this should be classified as a utility component. But I’d love for us to consider the naming if there are better ways to describe its purpose.
> 
> (Maybe @RichKummer has thoughts?)
> 
> Also we should label this a `feat` not a `chore` :D

ah, the framework agonistic utility, let me try that parallel. sounds like a good idea to me.

RichKummer:
`StringFormatter` seems a little generic to me, and could mean anything related to a string. I'm wondering if the name could incorporate `truncate/truncation` and `tooltip` to help convey the purpose? Those are the two key aspects to the utility

`TruncateWithTooltip`? There's also `TruncatedList` so maybe something more in line with that?

devadula-nandan:
i have updated this pr, and the changes include.

name change to `truncated string` and the tag name being `cds-truncated-string`.
introduced a new prop `with` which gives two options to present the full string

RichKummer:
Hey @devadula-nandan ! 

**Tooltip**
- For this use case, I think `Tooltip` is more appropriate. The `DefinitionToolip` is built "to provide additional information about a particular term or phrase in text content" (from the [Storybook](https://react.carbondesignsystem.com/?path=/docs/components-definitiontooltip--overview)). There's also an impact on the actual tooltip styling (the DefinitionTooltip has a slimmer container.
- That said, the description still be focusable if it's truncated with the tooltip. Does explicitly adding the tabindex have a negative effect on adopting this?

**Expand**
This is looking great, and mirrors the Tearsheet specs. Two things that I'm looking at:
- I can’t tell from the build, but I’m wondering if we could apply the [[Carbon motion package](https://carbondesignsystem.com/elements/motion/code/)](https://carbondesignsystem.com/elements/motion/code/) to provide some motion out of the box. Using the [[Motion generator](https://ibm.github.io/motion/)](https://ibm.github.io/motion/), it looks like Productive motion (UI) using `Easing: cubic-bezier(0.2, 0.2, 0.38, 0.9)` might work for the use case. This could also become a follow-on enhancement.
- For the "read more" indication, it looks like the actual clickable area extends to the left of the "read more" label, including an area with a gradient fade-out. I don't think we necessarily need the fade-out or the expanded clickable area (shown below in the pink border)
Screenshot 2025-05-15 at 11 45 52 AM

devadula-nandan:
> * That said, the description still be focusable if it's truncated with the tooltip. Does explicitly adding the tabindex have a negative effect on adopting this?

it could be an accessibility choice, and we can pass this to accessibility team to get the expected behavior, ill make it tabbable in this pr for now.

> * I can’t tell from the build, but I’m wondering if we could apply the [[Carbon motion package](https://carbondesignsystem.com/elements/motion/code/)](https://carbondesignsystem.com/elements/motion/code/) to provide some motion out of the box. Using the [[Motion generator](https://ibm.github.io/motion/)](https://ibm.github.io/motion/), it looks like Productive motion (UI) using Easing: cubic-bezier(0.2, 0.2, 0.38, 0.9) might work for the use case. This could also become a follow-on enhancement.

i couldn't find the corresponding tokens in the library for the exact vales mentioned, so i have used the nearest available match.
```
// transition: max-height 107ms cubic-bezier(0.2, 0.2, 0.38, 0.9); // recommended
transition: max-height $duration-fast-02 motion(standard, productive); // nearest match for the token
```

> * For the "read more" indication, it looks like the actual clickable area extends to the left of the "read more" label, including an area with a gradient fade-out. I don't think we necessarily need the fade-out or the expanded clickable area (shown below in the pink border)

removed the gradient as per discussion

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.ts
```diff
@@ -0,0 +1,214 @@
...
+          ? '0'
+          : undefined}
+      >
+        ${this.value} ${this._renderToggleButton()}
+      </div>
+    `;
+
+    return this.with === 'tooltip' && this._isOverflowing
+      ? html`
+          <cds-tooltip align="bottom" enter-delay-ms="0" leave-delay-ms="0">
```
elycheea, line 197: I feel like we should support both `autoAlign` and a manual `tooltipAlign` when using the tooltip.

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.ts
```diff
@@ -0,0 +1,214 @@
...
+   * The label on expand button.
+   */
+  @property({ attribute: 'expand-label', type: String, reflect: true })
+  expandLabel = '...more';
+
+  /**
+   * The label on the collapse button.
+   */
+  @property({ attribute: 'collapse-label', type: String, reflect: true })
+  collapseLabel = '...less';
```
elycheea, line 51: I recommend we stick with “View” or “Read” as our defaults?

```suggestion
  @property({ attribute: 'expand-label', type: String, reflect: true })
  expandLabel = 'View more';

  /**
   * The label on the collapse button.
   */
  @property({ attribute: 'collapse-label', type: String, reflect: true })
  collapseLabel = 'View less';
```

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.scss
```diff
@@ -0,0 +1,114 @@
...
+.#{$block-class}_button-expand {
+  position: absolute;
+  background-color: $background;
+  inset-block-end: 0;
+  inset-inline-end: 0;
+
+  &.#{$block-class}_button-layered {
+    background-color: $layer;
+  }
+
```
elycheea, line 92: We can try using generated content `::before/::after` to include the ellipses before the expand?

image

image

It’s still included in the click but might be a better in-between since I do think we’d expect to see an ellipsis for truncated text.

elycheea: Found a similar example from Dell Design System although their implementation is a bit more manual — a little akin to the non-linear reading component actually? 🤔 

https://vanilla.delldesignsystem.com/2.46.2/index.html?path=/docs/components-view-more-less--inline

elycheea (changes requested):

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.mdx
```diff
@@ -0,0 +1,56 @@
+import { ArgTypes, Markdown, Meta } from '@storybook/blocks';
+import { Canvas } from '@storybook/addon-docs';
+import { cdnJs, cdnCss } from '../../globals/internal/storybook-cdn';
+import * as TruncatedText from './truncated-text.stories';
+
+<Meta of={TruncatedText} />
+
+# TruncatedText
+
+The `TruncatedText` component truncates text based on a specified number of
```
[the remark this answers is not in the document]
devadula-nandan: agree with changing the wording to utility.

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  /**
+   * The method to display the full text when truncated. Options are "tooltip" or "expand". if not passed, the text would just be truncated with ellipsis.
+   */
+  @property({ type: String, reflect: true }) with: 'tooltip' | 'expand' | null =
+    null;
```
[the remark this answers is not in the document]
devadula-nandan: defaulted to tooltip

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.ts
devadula-nandan: added support

devadula-nandan:
the align right seems to be adding some unnecessary styles, it could be adding some rtl direction.

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.scss
devadula-nandan: 👀 looks like a lot of manual setup, also we have to decide at what character we want to truncate and wrap them in spans, which doesn't give the flexibility of having consistent heights for layouts having height constraints, at variable sizes.

in truncated text implementation, when we are truncating by number of lines. i had this challenge of showing the truncation on the text while accounting for `view more` button, which was technically not feasible as webkit line clamp doesn't provide any offset to account for inline button size. a common way design systems tackled this scenario was to add a subtle gradient before the truncation to avoid harsh cut on the text.
we have currently removed it as per the design. but we can add it if we wanted to.
by increasing`0%` in the end to some `20%`
in this line https://github.com/carbon-design-system/ibm-products/pull/7397/files#diff-61d4b66abd22cfa87c99ddcaa80fb43747db3dbd02ea0fecfbb79d5198dffb04R31

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.scss
```diff
@@ -90,6 +87,18 @@ $block-class: #{$prefix}--truncated-text;
...
 
+  &::before {
+    position: absolute;
+    block-size: 100%;
+    color: $text-primary;
+    content: '...';
+    inline-size: 1em;
+    inset-block-start: 0;
+    inset-inline-start: calc(-1em - 0.2em);
+    padding-inline-start: 0.2em;
```
elycheea: Any reason we’re using `em` instead of `rem` here? If we’re using these specific numbers over tokens, we should also comment on how we derived these; e.g. avoid magic numbers!

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.ts
```diff
@@ -0,0 +1,232 @@
...
+   * Specify how the tooltip should align with the content.
+   */
+  @property({ reflect: true, type: String })
+  align = 'top';
+
+  /**
+   * Specify whether a auto align functionality should be applied
+   */
+  @property({ type: Boolean, reflect: true })
+  autoalign = false;
```
elycheea, line 51: Small thing but noticed that React uses `autoAlign` but WC looks like it uses `autoalign`. 👀

elycheea (commented):
A few small comments.

Also wonder if we’Ll need to move some of the styles to `@carbon/ibm-products-styles` once we add the React version.

elycheea (dismissed):
Actually updating to approve — the comments I have are pretty minor and I think can be considered non-blocking. We can definitely update when we update the React implementation too.

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.scss
devadula-nandan: i intended `em` particularly in this case as this utility (or component) will be used in various html text elements, that has varying font sizes, so setting `padding-inline-start: 0.2em` would set the tiny padding between `...` and the text to `0.2em` and it scales as per the context (parent font size) its used in. so a tiny text will have tiny padding, and larger text will have larger padding.

there will be cases where the font sizes are modified in an application.
cases where custom typography mixins could be applied to certain elements.

in such cases, if we use rem, it will always be derived from the root, but em derive the units from the parent.

added the comment and removed padding as we are not using the gradient

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.ts
devadula-nandan: According to Web Component conventions, properties in the JavaScript context should typically be camelCase (e.g., autoAlign), while their corresponding HTML attributes should be kebab-case (auto-align). In the case of carbon/web-components, the `autoalign` property does not follow this convention on tooltip component. i have just followed carbon for consistency. but If this is not a deliberate design decision, it might be worth raising a GitHub issue for clarification or consistency.

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.scss
elycheea: I actually wonder if `ch` would be better alternative then? But can do that when we move styles to `@carbon/ibm-products-styles` for the React version.

packages/ibm-products-web-components/examples/truncated-text/src/styles.scss
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+@use '@carbon/styles/scss/reset';
+@use '@carbon/styles/scss/theme';
+@use '@carbon/styles/scss/themes';
+
+:root {
+  @include theme.theme(themes.$white);
+  background-color: var(--cds-background);
+  color: var(--cds-text-primary);
```
matthewgallo: These should be using sass variables

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.scss
```diff
@@ -0,0 +1,121 @@
...
+
+@use 'sass:map';
+@use '@carbon/styles/scss/utilities' as *;
+@use '@carbon/themes' as *;
+@use '@carbon/styles/scss/spacing' as *;
+@use '@carbon/styles/scss/theme' as *;
+@use '@carbon/styles/scss/motion' as *;
+
+$prefix: 'c4p';
+$carbon-prefix: 'cds';
```
matthewgallo: Can you change this to not hard code the `cds` prefix. It can be accessed from the styles config

```scss
@use '@carbon/styles/scss/config'

.#{config.$prefix}--some-carbon-component {
  ...
}
```

packages/ibm-products-web-components/examples/truncated-text/src/styles.scss
devadula-nandan: looks like all other examples still have them, and coming while cloning a component 👀 , fixed in this.

packages/ibm-products-web-components/src/components/truncated-text/truncated-text.scss
devadula-nandan: looks like all other examples still have them, and coming while cloning a component 👀 , fixed in this.

elycheea (approved):

PR #7403, opened 2025-04-29
fix(InterstitialScreen): resolve accessibility errors

### LINKED ISSUE #7263: [a11y]: two InterstitialScreen component Accessibility Checker errors (probably one-liner)
### Package

@carbon/ibm-products

### Browser

Chrome

### Operating System

MacOS

### Package version

v2.63.0

### React version

v17.0.2

IBM Equal Access Accessibility Checker, and for the first issue also the one built-in into storybook

### Assistive technology

_No response_

### Description

There's an error when running IBM Equal Accessibility Checker on site with the current version of the InterstitialScreen component, regarding element with class
c4p--carousel__elements
The violation is "The scrollable element <div> with non-interactive content is not tabbable"
One way to remove this violation is by giving it tabindex="0".

And there's another error in element with class
c4p--interstitial-screen--body
The violation is "The tabbable element does not have a valid widget role"
which can be mitigated by removing its tabindex="0"

The first (c4p--carousel__elements) one is visible in https://ibm-products.carbondesignsystem.com/?path=/story/experimental-onboarding-interstitial-screen-interstitialscreen--interstitial-screen-full-screen-no-media-with-cards&globals=viewport:basic
You can see it in the Accessibility tab as "Scrollable region must have keyboard access" 

So the issue seems to be solvable by moving that tabindex="0" a few levels deeper from c4p--interstitial-screen--body to c4p--carousel__elements.

### WCAG 2.1 Violation

Success Criterion 4.1.2: Name, Role, Value

### Reproduction/example

https://stackblitz.com/edit/github-m8ct9l9j-mii4sxzt?file=src%2FExample%2Fimage.js

### Steps to reproduce

it appears in our case when IntersititalScreen is visible and IBM Equal Access Accessibility Checker is run. In the stackblitz IBM Equal Access Accessibility Checker probably doesn't check contents of the iframe, sadly.

Here's a screen of what I get (these are the last two violations I'm tasked to investigate/fix):

Image

### Code of Conduct

makafsal:
Closes #7263 

#### What did you change?
Conditionally added `tabindex` and `role`

Accessibility Checker


packages/ibm-products/src/components/Carousel/Carousel.tsx
```diff
@@ -352,6 +354,18 @@ const Carousel = React.forwardRef<HTMLDivElement, CarouselProps>(
...
+    useIsomorphicEffect(() => {
+      if (scrollRef?.current) {
+        setIsScrollableElement(
+          scrollRef.current.scrollHeight > scrollRef.current.clientHeight
+        );
+      }
+    }, []);
+
+    const scrollableElementAttributes = isScrollableElement
+      ? { 'aria-label': 'Carousel element', tabIndex: 0, role: 'region' }
```
matthewgallo, line 366: There should be a way to pass this `aria-label` in to the component so it is not hard coded. I'll open a follow up issue for this.

matthewgallo (approved):

matthewgallo (changes requested):
I'm noticing an AVT test is failing because of these changes 
image

amal-k-joy (changes requested):
Looks like the error is still in coachmarkOverlayElements 
image

packages/ibm-products/src/components/Carousel/Carousel.tsx
```diff
@@ -352,6 +359,14 @@ const Carousel = React.forwardRef<HTMLDivElement, CarouselProps>(
       }
     }, [rightFadedEdgeRef, rightFadedEdgeColor]);
 
+    useIsomorphicEffect(() => {
```
amal-k-joy, line 362: do we need `isScrollableElement` state and this logic now ?

amal-k-joy (changes requested):

amal-k-joy (approved):
LGTM

PR #7415, opened 2025-04-30
chore: move products storybook config from core to products

### LINKED ISSUE #6362: move storybook config from core to ibm-products
after some initial discussion, we agreed to move the storybook config from `core` to `ibm-products` to help simplify the mono-repo

- [x] Move the storybook `/` from packages `core` to the `ibm-products` 
  - [x] Remove `lottie-web` as a dependency
- [x] Double check any scripts/netlify stuffs on the core
  - [x] Add `lottie-web` as a dependency only during netlify build?
- [x] Review and update any Github workflow as needed

davidmenendez:
Closes #6362

### DIFF
--- a/.percy.yml
+++ b/.percy.yml
@@ -10,6 +10,7 @@ discovery:
 storybook:
   include: ['/IBM Products/']
   exclude: [
+      '/Internal/',
       '/Datagrid/*',
       '/DataSpreadsheet/',
       'Cascade: With Grid',


packages/ibm-products/src/components/Datagrid/Datagrid.stories.jsx
```diff
@@ -36,7 +36,7 @@ import { Wrapper } from './utils/Wrapper';
 import DocsPage from './Datagrid.docs-page';
 import { getBatchActions } from './utils/getBatchActions';
 import { StatusIcon } from '../StatusIcon';
-import { Annotation } from '../../../../core/.storybook/Annotation';
+import { Annotation } from '../../../../ibm-products/.storybook/Annotation';
```
matthewgallo, line 39: I'm seeing a lot of imports where you go all the way out to `/ibm-products` but you can just do `../../../.storybook/Annotation` since you're already in `/ibm-products`.

matthewgallo (changes requested):
Left a comment around some imports. Also, all of the storybook packages need to be moved from `packages/core/package.json` into `packages/ibm-products/package.json`. This is also true for the percy packages, that's why [Netlify is currently failing](https://app.netlify.com/sites/carbon-for-ibm-products/deploys/6812848fb15a5d0008cb6c2f#L3180) because it can't find the percy packages in `packages/ibm-products/package.json`.

We will also need to change the "Publish directory" within our Netlify build config after this is merged as it currently points to `packages/core/storybook-static`.

packages/ibm-products/src/components/Datagrid/Datagrid.stories.jsx
davidmenendez: ahh yes. it's so nice that sometimes vscode chooses not to update the paths when you move something 🙄

davidmenendez:
> Left a comment around some imports. Also, all of the storybook packages need to be moved from `packages/core/package.json` into `packages/ibm-products/package.json`. This is also true for the percy packages, that's why [Netlify is currently failing](https://app.netlify.com/sites/carbon-for-ibm-products/deploys/6812848fb15a5d0008cb6c2f#L3180) because it can't find the percy packages in `packages/ibm-products/package.json`.
> 
> We will also need to change the "Publish directory" within our Netlify build config after this is merged as it currently points to `packages/core/storybook-static`.

yeah totally. i knew that debugging this part would probably need to happen once deploying happens because despite cleaning and reinstalling dependencies everything still worked locally regardless of which package.json the dependencies reside in. lots of fun quirks with this PR haha

matthewgallo (approved):
The only other thing I was concerned about was the `lottie-web` dependency remaining within `packages/core/package.json` might not work now that storybook is configured from `packages/ibm-products`, but I ran it locally after the storybook build script via `npx http-server packages/ibm-products/storybook-static` and the lottie animation on the `Welcome` story loads as expected.

I think we eventually could remove `packages/core` entirely but we can revisit that once we get storybook setup, nice work! 🎉

matthewgallo:
Storybook is failing since the config is still expecting the publish directory to be `packages/core/storybook-static`. Right before we merge this, we should update the Netlify config to point to `packages/ibm-products/storybook-static`.

matthewgallo (changes requested):
One last thing I just noticed, the `Internal` directory should not be included.
Screenshot 2025-05-02 at 11 28 09 AM

davidmenendez:
in the last commit i changed the `.stories.jsx` extension to `.internal.stories.jsx` for internal stories. they no longer show up in the sidebar.

davidmenendez:
netlify is failing, but it's back to just updating the config from the netlify UI 👍

matthewgallo:
Nice! My only other comment would be to ignore any "Internal" stories from percy.

Issue #7298, opened 2025-04-14
Visual Bug: Simple page header with breadcrumb


rachels01:
### Package

Carbon for IBM Products

### Description

**The issue:**
- Currently in Storybook, the Simple Page Header With Breadcrumb is misaligned when scrolling. Upon scroll, both the breadcrumb and any buttons should be vertically centered. 

![Image](https://github.com/user-attachments/assets/dfa5a8be-2bf9-4e18-bcfb-7f4aeb555843)

![Image](https://github.com/user-attachments/assets/ff9f2c8d-c7c0-4902-8268-b055404b6600)

### Component(s) impacted

Simple Page Header With Breadcrumb

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

v11

### Suggested Severity

None

### Product/offering

Carbon for Products

### CodeSandbox or Stackblitz example

https://ibm-products.carbondesignsystem.com/?path=/story/ibm-products-components-page-header-pageheader--with-tabs-and-tags&globals=viewport:basic

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

rachels01:
Note: 
This issue was originally flagged in this message: https://ibm-casdesign.slack.com/archives/CQGR0HC05/p1744652907942069

PR #7447, opened 2025-05-06
fix(PageHeader): page actions without title (collapseTitle=true)

### LINKED ISSUE #7277: [PageHeader]: pageActions not visible with collapseTitle
### Package

Carbon for IBM Products

### Description

Earlier (in v1), we could show `pageActions` with `collapseTitle=true`, in v2, they're not visible.

Also, this is an unrelated secondary issue, we can't show `pageActions` with collapseTitle=true unless we also provide `actionBarItems` which I think shouldn't be required if you just need to show pageActions.

Root cause:
Image

Potential fix:
Image

### Component(s) impacted

PageHeader

### Browser

Chrome

2.63.0

### Suggested Severity

### Product/offering

IBM Verify

https://stackblitz.com/edit/c11-ts-template-keajhy3m?file=src%2FApp.tsx

Image

_No response_

### Code of Conduct

makafsal:
Closes #7277 #7298 

#### What did you change?
Changed styles

`yarn test`
Storybook


amal-k-joy (approved):
LGTM

packages/ibm-products/src/components/PageHeader/PageHeader.tsx
```diff
@@ -1035,7 +1038,7 @@ export let PageHeader = React.forwardRef(
                             />
                           </>
                         ) : (
-                          widthIsNarrow &&
+                          (widthIsNarrow || pageActions) &&
```
s100, line 1041: @ntkibm This change seems to result in a strange scenario where `thePageActions(...)` renders twice in our case, once in the breadcrumb row (invisibly??) and once more in the title row. This has caused some of our tests to break because they assume that the page actions will only appear in the DOM once. I'm not sure whether to raise a bug for this, it's possible that we are misusing the component. Can you explain in more detail what this change does please?

ntkibm: hey @s100, you probably want to check with @makafsal who is the author of this pr.. i'm the reporter of the issue.. i don't have much idea about the internal code.

PR #7463, opened 2025-05-07
feat(emptystate): automatically use right heading level

### LINKED ISSUE #7459: [Feature Request]: EmptyState: automatically use correct heading level
### The problem

As follows from https://www.w3.org/WAI/tutorials/page-structure/headings/, an EmptyState's heading level needs to vary depending on where the EmptyState is shown.  If the page is nothing but the EmptyState component, then the heading should be H1.   If EmptyState is in a tearsheet, then it should be H4.   If it's in a section of the page, then it depends on the level of the nearest ancestor heading.  Etc.

In theory, this means that the application is responsible for instantiating the EmptyState as

```jsx
<EmptyState headingAs="h1"> ... </EmptyState>
```

or

```jsx
<EmptyState headingAs="h2"> ... </EmptyState>
```

etc.

In practice, it means that most EmptyState components will just render with its default `<h3>` heading, even if another heading level is more appropriate.

### The solution

Ideally, EmptyState would leverage Section and Heading, like I've done for other @carbon/react and @carbon/ibm-products components, and that would be the end of the story.

However, EmptyState is more complicated than other components (ex: Card) because sometimes it's shown full-page, in which case it should ideally use an H1 heading.

The best solution I can think of is to automatically set the heading level for the non-full-page cases, i.e. whenever there's a parent heading logically above the EmptyState component.  But for the full page case, still require the app to manually specify headingAs:

```jsx
<EmptyState headingAs="h1"> ... </EmptyState>
```

I'm open to any other better ideas.

### Design link

_No response_

### Links to other materials

See https://carbondesignsystem.com/patterns/empty-states-pattern/ for demonstrations of where the EmptyState can occur.

And see https://www.w3.org/WAI/tutorials/page-structure/headings/ for general information on how heading hierarchies are supposed to work in HTML.

### Owner/main maintainer(s)

wkeese

_No response_

### Product/offering

IKC

### Business priority

None

### Code of Conduct

wkeese:
Closes #7459.

Enhance `EmptyState` to use the right heading level automatically, rather than forcing the application to specify `headingAs`.

Note that you still need to specify `headingAs="h1"` for full-page `EmptyState`, i.e. when there is no higher-level heading above the `EmptyState`.

These changes match the heading updates I've been making to other components to use `Section` and `Heading`.

Also, fix strange code in `EmptyStateContent` that was confused about when to use `Section` and when to use `Heading`.

#### What did you change?

`EmptyState`'s wrapper node is a `Section`, and it uses `Heading` for the heading.  Unless overridden by `headingAs`.

Storybook, including a temporary change to explicitly specify `headingAs`.

### DIFF
--- a/packages/ibm-products/src/components/EmptyStates/EmptyState.tsx
+++ b/packages/ibm-products/src/components/EmptyStates/EmptyState.tsx
@@ -31,10 +31,9 @@ enum sizes {
 }
 
 // Default values for props
-export const defaults: { position: string; size: sizes; headingAs: string } = {
+export const defaults: { position: string; size: sizes } = {
   position: 'top',
   size: sizes.lg,
-  headingAs: 'h3',
 };
 
 export interface EmptyStateAction extends ButtonProps<React.ElementType> {
@@ -79,7 +78,9 @@ export interface EmptyStateProps {
   link?: CustomLink;
 
   /**
-   * Empty state headingAs allows you to customize the type of heading element
+   * Customize the heading element.  Set to "h1" when EmptyState is full page, i.e. there is no higher header.
+   * Otherwise, you normally don't need to specify this: EmptyState will automatically pick the right heading level
+   * (h2-h6) by leveraging Section and Heading.
    */
   headingAs?: (() => ReactNode) | string | ElementType;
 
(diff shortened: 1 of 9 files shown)


matthewgallo (approved):
Thanks as always @wkeese for your contribution! 🎉

davidmenendez (approved):

Issue #7461, opened 2025-05-07
InterstitialScreen: accessibility testing issue set 2


amal-k-joy:
These are the set 2 of issues raised by Raghu on his accessibility testing.

Issue   title/Summary | Issue Description | Recommended Fix | User Impact | Developer comments
-- | -- | -- | -- | --
Non Interactive element Modal Div receiving  Tab focus . | Modal Div   receiving  Tab focus .:     Tab focus on Non interactive controls    such as Text , Div etc.  will   make  the Keyboard /Screen reader users   confuse. | Avoid Tab focus   from Non Interactive Elements such as    Div , Text etc. | When   noninteractive elements like texts, graphics or icons receive tab focus,   screen reader users can wrongly assume them to be interactive controls and   can even try to activate them. |  
Content modification on the Modal due to activation on Next   /Previous Buttons are not conveyed  to   screen reader users. | Content   modification on the Modal due to activation on Next /Previous Buttons are not   conveyed  to screen reader users.     By clicking  Next /Previous  content in the Modal got updated - However   , screen reader   users does not get   any clue on this content modifications. | Make   the dynamic  content updates  indicated to screen reader users  by using appropriate technics :      For example : Providing  a  status indication  for screen reader users  such as - "Step 3    Provide your experience  Details " Step 4 - "Provide   your  Qualificational  information " etc.      while moving to these steps - screen reader should auto announce the   Topic  of the Steps | If   the screen reader fails to identify a dynamic content change, it affects   users who are blind as they may not know about the change in the page. |  
Interactive components with in the Modal not reachable  by Tab key    consistently . | Modal   controls(step1) are not reachable by Tab consistently - also  once the    Elements receives focus - Tab order not following the visual   order  of the  components , which  how these are placed in the  Modal. | Make   the  Modal controls  reachable by Keyboard alone , also Tab   order expected to follow the visual order of the control  consistently. | When   the standard keyboard navigation method is not supported, it impacts the   user’s ability to interact with page controls. This defect affects screen   reader and keyboard-only users. |  
[...]

Issue #7460, opened 2025-05-07
InterstitialScreen: accessibility testing issue set 2


amal-k-joy:
These are the set 2 of issues raised by Raghu on his accessibility testing.

[Carbon for IBM Components_Interstitialscreen_12thMay_Updated.xlsx](https://github.com/user-attachments/files/20186087/Carbon.for.IBM.Components_Interstitialscreen_12thMay_Updated.xlsx)

Issue   title/Summary | Issue Description | Recommended Fix | User Impact | Developer comments
-- | -- | -- | -- | --
After closing the  modal,   focus is not returning to the triggered element | After closing   the  modal, focus is not returning to   the triggered element | After closing   the  modal, focus  expected to set back to the  triggered element | After a user   action, if the focus randomly jumps to another spot on the web page, it’s   tedious to navigate back to the relevant element. This negatively affects the   experience of keyboard-only users and those with cognitive disabilities. | Introduced a new   prop `launcherButtonRef` to achieve this similar to components
Screen reader not announced the status  for the    Step indicators -  where  the status of each steps  indicated visually. | Status   indicators are presented as Buttons  -   here  focusing on each Step indicators   Screen reader not announced the status    -  where  the status of each steps  indicated visually. The announcing status   is not clearer to the screen reader user. | Make   the status   indication  available for screen reader users as well   :     Focusing on each  Stepper Elements  screen reader expected to announce  its    status .     For example : Step 1  completed        Step 2  IN progress /Current      Step 3  Not completed  so on | When   the name, role or state of an element is not conveyed appropriately, it   prevents screen reader users to interact smoothly with the page element. | Step   change is announced to user using `aria-live`
[...]

PR #7478, opened 2025-05-09
fix: interstitial accessibility issue fixes

### LINKED ISSUE #7458: InterstitialScreen: accessibility testing issue set 1
These are the set 1 of issues raised by Raghu on his accessibility testing.

[Carbon for IBM Components_Interstitialscreen_12thMay_Updated.xlsx](https://github.com/user-attachments/files/20186087/Carbon.for.IBM.Components_Interstitialscreen_12thMay_Updated.xlsx)

Issue   title/Summary | Issue Description | Recommended Fix | User Impact | Developer comments
-- | -- | -- | -- | --
reading   order of Close Button is not  following   the visual order  of the Element. | On Arrow key navigation  , "Close Button " announced after   the H2  and H3 level headings  (which mentioned below) , Here  Screen reader  reading order is not matching with Logical   order . | Recommend to  make this "Close Button "   announced  as the First Element in the   Modal . | If the components are  not announced in a  Meaningful order /In an order How its   visually placed in the  screen - its   may distract its   meaning , How it's   suppose to be understood, impact    screen reader users. | This is a known behaviour of   carbon compose modal. Ref   [18262](https://github.com/carbon-design-system/carbon/issues/18262). Close button is placed after   the heading elements.
Screen   reader is announcing the  Elements   which is not visually available in the screen | Screen reader is announcing   the  Elements which is not visually   available in the screen - creating confusion    for Screen reader users.     Note : This element is not  keyboard   operable /Tab key reachable . | Screen reader is not expected to   announce the  Elements which is not   visually available in the screen :     Make sure  the visually hidden   elements are hidden from screen reader detection as well . | If the screen reader identifies   the visually hidden content, it tends to mislead users who are blind. | Behaviour of compose modal to trap focus via sentinal elements. Fixed by using the carbon _Feature flag_ (`enableExperimentalFocusWrapWithoutSentinels`)
Heading   level hierarchy  is not maintained        1. H1 Placed after H2 and H3 Level Headings .     2. H1  is not assigned  for the Main Headline  of the Modal. | Heading level hierarchy  is not maintained- Screen Reder  users may struggle to identify the content   structure of the page . | For the benefits of Screen Reder   users :     1. Maintain heading levels    hierarchy       2. H1 for the Main Header for the Modal    at the Top , H2 for the Sub Headlines of H1, H3 for the subhead lines   of H2  and so on | If page headings are not in the   right sequence, screen reader users find it difficult to comprehend the   hierarchical structure of the page content. | Fixed. Title and subTitle changed   to h1 and h2

amal-k-joy:
Closes #7458, #7460, #7461

{{short description}}

### DIFF
--- a/packages/ibm-products/src/components/Carousel/Carousel.tsx
+++ b/packages/ibm-products/src/components/Carousel/Carousel.tsx
@@ -230,8 +230,21 @@ const Carousel = React.forwardRef<HTMLDivElement, CarouselProps>(
     }, [handleOnScroll]);
 
     const handleScrollToView = useCallback((itemNumber) => {
+      updateAriaHidden(itemNumber);
       childElementsRef.current[itemNumber].scrollIntoView();
     }, []);
+
+    const updateAriaHidden = (itemNumber) => {
+      //aria-hidden need to updated based on the active item, otherwise screen reader will reset to first item while
+      //interact with element via Control + Option + Down Arrow
+      childElementsRef.current?.forEach((item, idx) => {
+        if (idx === itemNumber) {
+          item.setAttribute('aria-hidden', false);
+        } else {
+          item.setAttribute('aria-hidden', true);
+        }
+      });
+    };
 
     // Trigger a callback after first render (and applied CSS).
     useEffect(() => {
@@ -242,6 +255,7 @@ const Carousel = React.forwardRef<HTMLDivElement, CarouselProps>(
       setTimeout(() => {
         // But, because we are making calculations based on the final,
         // applied CSS, we must wait for one more "tick".
+        updateAriaHidden(0);
         handleOnScroll();
       }, 0);
       // eslint-disable-next-line react-hooks/exhaustive-deps
--- a/packages/ibm-products/src/components/InterstitialScreen/InterstitialScreen.tsx
+++ b/packages/ibm-products/src/components/InterstitialScreen/InterstitialScreen.tsx
@@ -77,6 +77,11 @@ export interface InterstitialScreenProps {
    * Function to call when the close button is clicked.
    */
   onClose?: (value: ActionType) => void;
+
+  /**
+   * Provide a ref to return focus to once the interstitial is closed.
+   */
+  launcherButtonRef?: RefObject<any>;
 }
 
 // Define the type for InterstitialScreen, extending it to include Header
@@ -128,6 +133,7 @@ export let InterstitialScreen = React.forwardRef<
     interstitialAriaLabel = 'Interstitial screen',
     isFullScreen = false,
     isOpen = false,
+    launcherButtonRef,
     onClose,
     ...rest
   } = props;
@@ -174,7 +180,12 @@ export let InterstitialScreen = React.forwardRef<
     // for modal only, "is-visible" triggers animation
     setIsVisibleClass(!isFullScreen && isOpen ? 'is-visible' : null);
     nextButtonRef?.current?.focus();
-  }, [isFullScreen, isOpen]);
+    if (!isOpen && launcherButtonRef) {
+      setTimeout(() => {
+        launcherButtonRef.current.focus();
+      }, 0);
+    }
+  }, [launcherButtonRef, isFullScreen, isOpen]);
 
   // hitting escape key also closes this component
   useEffect(() => {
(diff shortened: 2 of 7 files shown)


packages/ibm-products/src/components/Guidebanner/Guidebanner.tsx
```diff
@@ -150,6 +150,7 @@ export let Guidebanner = React.forwardRef<HTMLDivElement, GuidebannerProps>(
           onScroll={(scrollPercent) => {
             setScrollPosition(scrollPercent);
           }}
+          isScrollMode={true}
```
makafsal, line 153: Why is this always `true`?

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreen.tsx
```diff
@@ -77,6 +80,11 @@ export interface InterstitialScreenProps {
    * Function to call when the close button is clicked.
    */
   onClose?: (value: ActionType) => void;
+
+  /**
+   * Provide a ref to return focus to once the interstitial is closed.
+   */
+  launcherButtonRef?: RefObject<any>;
```
makafsal, line 87: Instead of `any`, would `HTMLElement` be fine?

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenBody.tsx
```diff
@@ -51,7 +51,7 @@ const InterstitialScreenBody = React.forwardRef<
   const blockClass = `${pkg.prefix}--interstitial-screen`;
   const bodyBlockClass = `${blockClass}--internal-body`;
 
-  const [isMultiStep, setIsMultiStep] = useState(false);
+  const [stepType, setStepType] = useState('');
```
makafsal, line 54: What about giving a type for `stepType`? 
`type StepType = 'multi' | 'single'`

makafsal (changes requested):

packages/ibm-products/src/components/Guidebanner/Guidebanner.tsx
amal-k-joy: thats is true only for `Guidebanner`, since in Guidebanner, multiple carousal items will be visible at the same time. So here is used more like a scrolling behaviours. 
For other components like `interstitial` , it will be false.

packages/ibm-products/src/components/Carousel/Carousel.tsx
```diff
@@ -230,9 +241,46 @@ const Carousel = React.forwardRef<HTMLDivElement, CarouselProps>(
     }, [handleOnScroll]);
 
     const handleScrollToView = useCallback((itemNumber) => {
+      updateAriaHiddenTabIndex(itemNumber);
       childElementsRef.current[itemNumber].scrollIntoView();
+      // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
+    const getFocusableElements = (container) => {
```
makafsal, line 249: Try to utilize the `getFocusable` method from the `useFocus` hook. You can only consume all the focusable elements from the container by spreading `allElements` from the `useFocus` hook.

```sh
const { allElements } = useFocus(modalRef);
```

It only invoke the `getFocusable` method, so no need to worry about the unwanted features of `useFocus` hook.

packages/ibm-products/src/components/InterstitialScreen/InterstitialScreen.tsx
```diff
@@ -174,7 +183,12 @@ export let InterstitialScreen = React.forwardRef<
     // for modal only, "is-visible" triggers animation
     setIsVisibleClass(!isFullScreen && isOpen ? 'is-visible' : null);
     nextButtonRef?.current?.focus();
-  }, [isFullScreen, isOpen]);
+    if (!isOpen && launcherButtonRef) {
+      setTimeout(() => {
+        launcherButtonRef.current.focus();
```
makafsal, line 188: I think better to include a test for this.

makafsal (changes requested):

packages/ibm-products/src/components/Carousel/Carousel.tsx
amal-k-joy: We cannot use `useFocus` hook here since i need to get the elements dynamically based on a logic.

makafsal (approved):
LGTM

PR #7494, opened 2025-05-13
feat: add storybook (react and wc) amplitude instrumentation

matthewgallo:
Closes https://github.com/carbon-design-system/carbon-for-products-website/issues/534

This PR will enable Amplitude instrumentation in our React and web component storybooks. I chose to use the cdn option so we don't have to worry about the amplitude dependency, especially now that our storybook is included in `package/ibm-products`.

I included the change in `/.storybook/main.js` instead of `manager-head.html` because you can't access environment variables from `manager-head.html`. The contents provided in the `manageHead` property callback are the contents of `manager-head.html` so it still works the same. This allows us to only send events in production by checking the value of `process.env.NODE_ENV`.

#### What did you change?
```
cspell.json
packages/ibm-products/.storybook/main.js
packages/ibm-products-web-components/.storybook/main.ts
```
Verified event was sent via browser extension and from within the Amplitude project created


davidmenendez (approved):

davidmenendez (dismissed):

davidmenendez (approved):

jlongshore (approved):
LGTM

Issue #7513, opened 2025-05-15
Investigate flaky Percy tests


matthewgallo:
Our PRs consistently report visual changes from Percy and most often this is because of flaky percy tests. https://github.com/carbon-design-system/ibm-products/pull/7483 includes some changes to hopefully resolve this by utilizing `waitForSelector` in the global percy config to make sure percy doesn't take snapshots until storybook has loaded.

PR #7520, opened 2025-05-16
test: try `.story-wrapper` for percy waitForSelector

matthewgallo:
Closes https://github.com/carbon-design-system/ibm-products/issues/7513 (already closed, but I noticed Percy still took a snapshot of the storybook loading spinner for `Saving` so trying a different `waitForSelector`)

I updated the `waitForSelector` value to target the `.story-wrapper` element which doesn't even get included in the DOM until the story has loaded. I'm hoping switching to this selector helps because I noticed that in one of the Percy builds, the `Saving` component received a snapshot of the storybook loader which is not what we would expect.

#### What did you change?
`.percy.yml`
Checking the percy flagged snapshots

### DIFF
--- a/.percy.yml
+++ b/.percy.yml
@@ -12,7 +12,7 @@ storybook:
   # waits for data-carbon-theme attribute to exist before taking snapshot
   # if storybook is still loading, percy could take a snapshot of the storybook
   # loading spinner and cause flaky tests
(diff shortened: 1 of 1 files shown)


amal-k-joy (approved):
@matthewgallo 
Can you link an issue to the PR?

PR #7531, opened 2025-05-19
fix: include .internal.stories to ignore stories glob

davidmenendez:


matthewgallo (approved):
I like the cleanup in here too 👏

anamikaanu96 (approved):

PR #7540, opened 2025-05-21
chore: change-logs repeated in all packages

### LINKED ISSUE #7437: Commits repeated in Web Components changelog
While reviewing #7436, noticed that our [Web Components changelogs](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products-web-components/CHANGELOG.md) have been repeating.

The release notes in [**Releases**](https://github.com/carbon-design-system/ibm-products/releases) are correct so this issue may be related to the PR action.

### Examples

- The full-page error was originally released in 0.13.0 but appears again in the changelogs for 0.14.0 and 0.15.0.
- The `es-custom` has appeared from 0.11.0 through latest
- “web-components: specify files/dirs to include in package” appeared from 0.8.0 through 0.11.0.

### Acceptance criteria

- [x] Investigate and address cause of repeated release notes
- [ ] Based on [**Releases**](https://github.com/carbon-design-system/ibm-products/releases), correct the WC release notes and remove the incorrectly repeated items

### observations
Seems that , not just webcomponents change log, all changelogs are being repeated since [v2.60.0-rc.0](https://github.com/carbon-design-system/ibm-products/pull/7524/files#diff-8f035289d274d6d70d7fcda70c304e870ea8163ef97c0c34e764bd38d6e389b4R6). Also noticed that we had a 2.60.0-rc.1, that would have triggered the accumulation of logs. This seems like a issue with lerna change log creation.

### Potential fix
We are currently using get-changelog.js as part of [Create github tag and PR](https://github.com/carbon-design-system/ibm-products/blob/main/.github/workflows/create-release-tag-and-pr.yml) workflow to generate release notes.
We could use the same script to write those logs into respective changelog.md files as demonstrated in this [draft PR](https://github.com/carbon-design-system/ibm-products/pull/7540) and disable lerna for changelogs.

We can also consider not maintaining `changelog.md` , since core doesn't maintain those.

amal-k-joy:
Closes #7437 

[Change-logs](https://github.com/carbon-design-system/ibm-products/pull/7524/files#diff-8f035289d274d6d70d7fcda70c304e870ea8163ef97c0c34e764bd38d6e389b4R6) seem to be accumulating and repeating since `v2.62.0` in all packages. Lerna used generate and update changelog.md during releases. Looks like some glitch happened after a subsequent RC release `2.62.0-rc.1`.

Now `changelog.md` is updated using a manual script `get-changelog.js` while running the workflow  `Create github tag and PR `during release. We are already using the same script for generating changelog in [release](https://github.com/carbon-design-system/ibm-products/releases) page. Now I am reusing this scripts to update `changelog.md` as well.

#### What did you change?
Updated `get-changelog.js` to include logic to write the logs to `changelog.md` files
Updated `Create github tag and PR` to include action to update changelogs
Changelogs are generating in local as expected. But fullflow can be tested after merging.

### DIFF
--- a/.github/workflows/create-release-tag-and-pr.yml
+++ b/.github/workflows/create-release-tag-and-pr.yml
@@ -45,9 +45,29 @@ jobs:
 
       # Create a new branch off the release branch for the PR, we do this to ensure the release branch
       # is not affected
+
+      - name: Setup Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: '20.x'
+          cache: 'yarn'
+
+      - name: Install dependencies
+        run: yarn install
+
       - name: Create branch
         run: |
           git checkout -b chore/${{ github.event.inputs.tag }}-release
+
+      - name: Generate CHANGELOG.md
+        run: |
+          node ./scripts/get-changelog.js -f ${{ github.event.inputs.previous-tag }} -t ${{ github.event.inputs.tag }}
+
+      - name: Commit changelog
+        run: |
+          git add CHANGELOG.md
+          git commit -m "docs(changelog): update for ${{ github.event.inputs.tag }}" || echo "No changes to commit"
+
           git push origin chore/${{ github.event.inputs.tag }}-release
       # Create PR with the new branch
       - name: Create Pull Request
--- a/.github/workflows/create-release-tag-and-pr.yml
+++ b/.github/workflows/create-release-tag-and-pr.yml
@@ -61,7 +61,7 @@ jobs:
 
       - name: Generate CHANGELOG.md
         run: |
-          node ./scripts/get-changelog.js -f ${{ github.event.inputs.previous-tag }} -t ${{ github.event.inputs.tag }}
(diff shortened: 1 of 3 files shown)


matthewgallo (approved):

.github/workflows/create-release-tag-and-pr.yml
```diff
@@ -45,9 +45,29 @@ jobs:
...
         run: |
           git checkout -b chore/${{ github.event.inputs.tag }}-release
+
+      - name: Generate CHANGELOG.md
+        run: |
+          node ./scripts/get-changelog.js -f ${{ github.event.inputs.previous-tag }} -t ${{ github.event.inputs.tag }} -w true
+
+      - name: Commit changelog
+        run: |
+          git add CHANGELOG.md
```
elycheea, line 68:
```suggestion
          git add .
```

I think this should actually be `git add .` or maybe `git add '**/CHANGELOG.md'` to scope it down.

amal-k-joy: Good catch.
I have updated accordingly

PR #7553, opened 2025-05-22
chore: overflow handler web component stories

### LINKED ISSUE #7453: Add stories and documentation for OverflowHandler
Now that the utilities section in the [web components storybook](https://ibm-products-web-components.netlify.app/?path=/docs/experimental-utilities-overflowhandler--docs) has been merged we can start working on adding the stories and documentation.

davidmenendez:
Closes #7453

### DIFF
--- a/packages/ibm-products-web-components/src/utilities/overflow-handler/overflow-handler.stories.ts
+++ b/packages/ibm-products-web-components/src/utilities/overflow-handler/overflow-handler.stories.ts
@@ -8,17 +8,18 @@
  */
 
 import { html } from 'lit';
+import { render } from 'lit-html';
 import styles from './story-styles.scss?lit';
 import '@carbon/web-components/es/components/tag/index.js';
 import '@carbon/web-components/es/components/slider/index.js';
 
 const storyPrefix = 'overflow-handler-stories__';
 
-const getTags = (n: number = 5) => {
+const getItems = (n: number = 5) => {
   return Array(n)
     .fill(null)
     .map((_, idx) => ({
-      label: `tag-${idx + 1}`,
+      label: `item-${idx + 1}`,
     }));
 };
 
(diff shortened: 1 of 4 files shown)


davidmenendez:
@jlongshore & @devadula-nandan when you have a change to review please take a look. thanks!

devadula-nandan:
Hi @davidmenendez, 
the dimension control doesn't seem to be working. could you check with that? as per our discussion the slider is supposed to change the height and flex direction should be set to column in order for that to work.

also by making the resizable container styles
resize: auto;
overflow: auto;
could get rid of the sliders and all their states that handle width and height.
i remember rich sharing a design for presenting resizable containers in stories

the mentioning of data-fixed attribute was missing in the docs page.
here is an example of `data-fixed` attribute in action
https://stackblitz.com/edit/vitejs-vite-j8linif7?file=src%2Ftagset.ts,src%2Factionbar.ts,index.html

the docs mentions 
>For details on implementation, please refer to the example code.

which means patterns that can be built using this utility i.e. `ActionBar`, `TagSet`, etc. will be provided as examples if i am right?

devadula-nandan (commented):
Hi @davidmenendez, 

the dimension control doesn't seem to be working. could you check with that? as per our discussion the slider is supposed to change the height and flex direction should be set to column, and height 100% of parent in order for that to work.

the mentioning of data-fixed attribute was missing in the docs page.
here is an example of `data-fixed` attribute in action
https://stackblitz.com/edit/vitejs-vite-j8linif7?file=src%2Ftagset.ts,src%2Factionbar.ts,index.html

the docs mentions 
>For details on implementation, please refer to the example code.

which means patterns that can be built using this utility i.e. `ActionBar`, `TagSet`, etc. will be provided as examples if i am right?

davidmenendez:
@devadula-nandan i knew i had forgotten something haha. thank you for the reminder. i will update now!

in regards to the example code comment: i think how we want to display "recipe" code is still being determined. right now i don't think we need to explicitly re-write those patterns exactly as examples since we're demonstrating different use cases with avatars and tags. if we determine we want to explicitly demonstrate that code we can add that in a follow up PR.

devadula-nandan:
> @devadula-nandan i knew i had forgotten something haha. thank you for the reminder. i will update now!
> 
> in regards to the example code comment: i think how we want to display "recipe" code is still being determined. right now i don't think we need to explicitly re-write those patterns exactly as examples since we're demonstrating different use cases with avatars and tags. if we determine we want to explicitly demonstrate that code we can add that in a follow up PR.

great, i was asking because, all those example patterns are built in stackblitz, while the work on overflowHandler has happened, to verify the use cases, and are ready to be moved into patterns examples until we decide to do.
showing a coded example would always helps in discoverability of this utility.

anyways here's the `assumed to be` expected behavior for the dimension control
https://deploy-preview-7611--ibm-products-web-components.netlify.app/?path=/story/experimental-utilities-overflowhandler--with-tags
and the changes that could potentially achieve it. in case it helps.
https://github.com/carbon-design-system/ibm-products/pull/7611/commits/a6dbd552044e59d5d07d9a539647b1ba2bdd3289

davidmenendez:
@devadula-nandan changes added 👍

devadula-nandan (approved):
@davidmenendez, added few final changes, to make it complete. pls check once.

elycheea (commented):
Summarizing comments from Slack —
- Would be nice if we also showed the “Hidden items” section like we do in React, especially since the docs do reference that you can render hidden items elsewhere.
- The docs actually feel like they could/should belong on the README for the overflowHandler. Maybe we can followup and add them to the utility?
- In terms of the IA, it actually feels like this could be a whole “Items overflow” pattern, which might feel more straightforward on the Storybook than documenting the utility from outside our library — which you brought up initially too but had to see it to understand your concerns about it.

None of this is blocking for the PR but could be followup work, especially depending on design feedback.

devadula-nandan:
>Would be nice if we also showed the “Hidden items” section like we do in React, especially since the docs do reference that you can render hidden items elsewhere.

the hidden and visible count is shown in actions tab as well but.

i think there is still a way, inspite of the the web component story template limitation compared to react templates supporting reactive states. not sure if it really works. but will see if i can find it and add in a follow up.

the idea is, we need to show the same thing from visible (clone it on a different node), and scope the styles to show items that have the hidden attribute.

but none of it is a blocker, and this is good to merge.

PR #7557, opened 2025-05-23
feat(notification panel): create notification panel wc

### LINKED ISSUE #7452: Create initial Notifications panel (web component)

sangeethababu9223:
Closes #7452 

Create initial Notification panel (web component)

#### What did you change?
- Created Notification panel (`c4p-notification-panel`) and Notification (`c4p-notification`) Components
- Header section is kept as same as React version - Passing props
- Today and Previous sections can be given as slots - This takes away the sorting logic and inner element styling from the component itself. Any styling/Sorting logic will have to be handled by user. (The styling is kept in storybook style file)
-  In Notification component, Title and and Description are given as slots, this gives flexibility to add html elements to the title and description instead of just text. In these sections too, the styling is expected to be handled by user. (Basic styles are added in story style file)
- Footer is a single slot and user an choose to add elements and style it as needed. (Has kept basing styling in storybook assets)
- Custom events in Notification Panel Component: `c4p-notification-dismiss-all`, `c4p-notification-donot-disturb-change`, `c4p-notification-click-outside` and `c4p-notification-panel-beingclosed`.
- Custom events in Notification: `c4p-notification-dismiss`
- Default component for footer slot added - `c4p-notification-footer`

Storybook


packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
```diff
@@ -0,0 +1,372 @@
...
+    } = this;
+    const classes = classMap({
+      [`${blockClass}__container`]: true,
+      [`${blockClass}__entrance`]: open,
+      [`${blockClass}__exit`]: !open,
+    });
+
+    return html`
+      <div role="dialog" tabindex="0" class=${classes}>
+        <div>
```
amal-k-joy: Does this wrapper required?

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
```diff
@@ -0,0 +1,372 @@
...
+              label-a=${doNotDisturbLabel}
+              label-b=${doNotDisturbLabel}
+              aria-label=${doNotDisturbLabel}
+              @cds-toggle-changed=${handleToggle}
+            ></cds-toggle>
+          </div>
+          <div class="${blockClass}__main-section">
+            ${_hasTodayContent
+              ? html`
+                  <cds-heading class="${blockClass}__time-section-label">
```
amal-k-joy: Can you validate on the heading level? Currently its rendering as h1 where as in react its h3. Check the feasibility user passed levels

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
```diff
@@ -0,0 +1,372 @@
...
+  @HostListener('document:click')
+  // @ts-ignore: The decorator refers to this method but TS thinks this method is not referred to
+  private _handleClick = (event: MouseEvent) => {
+    const target = event.target as Node;
+
+    if (!this.open || this.contains(target)) {
+      return;
+    }
+    // Check if the target is an actionable element
+    const actionableSelectors = [
```
amal-k-joy, line 346: What this logic is supposed to do?
When i kept open the panel and clicked trigger button , it closed the panel.


Also should we remove the event listener after panel is closed or is it already happening?

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
```diff
@@ -0,0 +1,372 @@
...
+    this.dispatchEvent(
+      new CustomEvent(
+        (
+          this.constructor as typeof CDSNotificationPanel
+        ).eventDonotDisturbChange,
+        init
+      )
+    );
+  }
+  @HostListener('document:keydown')
```
amal-k-joy: do we need to listen to `document` `keydown` or just listen to the `panel` `keydown`?

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
matthewgallo: Elysia and I were talking about this yesterday as well. Is there a need to use `cds-heading` here? I don't think the notification needs a heading level at all, it's possibly more an issue of removing that in the React implementation.

This would also solve the style issues you discussed during our eng huddle.

sangeethababu9223: Hey @matthewgallo,
Yes, If we don't have to use `cds-heading` then it would save us a lot of trouble with the styling. But would it be okay to directly use an h3 tag instead?

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.mdx
```diff
@@ -0,0 +1,124 @@
...
+                  .open=${args.open}
+                  @c4p-notification-click=${item.onNotificationClick}
+                  @c4p-notification-dismiss=${() => {
+                    notificationSingleDismiss(item.id, 'today');
+                  }}
+                  type=${item.type}
+                  unread=${item.unread}
+                  .timestamp=${item.timestamp}
+                >
+                  <cds-heading
```
amal-k-joy: I think this itself is better approach to keep the `title` slot and I would move the `title` styling to storybook level rather than from component except margin-block-end. This will allow user to use heading of any level or anything based on their requirement.
`:slotted(.c4p--notifications-panel__notification__notification-title) {....`

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.scss
```diff
@@ -0,0 +1,66 @@
...
+$carbon-prefix: 'cds';
+@use '@carbon/ibm-products-styles/scss/components/NotificationsPanel/index' as *;
+
+$block-class: #{$prefix}--notifications-panel;
+$block-class-action-set: #{$prefix}--action-set;
+
+:host(#{$prefix}-notification-panel) {
+  .c4p--notifications-panel__do-not-disturb-toggle {
+    display: block;
+    margin-block-start: -$spacing-05;
```
amal-k-joy, line 32: Looks like carbon is unnecessarily keeping this element even though `labelText` prop is not passed, which caused an extra spacing that made you to give -ve margin. I think we can raise an issue in carbon to remove this.

image

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
amal-k-joy: I too think that a better way to avoid lot of style overrides through variables. Probably you can start from h2 for **Notifications** and hierarchically down.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.scss
```diff
@@ -0,0 +1,66 @@
...
+  --#{$carbon-prefix}-heading-06-font-weight: --#{$carbon-prefix}-label-01-font-weight;
+  --#{$carbon-prefix}-heading-06-line-height: --#{$carbon-prefix}-label-01-line-height;
+  --#{$carbon-prefix}-heading-06-letter-spacing: --#{$carbon-prefix}-label-01-letter-spacing;
+}
+
+.#{$block-class}__bottom-actions-container {
+  position: sticky;
+  z-index: 2;
+  background-color: $layer-01;
+  block-size: 2.5rem;
```
amal-k-joy: do we have a carbon token instead of using absolute values?
Applicable in all places.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.stories.ts
```diff
@@ -0,0 +1,311 @@
...
+        setDataToday([...filteredData]);
+      } else {
+        const filteredData = dataPrevious.filter(
+          (obj) => obj.id !== notificationId
+        );
+        setDataPrevious([...filteredData]);
+      }
+    };
+    const clickOutside = (event) => {
+      setOpenPanel(false);
```
amal-k-joy, line 152: do we need to close from here as it already closed on outside click?

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.stories.ts
```diff
@@ -0,0 +1,311 @@
...
+        @c4p-notification-click-outside=${clickOutside}
+      >
+        ${dataToday.length > 0
+          ? html`
+              ${dataToday.map((item) => {
+                return html`
+                  <c4p-notification
+                    slot="today"
+                    .open=${args.open}
+                    @c4p-notification-click=${item.onNotificationClick}
```
amal-k-joy: do we need to have custom event for click? I think it should by default emit `click` event
` @click=${item.onNotificationClick}`

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
```diff
@@ -0,0 +1,372 @@
...
+            <slot name="previous"></slot>
+          </div>
+          <div class="${blockClass}__bottom-actions-container">
+            <slot name="footer"></slot>
+          </div>
+        </div>
+      </div>
+    `;
+  }
+  private _markFirstNotification() {
```
amal-k-joy, line 234: Why do we need to add `first-notification` ? Can't user easily select via css selectors?

packages/ibm-products-web-components/src/components/notification-panel/notification.scss
```diff
@@ -0,0 +1,142 @@
...
+:host(#{$prefix}-notification) {
+  position: relative;
+  display: flex;
+  align-items: flex-start;
+  padding: $spacing-05;
+  border: 0;
+  background-color: $layer-01;
+  cursor: pointer;
+  inline-size: 100%;
+  min-block-size: 6.25rem;
```
amal-k-joy, line 36: replace with carbon tokens

packages/ibm-products-web-components/src/components/notification-panel/notification.ts
```diff
@@ -0,0 +1,233 @@
...
+        size="sm"
+        class="${blockClass}__dismiss-single-button"
+        @click=${dismissSingleNotification}
+      >
+        ${Close16({ slot: 'icon' })}
+      </cds-button>
+    `;
+  }
+
+  @HostListener('document:keydown')
```
amal-k-joy: Do we need to listen to document keydown ?

amal-k-joy (changes requested):

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
sangeethababu9223: Hey @amal-k-joy,
We could change it to listen to panel only, but shouldn't it listen to the document if we want to have the same behaviour as clickoutside for pressing `Esc` key too?

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.scss
sangeethababu9223: For `block-size` and  `min-block-size` values are directly used instead of tokens in the React stylesheet.
I have used the same here too.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
sangeethababu9223: No.  `:first-of-type`/`:first-child` doesn't work because of the whole the whole light DOM/ shadow DOM issue. 
One thing I could try to do is move this style to storybook. 

This is one limitation I was talking about when we use slots.

packages/ibm-products-web-components/src/components/notification-panel/notification.scss
sangeethababu9223: In the `ibm-products-styles/scss/components/NotificationsPanel/_notifications-panel.scss`, the value for `min-block-size` is given directly without tokens. That is why the same is added here too.

packages/ibm-products-web-components/src/components/notification-panel/notification.ts
sangeethababu9223: I'll update this.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
sangeethababu9223: This checks if an actionable element was clicked and if so it doesn't return focus to trigger item. But closes panel.
It still needs some more changes to handle Core carbon's actionable elements.

With the event handler, we don't need to remove the listener, it is happening under the hood with `@HostListener` and `HostListenerMixin`.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
amal-k-joy: I think esc should work since the panel is open. Can you check.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
sangeethababu9223: Removed the extra div here.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.mdx
sangeethababu9223: It is safe for us to keep `:slotted()`. This gives a very basic styling. But it has the least precedence. So if a user styles it from their end in light DOM, this will get overwritten.

packages/ibm-products-web-components/src/components/notification-panel/notification.ts
sangeethababu9223: Done

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.stories.ts
sangeethababu9223: You're right. I had added tis since custom elements doesn't take pressing `enter` key as the same as click event. 
This is handled separately now. And we can use just `@click` on the component instead of a custom event.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
sangeethababu9223: I have slightly modified the implementation so that `keydown` is detected in the component only.
This means that `keydown` won't be detected outside of the component.
But it also matches the React version, So I hope it should be okay.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
elycheea: I’m not certain we need to use a heading level at all here. I think a `<span>` or `<p>` would be perfectly acceptable for a notification panel. Even the notification components in Carbon use `<div>`s instead of heading, and I expect these should be comparable in semantic structure. :)

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.scss
sangeethababu9223: I've created an issue in Cor for this: [#19642](https://github.com/carbon-design-system/carbon/issues/19642)

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.stories.ts
sangeethababu9223: We can't directly change arg in stories from the `clickOutside` function. I tried with `updateArgs` and `context.updateArgs`, but it doesn't work.
So I had to use the state option provided by storybook. I'm updating the `openPanel` state here as it is mapped to the `open` property in notification panel in the story.

 <c4p-notification-panel
        .triggerButtonRef=${triggerButton}
        .open="${openPanel}"
        title-text="${args.titleText}" `
....
</c4p-notification-panel>

If I don't update it, then after user closes the panel with `clickOutside`, they will have to click on trigger button twice for it to open. This is because the `openPanel` needs to be updated for the story to realise the value of `open` is changed.
So I've to update the `openPanel` state.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
sangeethababu9223: I've updated this to include core carbon's actionable elements too.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.ts
sangeethababu9223: I've removed the `cds-heading` and used `h2`, `h3` and `h4` tags. And it does take away the needs to overwrite a lot of styling.
One catch is there is one area where were using `cds-heading` in slot. And for that we were using css variables to overwrite the styles. Since this is a custom component, when we add value for these css variable, the h1/h2/h3 element would pick the value from there. Hence we can use `:slotted` in our component to style these sections. But when we are using h1/h2/h3/div tags, light DOM styling automatically has the precedence.
Hence we have to style from storybook styles. And the user will have to do the same from their end.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.scss
amal-k-joy: @sangeethababu9223 , Do we need to keep that -ve margin , since it will automatically fix when core issue is fixed?

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.scss
amal-k-joy: @sangeethababu9223 
But since we have token for 2.5rem : $spacing-08 , we could easily replace it.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.scss
sangeethababu9223: I can remove that. But the layout would look a bit off until the issue is fixed. Should we wait till core fixes this and then remove the negative margin?

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.scss
sangeethababu9223: Done

packages/ibm-products-web-components/src/components/notification-panel/_story-assets/index.ts
elycheea, line 1: Are we using this file? Or was this accidentally left in?

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.mdx
```diff
@@ -0,0 +1,108 @@
...
+### JS (via import)
+
+```javascript
+import '@carbon/ibm-products-web-components/es/components/notification-panel/index.js';
+```
+
+### HTML
+
+```html
+      <c4p-notification-panel
```
elycheea: We can probably reduce these indentations a little bit?

First line gets moved back in the docs page, but the rest look a little too far indented.

image

packages/ibm-products-web-components/src/components/notification-panel/NotificationsPanel_data.js
```diff
@@ -0,0 +1,229 @@
...
+  },
+];
+
+export const extraData = {
+  id: uuidv4(),
+  type: 'error',
+  title: 'Cluster unreachable',
+  description:
+    'Not able to establish connection with provided cluster. Please check your logs and memory allocation to resolve this issue further.',
+  timestamp: dayBeforeYesterday,
```
elycheea: We should use the current time for this since I didn’t notice the new notifications getting added until I realized they were getting added to the very end of the notifications. These should ideally update from the top of the list. 

Are we also going to add a badge indicator when new notifications are added?

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.stories.ts
```diff
@@ -0,0 +1,310 @@
...
+            </div>
+          </cds-header-panel>
+          <cds-header-global-action
+            aria-label="Notification"
+            tooltip-text="Notification"
+            id="trigger-button"
+            @click="${toggleButton}"
+          >
+            ${Notification20({ slot: 'icon' })}
+          </cds-header-global-action>
```
elycheea: Just a part of the stories, but we should also include the unread notification example here. I’m fine if we want to open that as a followup issue though.

elycheea (commented):
Some minor things but did we want to add a little documentation about the default footer component versus using a custom one? It may be apparent to people more familiar with Web Components but could be helpful for people newer to it (like me!).

packages/ibm-products-web-components/src/components/notification-panel/_story-assets/index.ts
sangeethababu9223: Thank you for catching that. It is not being used. I've removed it now.

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.mdx
sangeethababu9223: Done

sangeethababu9223:
> Some minor things but did we want to add a little documentation about the default footer component versus using a custom one? It may be apparent to people more familiar with Web Components but could be helpful for people newer to it (like me!).

Hey @elycheea,
I've updated the doc section with details on the slots. It would definitely make it more understandable to adopters.
Thank you for pointing this out!

packages/ibm-products-web-components/src/components/notification-panel/notification-panel.stories.ts
sangeethababu9223: Hey @elycheea,
I've updated the story with the unread notifications. Every new notification added, by default will be marked as unread. 
I was a bit unclear on the behaviour initially since nothing was visually different for unread notifications in React. But when I looked further into it, I could see that this unread status is being overridden in story even before user has seen it.
When I disabled this, what I could see is the `unread` status is reset once the panel is closed. I've implemented the same here in WC version too. 
One issue is that `:slotted()` styling of the title gets overridden by light DOM basic styles. So I had to add the unread style to storybook stylesheet too. (So user will also have to do the same)

packages/ibm-products-web-components/src/components/notification-panel/NotificationsPanel_data.js
sangeethababu9223: Hey @elycheea,
I've updated the logic to add new notifications to the top instead. 
Also has added an option to show a different icon when a new unread notification is available.

elycheea (approved):
Your doc updates look great! Rest looks good! Thanks for the quick changes.

amal-k-joy (approved):

PR #7576, opened 2025-05-27
fix: options-tile toggle button

### LINKED ISSUE #7544: [a11y] OptionsTile (wc): expand/collapse click event should occur on an interactive element
### Package

@carbon/ibm-products

### Browser

Chrome

### Operating System

MacOS

### Package version

Latest web component package

### React version

N/A

IBM Equal Access Checker

### Assistive technology

_No response_

### Description

The expand/collapse behavior is not accessible via keyboard, the click event should happen on an interactive element.

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://ibm-products-web-components.carbondesignsystem.com/?path=/story/experimental-optionstile--default

### Steps to reproduce

See [story](https://ibm-products-web-components.carbondesignsystem.com/?path=/story/experimental-optionstile--default) for options tile

### Code of Conduct

davidmenendez:
Closes #7544 

converts the chevron icon to an icon button


matthewgallo (commented):
Looks like you need to update your tests now that the click event has moved to the button.

davidmenendez:
@matthewgallo thanks for pointing that out. fixt!

elycheea (commented):
Should this not use `details` and `summary` rather than a single interactive icon?

In the React version, the entire header/summary tile is selectable/interactive. I would argue that this difference would still be a parity issue although I don’t remember if there was an intentional decision _not_ to use `details`/`summary` for this version?

davidmenendez:
> Should this not use `details` and `summary` rather than a single interactive icon?
> 
> In the React version, the entire header/summary tile is selectable/interactive. I would argue that this difference would still be a parity issue although I don’t remember if there was an intentional decision _not_ to use `details`/`summary` for this version?

@elycheea it looks like there's still some issue with this implementation. namely the problem i've been concerned about for a while, which is the button click within the header click interaction. [this article](https://www.scottohara.me/blog/2022/09/12/details-summary.html), which i've referenced in the past when looking into this specific issue, makes mention of this issue in the "summary can have nested interactive elements and headings" section.

the article also makes some mention of accessibility related issues with styling the widget to remove the default triangle indicator.

out of curiosity i created a simple version of this with a button inside a summary element to demonstrate keyboard accessibility. i was able to tab between the two elements normally, but i found i was unable to accomplish this same task with voice-over controls. in fact, voice-over wouldn't even acknowledge the button's existence. it simple read the text content of the entire summary element with no indicator that there was a button child.

i thought this was something we had discussed during the planning phase, but i can't actually find any evidence to support that haha. perhaps it was discussed elsewhere and i'm drawing a blank on when / where that was. we should set up some time to discuss how we want to address these concerns with accessibility and parity.

davidmenendez:
based on our conversation with the accessibility guild today, it looks like my concerns will have to be looked at and addressed, but in an effort to enforce parity between the web components and the react components i went ahead and made the changes that @elycheea suggested. any fixes or enhancements that need to be done to address the accessibility concerns will have to be done to both anyways, so i think it should be ok to move forward with these changes for now and then open a separate issue to address the accessibility changes in both places when the fix for that has been determined.

devadula-nandan (approved):
LGTM

matthewgallo (approved):

PR #7600, opened 2025-06-02
feat(tearsheet): add decorator option

### LINKED ISSUE #7331: Add decorator prop to Tearsheet

sangeethababu9223:
Closes #7331 

Add decorator prop to Tearsheet

#### What did you change?
Added support for `decorator` slot. This is then mapped to slot `ai-label` in Modal component.
Added decorator control in storybook 

Storybook

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products-web-components/src/components/tearsheet/tearsheet.scss
+++ b/packages/ibm-products-web-components/src/components/tearsheet/tearsheet.scss
@@ -102,8 +102,11 @@ $motion-duration: $duration-moderate-02;
   .#{$block-class}__influencer[wide] {
     @extend .#{$block-class}__influencer--wide;
   }
+  ::slotted([slot='decorator']),
   ::slotted(#{$carbon-prefix}-slug) {
+    position: absolute;
     display: flex;
+    inset-block-start: to-rem(10px);
     inset-inline-end: 0;
     /* stylelint-disable-next-line carbon/layout-use */
     margin-block: 6px;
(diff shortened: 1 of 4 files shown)


matthewgallo (approved):
I noticed that storybook displays the `type` as `number` when we configure the controls with dropdown items. We may need to update the `argTypes` in order to avoid this. It was already present in other props so don't think we need to block this PR because of it.

image

packages/ibm-products-web-components/src/components/tearsheet/utils.ts
```diff
@@ -295,3 +297,40 @@ export const getSlug = (index) => {
       return null;
   }
 };
+// cspell: disable
+export const getDecorator = (decorator) => {
+  switch (decorator) {
+    case 1:
```
davidmenendez: can we change the `decorator` value to something slightly more descriptive? i know this is just for storybook and testing, but if i'm looking at this i'm not sure what the significance of the value 1 or 2 is. maybe just use constants to define these like `const NO_DECORATOR` or something

sangeethababu9223: Hey @davidmenendez,
This is updated.

davidmenendez (dismissed):

packages/ibm-products-web-components/src/components/tearsheet/tearsheet.scss
```diff
@@ -255,11 +265,13 @@ $motion-duration: $duration-moderate-02;
   }
 
   .#{$block-class}__header[has-close-icon],
-  .#{$block-class}__header[has-slug] {
+  .#{$block-class}__header[has-slug],
+  #{$block-class}__header[has-decorator] {
```
matthewgallo: I think this is meant to be a class selector?
```suggestion
  .#{$block-class}__header[has-decorator] {
```

sangeethababu9223: Yes. Thank you for catching that!
I've updated the style.

amal-k-joy (approved):
LGTM

PR #7608, opened 2025-06-02
fix: canary release workflow

### LINKED ISSUE #6729: Canary releases fail after `lerna` upgrade from `8.0.2` to `8.1.9`
### Package

Carbon for IBM Products

### Description

Our canary releases now fail after the `lerna` upgrade. Support for provenance publishing seems to have possibly changed between these two versions. No other release types are affected, just canary.

### Component(s) impacted

N/A

### Browser

_No response_

Canary versions

### Suggested Severity

None

### Product/offering

N/A

N/A

Look at failing canary release actions, ie https://github.com/carbon-design-system/ibm-products/actions/runs/12813203513/job/35726721766

_No response_

### Code of Conduct

davidmenendez:
Closes #6729

the last several [canary release failures ](https://github.com/carbon-design-system/ibm-products/actions/workflows/release-canary.yml)show the following in the publish log

```
lerna WARN notice Package failed to publish: @carbon/ibm-products-styles
lerna ERR! E422 Error verifying sigstore provenance bundle: Failed to validate the provenance subject against the package name, version and tarball integrity. Please try re-publishing with the latest stable version of npm
```

unfortunately i'm unable to find anything with this error message specifically. right now the only thing i'm attempting to do is up update the node version used in the canary publish workflow from 20 to 24

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
-          node-version: '20.x'
+          node-version: '22.x'
           registry-url: https://registry.npmjs.org
           cache: 'yarn'
 
@@ -47,7 +47,5 @@ jobs:
 
       # Publish release with canary tag to NPM using Lerna
       - name: Publish canary release of packages
-        env:
-          NPM_CONFIG_PROVENANCE: false
         run: |
           yarn lerna publish --canary --preid canary --dist-tag canary --yes
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
-          node-version: '22.x'
+          node-version: '24.x'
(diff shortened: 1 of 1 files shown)


davidmenendez:
@matthewgallo when you get a chance can you take a look and let me know what you think or any thoughts you have regarding this? the log in the original issue is no longer available, so i don't know if the error message was different.

the only direct reference i can find to part of this error message is in this [workflow file](https://github.com/webdriverio/webdriverio/blob/eafdc059bd009ffdbbfe5a789e9096765aa93e58/.github/workflows/publish.yml#L24) which suggests disabling provenance

matthewgallo (approved):
We certainly can try with the updates node version. I also stumbled upon that same issue you linked. If it still doesn't work, we can just reopen #6729. I'm really not sure what's going on either, I was thinking maybe there could be a new `lerna` version we could use since this broke after upgrading that package previously. But there haven't been a lot of changes coming from `lerna` so I don't think that would help.

davidmenendez:
> We certainly can try with the updated node version. I also stumbled upon that same issue you linked. If it still doesn't work, we can just reopen #6729. I'm really not sure what's going on either, I was thinking maybe there could be a new `lerna` version we could use since this broke after upgrading that package previously. But there haven't been a lot of changes coming from `lerna` so I don't think that would help.

i was able to find reference in lerna's github to being able to support parity, but that was introduced in like v6 or something 2 years ago haha. hopefully updating node help and if it doesn't it might just be an issue i can open up with lerna directly.

PR #7640, opened 2025-06-09
fix: use carbon prefix directly from `@carbon/styles` package config

### LINKED ISSUE #7639: Custom Carbon prefix (`cds`): update sass files
Sass files should reference `config.$prefix` rather than using our own defined `$carbon-prefix` from within our own package, otherwise custom prefixing with Carbon won't entirely work as any styles referencing `@carbon/react` component classes within our components, will still be using `cds` rather than the custom prefix.

### Acceptance criteria

- [ ] Replace any usage of `c4p-settings.$carbon-prefix` with `config.$prefix` from the carbon styles config

For example:
```scss
@use '@carbon/styles/scss/config';

.#{$block-class} .#{config.$prefix} {
  ...
  ...
}
```

matthewgallo:
Closes #7639 

Our current approach of directly referencing `c4p-settings.$carbon-prefix` causes our library to not support custom prefixing with Carbon. If someone attempts to add custom prefixing with base Carbon, the custom prefix won't be applied to any of the Carbon components that we reference directly within our component styles.

Instead, we should be accessing the carbon styles config module directly so that if the prefix is being passed a custom value, we're also using it.

#### What did you change?
Any sass file in our project that uses `c4p-settings.$carbon-prefix`
Manually added a custom Carbon prefix to our storybook
`packages/ibm-products/.storybook/preview.js`
Screenshot 2025-06-09 at 11 02 30 AM
`packages/ibm-products/.storybook/index.scss`
Screenshot 2025-06-09 at 11 02 19 AM

#### PR Checklist

- [ ] Wrote passing tests that cover this change, will investigate the best way to test for this

our contributing docs.

### DIFF
--- a/packages/ibm-products-styles/src/components/APIKeyModal/_api-key-modal.scss
+++ b/packages/ibm-products-styles/src/components/APIKeyModal/_api-key-modal.scss
@@ -1,5 +1,5 @@
 //
-// Copyright IBM Corp. 2021, 2022
+// Copyright IBM Corp. 2021, 2025
 //
 // This source code is licensed under the Apache-2.0 license found in the
 // LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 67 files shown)


makafsal (approved):
LGTM 🚀

makafsal:
Moved the changes to this [PR](https://github.com/carbon-design-system/ibm-products/pull/8904)

PR #6552, opened 2024-12-05
chore(deps): update to react 19

matthewgallo:
Closes #

{{short description}}

PR #7681, opened 2025-06-16
chore: init update to react 19

### LINKED ISSUE #7057: Update to React 19

davidmenendez:
Closes #7057


elycheea:
@davidmenendez any reason why we opened a new PR instead of resolving the conflicts from #6552? It shouldn’t be too outdated since Matt and Amal have both tried to keep it somewhat in sync.

davidmenendez:
> @davidmenendez any reason why we opened a new PR instead of resolving the conflicts from #6552? It shouldn’t be too outdated since Matt and Amal have both tried to keep it somewhat in sync.

since that PR was 2 months old i figured it would be easier to start fresh, especially considering certain large changes like the storybook config. but if it's preferable to continue there i can totally just close this and continue working there 👍

elycheea:
@davidmenendez No, it might actually work out better to have a fresh PR, especially if you’ve been able to incorporate a lot of the changes from Matt and Amal’s original PRs. I would watch out for the Storybook PR coming up though.

davidmenendez:
the CI failures aren't even related to the web-component package, so please ignore those.

matthewgallo (changes requested):
All of the changes within `examples` need to pin `react-is` to v19 via resolutions. I pulled this down locally and noticed that the version of `react-is` for the examples is v16 which won't be compatible with `react@19`.

matthewgallo:
Also, `NavItemLink` and `RowSizeRadioGroup` both use `defaultProps`, which is no longer supported in React 19.

davidmenendez:
i think this should now address everything in the [breaking changes section of the migration guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#new-deprecations) 👍

matthewgallo (dismissed):
Nice work! 🎉  AVT/VRT are failing but #7732 should address that. After that is merged, we can merge `main` into this branch and check it again

amal-k-joy:
@davidmenendez , Looks like there is an issue with Condition builder functionality. Can we hold this some time when i figure this out. It appears that the Popover component is triggering an extra onRequestClose callback compared to before, which is causing subsequent popovers to not stay open.

davidmenendez:
> @davidmenendez , Looks like there is an issue with Condition builder functionality. Can we hold this some time when i figure this out. It appears that the Popover component is triggering an extra onRequestClose callback compared to before, which is causing subsequent popovers to not stay open.

interesting. ok we can follow up with that in a separate issue 👍

e2e/components/Tearsheet/Tearsheet-test.avt.e2e.js
elycheea, line 1: Can we double check if we still need to comment these out?

davidmenendez:
@elycheea @matthewgallo was able to fix those tearsheet e2e tests just fyi 👍

amal-k-joy:
> > @davidmenendez , Looks like there is an issue with Condition builder functionality. Can we hold this some time when i figure this out. It appears that the Popover component is triggering an extra onRequestClose callback compared to before, which is causing subsequent popovers to not stay open.
> 
> interesting. ok we can follow up with that in a separate issue 👍

I have opened an [issue](https://github.com/carbon-design-system/ibm-products/issues/7747) and raised PR to address this. We merge it after this

davidmenendez:
ok awesome. thanks @amal-k-joy

PR #7707, opened 2025-06-19
fix(notification panel): clickoutside return focus to trigger button

### LINKED ISSUE #7377: Notification Panel: When clicked on Interactive elements outside Panel the focus returns to Trigger button
Notification Panel has a functionality that returns the focus to the trigger button when user clicks outside the panel. This is acceptable when the user is not clicking on any other interactive elements like a button or link. But it was identified that even when user clicks on an Interactive elements, the focus is being returned to the trigger button instead of staying on the clicked item.

For the time being this behaviour is disabled in [PR](https://github.com/carbon-design-system/ibm-products/pull/7344)  for https://github.com/carbon-design-system/ibm-products/issues/6902 issue.

But we need to find alternative solutions. 

As suggested by @elycheea/ @szinta, one possible solution would to separately handle clicks on Interactive elements while allowing focus return to trigger button for click on anything outside of the Panel that is non interactive.

Or as suggested by @matthewgallo / @makafsal, we could let user handle the behaviour with`onClickOutside ` function available. 

**Acceptance criteria for Handling focus return in the Component itself**
 - [x] Modify the click-outside handler logic to check if the target element for the click event is an actionable item like a `button`, `link`, `input` etc..., If not, return focus to the trigger button. Otherwise keep the focus on the actionable element itself.
 - [x] Add avt tests for the interaction

Something like this:
<pre>
      useClickOutside(ref || notificationPanelRef, (target) => {
      const element = target as HTMLElement;
      const isActionable =
        element instanceof HTMLButtonElement ||
        element instanceof HTMLAnchorElement ||
        element instanceof HTMLInputElement ||
        element instanceof HTMLSelectElement ||
        element instanceof HTMLTextAreaElement ||
        element.tabIndex >= 0 ;

      if (!isActionable) {
        setTimeout(() => {
          triggerButtonRef?.current?.focus();
        }, 100);
      }
      onClickOutside?.();
    });
</pre>

sangeethababu9223:
Closes #7377

When clicked outside the notification panel, the focus should return to trigger button only if the clicked element is not an actionable element like a button, input, link etc.

#### What did you change?

- Modified the `useClickOutside` hook callback to check if the clicked element is an actionable element and focus return to trigger based on that.
- Added a function to check if the clicked element is actionable.
- Updated Unit test for the change.
- updated avt test to match the change.

Storybook

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.tsx
+++ b/packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.tsx
@@ -409,7 +409,13 @@ export let NotificationsPanel = React.forwardRef(
       setAllNotifications(data);
     }, [data]);
 
-    useClickOutside(ref || notificationPanelRef, () => {
+    useClickOutside(ref || notificationPanelRef, (target) => {
+      const element = target as HTMLElement;
+      if (!isActionableElement(element)) {
+        setTimeout(() => {
+          triggerButtonRef?.current?.focus();
+        }, 100);
+      }
       onClickOutside?.();
     });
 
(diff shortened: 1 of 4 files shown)


davidmenendez (approved):

packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.tsx
```diff
@@ -683,6 +689,45 @@ export let NotificationsPanel = React.forwardRef(
       );
     };
 
+    const isActionableElement = (el: HTMLElement | null): boolean => {
```
davidmenendez, line 692: i really like what you've done here. i think there's utility for this outside of `NotificationsPanel`. we should add it as a util after this gets merged in. it'll also make it easier to expand on the testing 🎉

sangeethababu9223: Thanks @davidmenendez. We can do that.

devadula-nandan (approved):
🚀

PR #7479, opened 2025-05-09
feat(openStackblitz): about modal and export Modal

anamikaanu96:
Closes #7175

Implemented open stackblitz in About Modal and Export Modal

#### What did you change? 
Created 
packages/ibm-products/src/components/AboutModal/AboutModal.mdx and 
packages/ibm-products/src/components/ExportModal/ExportModal.mdx

Updated
packages/ibm-products/src/components/AboutModal/AboutModal.stories.jsx
packages/ibm-products/src/components/ExportModal/ExportModal.stories.jsx

#### How did you test and verify your work? yarn storybook and open stackblitz (Without the changes to codePreviewer.tsx from [7365](https://github.com/carbon-design-system/ibm-products/pull/7365), this PR cannot be tested in the deployment preview.)

**Note: This PR should only be merged after [7365](https://github.com/carbon-design-system/ibm-products/pull/7365)**
Therefore, marking it as a draft until [7365](https://github.com/carbon-design-system/ibm-products/pull/7365) is merged.

PR #7712, opened 2025-06-19
fix(AboutModal): update stories to import local assets

matthewgallo:
Fixes broken images in about modal story

#### What did you change?
Update about modal story to import tech logos from local story assets directory.
Storybook
#### PR Checklist

our contributing docs.


anamikaanu96 (commented):
I think it was working when I changed to hostedUrls in the [PR](https://github.com/carbon-design-system/ibm-products/pull/7479) 🤔

Now its working as expected in Storybook, but not in StackBlitz due to missing asset imports.

To ensure it works in both environments, as mentioned earlier, we could collaborate with a designer to optimize and reduce the file sizes. For example, the Grafana and Ansible logos exceed 70KB, which is likely why they are served via hosted URLs instead of being inlined as Base64.

Base64 is not causing any issue for other two logo's.

https://github.com/carbon-design-system/ibm-products/pull/7479#issuecomment-2900132033

makafsal:
> I think it was working when I changed to hostedUrls in the [PR](https://github.com/carbon-design-system/ibm-products/pull/7479) 🤔
> 
> Now its working as expected in Storybook, but not in StackBlitz due to missing asset imports.
> 
> To ensure its works in both environments, as mentioned earlier, we could collaborate with a designer to optimize and reduce the file sizes. For example, the Grafana and Ansible logos exceed 70KB, which is likely why they are served via hosted URLs instead of being inlined as Base64.
> 
> Base64 is not causing any issue for other two logo's.
> 
> [#7479 (comment)](https://github.com/carbon-design-system/ibm-products/pull/7479#issuecomment-2900132033)

So we don't need this change anymore? Can we close this PR?

makafsal:
These changes are already covered in this [PR](https://github.com/carbon-design-system/ibm-products/pull/7479). Hence, closing this PR.

Issue #7700, opened 2025-06-18
`createOverflowHandler` should account for padding/margins and maintain item order


matthewgallo:
To maintain more predictability in it's output, the `createOverflowHandler` should account for padding and margins. It should also maintain the order of items as the order is important some cases, particularly breadcrumb order.

matthewgallo:
This is required for overflow behavior within the new `PageHeader`.

PR #7720, opened 2025-06-23
feat: wc add overflow patterns

### LINKED ISSUE #7637: Add all the patterns that can be built through overflowHandler from @carbon/utilities
- [x] Action bar
  - [x] base story
  - [ ] ~story with fixed action ??~
- [x] Tagset
  - [x] with normal overflow popover
  - [x] with many tags (overflow popover and modal)
  - [x] with close (dismissable tags)
- [x] User avatars
  - [x] user avatars overflow with popover
  - [x] condensed useravatars (overlaying) ??
- [ ] #7776
- [x] breadcrumbs

devadula-nandan:
Closes #7637 

overflow patterns initial commit

#### PR Checklist

our contributing docs.


packages/ibm-products-web-components/examples/set-of-breadcrumbs/src/overflowHandler.ts
devadula-nandan, line 1: @matthewgallo this file needs to be removed once we get your pr [#7700](https://github.com/carbon-design-system/ibm-products/issues/7700) in utilities merged. or do we remove this in this pr. and time it to be merged when the utility is updated?

davidmenendez (changes requested):
looks like something is slightly off calculation because the last item extends beyond the parent container 🤔 

Screenshot 2025-07-07 at 6 34 34 PM
Screenshot 2025-07-07 at 6 34 16 PM

devadula-nandan:
> looks like something is slightly off calculation because the last item extends beyond the parent container 🤔

Hi David, this happens when the count of actions increased through controls, and the handler isn't instantiated at the right time after the data updates internally. i have added few tweaks that should fix this. however set of actions are supposed to be fixed items, which should properly render with the overflow behavior in the first render without any issues.

BTW
in an ideal use case for tagset and user-avatars, i could imagine an extreme usage where the application queries a database that might have thousands of items. and its not recommended to render too many items in the dom even if they are hidden through display none by the handler they would still be in the dom anyways and would lead to performance issues. the application would be most efficient / responsive, if it renders a safe capped amount of items in the dom, which can be handled by the utility and the dom effectively. and render the remaining in the modal if there are toomany. this could be achieved in implementation side. do u think this must be mentioned in docs? if so the wording for it?

davidmenendez:
@devadula-nandan that all makes sense. i definitely don't think it's worth introducing any performance issues if this is just a storybook related issue. i think we can move forward 👍

davidmenendez (approved):
LGTM! great work 👏

matthewgallo (approved):

PR #7732, opened 2025-06-25
fix(avt): update notification panel avt story urls and fix percy config issue

matthewgallo:
Updates urls for two avt tests from `NotificationsPanel`, these are causing avt tests to fail after the storybook IA changes. Also noticed that in our `percy.yml` config file, we were still only including snapshots within `IBM Products` directory, which doesn't exist after the storybook IA changes (this is what is causing the `vrt-runner` to fail). I've updated it to include snapshots for `Components`, `Patterns`, and `Utilities`.

#### What did you change?
```
e2e/components/NotificationsPanel/NotificationsPanel-test.avt.e2e.js
```
Will verify in this PR
#### PR Checklist

our contributing docs.

### DIFF
--- a/.percy.yml
+++ b/.percy.yml
@@ -13,7 +13,7 @@ storybook:
   # if storybook is still loading, percy could take a snapshot of the storybook
   # loading spinner and cause flaky tests
   waitForSelector: '.story-wrapper'
-  include: ['/IBM Products/']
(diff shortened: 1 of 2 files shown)


davidmenendez (approved):

elycheea (approved):
Omg. Good catches @matthewgallo. I’m surprised these didn’t get caught in the original PR when I tested. 🙈

PR #7742, opened 2025-06-27
fix: for checkbox labels in the custom columns

### LINKED ISSUE #7725: [Customize columns in Data grid regression]: Tests failing after upgrade to v2.68.0
### Package

Carbon for IBM Products

### Description

Our unit tests have started to fail after picking up `"@carbon/ibm-products": "2.68.0"`. The point they are failing at is when we are using react testing library to look at one of the checkboxes to see if it is selected or not. RTL is unable to find the checkbox. 

Using this repos storybook as an example, in the following screen shot the tests can find the checkbox called "Column name" using code like:
```
      const selectAllCheckBox = screen.getByRole('checkbox', {
        name: 'Column name'
      })
```
however it would be unable to find rest of the rows (e.g. it fails to find "Row Index") using:
```
      const rowIndexCheckBox = screen.getByRole('checkbox', {
        name: 'Row Index'
      })
```

Image

This was working on `v2.67.0` and previous versions.

### Component(s) impacted

Customize columns in Data grid. Storybook link: https://ibm-products.carbondesignsystem.com/?path=/story/deprecated-datagrid-datagrid-columncustomization--column-customization-usage-story&globals=viewport:basic

### Browser

_No response_

v2.68.0

### Suggested Severity

None

### Product/offering

IBM App Connect

https://ibm-products.carbondesignsystem.com/?path=/story/deprecated-datagrid-datagrid-columncustomization--column-customization-usage-story&globals=viewport:basic

I suspect that extending this test https://github.com/carbon-design-system/ibm-products/blob/fb2e4646a90c302cc21775cc525e9dcf86ae57f6/packages/ibm-products/src/components/Datagrid/Datagrid.test.js#L2156-L2158 to also include a check for the rows being displayed e.g. 
```
      const rowIndexCheckBox = screen.getByRole('checkbox', {
        name: 'Row Index'
      })
```
would demonstrate the issue

_No response_

### Code of Conduct

jlongshore:
Closes #7725 (potentially)

**UPDATE: 6/27/2025**
Removing highlighted text due to potential a11y issues. Will revisit that at another time. 
------

Current functionality is breaking apart the column titles and placing them alongside empty strong tags in version 2.68.0... 
image

Whereas in version 2.67.0, the entire string was not broken up  
image

There was a [change](https://github.com/carbon-design-system/ibm-products/commit/06cb10fba9ce6e19cc8058a73eb4e2d428ca1b29#diff-0e74900bc4c0c0d0d8ae5de6d68ba1798aaee9e15068b68127dc4dd387851d8b) made to the DraggableItemsList.js file

#### What did you change?
Passed the entire column header to the `labelText` prop of the Checkbox component if the "filterString" is empty, otherwise pass the "highlightedText"

Developer console, inspecting the element visually

This screenshot shows no filter text being applied, and the empty strong tags and individual characters are not visible.
image

This screenshot shows a filter text applied, and the strong tags wrapped around the matching text only in a more logical way.
image

#### PR Checklist
(Deprecated component)

~- [ ] Updated documentation and storybook examples~
~- [ ] Wrote passing tests that cover this change~
~- [ ] Addressed any impact on accessibility (a11y)~
~- [ ] Tested for cross-browser consistency~

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/Datagrid/Datagrid/addons/CustomizeColumns/DraggableItemsList.js
+++ b/packages/ibm-products/src/components/Datagrid/Datagrid/addons/CustomizeColumns/DraggableItemsList.js
@@ -209,6 +209,16 @@ export const DraggableItemsList = ({
         >
           {visibleCols.map((colDef) => {
             const colHeaderTitle = getNodeTextContent(colDef.Header);
+            // const parts = colHeaderTitle.split(
+            //   new RegExp(`(${filterString})`, 'gi')
+            // );
+            // const highlightedText = parts.map((part) =>
+            //   part.toLowerCase() === filterString.toLowerCase() ? (
+            //     <strong>{part}</strong>
+            //   ) : (
+            //     part
+            //   )
+            // );
             const isFrozenColumn = !!colDef.sticky;
             const isDisabled = colDef.disabled;
 
@@ -218,7 +228,7 @@ export const DraggableItemsList = ({
                 disabled={isDisabled || isFrozenColumn}
                 onChange={(_, { checked }) => onSelectColumn(colDef, checked)}
                 id={`${blockClass}__customization-column-${colDef.id}`}
-                labelText={colHeaderTitle}
+                labelText={colHeaderTitle} //filterString ? highlightedText : colHeaderTitle
                 className={`${blockClass}__customize-columns-checkbox`}
                 onKeyDown={(event) => handleCheckboxKeydown(event, colDef)}
               />


davidmenendez:
interesting issue here. @jlongshore i have some accessibility concerns with how these labels are being rendered in individual pieces in this current implementation. i think for now we should not be using `highlightedText` until we have a chance to evaluate accessibility here. comment it out and just use `colHeaderTitle`. i'll open a follow up issue once this is merged.

jonrober-80:
Thanks for addressing this. I saw the the description talk about it "potentially" being a fix. I think this could be recreated in an existing unit test (`Customizing Columns disable save button when un-select all columns`) with a couple of extra lines. 

You have this test - [https://github.com/carbon-design-system/ibm-products/blob/fb2e4646a90c302cc21775cc[…]/packages/ibm-products/src/components/Datagrid/Datagrid.test.js](https://github.com/carbon-design-system/ibm-products/blob/fb2e4646a90c302cc21775cc525e9dcf86ae57f6/packages/ibm-products/src/components/Datagrid/Datagrid.test.js#L2156-L2158)
but it's only testing for the option called "Column name" which seems like it has different properties to everything else (it acts like a 'select all' option).
I suspect that if you were to add a check after that one to look for:
```
      const rowIndexCheckBox = screen.getByRole('checkbox', {
        name: 'Row Index'
      })
```
then this would should that same issue that I was seeing when I raised the issue.

PR #7755, opened 2025-06-30
feat(TruncatedText): init TruncatedText react component

### LINKED ISSUE #7504: [Feature Request]: TruncatedText (React)
### The problem

Replacement for StringFormatter

### The solution

https://deploy-preview-7397--ibm-products-web-components.netlify.app/?path=/docs/experimental-utilities-truncatedtext--docs&globals=theme:g100

### Design link

_No response_

### Links to other materials

_No response_

### Owner/main maintainer(s)

IBM Products

_No response_

### Product/offering

_No response_

### Business priority

None

### Code of Conduct

### Acceptance criteria
- [x] Align with design on truncation functionality.
    - [ ] If design supports multiple ways to truncate, `TruncatedText` to support "with" prop to allow various ways ([see PR](https://deploy-preview-7397--ibm-products-web-components.netlify.app/?path=/docs/experimental-utilities-truncatedtext--docs&globals=theme:g100))
- [x] Replicate source code
- [x] Replicate styles
- [x] Replicate stories
- [x] Replicate testing

Should also follow up with #6901 to deprecate `StringFormatter` once merged

davidmenendez:
Closes #7504

### DIFF
--- a/packages/ibm-products-styles/src/components/TruncatedText/_carbon-imports.scss
+++ b/packages/ibm-products-styles/src/components/TruncatedText/_carbon-imports.scss
@@ -4,7 +4,3 @@
 // This source code is licensed under the Apache-2.0 license found in the
 // LICENSE file in the root directory of this source tree.
 //
-.truncated-text-stories__viewport {
-  inline-size: 100%;
-  max-inline-size: 300px;
-}
--- a/packages/ibm-products-styles/src/components/TruncatedText/_index-with-carbon.scss
+++ b/packages/ibm-products-styles/src/components/TruncatedText/_index-with-carbon.scss
@@ -4,7 +4,6 @@
 // This source code is licensed under the Apache-2.0 license found in the
 // LICENSE file in the root directory of this source tree.
 //
-.truncated-text-stories__viewport {
-  inline-size: 100%;
-  max-inline-size: 300px;
-}
+
+@use './carbon-imports';
+@use './truncated-text';
--- a/packages/ibm-products-styles/src/components/TruncatedText/_index.scss
+++ b/packages/ibm-products-styles/src/components/TruncatedText/_index.scss
@@ -4,7 +4,5 @@
 // This source code is licensed under the Apache-2.0 license found in the
 // LICENSE file in the root directory of this source tree.
 //
-.truncated-text-stories__viewport {
-  inline-size: 100%;
-  max-inline-size: 300px;
-}
+
+@use './truncated-text';
--- a/packages/ibm-products-styles/src/components/TruncatedText/_truncated-text.scss
+++ b/packages/ibm-products-styles/src/components/TruncatedText/_truncated-text.scss
@@ -1,0 +1,50 @@
+//
+// Copyright IBM Corp. 2025
+//
+// This source code is licensed under the Apache-2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+//
+
+@use '../../global/styles/project-settings' as c4p-settings;
+@use '../../global/styles/mixins';
+@use '@carbon/styles/scss/theme' as *;
+
+$block-class: #{c4p-settings.$pkg-prefix}--truncated-text;
+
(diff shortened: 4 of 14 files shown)


davidmenendez:
the only thing left here is to implement the animation, but i think we might just want to do that in a follow up. `line-clamp` suffers from a similar inability to animate from `none` to something. similar to `height:auto` to something with height or vice versa.

everything else is there though.

devadula-nandan (dismissed):
isn't the goal is to show the tooltip only if it truncates? this was the exact behavior previously from string formatter. and we had to make it dynamic and consistent across browsers for reusability. #7345 
reference slack [thread](https://ibm-studios.slack.com/archives/CQGR0HC05/p1744316174806799?thread_ts=1744315440.108449&cid=CQGR0HC05). 

https://github.com/user-attachments/assets/cb1a0492-f309-4e0a-afbc-ea4ba0960d23

the tooltip variant couldn't be navigated through keyboard with tab key if its truncating. which makes it inaccessible for keyboard users.

for the expand variant
have we decided if the view more or less buttons needs to be hidden for screen readers? as they being redundant for the text being fully accessible? 
if we try to hide them in screen readers with tab index -1, they are not keyboard accessible which is kind of dead lock situation.

should the buttons not render if the text is not truncating?

davidmenendez:
@devadula-nandan you're absolutely correct. thank you for pointing out that glaring omission. i'll get that fixed today.

davidmenendez:
@devadula-nandan `resizeObserver` has been added. the component should now react properly to `lines` and width

https://deploy-preview-7755--carbon-for-ibm-products.netlify.app/?path=/story/utilities-truncatedtext--with-expand&globals=viewport:basic

davidmenendez:
@devadula-nandan made some additional updates. should be ready for review now

amal-k-joy (approved):
LGTM 🎉

davidmenendez:
quick update: i meant to open the parity updates in a separate branch, but accidentally did it here, which i guess works out because the parity PR wouldn't be merged till the initial react version PR was merged anyways. so now it's a two in one.

anamikaanu96 (changes requested):
Looks good to me!🎉 

But few minor issues

- If we are not providing `align` prop, its not taking the default value `top`.

Screenshot 2025-07-17 at 3 14 52 PM

- Also instead of giving `align` type as `string` can we use `PopoverAlignment` from `@carbon/react`?

- It seems like auto align is not working properly.
Screenshot 2025-07-17 at 3 33 35 PM

davidmenendez:
@anamikaanu96 thanks for pointing out the alignment types. updated to use the carbon exported typing. i think the `autoAlign` is an implementation issue in the specific context of storybook because the tooltip is just really big. if you pop it out of storybook into the individual iframe it appears to work better.

makafsal (commented):
Although you've already implemented the React component, I'm not insisting on using `lit-react-wrapper`, but just curious why we don't use the [@lit/react](https://lit.dev/docs/frameworks/react/) wrapper to create the React component, as the truncated-text web component already exists?

davidmenendez:
> Although you've already implemented the React component, I'm not insisting on using `lit-react-wrapper`, but just curious why we don't use the [@lit/react](https://lit.dev/docs/frameworks/react/) wrapper to create the React component, as the truncated-text web component already exists?

@makafsal i'm definitely not opposed to it. i don't believe this is something we've really looked at in-depth. i know we've talked about it somewhat. i agree that i think it's something we should consider though. do you know if there's an issue to investigate this more in-depth? if it's something we can utilize effectively i think it's worth considering.

elycheea:
@makafsal @davidmenendez I would rather we start with the wrappers in our Labs experiences before fully investing in it for our core library. Let’s continue to explore it on style picker  and other new components with introduce through Labs. We want to understand how it performs and our adopters’ experience with wrapped components before we bring this practice over.

makafsal (approved):

PR #7791, opened 2025-07-03
feat: add support for  children prop for InterstitialScreen.Body

### LINKED ISSUE #7790: InterstitialScreen: support static children for InterstitialScreen.Body
Currently `InterstitialScreen.Body` accepts `contentRenderer` callback prop that will return the content to render in body. 
This enhancement will add support to children prop for `InterstitialScreen.Body `,if user want to go with some simple static content that doesn't need to access internal state and also offers more performance by reducing re-renders

amal-k-joy:
Closes #7790 

Currently InterstitialScreen.Body accepts contentRenderer callback prop that will return the content to render in body.
This enhancement will add support to children prop for InterstitialScreen.Body ,if user want to go with some simple static content that doesn't need to access internal state and also offers more performance by reducing re-renders

#### What did you change?
add support for children

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenBody.tsx
+++ b/packages/ibm-products/src/components/InterstitialScreen/InterstitialScreenBody.tsx
@@ -32,6 +32,10 @@ type contentRendererArgs = {
 };
 export interface InterstitialScreenBodyProps {
   /**
+   * Pass you static content as children
+   */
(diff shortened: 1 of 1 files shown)


matthewgallo:
Might be worth calling this out in the docs that you can use either `contentRenderer` or `children` but not both.

PR #7812, opened 2025-07-08
fix: percy

davidmenendez:
updates to percy dependencies


elycheea:
@davidmenendez Can you open a follow up issue for looking into the tests we skipped os we can investigate if it’s related to animation timing and remove these in a future sprint?

matthewgallo (approved):

davidmenendez:
@elycheea done and done 👍 https://github.com/carbon-design-system/ibm-products/issues/7826

Issue #7788, opened 2025-07-03
[Notification Panel]: Panel opens initially but closes instantly upon reopening in Storybook


anamikaanu96:
### Package

Carbon for IBM Products

### Description

Initially, the panel opens as expected, but after closing it, it is opening and suddenly closing.

### Component(s) impacted

Notification Panel

### Browser

Chrome

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

v2.70.0-rc.0

### Suggested Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Product/offering

Carbon for IBM Products

### CodeSandbox or Stackblitz example

https://ibm-products.carbondesignsystem.com/?path=/story/components-notificationspanel--default&globals=viewport:basic

### Steps to reproduce the issue (if applicable)

1. Close the notification Panel
2. Reopen

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [ ] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

PR #7830, opened 2025-07-10
fix(notification panel): clickoutside handler function exclude trigger button

### LINKED ISSUE #7828: [NotificationsPanel]: Resolve useClickOutside bug
[original issue](https://github.com/carbon-design-system/ibm-products/issues/7788) that was tied to temporarily removing the code.

the `useClickOutside` functionality added in `NotificationsPanel` contains a bug that prevents the panel from opening on open requests. it works initially, but fails afterwards. the code needs to be fixed and re-implemented.

acceptance criteria

- [ ] review the requirements in the [original issue](https://github.com/carbon-design-system/ibm-products/issues/7377)
- [ ] review the [original implementation details](https://github.com/carbon-design-system/ibm-products/pull/7707)
- [ ] un-comment out the [code](https://github.com/carbon-design-system/ibm-products/blob/release/v2.70.1/packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.tsx#L412) make necessary adjustments
- [ ] remove the `skip` flag from the tests in `NotificationsPanel.tests.js` and `NotificationsPanel-test.avt.e2e.js`
- [ ] verify the functionality is working properly with passing tests

sangeethababu9223:
resolves #7828 
Exclude trigger button from `clickoutside` function
Ref: #7788
Note: Some avt tests were failing and had to update the test to resolve those issues.


packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.tsx
```diff
@@ -409,15 +410,29 @@ export let NotificationsPanel = React.forwardRef(
...
-    //   onClickOutside?.();
-    // });
+    useEffect(() => {
+      const button = triggerButtonRef?.current;
+      const handlePointerDown = () => {
+        pointerInsideTrigger.current = true;
+      };
+      button?.addEventListener('pointerdown', handlePointerDown, true);
+      return () => {
+        button?.removeEventListener('pointerdown', handlePointerDown, true);
```
amal-k-joy: i think we use `click` event instead of `pointerdown`.
Also shall we rename `pointerInsideTrigger` to `isClickOnTrigger` ?

sangeethababu9223: Thanks @amal-k-joy, that makes sense.
I've updated this.

davidmenendez:
@sangeethababu9223 apologies as i didn't expect you to pick this up since you're on the new squad. i only originally tagged you to get your thoughts on the matter haha. if you don't have time for this please don't feel like you need to complete it!

amal-k-joy:
> i'm not sure we need the `isClickOnTrigger` ref. could an implementation like this work?
> 
> ```
>   useClickOutside(ref || notificationPanelRef, (target) => {
>     const trigger = triggerButtonRef?.current;
>     if (trigger) { // check for a trigger
>       if (trigger.contains(target)) { // avoid firing the trigger twice
>         return;
>       }
>       trigger.focus();
>     }
>     onClickOutside?.();
>   });
> 
>   const handleKeydown = (event) => {
>     if (event.key === 'Escape') {
>       onClickOutside?.();
>       triggerButtonRef?.current?.focus();
>     }
>   };
> ```

@davidmenendez ,
This was the intended solution initially, but it didn’t work as expected. The issue is that event.target refers to the SVG icon inside the trigger button. Since the icon can change dynamically (e.g., switching to a different SVG in this case), the condition becomes unreliable and fails when the icon changes.

davidmenendez:
@amal-k-joy thank you for the explanation. that makes sense.

amal-k-joy (approved):
LGTM

sangeethababu9223:
> @sangeethababu9223 apologies as i didn't expect you to pick this up since you're on the new squad. i only originally tagged you to get your thoughts on the matter haha. if you don't have time for this please don't feel like you need to complete it!

That's completely fine @davidmenendez. Since I worked on this one, I figured it'd be easier to just fix it myself! 😄

PR #7832, opened 2025-07-10
fix: layout breakage due to extra nodes added to Tearsheet

wkeese:
Refs #3856, #7730.

The extra nodes added in #7730 break our Tearsheet layout.  As an example, our edit tearsheet now looks like this:

edit mde

I guess it breaks stuff like when the top node of the content has `height: 100%`.   That kind of CSS stopped working due to the extra node.

While it may be possible for existing applications to adjust their own CSS to avoid this problem, it would be better if existing applications just continued to work.   In other words, this PR aims to minimize the breakage from #7730.

#### What did you change?

Don't add extra nodes unless necessary, and when they are necessary, style them with `display: contents`.

Tested locally against our application.  It's working again (or at least better than before).

#### PR Checklist

our contributing docs.


davidmenendez (approved):
Thanks Bill. Appreciate your assistance on this!

wkeese:
As I side note (as I wrote to @szinta on Slack), I think these `<Wrap>` helpers are counterproductive.  For example, the current code is:

```tsx
<Wrap
  className={cx({
    [`${bc}__influencer`]: true,
    [`${bc}__influencer--wide`]: influencerWidth === 'wide',
  })}
  neverRender={influencerPosition === 'right'}
  element={SectionLevel3}
>
  <Wrap element={Layer} className={`${bc}__layer`}>
    {influencer}
  </Wrap>
</Wrap>
```

But IIUC it could be rewritten as:

```tsx
{influencer && influencerPosition === "left" ? (<SectionLevel3
  className={cx({
    [`${bc}__influencer`]: true,
    [`${bc}__influencer--wide`]: influencerWidth === 'wide',
  })}
>
  <Layer className={`${bc}__layer`}>
    {influencer}
  </Layer>
</SectionLevel3>) : null}
```

The second one is clearer and shorter to me.

That's not the critical issue though, so I didn't change it in this PR.

Mikadv:
Can't wait for it :)

PR #7858, opened 2025-07-16
fix(PageHeader): add workaround for tab focus state with scroll buttons

### LINKED ISSUE #7816: [Bug] PageHeader default tabs overflow
### 1.  Default tabs layering
It was noted that within the [PageHeader component](https://react.carbondesignsystem.com/?path=/story/patterns-unstable-pageheader--default&globals=theme:g10), that the overflow scroll chevron for default tabs has the wrong background color. It should match the same color as the layer it sits on. In the case of page header, the background color should be layer-01. The default tabs themselves do not have a background color so it sits transparently on the layer it's placed on. 

Maybe it's inheriting the layer above since it's placed on it? Or maybe it's always set to background?

Header overflow tabs example:
Image

2. When tabs go into overflow mode, they loose their 1px gap between each tab and overlaps focus state by 1 px in the bottom right corner. They should retain the 1px gap.
Focus state:

Image

matthewgallo:
Closes #7816

Adds a small style workaround until this is addressed in `@carbon/react` for the focus state border on the tab getting clipped.

#### What did you change?
```
packages/ibm-products-styles/src/components/PageHeader/_page-header.scss
```
Storybook
#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products-styles/src/components/PageHeader/_page-header.scss
+++ b/packages/ibm-products-styles/src/components/PageHeader/_page-header.scss
@@ -917,6 +917,10 @@ $duration: 1000ms;
     }
   }
 
+  .#{$carbon-prefix}--tabs .#{$carbon-prefix}--tabs__nav-link {
+    margin-inline-end: $spacing-01;
+  }
+
   .#{$block-class}__tags {
     display: flex;
     align-items: center;


packages/ibm-products-styles/src/components/PageHeader/_page-header.scss
```diff
@@ -917,6 +917,10 @@ $duration: 1000ms;
     }
   }
 
+  .#{$carbon-prefix}--tabs .#{$carbon-prefix}--tabs__nav-link {
+    margin-inline-end: $spacing-01;
+  }
+
```
elycheea, line 926: If this is meant to be a temporary workaround, let’s add a comment linking to the [issue from Core](https://github.com/carbon-design-system/carbon/issues/19942).

PR #7859, opened 2025-07-16
chore: remove core package

### LINKED ISSUE #7604: Identify what remaining parts of `packages/core` can be moved into `packages/ibm-products/.storybook` and move them
Most of the contents of `packages/core` was strictly used for storybook previously. Now that our storybook config has moved into `packages/ibm-products` we should identify and move any remaining components still within `packages/core`.

After this work, we should be able to delete `packages/core` entirely as it will no longer be necessary. This may involve adding `lottie-web` as a dependency during our storybook build step since it's required for the welcome page animation but we _do not_ want to add it to our package.json in `packages/ibm-products`.

davidmenendez:
Closes #7604 

moves anything that is still being used from `packages/core`, moves it into `packages/ibm-products` and deletes `packages/core`


amal-k-joy (approved):
LGTM 🎉

makafsal (approved):
LGTM

matthewgallo (approved):

PR #7867, opened 2025-07-21
feat: creating webcomponent interstitialScreen intial commit

amal-k-joy:
Closes #

{{short description}}

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body-item.ts
+++ b/packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body-item.ts
@@ -1,0 +1,57 @@
+/**
+ * @license
+ *
+ * Copyright IBM Corp. 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { html, LitElement, PropertyValues } from 'lit';
+import { prefix } from '../../globals/settings';
+import HostListenerMixin from '@carbon/web-components/es/globals/mixins/host-listener.js';
+import { carbonElement as customElement } from '@carbon/web-components/es/globals/decorators/carbon-element.js';
+
+import styles from './interstitial-screen-body-item.scss?lit';
+import { property } from 'lit/decorators.js';
+import {
+  interstitialDetailsSignal,
+  updateInterstitialDetailsSignal,
+} from './interstitial-screen-context';
+
+@customElement(`${prefix}-interstitial-screen-body-item`)
+class CDSInterstitialScreenBodyItem extends HostListenerMixin(LitElement) {
+  @property({ reflect: true })
+  stepTitle: string = '';
+
+  protected firstUpdated(_changedProperties: PropertyValues): void {
+    const slot = this.shadowRoot?.querySelector('slot') as HTMLSlotElement;
+    slot.addEventListener('slotchange', this.updateStepDetails);
+
+    this.updateStepDetails();
+  }
+
+  private updateStepDetails() {
+    const randomId = Math.random().toString(36).slice(2);
+    const stepKey = `${this.stepTitle?.replace(/\s+/g, '') || randomId}`;
+    const newStep = {
+      stepTitle: this.stepTitle,
+      id: this.id ?? stepKey,
+    };
+
+    const exists = interstitialDetailsSignal
+      .get()
+      .stepDetails?.some((step) => step.stepTitle === newStep.stepTitle);
+
+    if (!exists && newStep.stepTitle) {
+      updateInterstitialDetailsSignal({ name: 'stepDetails', detail: newStep });
+    }
+  }
+
+  render() {
+    return html` <slot></slot> `;
+  }
+
+  static styles = styles;
+}
+export default CDSInterstitialScreenBodyItem;
--- a/packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body.scss
+++ b/packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body.scss
@@ -1,0 +1,30 @@
+/**
+ * @license
+ *
+ * Copyright IBM Corp. 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+@use './interstitial-screen';
+@use '../../utilities/carousel/index' as carousel;
+@use '@carbon/styles/scss/spacing' as *;
+
+@include carousel.wrapperStyles();
+
+$prefix: 'c4p';
+
+.#{$prefix}--interstitial-screen--body {
+  block-size: 100%;
+}
+.#{$prefix}--interstitial-screen__carousel,
+.#{$prefix}--interstitial-screen__contentWrapper {
+  display: flex;
+}
+
+:host {
+  display: flex;
+  flex-grow: 1;
+  /* stylelint-disable-next-line declaration-no-important */
+  padding-block-end: 0 !important;
+}
--- a/packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body.ts
+++ b/packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body.ts
@@ -1,0 +1,118 @@
+/**
+ * @license
+ *
+ * Copyright IBM Corp. 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { LitElement, html } from 'lit';
+import { state } from 'lit/decorators.js';
+import { prefix } from '../../globals/settings';
+import '@carbon/web-components/es/components/modal/index.js';
+import HostListenerMixin from '@carbon/web-components/es/globals/mixins/host-listener.js';
+import { carbonElement as customElement } from '@carbon/web-components/es/globals/decorators/carbon-element.js';
+import styles from './interstitial-screen-body.scss?lit';
+import { InitCarousel, initCarousel } from '../../utilities/carousel';
+import { ref, createRef } from 'lit/directives/ref.js';
+import {
+  interstitialDetailsSignal,
+  updateInterstitialDetailsSignal,
+} from './interstitial-screen-context';
+
+const blockClass = `${prefix}--interstitial-screen`;
+
+@customElement(`${prefix}-interstitial-screen-body`)
+class CDSInterstitialScreenBody extends HostListenerMixin(LitElement) {
+  @state()
+  stepType: 'single' | 'multi' | undefined = 'multi';
+
+  private carouselAPI!: InitCarousel;
+  private carouselElement = createRef<HTMLElement>();
+
+  firstUpdated(): void {
+    const slot = this.shadowRoot?.querySelector('slot') as HTMLSlotElement;
+
+    const updateStepMode = () => {
+      const assigned = slot.assignedElements({ flatten: true });
+
+      if (assigned.length === 1) {
+        this.stepType = 'single';
+      } else if (assigned.length > 1) {
+        this.stepType = 'multi';
+        //initialize carousel for multi-step
+        this._initCarousel();
+      }
+    };
+
+    // Run once after initial slot content
+    updateStepMode();
+  }
+
+  private _initCarousel() {
+    this.carouselAPI = initCarousel(this.carouselElement.value as HTMLElement, {
+      onViewChangeEnd: this.onViewChangeEnd,
+      excludeSwipeSupport: true,
+    });
+    interstitialDetailsSignal.set({
+      ...interstitialDetailsSignal.get(),
+      carouselAPI: this.carouselAPI,
+    });
+  }
+
+  private onViewChangeEnd = ({ currentIndex, lastIndex, totalViews }) => {
+    updateInterstitialDetailsSignal({
+      name: 'currentStep',
+      detail: currentIndex,
+    });
+
+    this.dispatchEvent(
+      new CustomEvent(
+        (
+          this.constructor as typeof CDSInterstitialScreenBody
+        ).eventOnViewChangeEnd,
+        {
+          bubbles: true,
+          cancelable: true,
+          composed: true,
+          detail: {
+            currentStep: currentIndex,
+            totalStepCount: totalViews,
+            lastStep: lastIndex,
+          },
+        }
+      )
+    );
+  };
+
+  render() {
+    return html`
+      <div class="${blockClass}--body">
+        ${this.stepType === 'multi'
+          ? html`<div
+              ${ref(this.carouselElement)}
+              class="${blockClass}__carousel"
+            >
+              <slot></slot>
+            </div>`
+          : this.stepType === 'single'
+            ? html`<div class="${blockClass}__contentWrapper">
+                <slot></slot>
+              </div>`
+            : ''}
+      </div>
+    `;
+  }
+
+  static styles = styles;
+
+  /**
+   * The name of the custom event fired just after the action.
+   */
+  static get eventOnViewChangeEnd() {
+    return `${prefix}-on-after-action`;
+  }
+}
+
+export default CDSInterstitialScreenBody;
--- a/packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.scss
+++ b/packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.scss
@@ -1,0 +1,32 @@
+/**
+ * @license
+ *
+ * Copyright IBM Corp. 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+@use './interstitial-screen';
+@use '@carbon/styles/scss/spacing' as *;
+
+$prefix: 'c4p';
+$carbon-prefix: 'cds';
+
+.#{$prefix}--interstitial-screen--footer-controls {
+  display: flex;
+}
+#{$carbon-prefix}-button {
+  block-size: 100%;
+}
+#{$carbon-prefix}-inline-loading {
+  position: absolute;
+  inset-block-start: 0;
+}
+.#{$prefix}--interstitial-screen--skip-btn {
+  /* stylelint-disable-next-line declaration-no-important */
+  padding-inline-start: 0 !important;
+
+  &::part(button) {
+    padding-inline-start: $spacing-07;
+  }
+}
--- a/packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.ts
+++ b/packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.ts
@@ -1,0 +1,256 @@
+/**
+ * @license
+ *
+ * Copyright IBM Corp. 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { html, LitElement, nothing, PropertyValues } from 'lit';
+import { property, state } from 'lit/decorators.js';
+import { prefix } from '../../globals/settings';
+import HostListenerMixin from '@carbon/web-components/es/globals/mixins/host-listener.js';
+import { carbonElement as customElement } from '@carbon/web-components/es/globals/decorators/carbon-element.js';
+import '@carbon/web-components/es/components/button/index.js';
+import styles from './interstitial-screen-footer.scss?lit';
+import { interstitialDetailsSignal } from './interstitial-screen-context';
+import { SignalWatcher } from '@lit-labs/signals';
+import '@carbon/web-components/es/components/inline-loading/inline-loading.js';
+import { CDSModalFooter } from '@carbon/web-components/es/index.js';
+import ArrowRight from '@carbon/web-components/es/icons/arrow--right/16.js';
+
+const blockClass = `${prefix}--interstitial-screen`;
+const headerBlockClass = `${blockClass}--internal-header`;
+
+export type ActionType = 'close' | 'start' | 'skip' | 'back' | 'next';
+
+@customElement(`${prefix}-interstitial-screen-footer`)
+class CDSInterstitialScreenFooter extends SignalWatcher(
+  HostListenerMixin(CDSModalFooter)
+) {
+  @property({ reflect: true })
+  nextButtonLabel: string = 'Next';
+
+  @property({ reflect: true })
+  previousButtonLabel: string = 'Back';
+
+  @property({ reflect: true })
+  skipButtonLabel: string = 'Skip';
+
+  @property({ reflect: true })
+  startButtonLabel = 'Get Started';
+
+  @property({ type: Boolean, reflect: true, attribute: 'async-action' })
+  asyncAction;
+
+  @state()
+  loadingAction;
+
+  private _handleUserInitiatedClose(
+    triggeredBy: EventTarget | null | ActionType
+  ) {
+    const init = {
+      bubbles: true,
+      cancelable: true,
+      composed: true,
+      detail: {
+        triggeredBy,
+      },
+    };
+    if (
+      this.dispatchEvent(
+        new CustomEvent(
+          (
+            this.constructor as typeof CDSInterstitialScreenFooter
+          ).eventBeforeClose,
+          init
+        )
+      )
+    ) {
+      this.dispatchEvent(
+        new CustomEvent(
+          (this.constructor as typeof CDSInterstitialScreenFooter).eventClose,
+          init
+        )
+      );
+    }
+  }
+
+  private handleSkip = () => this.handleAction('skip');
+  private handleStart = () => this.handleAction('start');
+  private handleClickNext = () => this.handleAction('next');
+  private handleClickPrev = () => this.handleAction('back');
+
+  private handleAction = async (actionType: ActionType) => {
+    this.loadingAction = actionType;
+
+    const currentStep = interstitialDetailsSignal.get().currentStep;
+    const stepCount = interstitialDetailsSignal.get().stepDetails.length;
+
+    let resolvePromise;
+    const proceedPromise = new Promise<boolean>((resolve) => {
+      resolvePromise = resolve;
+    });
+
+    const customEvent = new CustomEvent(
+      (
+        this.constructor as typeof CDSInterstitialScreenFooter
+      ).eventOnBeforeAction,
+      {
+        bubbles: true,
+        composed: true,
+        cancelable: true,
+        detail: {
+          step: currentStep,
+          stepCount: stepCount,
+          proceed: (allow: boolean | Promise<boolean>) => {
+            Promise.resolve(allow).then(resolvePromise);
+          },
+        },
+      }
+    );
+
+    const eventNotCanceled = this.dispatchEvent(customEvent);
+
+    if (!eventNotCanceled) {
+      return;
+    }
+
+    const canProceed = this.asyncAction ? await proceedPromise : true;
+
+    this.loadingAction = '';
+
+    if (canProceed) {
+      const carouselAPI = interstitialDetailsSignal.get()?.carouselAPI;
+      if (actionType == 'next') {
+        carouselAPI.next();
+      } else if (actionType === 'back') {
+        carouselAPI.prev();
+      } else {
+        this._handleUserInitiatedClose?.(actionType);
+      }
+    }
+  };
+
+  getFooterContent() {
+    const stepDetails = interstitialDetailsSignal.get().stepDetails;
+    const currentStep = interstitialDetailsSignal.get().currentStep;
+    const isMulti = stepDetails?.length > 0;
+    const progStepCeil = stepDetails?.length - 1;
+    return html`
+      <div class="${blockClass}--footer">
+        ${isMulti
+          ? html`
+              <cds-button
+                class="${blockClass}--skip-btn"
+                kind="ghost"
+                size="lg"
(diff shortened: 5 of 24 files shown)


packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body-item.ts
```diff
@@ -0,0 +1,64 @@
...
+@customElement(`${prefix}-interstitial-screen-body-item`)
+class CDSInterstitialScreenBodyItem extends HostListenerMixin(LitElement) {
+  /**
+   * This will serve the labels for each step
+   */
+  @property({ reflect: true })
+  stepTitle: string = '';
+
+  protected firstUpdated(_changedProperties: PropertyValues): void {
+    const slot = this.shadowRoot?.querySelector('slot') as HTMLSlotElement;
```
makafsal: We can use `@query()` decorator instead of `querySelector()`.

[Reference](https://lit.dev/docs/components/shadow-dom/#@query-@queryall-and-@queryasync-decorators)

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body-item.ts
```diff
@@ -0,0 +1,64 @@
...
+    slot.addEventListener('slotchange', this.updateStepDetails);
```
makafsal: Here, we can use the `@slotchange` event. Attach the `@slotchange` event to the `slot` like this:

```js
  <slot @slotchange=${this.updateStepDetails}></slot>
```

[Reference](https://lit.dev/docs/components/shadow-dom/#accessing-slotted-children)

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body-item.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    this.updateStepDetails();
+  }
+
+  private updateStepDetails() {
+    const randomId = Math.random().toString(36).slice(2);
```
makafsal: > Just a suggestion

Why don't we use `crypto`? It is a built-in API and cleaner than this.

```js
  crypto.randomUUID()
```

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body-item.ts
```diff
@@ -0,0 +1,64 @@
...
+      .get()
+      .stepDetails?.some((step) => step.stepTitle === newStep.stepTitle);
+
+    if (!exists && newStep.stepTitle) {
+      updateInterstitialDetailsSignal({ name: 'stepDetails', detail: newStep });
+    }
+  }
+
+  render() {
+    return html` <slot></slot> `;
```
makafsal: This will become like this:

```sh
  <slot @slotchange=${this.updateStepDetails}></slot>
```

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body.scss
```diff
@@ -0,0 +1,30 @@
...
+.#{$prefix}--interstitial-screen__carousel,
+.#{$prefix}--interstitial-screen__contentWrapper {
+  display: flex;
+}
+
+:host {
+  display: flex;
+  flex-grow: 1;
+  /* stylelint-disable-next-line declaration-no-important */
+  padding-block-end: 0 !important;
```
makafsal, line 29: > A suggestion

If possible, avoid the use of `!important`.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body.ts
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * interstitial-screen-body for body section
+ * @element c4p-interstitial-screen-body
+ * @fires c4p-on-before-step-change - The name of the custom event fired at the start of  the step change.
+ * @fires c4p-on-after-step-change -  The name of the custom event fired at the end of  the step change.
+ */
+@customElement(`${prefix}-interstitial-screen-body`)
+class CDSInterstitialScreenBody extends HostListenerMixin(LitElement) {
+  @state()
+  stepType: 'single' | 'multi' | undefined = 'multi';
```
makafsal: Do we need `undefined` here?

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  private carouselAPI!: InitCarousel;
+  private carouselElement = createRef<HTMLElement>();
+
+  firstUpdated(): void {
+    const slot = this.shadowRoot?.querySelector('slot') as HTMLSlotElement;
```
makafsal, line 41: Use `@query()`

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body.ts
```diff
@@ -0,0 +1,150 @@
...
+      const assigned = slot.assignedElements({ flatten: true });
+
+      if (assigned.length === 1) {
+        this.stepType = 'single';
+      } else if (assigned.length > 1) {
+        this.stepType = 'multi';
+        //initialize carousel for multi-step
+        this._initCarousel();
+      }
+    };
```
makafsal: I think we can either remove this from a function or move this function declaration out of the `firstUpdated` lifecycle method.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body.ts
```diff
@@ -0,0 +1,150 @@
...
+              ${ref(this.carouselElement)}
+              class="${blockClass}__carousel"
+            >
+              <slot></slot>
+            </div>`
+          : this.stepType === 'single'
+            ? html`<div class="${blockClass}__contentWrapper">
+                <slot></slot>
+              </div>`
+            : ''}
```
makafsal: If we remove `undefined` from `stepType`, we can avoid chaining to `''`.

```js
  ${this.stepType === 'multi' ? html` ... ` : html` ... `}
```

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.scss
```diff
@@ -0,0 +1,32 @@
...
+#{$carbon-prefix}-button {
+  block-size: 100%;
+}
+#{$carbon-prefix}-inline-loading {
+  position: absolute;
+  inset-block-start: 0;
+}
+.#{$prefix}--interstitial-screen--skip-btn {
+  /* stylelint-disable-next-line declaration-no-important */
+  padding-inline-start: 0 !important;
```
makafsal, line 35: Try to avoid `!important`.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.ts
```diff
@@ -0,0 +1,260 @@
...
+  private handleSkip = () => this.handleAction('skip');
+  private handleStart = () => this.handleAction('start');
+  private handleClickNext = () => this.handleAction('next');
+  private handleClickPrev = () => this.handleAction('back');
+
+  private handleAction = async (actionType: ActionType) => {
+    this.loadingAction = actionType;
+
+    const currentStep = interstitialDetailsSignal.get().currentStep;
+    const stepCount = interstitialDetailsSignal.get().stepDetails.length;
```
makafsal: > Suggestion

Can be refactored, so could avoid multiple `get()`.

```js
const { currentStep, stepDetails } = interstitialDetailsSignal?.get();
const stepCount = stepDetails?.length;
```

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.ts
```diff
@@ -0,0 +1,260 @@
...
+        carouselAPI.prev();
+      } else {
+        this._handleUserInitiatedClose?.(actionType);
+      }
+    }
+  };
+
+  getFooterContent() {
+    const stepDetails = interstitialDetailsSignal.get().stepDetails;
+    const currentStep = interstitialDetailsSignal.get().currentStep;
```
makafsal: Inside this function, getting values from `signal` can be refactored.

```js
const { stepDetails, currentStep, disableActions } = interstitialDetailsSignal.get();
const { start, next, back, skip } = disableActions;
```

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.ts
```diff
@@ -0,0 +1,260 @@
...
+        ${isMulti
+          ? html`
+              <cds-button
+                class="${blockClass}--skip-btn"
+                kind="ghost"
+                size="lg"
+                title="${this.skipButtonLabel}"
+                @click="${this.handleSkip}"
+                ?disabled="${interstitialDetailsSignal.get().disableActions
+                  ?.skip}"
```
makafsal: By refactoring the `.get()` method and spreading `disableActions`, we can change this to this:

```js
 ?disabled="${skip}"
```

More readability.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.ts
```diff
@@ -0,0 +1,260 @@
...
+          : ''}
+        <div class="${blockClass}--footer-controls">
+          ${isMulti && currentStep > 0
+            ? html`
+                <cds-button
+                  class="${blockClass}--prev-btn"
+                  kind="secondary"
+                  size="lg"
+                  title="${this.previousButtonLabel}"
+                  ?disabled="${interstitialDetailsSignal.get().disableActions
```
makafsal: Can change this to this:

```js
 ?disabled="${back}"
```

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.ts
```diff
@@ -0,0 +1,260 @@
...
+              `
+            : nothing}
+          ${isMulti && currentStep < progStepCeil
+            ? html`
+                <cds-button
+                  class="${blockClass}--next-btn"
+                  kind="primary"
+                  size="lg"
+                  title="${this.nextButtonLabel}"
+                  ?disabled="${interstitialDetailsSignal.get().disableActions
```
makafsal: Can change this to this:

```js
 ?disabled="${back}"
```

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.ts
```diff
@@ -0,0 +1,260 @@
...
+            : nothing}
+          ${(isMulti && currentStep === progStepCeil) || !isMulti
+            ? html`
+                <cds-button
+                  class="${blockClass}--start-btn"
+                  kind="primary"
+                  size="lg"
+                  title="${this.startButtonLabel}"
+                  ?disabled="${interstitialDetailsSignal.get().disableActions
+                    ?.start}"
```
makafsal:
```js
 ?disabled="${start}"
```

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-header.ts
```diff
@@ -0,0 +1,156 @@
...
+    const currentStep = interstitialDetailsSignal.get().currentStep;
+    if (index === currentStep) {
+      return 'current';
+    } else if (index < currentStep) {
+      return 'complete';
+    } else if (index > currentStep) {
+      return 'incomplete';
+    } else {
+      return 'invalid';
+    }
```
makafsal:
```suggestion
    
      return 'invalid';
    
```

Can avoid `else {}` branching.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-header.ts
```diff
@@ -0,0 +1,156 @@
...
+                            label="${eachStep.stepTitle}"
+                            key="${eachStep.id}"
+                            state="${this.getStepState(index)}"
+                          ></cds-progress-step>`
+                      )}
+                    </cds-progress-indicator>
+                  </div>
+                `
+              : html``
+          }
```
makafsal: Move to a separate function for more readability.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-header.ts
```diff
@@ -0,0 +1,156 @@
...
+    const stepDetails = interstitialDetailsSignal.get().stepDetails;
+
+    return html`
+      ${(this.headerTitle || this.headerSubTitle) &&
+      html`
+        <div class="${blockClass}--titleContainer">
+          ${this.headerTitle && html`<h1>${this.headerTitle}</h1>`}
+          ${this.headerSubTitle && html`<h2>${this.headerSubTitle}</h2>`}
+          </div>
+          <slot></slot>
```
makafsal: Move this to a separate function and use `if` to conditionally render, for better readability.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-header.ts
```diff
@@ -0,0 +1,156 @@
...
+                `
+              : html``
+          }
+        </div>
+      `}
+    `;
+  }
+
+  render() {
+    const isFullScreen = interstitialDetailsSignal.get().isFullScreen;
```
makafsal:
```suggestion
    const { isFullScreen } = interstitialDetailsSignal.get();
```

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.ts
```diff
@@ -0,0 +1,216 @@
...
+  @property({ type: Boolean, reflect: true, attribute: 'fullscreen' })
+  isFullScreen: boolean = false;
+  /**
+   * Specifies whether the component is currently open.
+   */
+  @property({ type: Boolean, reflect: true })
+  open: boolean = false;
+
+  @state()
+  stepDetails: Array<{ stepTitle: string; name?: string }> = [];
```
makafsal, line 62: Can we define a separate type?

```suggestion
interface StepDetail { stepTitle: string; name?: string }

  stepDetails: Array<StepDetail> = [];
```

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.addEventListener(`${prefix}-request-close`, this._handleClose);
+  }
+  firstUpdated() {
+    // This has to do since cds-modal does not accept cds-body inside a slotted children.It will append it explicitly append cds body
+    this.header = this.querySelector('c4p-interstitial-screen-header');
+    this.body = this.querySelector('c4p-interstitial-screen-body');
+    this.footer = this.querySelector('c4p-interstitial-screen-footer');
```
makafsal: Use `@query()` decorator instead.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.ts
```diff
@@ -0,0 +1,216 @@
...
+        this.dispatchInItializeEvent();
+      }
+
+      this._wasOpen = isOpen;
+    }
+  }
+
+  private dispatchInItializeEvent = () => {
+    setTimeout(() => {
+      const carouselAPI = interstitialDetailsSignal.get().carouselAPI;
```
makafsal:
```suggestion
      const { carouselAPI } = interstitialDetailsSignal.get();
```

packages/ibm-products-web-components/src/utilities/carousel/carousel.ts
```diff
@@ -0,0 +1,289 @@
...
+        }
+
+        const itemsWrapper = carouselContainer.querySelector(
+          `.${prefix}__itemsWrapper`
+        ) as HTMLElement;
+        if (itemsWrapper) {
+          itemsWrapper.style.blockSize = `${itemHeightSmallest}px`;
+        }
+      }
+    }, 10); // Delay to ensure layout is rendered before measuring
```
makafsal: An arbitrary delay in `setTimeout` may cause a race condition.

makafsal (changes requested):

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body-item.ts
amal-k-joy: This query is locally scoped to this function, so i think we can keep it as it is.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body.scss
amal-k-joy: We need this to avoid override from cds-modal styles
```
:host(cds-modal-body) ::slotted(:last-child) {
    padding-block-end: 2rem;
}
```

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body.ts
amal-k-joy: This query is locally scoped to this function, so i think we can keep it as it is.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.ts
amal-k-joy: querySelector wouldnt work here .
The header, body, and footer elements are light DOM children passed into <c4p-interstitial-screen>. But:@query() only works for elements inside the component's own Shadow DOM.

packages/ibm-products-web-components/src/utilities/carousel/carousel.ts
amal-k-joy: Modified the logic a bit

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.scss
amal-k-joy: I am afraid we cant , some overriders are required

makafsal (dismissed):
Awesome 🚀

packages/ibm-products-web-components/src/utilities/carousel/_index.scss
```diff
@@ -0,0 +1,105 @@
...
+  @include wrapperStyles;
+  @include viewStyles;
+}
+
+/// Combined for React
+@mixin carouselClassStyles {
+  @include keyframes;
+  @include wrapperStyles;
+  @include viewStyles;
+}
```
elycheea: Since these are currently the same, do we actually need two separate mixins? Or is this mostly so we have the flexibility later if we do need more custom styling per technology/framework? 🤔

packages/ibm-products-web-components/src/components/interstitial-screen/story-styles.scss
```diff
@@ -0,0 +1,20 @@
...
+$interstitial-block-class: #{c4p}--interstitial-screen;
+$block-class: #{c4p}--interstitial-screen-view-module;
+
+.#{$block-class}--enableTag {
+  margin-block-start: $spacing-07;
+}
+
+.c4p--interstitial-screen-view-module {
+  margin-block: $spacing-07;
+}
```
elycheea, line 22: Looks like we’re missing additional styles for the view module heading. I think this should be using `heading-05` based on the React version.

image

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.ts
```diff
@@ -0,0 +1,257 @@
...
+
+  getFooterContent() {
+    const { stepDetails, currentStep, disableActions } =
+      interstitialDetailsSignal.get();
+    const { start, next, back, skip } = disableActions;
+    const isMulti = stepDetails?.length > 0;
+    const progStepCeil = stepDetails?.length - 1;
+    return html`
+      <div class="${blockClass}--footer">
+        ${isMulti
```
davidmenendez, line 159: there's a lot of repetition here for settings the button props dynamically. could we possibly simplify this so we can avoid having so much conditional code in the render? maybe something along the lines of...

```
const getBtnProps = () => {
  const defaultProps = { size: 'lg' ... };
  if (condition1) {
    return { ...defaultProps, kind: condition1-kind }
  }
}
...
<cds-button {...btnProps} />
```

just some way to avoid having so much redundant code

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-header.ts
```diff
@@ -0,0 +1,164 @@
...
+      )
+    );
+  }
+
+  getProgressIndicatorContent(stepDetails) {
+    return html`
+      <div class="${blockClass}--progress">
+        <cds-progress-indicator>
+          ${stepDetails.map(
+            (eachStep, index) =>
```
davidmenendez: can we rename `eachStep` to just `step`?

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.ts
davidmenendez: i think we should use using `prefix` here, right?

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.ts
```diff
@@ -0,0 +1,216 @@
...
+        )
+      );
+    }
+  }
+
+  //template methods
+
+  renderFullScreen() {
+    return this.open
+      ? html`<div class="${blockClass} ${blockClass}--full-screen" role="main">
```
davidmenendez: i don't think we should be explicitly adding this role without the ability to a user to provide their own

davidmenendez (changes requested):
looks great! have some comments 👍

davidmenendez:
also, i don't think it needs to be done here, but i would love to see testing implemented for the swipe utility!

packages/ibm-products-web-components/src/utilities/carousel/_index.scss
```diff
@@ -0,0 +1,105 @@
+/**
+ * Copyright IBM Corp. 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+@use '@carbon/styles/scss/motion' as *;
+
+$prefix: 'carousel' !default;
+$animateTime: $duration-moderate-01;
```
elycheea: I wonder if we want to reduce the timing a little bit? The current animation feels faster than the React version’s.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-body.scss
elycheea: I too would love if we can avoid use of `!important` but I can see how shadow dom complicates it here. 🫠 Probably fine for now but should definitely keep investigate if there’s other ways we can increase specificity.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.stories.ts
```diff
@@ -0,0 +1,503 @@
...
+                  block-size: 100%;
+                  display: contents;
+                }
+              </style>
+
+              <div class="${blockClass}--footer">
+                <cds-button
+                  class="${blockClass}--skip-btn"
+                  kind="ghost"
+                  size="lg"
```
elycheea, line 367: I think these should all use the larger button size? It also might help to differentiate the text here like the React version does because at first I didn’t realize what made these buttons custom other than the size, which felt more like a bug.

```suggestion
                  size="xl"
```

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.stories.ts
```diff
@@ -0,0 +1,503 @@
...
+                          title="Back"
+                          @click=${handleClickPrev}
+                        >
+                          Back
+                        </cds-button>
+                      `
+                    : nothing}
+                  ${stepLength > 0 && currentStep < progStepCeil
+                    ? html`
+                        <cds-button
```
elycheea, line 389: image

The next button here is a bit shorter than typical. Can we add some story styles to make the sizing feel consistent?

elycheea (commented):
Mostly some minor details. The syntax and structure feels so foreign to me still. 😹 Especially `nothing` 🤯 

- Missing some story styles (non-blocking but nice to address).
- Some small suggestions for improving the `Custom action buttons` story.
- Avoiding `!important` seems trickier here. We might not have alternatives for now, but definitely want us to keep our eyes on it.
- Carousel timing seems a bit fast.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.ts
elycheea: I don’t mind this one in full screen, but maybe this isn’t as appropriate if these are also dismissible.

elycheea:
@davidmenendez I think that would be on `@carbon/utilities` since we’ve just copied/brought it over from what’s currently in review?

packages/ibm-products-web-components/src/utilities/carousel/_index.scss
amal-k-joy: You are right, initially it was separate mixins for both webcomponents and react and later I combined it and then I missed to update the output mixin.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.ts
amal-k-joy: I agree that avoiding redundancy is generally good, but in this case, `size` is the only common prop, and the rest need to be configured conditionally. I'm concerned that merging them might actually reduce readability, and we wouldn't gain much from it.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.ts
amal-k-joy: @elycheea , @davidmenendez 
 That’s a good observation. I was wondering — do we really need to support dismissing a fullscreen component? In most cases, the user can handle navigation or conditional rendering themselves. Also, createFullpage doesn’t include a dismiss feature either.

If we’re aligned on this, I can open an issue to remove the dismiss logic from both the React and Web Component implementations.

packages/ibm-products-web-components/src/utilities/carousel/_index.scss
amal-k-joy: Changed to `duration-moderate-02`

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.stories.ts
amal-k-joy: Custom button story styles was bit off, since footer was not rendering inside `cds-modal-footer`. Corrected it

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.ts
davidmenendez: oh wow i totally missed that full screen mode doesn't have a close icon / mechanism. actually if that's the case then yeah i think the role is fine, although i still don't think it's a bad idea to allow that to be configurable by the consumer if they so choose to change it.

and no, to your point: something like this should be left to the consumer to implement how they toggle the open and closed state.

davidmenendez:
> @davidmenendez I think that would be on `@carbon/utilities` since we’ve just copied/brought it over from what’s currently in review?

oh i didn't realize that. agreed 👍

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.ts
elycheea: Hmm, I don’t see the close icon in the [usage guidance](https://pages.github.ibm.com/carbon/ibm-products/components/onboarding/interstitial-screen/usage/) either. I think we can open an issue to remove the close icon, but might want to check with @bndege2 or @susanacr2 as well.

I still think we should remove the `main` role (or allow it to be configurable) though if this comes up as an overlay, it would be safer.

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen-footer.ts
davidmenendez: fair enough 👍

davidmenendez (approved):

elycheea:
@amal-k-joy I see you resolved and responded to some of my comments. Did you perhaps forget to push your changes? 😅

amal-k-joy:
> @amal-k-joy I see you resolved and responded to some of my comments. Did you perhaps forget to push your changes? 😅

@elycheea ,
my bad! i just realised, push has failed after commit.  Thanks for pointing this out

packages/ibm-products-web-components/src/components/interstitial-screen/interstitial-screen.ts
amal-k-joy: Thats sounds right
i have removed the role being set internally, now user can set the role on the host.

davidmenendez (approved):

PR #7869, opened 2025-07-21
Canary removal

matthewgallo:
Closes #

{{short description}}

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/.storybook/manager.js
+++ b/packages/ibm-products/.storybook/manager.js
@@ -44,15 +44,15 @@ addons.setConfig({
   theme: theme,
   sidebar: {
     renderLabel: (item) => {
-      const isUnstable = !!(
-        item.type === 'component' && checkCanaryStatus(item.name)
-      );
-      const dark = window.matchMedia('(prefers-color-scheme: dark)').matches;
-      if (isUnstable) {
-        // Canary tag is applied if `item.name` (`displayName` of the component) is confirmed
-        // to be canary/not yet released via the checkCanaryStatus utility
-        return renderComponentLabel(item.name, dark);
-      }
+      // const isUnstable = !!(
+      //   item.type === 'component' && checkCanaryStatus(item.name)
+      // );
+      // const dark = window.matchMedia('(prefers-color-scheme: dark)').matches;
+      // if (isUnstable) {
+      //   // Canary tag is applied if `item.name` (`displayName` of the component) is confirmed
+      //   // to be canary/not yet released via the checkCanaryStatus utility
+      //   return renderComponentLabel(item.name, dark);
+      // }
 
       return item.name;
     },
--- a/packages/ibm-products/.storybook/preview.js
+++ b/packages/ibm-products/.storybook/preview.js
@@ -129,7 +129,7 @@ const parameters = {
         'Patterns',
         ['*', 'Prebuilt patterns'],
         'Utilities',
-        'Experimental',
+        'Preview',
         ['*', 'FeatureFlags'],
         'Deprecated',
       ],
--- a/packages/ibm-products/src/components/APIKeyModal/APIKeyModal.tsx
+++ b/packages/ibm-products/src/components/APIKeyModal/APIKeyModal.tsx
@@ -50,7 +50,7 @@ const defaults = {
   customSteps: [],
 };
 
-export let APIKeyModal: React.FC<APIKeyModalProps> = forwardRef(
+export const APIKeyModal: React.FC<APIKeyModalProps> = forwardRef(
   (
     {
       // The component props, in alphabetical order (for consistency).
@@ -416,7 +416,6 @@ const downloadRequiredProps = (type) =>
   isRequiredIf(type, ({ hasDownloadLink }) => hasDownloadLink);
 
 // Return a placeholder if not released and not enabled by feature flag
-APIKeyModal = pkg.checkComponentEnabled(APIKeyModal, componentName);
 
 export const deprecatedProps = {
   /**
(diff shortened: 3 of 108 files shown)


packages/ibm-products/src/components/index.js
```diff
@@ -32,31 +45,78 @@ export {
...
+
+/**
+ * Non-stable components
+ * draft | previewCandidate | preview
+ */
+
+// Preview
+export { Coachmark as preview__Coachmark } from './Coachmark/next/Coachmark';
+export { DataSpreadsheet as preview__DataSpreadsheet } from './DataSpreadsheet';
+export { Decorator as preview__Decorator } from './Decorator';
```
elycheea: Let’s move these down to preview candidates.

DataSpreadsheet is on the hook for deprecation — preview would suggest it’s approaching stability. We might not want to encourage this so strongly here.

Decorator may want a composability review before we flag it as preview.

packages/ibm-products/src/components/index.js
```diff
@@ -32,31 +45,78 @@ export {
...
+  useFeatureFlag as preview__useFeatureFlag,
+  useFeatureFlags as preview__useFeatureFlags,
+} from './FeatureFlags';
+export * as preview__PageHeader from './PageHeader/next';
+export { SearchBar as preview__SearchBar } from './SearchBar';
+export {
+  Toolbar as preview__Toolbar,
+  ToolbarButton as preview__ToolbarButton,
+  ToolbarGroup as preview__ToolbarGroup,
+} from './Toolbar';
```
elycheea: We’re treating Toolbar as more of a pattern going forward with its potential to move to Core. We could recategorize this as preview candidate and add the React pattern code too.

I think we were uncertain if search bar would go to Core too so maybe keep it preview candidate as well.

packages/ibm-products/src/components/index.js
```diff
@@ -7,10 +7,21 @@
...
 export { Cascade } from './Cascade';
-export { ComboButton, ComboButtonItem } from './ComboButton';
+export { Checklist } from './Checklist';
+export { Coachmark } from './Coachmark';
+export { CoachmarkBeacon } from './CoachmarkBeacon';
+export { CoachmarkButton } from './CoachmarkButton';
+export { CoachmarkFixed } from './CoachmarkFixed';
+export { CoachmarkOverlayElement } from './CoachmarkOverlayElement';
+export { CoachmarkOverlayElements } from './CoachmarkOverlayElements';
+export { CoachmarkStack } from './CoachmarkStack';
```
elycheea: Don’t think Coachmark v1 should be under stable? I think we considered preview candidate, but I know the renames will create a bit of whiplash too.

packages/ibm-products/src/components/index.js
```diff
@@ -20,6 +31,8 @@ export {
   CreateTearsheetDivider,
 } from './CreateTearsheet';
 export { CreateTearsheetNarrow } from './CreateTearsheetNarrow';
+export { DelimitedList } from './DelimitedList';
```
elycheea: Let’s move DelimitedList to preview candidate. I think this one is flagged for deprecation?

packages/ibm-products/src/components/index.js
```diff
@@ -32,31 +45,78 @@ export {
...
 export { SidePanel } from './SidePanel';
 export { SingleAddSelect } from './SingleAddSelect';
 export { StatusIcon } from './StatusIcon';
+export { StringFormatter } from './StringFormatter';
+export { TagOverflow } from './TagOverflow';
 export { TagSet } from './TagSet';
 export { Tearsheet, TearsheetNarrow } from './Tearsheet';
-export { Toolbar, ToolbarButton, ToolbarGroup } from './Toolbar';
-export { UserProfileImage } from './UserProfileImage';
+export { TruncatedList } from './TruncatedList';
```
elycheea: Is this one not also flagged for potential deprecation or have we not decided on it?

packages/ibm-products/src/components/index.js
matthewgallo: I wasn't fully sure how we want to handle `Coachmark` since the new version is already exported as `preview__Coachmark`. It felt weird having the existing coachmark be preview candidate, but the new one is preview. But maybe it does make sense to have it ahead in terms of status because of the increased composability.

packages/ibm-products/src/components/index.js
matthewgallo: I confused this with truncated text in my head 😄

elycheea (approved):
LGTM! Thank you for taking this on! :shipit: 

Are we still timing this with Core’s renames?

matthewgallo:
Yes, it'd be great to still align on the renames across squads. Their [PR](https://github.com/carbon-design-system/carbon/pull/20376) has two approvals now. I have a [PR](https://github.com/carbon-design-system/carbon/pull/20377) to update the codemod to add import aliases to our components, this will limit the number of changes consumers have to make (no changes within jsx, just imports and all taken care of via the codemod). If we include the changes from this PR in our next release, it would be nice to get the codemod updates published to the `@carbon/upgrade` package.

davidmenendez (approved):

matthewgallo:
The [last commit](https://github.com/carbon-design-system/ibm-products/pull/7869/commits/5e900e7d458a7d10ccf703fbe0a7561607b902ad) updates the story files for preview candidates, so that the component is imported from the package index, meaning we import with the status prefix (just to ensure they're configured properly).

packages/ibm-products/.storybook/manager.js
```diff
@@ -44,15 +44,15 @@ addons.setConfig({
...
-      }
+      // const isUnstable = !!(
+      //   item.type === 'component' && checkCanaryStatus(item.name)
+      // );
+      // const dark = window.matchMedia('(prefers-color-scheme: dark)').matches;
+      // if (isUnstable) {
+      //   // Canary tag is applied if `item.name` (`displayName` of the component) is confirmed
+      //   // to be canary/not yet released via the checkCanaryStatus utility
+      //   return renderComponentLabel(item.name, dark);
+      // }
```
jlongshore: Should we remove these commented code lines?

PR #7908, opened 2025-07-28
feat(makeDraggable): utility added for drag functionality

### LINKED ISSUE #7909: Create a utility that enables HTML elements to be draggable.

anamikaanu96:
Closes #7909 

Implemented a utility which enables an HTML element to be draggable.

#### PR Checklist

our contributing docs.


makafsal (changes requested):
https://github.com/user-attachments/assets/72f8a2f4-1ab8-4eb4-b8a2-8a31398d73f4

Can you make the movement smoother like the mouse on the keyboard? Single pixel instead of 32px?

packages/ibm-products/src/global/js/utils/makeDraggable/makeDraggable.ts
```diff
@@ -0,0 +1,101 @@
...
+ * Makes a given element draggable using a handle element.
+ * @param el - The HTML element to move.
+ * @param handle - Optional HTML element to initiate the drag (e.g., header).
+ * @param focusableInHandle - Optional HTML element to focus on drag for keyboard interaction (e.g., Drag Icon).
+ */
+export function makeDraggable(
+  el: HTMLElement,
+  handle?: HTMLElement,
+  focusableInHandle?: HTMLElement
+): void {
```
makafsal: Instead of many parameters, a single `object` parameter with these values as props will give more configurability.

```suggestion
export function makeDraggable({
  el: HTMLElement,
  handle?: HTMLElement,
  focusableInHandle?: HTMLElement
}): void {
```

packages/ibm-products/src/global/js/utils/makeDraggable/makeDraggable.ts
```diff
@@ -0,0 +1,101 @@
...
+  ) {
+    el.style.position = 'relative';
+  }
+
+  let isDragging = false;
+  let offsetX = 0;
+  let offsetY = 0;
+
+  function onKeyDown(e: KeyboardEvent) {
+    const distance = e.shiftKey ? 128 : 32;
```
makafsal: I think a default value of `12px` will be smoother than `32px`.

Also, making this configurable will be more flexible.

Suggested prop name from @amal-k-joy: `dragStep`

anamikaanu96: I changed the default value of `dragStep` to `8px` and updated `shiftDragStep` from `128px` to `32px`.

davidmenendez (changes requested):
this looks great!

one thought about the implementation: this basically just looks like a hook. it's always used here in conjunction with `useEffect` so maybe it makes sense to refactor this to just be a hook instead `useDraggable` and then just pass in the various parameters

```
useDraggable({ el, handle, focusableInHandle });
```

was this something you considered previously?

anamikaanu96:
> one thought about the implementation: this basically just looks like a hook. it's always used here in conjunction with `useEffect` so maybe it makes sense to refactor this to just be a hook instead `useDraggable` and then just pass in the various parameters

Hey @davidmenendez, Thanks for reviewing!

It's currently implemented as a plain JS utility to keep it framework-agnostic, allowing it to be reused in both React and web components without being tied to React's lifecycle. I think it makes more sense to keep it as a utility rather than converting it into a hook.

davidmenendez (approved):
I agree with the decision to keep this framework agnostic. great job! i think this just needs some unit testing, but that can be done in a follow up issue & PR 👍

davidmenendez (dismissed):

makafsal (approved):
Overall, it's an impressive work 👏.

I've a few suggestions related to A11y, you can consider it as out of the scope of this PR.

1. We need a engage and disengage the state between drag. Something like in `Coachmark`, after focusing on the drag-handle user needs to press the `Enter` or `Space` key to engage in dragging, and after drag needs to press the `Enter` or `Space` key to confirm the drag position; otherwise press the `Esc` key to cancel the drag.

https://github.com/user-attachments/assets/ac57296f-9a32-4834-a110-98fd9560d6c1

2. When engaging, show it in the focus outline (like in `Coachmark`).
3. Also, provide appropriate `aria-roledescription` and `aria-label` for the dag-handle button. You can refer to the [dnd-kit example](https://master--5fc05e08a4a65d0021ae0bf2.chromatic.com/?path=/story/core-draggable-hooks-usedraggable--basic-setup).

PR #7913, opened 2025-07-29
feat(page-header): add breadcrumb overflow example

### LINKED ISSUE #7799: PageHeader: BreadcrumbBar - collapsible behavior
Implement collapsible behavior for BreadcrumbBar browser minimize

Image

**Breadcrumb**
- [x] breadcrumb area utilizes breadcrumb with overflow in smaller breakpoints
- [x] actions collapse into overflow menu

**Content actions**
- [x] ~Create sub component that can be used as a utility in rendering the content actions with overflowing behavior, similar to what has already been done with `PageHeader.ContentPageActions` which uses `createOverflowHandler` internally. This utility could also be used if implementing manually as well.~ We can actually just use `PageHeader.ContentPageActions` here and pass it to the `contentActions` prop of the `BreadcrumbBar`

matthewgallo:
Closes #7799 

Adds an example to demonstrate how to include overflowing behavior for the breadcrumbs within the breadcrumb bar of the page header. These changes are implementation only as I've included the example that @devadula-nandan built to show this behavior, not directly included within the page header component itself.

A local `overflowHandler.ts` util was added because the one from `@carbon/utilities` does not include padding/margin spacing in the over element size calculation. Until I submit a PR to update it within core, I've included a local version that includes those values.

The changes can be viewed from [this story](https://deploy-preview-7913--ibm-products-web-components.netlify.app/?path=/story/patterns-pageheader--tab-bar-with-tabs-and-tags).

#### What did you change?
```
packages/ibm-products-web-components/examples/set-of-breadcrumbs/index.html
packages/ibm-products-web-components/examples/set-of-breadcrumbs/package.json
packages/ibm-products-web-components/src/components/page-header/_story-assets/overflowHandler.ts
packages/ibm-products-web-components/src/components/page-header/_story-assets/set-of-breadcrumbs.scss
packages/ibm-products-web-components/src/components/page-header/_story-assets/set-of-breadcrumbs.ts
packages/ibm-products-web-components/src/components/page-header/page-header.scss
packages/ibm-products-web-components/src/components/page-header/page-header.stories.ts
```
Storybook
#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products-web-components/examples/set-of-breadcrumbs/index.html
+++ b/packages/ibm-products-web-components/examples/set-of-breadcrumbs/index.html
@@ -24,9 +24,9 @@ LICENSE file in the root directory of this source tree.
       </div>
     </section>
     <script type="module">
-      import { breadcrumbData } from './src/example-data.ts';
+      import { breadcrumbsData } from './src/example-data.ts';
       const setOfBreadcrumbs = document.getElementById('setOfBreadcrumbs');
-      setOfBreadcrumbs.breadcrumbData = breadcrumbData;
+      setOfBreadcrumbs.breadcrumbsData = breadcrumbsData;
     </script>
   </body>
 </html>
--- a/packages/ibm-products-web-components/examples/set-of-breadcrumbs/package.json
+++ b/packages/ibm-products-web-components/examples/set-of-breadcrumbs/package.json
@@ -20,8 +20,8 @@
   },
   "devDependencies": {
     "rimraf": "^3.0.2",
+    "typescript": "^5.5.3",
     "vite": "5.4.19",
-    "vite-plugin-lit-css": "^2.1.0",
-    "typescript": "^5.5.3"
+    "vite-plugin-lit-css": "^2.1.0"
   }
 }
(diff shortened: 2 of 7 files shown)


davidmenendez:
@matthewgallo when you go to the story you linked above the initial render looks like this

Screenshot 2025-07-29 at 8 57 45 AM

when you resize the window smaller and then back to default it looks like this

Screenshot 2025-07-29 at 8 57 54 AM

i'm assuming this behavior isn't intended, but please correct me if i'm wrong 😂

matthewgallo:
Ohh, yeah you're correct. That's not expected and looks to be Firefox related (Chrome and Safari seem fine), I'll check it out.

szinta (changes requested):
As per the story description, the Action list should collapse into a Overview menu, but it looks like that part is missed during implementation.

Also, I observed a couple of UI issues: (screenshot attached)
1. The overflow tooltip isn't visible.
2. Text overflow appears even when there's sufficient space to render the content.

image

Since this is an example, could we either add 3–4 breadcrumb items or use a long breadcrumb label as shown in the design to better demonstrate the overflow behavior when space is limited?

matthewgallo:
@szinta I made some changes so that the overflow tooltip is visible, added an extra breadcrumb, and increased the length of the title breadcrumb. I also included truncated text for the title breadcrumb as well.

Your second observation

> Text overflow appears even when there's sufficient space to render the content.

This is intended so that when you scroll there is space to display the title breadcrumb when it animates in. We can definitely chat with Tracey about it to confirm that's the expected interaction.

matthewgallo:
I was also going to look at the action list and tag collapsing in a separate PR to try to keep the PRs smaller.

PR #7837, opened 2025-07-14
feat(Coachmark): added coachmarkv2 tooltip variant

anamikaanu96:
Closes #7773 

Implemented Composable Coachmark's Tooltip variant using Beacon and Bubble

#### PR Checklist

As the author of this PR, before marking ready for review, confirm you:

- [x] Reviewed every line of the diff
- [ ] Updated documentation and storybook examples
- [ ] Wrote passing tests that cover this change
- [ ] Addressed any impact on accessibility (a11y)
- [ ] Tested for cross-browser consistency
- [x] Validated that this code is ready for review and status checks should pass

More details can be found in the [pull request](./CONTRIBUTING.md) section of
our contributing docs.

PR #7924, opened 2025-07-29
feat(coachmarkv2): added floating variant

### LINKED ISSUE #7843: Implement Coachmark Floating Variant
Will be taking after [Create a DraggableDialog component in Carbon Labs](https://github.com/carbon-design-system/ibm-products/issues/6927)

anamikaanu96:
Closes #7843 

Implemented Coachmark floating variant

**Should be merged after [feat(makeDraggable): utility added for drag functionality](https://github.com/carbon-design-system/ibm-products/pull/7908) and [feat(Coachmark): added coachmarkv2 tooltip variant](https://github.com/carbon-design-system/ibm-products/pull/7837)**

#### PR Checklist

our contributing docs.


packages/ibm-products/src/components/CoachmarkV2/Coachmark-v2.tsx
```diff
@@ -0,0 +1,221 @@
...
+   */
+  positionTune?: { x: number; y: number };
+  /**
+   * Specifies whether the component is floating or not.
+   */
+  isFloating?: boolean;
+}
+
+// Define the type for Coachmark, extending it to include Trigger and Content
+export type CoachmarkV2Component = React.ForwardRefExoticComponent<
```
davidmenendez: lets remove usage of `React.` and just move it to the import `ForwardRefExoticComponent` `FC` etc

packages/ibm-products/src/components/CoachmarkV2/Coachmark-v2.tsx
```diff
@@ -0,0 +1,221 @@
...
+ * Coachmarks are used to call out specific functionality or concepts
+ * within the UI that may not be intuitive but are important for the
+ * user to gain understanding of the product's main value and discover new use cases.
+ */
+export const CoachmarkV2 = React.forwardRef<HTMLDivElement, CoachmarkV2Props>(
+  (props, ref) => {
+    const {
+      children,
+      className,
+      ariaLabel = 'CoachmarkV2',
```
davidmenendez: we should avoid hard-coded strings like this when possible. this can be moved to the stories file.

packages/ibm-products/src/components/CoachmarkV2/Coachmark-v2.tsx
```diff
@@ -0,0 +1,221 @@
...
+      align = 'bottom',
+      isOpenByDefault = false,
+      positionTune,
+      isFloating,
+      ...rest
+    } = props;
+    const [open, setOpen] = React.useState(isOpenByDefault);
+    const triggerRef = useRef<HTMLElement>(null);
+    const internalRef = useRef<HTMLDivElement | null>(null);
+    const [contentRef, setContentRef] = React.useState<HTMLElement | null>(
```
davidmenendez: why use `useState` instead of `useRef` is we're setting a ref value?

packages/ibm-products/src/components/CoachmarkV2/Coachmark-v2.tsx
```diff
@@ -0,0 +1,221 @@
...
+   */
+  onClose?: () => void;
+  /**
+   * Where to render the Coachmark relative to its target.
+   */
+  align?: NewPopoverAlignment;
+  /**
+   * Fine tune the position of the target in pixels.
+   */
+  positionTune?: { x: number; y: number };
```
davidmenendez: i think we can shorten this to `position`

packages/ibm-products/src/components/CoachmarkV2/Coachmark-v2.tsx
```diff
@@ -0,0 +1,221 @@
...
+   */
+  className?: string;
+  /**
+   * The aria label applied to the Coachmark component
+   */
+  ariaLabel?: string;
+  /**
+   * Specifies whether the component is currently open.
+   */
+  isOpenByDefault?: boolean;
```
davidmenendez: i think we can shorten this to `defaultOpen`

packages/ibm-products/src/components/CoachmarkV2/Coachmark-v2.tsx
```diff
@@ -0,0 +1,221 @@
...
+   */
+  align?: NewPopoverAlignment;
+  /**
+   * Fine tune the position of the target in pixels.
+   */
+  positionTune?: { x: number; y: number };
+  /**
+   * Specifies whether the component is floating or not.
+   */
+  isFloating?: boolean;
```
davidmenendez: i think we can shorten this to `floating`

packages/ibm-products/src/components/CoachmarkV2/Coachmark-v2.tsx
```diff
@@ -0,0 +1,221 @@
...
+        el.style.transform = `translate(${x}px, ${y}px)`;
+      }
+    }, [positionTune]);
+
+    const setRef = (node: HTMLDivElement | null) => {
+      internalRef.current = node;
+      if (typeof ref === 'function') {
+        ref(node);
+      } else if (ref) {
+        (ref as React.MutableRefObject<HTMLDivElement | null>).current = node;
```
davidmenendez: `MutableRefObject` is deprecated. lets use `RefObject` here

packages/ibm-products/src/components/CoachmarkV2/Coachmark-v2.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+    return (
+      <CoachmarkV2Context.Provider
+        value={{
+          onClose,
+          open,
+          setOpen,
+          align,
+          triggerRef,
+          positionTune: positionTune ?? { x: 0, y: 0 },
```
davidmenendez: lets move the default value here to the prop declaration to simplify things

packages/ibm-products/src/components/CoachmarkV2/CoachmarkBeacon/CoachmarkBeacon.tsx
```diff
@@ -0,0 +1,133 @@
...
+ */
+export enum BEACON_KIND {
+  DEFAULT = 'default',
+}
+
+export interface CoachmarkButtonProps extends ButtonProps<React.ElementType> {
+  onClick?(): void;
+  onDoubleClick?(): void;
+  tabIndex?: number;
+  ['aria-bool']?: boolean;
```
davidmenendez: i _think_ this should be `aria-expanded`?

packages/ibm-products/src/components/CoachmarkV2/CoachmarkBeacon/CoachmarkBeacon.tsx
```diff
@@ -0,0 +1,133 @@
...
+ * Use beacon for the target prop of a Coachmark component.
+ */
+export const CoachmarkBeacon = React.forwardRef<
+  HTMLDivElement,
+  CoachmarkBeaconProps
+>((props, ref) => {
+  const {
+    label,
+    className,
+    kind = defaults.kind,
```
davidmenendez: the `defaults` object isn't necessary. we can just define the default values here

packages/ibm-products/src/components/CoachmarkV2/CoachmarkBubble/CoachmarkBubble.tsx
```diff
@@ -0,0 +1,186 @@
...
+  open: boolean;
+}
+
+const CoachmarkBubble = React.forwardRef<HTMLDivElement, BubbleProps>(
+  (props) => {
+    const {
+      children,
+      align,
+      target,
+      className: customClassName,
```
davidmenendez: can we just leave this as `className`?

packages/ibm-products/src/components/CoachmarkV2/CoachmarkContent.tsx
```diff
@@ -0,0 +1,168 @@
...
+  CoachmarkContentProps & React.RefAttributes<HTMLDivElement>
+> & {
+  Header: React.FC<ContentHeaderProps>;
+  Body: React.FC<ContentBodyProps>;
+};
+
+const CoachmarkContent = React.forwardRef<
+  HTMLDivElement,
+  CoachmarkContentProps
+>((props, ref) => {
```
davidmenendez: looks like this `ref` isn't being used

packages/ibm-products/src/components/CoachmarkV2/CoachmarkContent.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const { className = '', children, highContrast, dropShadow, ...rest } = props;
```
davidmenendez: looks like `rest` isn't being used

packages/ibm-products/src/components/CoachmarkV2/CoachmarkContent.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const coachmarkContentBlockClass = `${blockClass}--coachmark-content`;
+  const contentBodyClass = `${blockClass}--content-body`;
+  const { align, onClose, open, setOpen, triggerRef, setContentRef } =
```
davidmenendez: `onClose` isn't being used

packages/ibm-products/src/components/CoachmarkV2/CoachmarkContent.tsx
```diff
@@ -0,0 +1,168 @@
...
+    useContext(CoachmarkV2Context);
+  const [targetId, setTargetId] = useState<string | null>(null);
+
+  // setting targetId from triggerRef context value
+  useLayoutEffect(() => {
```
davidmenendez: this effect shouldn't be necessary. it looks like the stately value is just being derived from the ref object, which can be referenced at any time, so it shouldn't need to be kept in state. it could just be something like `const targetId = open ? triggerRef?.current?.id : null`

packages/ibm-products/src/components/CoachmarkV2/CoachmarkContent.tsx
```diff
@@ -0,0 +1,168 @@
...
+    };
+  }, [open, targetId, setOpen]);
+
+  useEffect(() => {
+    if (open && bubbleRef.current) {
+      const dragContainer = bubbleRef.current.querySelector(
+        `.${pkg.prefix}__bubble`
+      );
+      if (dragContainer instanceof HTMLElement) {
+        setContentRef(dragContainer);
```
davidmenendez: i think i mentioned this elsewhere, but we should probably just be setting this ref with `useRef` instead of `useState`

packages/ibm-products/src/components/CoachmarkV2/CoachmarkTrigger.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+  const userOnClick = (children as ReactElement<any>).props.onClick;
+
+  const composedOnClick = (event: React.MouseEvent) => {
+    setOpen?.(!open);
+    userOnClick?.(event);
+  };
+
+  // Clone the element and override onClick
+  childWithOnClick = cloneElement(children, {
```
davidmenendez: is this usage of `cloneElement` necessary? why can't these click handlers be applied to the outer div?

packages/ibm-products/src/components/CoachmarkV2/ContentHeader.tsx
```diff
@@ -0,0 +1,107 @@
...
+}
+
+export type EnrichedChildren = {
+  children: React.ReactNode;
+};
+
+const ContentHeader = React.forwardRef<HTMLDivElement, ContentHeaderProps>(
+  (props, ref) => {
+    const { className = '', closeIconDescription, ...rest } = props;
+    const { open, setOpen, onClose, contentRef, isFloating } =
```
davidmenendez: `open` isn't being used

packages/ibm-products/src/components/CoachmarkV2/ContentHeader.tsx
```diff
@@ -0,0 +1,107 @@
...
+        ref={handleRef}
+        className={cx(contentHeaderBlockClass, className)}
+      >
+        {isFloating && (
+          <Button
+            kind="ghost"
+            size="sm"
+            ref={dragRef}
+            renderIcon={Draggable}
+            iconDescription={'Drag'}
```
davidmenendez: hard-coded string

packages/ibm-products/src/components/CoachmarkV2/ContentHeader.tsx
```diff
@@ -0,0 +1,107 @@
...
+            hasIconOnly
+            className={`${contentHeaderBlockClass}--drag-icon`}
+          />
+        )}
+        <Button
+          kind="ghost"
+          size="sm"
+          renderIcon={Close}
+          iconDescription={closeIconDescription ?? 'Close'}
```
davidmenendez: we can remove the `?? 'close'` code since we shouldn't be providing any strings

davidmenendez (changes requested):
looks terrific! just a few comments & questions

packages/ibm-products/src/components/CoachmarkV2/Coachmark-v2.tsx
anamikaanu96: I have used `useState` to manage `contentRef` so it can be shared via context. This allows me to store the ref of `<CoachmarkV2.Content` and access it from both `<CoachmarkV2.Content>` and `<CoachmarkContent.Header>`. With `useRef`, I don’t think it would be possible to ensure the updated ref is available across components.

packages/ibm-products/src/components/CoachmarkV2/CoachmarkContent.tsx
anamikaanu96: I kept it as state because it needs to be shared between the parent and child components.

packages/ibm-products/src/components/CoachmarkV2/CoachmarkTrigger.tsx
amal-k-joy: We might not need to wrap the trigger inside <CoachmarkV2.Trigger>. Instead, we can follow the approach used in popovers—let the user handle the trigger interaction and control the open state via the open prop.
   ```
 //<CoachmarkV2.Trigger> 
          <Button
            id="CoachmarkBtn"
            kind="tertiary"
            size="md"
            label="Show information"
            renderIcon={Crossroads}
          >
            Show information
          </Button>
      //  </CoachmarkV2.Trigger>
```

File: packages/ibm-products-styles/src/components/CoachmarkV2/_bubble.scss [hunk 21]
```diff
@@ -0,0 +1,63 @@
+/**
+ * Copyright IBM Corp. 2025
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+@use '@carbon/react/scss/type' as *;
```
davidmenendez: it looks like this line is causing the build to fail

packages/ibm-products/src/components/CoachmarkV2/CoachmarkContent.tsx
davidmenendez: that value isn't be kept in context though. looks like it's just being passed to `CoachmarkBubble` below

```
target={`#${targetId}`}
```

you should still be able to derive the value like i showed above. can you try this and let me know if it works or not?

packages/ibm-products-styles/src/components/Coachmark/_coachmark.scss
```diff
@@ -29,3 +32,21 @@ $block-class: #{c4p-settings.$pkg-prefix}--coachmark;
 .#{$block-class} {
   position: relative;
 }
+
+// Onboarding/Coachmark/next
+
+$block-class-next: #{c4p-settings.$pkg-prefix}--coachmark__next;
+
+@include block-wrap(#{$block-class}__next) {
+  &.#{$block-class-next}--floating {
```
amal-k-joy, line 41: I was just wondering, since Bubble is used as an internal component, we can probably not render arrow for floating variant, rather than hiding via css.

packages/ibm-products/src/components/Coachmark/next/Coachmark/ContentHeader.tsx
```diff
@@ -0,0 +1,131 @@
...
+        contentRef &&
+        typeof handleRef === 'object' &&
+        handleRef !== null &&
+        'current' in handleRef &&
+        handleRef.current &&
+        dragRef.current
+      ) {
+        makeDraggable({
+          el: contentRef,
+          handle: handleRef.current,
```
amal-k-joy: As we discussed, please consider renaming the input names , may be in next PRs
handle ==> dragHandle,
 focusableInHandle ==> focusableDragHandle

packages/ibm-products/src/components/Coachmark/next/Coachmark/Coachmark.tsx
```diff
@@ -0,0 +1,257 @@
...
+        (el) => el.tabIndex >= 0 && !el.hasAttribute('disabled')
+      );
+
+      if (firstFocusable) {
+        triggerRef.current = firstFocusable;
+      }
+    }, [children]);
+
+    useIsomorphicEffect(() => {
+      const { x = 0, y = 0 } = position ?? {};
```
amal-k-joy, line 163: I’ve been thinking about the `position` prop in the Coachmark component. It feels like position should only be accepted by the Beacon, since that’s what defines the visual anchor or trigger. Ideally, the Coachmark itself shouldn’t require manual positioning — it should automatically anchor to its target (whether that’s the Beacon or another referenced element).

packages/ibm-products/src/components/Coachmark/next/Coachmark/Coachmark.stories.jsx
```diff
@@ -0,0 +1,208 @@
...
+
+  const handleClose = () => {
+    setIsOpen(false);
+  };
+
+  const handleButtonClick = () => {
+    setIsOpen((isOpen) => !isOpen);
+  };
+  return (
+    <Theme theme={carbonTheme}>
```
amal-k-joy, line 160: This structure feels a bit off to me — maybe it’s just personal preference, but I think only Coachmark should be the exported component if we are following Compound Component Pattern . Everything else, like  Content, Header, and Body, should be encapsulated within it.
```
<Coachmark open={true} onClose={handleClose}>
  <CoachmarkBeacon label="Help" onClick={handleBeaconClick} />
  <Coachmark.Content highContrast>
    <Coachmark.Content.Header
      closeIconDescription="Close"
      dragIconDescription="Drag"
      onClose={handleClose}
    />
    <Coachmark.Content.Body>
      <h2>Title</h2>
      <p>Description here.</p>
    </Coachmark.Content.Body>
  </Coachmark.Content>
</Coachmark>
```

packages/ibm-products/src/components/Coachmark/next/Coachmark/CoachmarkContent.tsx
```diff
@@ -0,0 +1,170 @@
...
+          !bubbleElement.contains(event.target as Node) &&
+          targetElement &&
+          !targetElement.contains(event.target as Node)
+        ) {
+          setOpen?.(false);
+        }
+      };
+
+      if (open) {
+        document.addEventListener('click', handleOutsideClick);
```
amal-k-joy, line 113: include keydown event to close component on Escape here

packages/ibm-products/src/components/Coachmark/next/Coachmark/ContentHeader.tsx
```diff
@@ -0,0 +1,131 @@
...
+      ...rest
+    } = props;
+    const { setOpen, onClose, contentRef, floating } =
+      useContext(CoachmarkContext);
+    const headerRef = useRef<HTMLHeadingElement | null>(null);
+    const dragRef = useRef<HTMLButtonElement | null>(null);
+    const handleRef = ref || headerRef;
+    const contentHeaderBlockClass = `${blockClass}--content-header`;
+    useEffect(() => {
+      const close = (e: KeyboardEvent) => {
```
amal-k-joy: Maybe we can move the logic for closing the component on Escape into CoachmarkContent.ts, since we're already listening to a window event there for outside click.

packages/ibm-products/src/components/Coachmark/next/Coachmark/ContentHeader.tsx
```diff
@@ -0,0 +1,131 @@
...
+        makeDraggable({
+          el: contentRef,
+          handle: handleRef.current,
+          focusableInHandle: dragRef.current,
+        });
+      }
+    });
+
+    return (
+      <CoachmarkBubbleHeader
```
amal-k-joy, line 79: `CoachmarkBubbleHeader` should accept children as well and we can keep it between drag handle and close icon. User can make use of this to render any custom actions or content.

```
 {floating && (
          <Button
            kind="ghost"
            size="sm"
            ref={dragRef}
            renderIcon={Draggable}
            iconDescription={dragIconDescription}
            hasIconOnly
            className={`${contentHeaderBlockClass}--drag-icon`}
          />
        )}
{children}
        <Button
          kind="ghost"
          size="sm"
          renderIcon={Close}
          iconDescription={closeIconDescription}
          hasIconOnly
          onClick={closeBubble}
        />
```

amal-k-joy (changes requested):

packages/ibm-products/src/components/Coachmark/next/Coachmark/Coachmark.tsx
anamikaanu96: Initially, I applied the `position` only to the `beacon`, but this caused the outer container to remain in its original position while only the beacon moved. Based on this, I believe that when we intend to move the entire `Coachmark`, the position should be applied to the main container. Applying it only to the `beacon` doesn't seem sufficient in this scenario.

amal-k-joy: In that case, I think we can remove the position prop from the Coachmark component. Since the target element (like the Beacon) is always controlled by the user and can be positioned freely in the DOM, the Coachmark should simply anchor itself to that target.
Any thoughts @matthewgallo , @davidmenendez ?

packages/ibm-products/src/components/Coachmark/next/Coachmark/Coachmark.stories.jsx
anamikaanu96: Thanks for figuring it out! 🙌

packages/ibm-products-styles/src/components/Coachmark/_coachmark.scss
anamikaanu96: I can handle this by adding a `floating` prop in `Bubble`, but I was thinking, if the bubble is moved to a stable component in the future, we’ll likely just do something like `import Bubble`. At that point, we’d need to go back and remove the prop and switch to using CSS instead.

amal-k-joy: In that case, shall we can open a feature request for Bubble at that time, just like `caret` prop in Popover. 
And now we can have this new prop for internal Bubble component .
@elycheea @matthewgallo , any thoughts?

packages/ibm-products/src/components/Coachmark/next/Coachmark/Coachmark.tsx
matthewgallo: I agree, ideally we remove it from the base `Coachmark` and ~only pass it to the beacon~ only have it used within the adopter's implementation.

packages/ibm-products-styles/src/components/Coachmark/_coachmark.scss
matthewgallo: I'm also feeling like it would be better to not render the arrow in the DOM if we don't need to. If we need to change this approach in the future, if bubble becomes it's own standalone component, we can revisit at that point.

packages/ibm-products/src/components/Coachmark/next/Coachmark/Coachmark.tsx
davidmenendez: agreed

packages/ibm-products/src/components/Coachmark/next/Coachmark/ContentHeader.tsx
anamikaanu96: changes done in [makeDraggable PR](https://github.com/carbon-design-system/ibm-products/pull/7908)

packages/ibm-products/src/components/Coachmark/next/Coachmark/Coachmark.tsx
amal-k-joy: Yes, the adopter will use a positioned Beacon or Button as the target. We don’t need to pass `position` to the Beacon either, since that can already be achieved on the adopter’s side.

amal-k-joy (approved):
Nicely done 🎉 🎉 
We can discuss and figure out if we need to keep `position` prop or remove it as discussed in the comment in a separate  PR.

Issue #7951, opened 2025-07-31
[tearsheet]: focus trap not working in tearsheet web-component


makafsal:
### Package

Carbon for IBM Products

### Description

Not even focusing on the first element, and overall, focusing is not working.

https://github.com/user-attachments/assets/74dc7953-f23a-4c77-8043-433bebc6e18a

### Component(s) impacted

Tearhsheet web component

### Browser

Chrome

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

v0.20.1

### Suggested Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Product/offering

Carbon IBM Products

### CodeSandbox or Stackblitz example

https://ibm-products-web-components.carbondesignsystem.com/?path=/story/components-tearsheet--default

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

stefan-cooper:
I am also seeing an issue related to focusing on elements in the Tearsheet component after bumping to ibm-products v2.64.0. Possibly from the changes in https://github.com/carbon-design-system/ibm-products/commit/6a17a20721daa5124ea2df6d408b231413fec007 ?

Seems to be fine in most cases, but I have some tests that use the focus which now no longer work/say that it is focused.

amal-k-joy:
Lets try to make use of `manageFocusTrap` utility for this.
`packages/ibm-products-web-components/src/utilities/manageFocusTrap/manageFocusTrap.ts`

matthewgallo:
Hey @stefan-cooper! This issue is specific to the tearsheet web component. If you're able to reproduce the focus issues you're finding in the react tearsheet feel free to open a new issue for us.

amal-k-joy:
We can probably hold this one until the [dialogue implementation](https://github.com/carbon-design-system/carbon/issues/20506) in modal by carbon. 
cc: @matthewgallo

Issue #7731, opened 2025-06-25
[Enhancement]: investigate remove or improve `useFocus` hook


makafsal:
### The problem

💭 **There are a few concerns about the `useFocus` hook:**

1. Why aren't we using the core's `wrapFocusWithoutSentinels` utility?
2. The `useFocus` is a React hook; it would be beneficial in WC if it were a vanilla JS utility.
3. Core uses `tabbable.js` library to list all tabbable items in an element.

### The solution

🎲 **Decision stage:**
- [ ] Investigate the possibility of using the core's `wrapFocusWithoutSentinels` utility.

🏃🏻‍♂️‍➡️ **If we can't use the `wrapFocusWithoutSentinels`, then**
- [ ] Research the best way to revamp `useFocus` hook into a framework-agnostic utility.
- [ ] Also, investigate the pros and cons of the `tabbable.js` library.

### Links to other materials

[wrapFocusWithoutSentinels](https://github.com/carbon-design-system/carbon/blob/6b474db8ca2b0cd29a3b35be5e371be6feea6ffc/packages/react/src/internal/wrapFocus.ts#L131)

### Owner/main maintainer(s)

Carbon for IBM Products

### Business priority

Low Priority = release date is not dependent on fix or not upcoming

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

davidmenendez:
this is a great start so far! thank you @makafsal 

## initial thoughts

IMO tab focus is one of those things that i don't think we should be expending so much time and effort on when other libraries exist. this is starting to reach `DataGrid` levels of why are we trying to reinvent the wheel? if core is already utilizing a third party library to handle this then i think we should be also or we should at least both align on a common framework / library to handle this functionality.

If core has some utility code that we can extend we should also definitely look into it.

[...]

PR #7954, opened 2025-07-31
fix(side-panel): resolve focus trap issue in side-panel web component

### LINKED ISSUE #7915: [side-panel]: Focus trap not working in the SlideOver side-panel
### Package

Carbon for IBM Products

### Description

Focus trap not working in the SlideOver side-panel:

https://github.com/user-attachments/assets/1e6f56fe-e4bc-43fe-af4e-7852e09c1815

### Component(s) impacted

side-panel web component

### Browser

Chrome

v0.20.0

### Suggested Severity

### Product/offering

Carbon for IBM Products Web Components

https://ibm-products-web-components.carbondesignsystem.com/?path=/story/components-sidepanel--slide-over

_No response_

_No response_

### Code of Conduct

makafsal:
Closes #7915 

Resolve focus trap issue

#### What did you change?
Changed the current implementation

Storybook

#### PR Checklist

our contributing docs.


makafsal:
> this is great! love the implementation.
> 
> i think we should work on creating a focus trap utility. we have a few different components where we're implementing focus trapping and it would be great to consolidate the functionality into a single function. that would also help to alleviate my other thought, which is testing. if we decouple this into it's own utility it'll make testing easier.
> 
> @makafsal do you know if there's been any discussion or an issue to create a focus trap utility? if not that's something we can work on and in the short term we should add some tests, but that can be done in a follow up issue / pr 👍

Yeah, I've already created one [issue](https://github.com/carbon-design-system/ibm-products/issues/7731) for utility-related tasks. I think we can have utility-related discussions over there.

Found focus-related issues in [Tearsheet](https://github.com/carbon-design-system/ibm-products/issues/7951) and [Modal (core)](https://github.com/carbon-design-system/carbon/issues/20073) web components as well.

packages/ibm-products-web-components/src/components/side-panel/side-panel.ts
```diff
@@ -177,63 +129,75 @@ class CDSSidePanel extends HostListenerMixin(LitElement) {
...
+  protected _handleHostKeydown = (event: KeyboardEvent) => {
+    if (event.key === 'Tab' && !this.slideIn) {
+      const { first: _firstElement, last: _lastElement } = this.getFocusable();
+
+      if (
+        event.shiftKey &&
+        (this.shadowRoot?.activeElement === _firstElement ||
+          document.activeElement === _firstElement)
+      ) {
+        event.preventDefault();
```
matthewgallo, line 194: Small thing, but since you're calling `event.preventDefault()` from both inside the `if` and `else` blocks, can you just move it outside so it's just called once?

makafsal: No, only want to `preventDefault` when the condition is satisfied; otherwise, we need the default behavior.

matthewgallo: Ahh I see, one check is for first element and the other is for the last element.

PR #7964, opened 2025-08-01
fix: include ellipsis in truncated-text

### LINKED ISSUE #7928: [Bug]: Add missing ellipsis to `c4p-truncated-text`
### Package

Carbon for IBM Products

### Description

The Web Components version of the TruncatedText utility component includes the `View more` on a new line but missed the ellipsis on the preceding line.

### Component(s) impacted

`c4p-truncated-text`

### Browser

_No response_

v2.71.1

### Suggested Severity

### Product/offering

CFIP

https://ibm-products-web-components.carbondesignsystem.com/?path=/story/utilities-truncatedtext--type-expand

Compare to React implementation

React | Web Components
----- | --------------
Image | Image

_No response_

### Code of Conduct

davidmenendez:
Closes #7928


davidmenendez:
@elycheea dumb question: should `...` have been added to `View less`? 😅

elycheea (changes requested):
image

@davidmenendez the ellipsis should be in the _body_ of the text, not in the label. 😂 

So “Modify ...” instead of “Modify the.”
No need to modify the default labels. 😅

davidmenendez:
> @davidmenendez the ellipsis should be in the _body_ of the text, not in the label. 😂

lmao what was i thinking 🤣

davidmenendez:
i also included a fix to address an issue when try to animate the height when going from line-clamp `none` to `{a value that causes the size of the element to shrink and not transition the way we expect it}`

elycheea (approved):
Looks good! Not to scope creep, but I wonder if we should name the stories `With tooltip` and `With expand` like in the React version. But not a blocker to this PR.

PR #7967, opened 2025-08-01
fix(EmptyState): Modified subtitle container to be a div

### LINKED ISSUE #7490: EmptyState subtitle with ToggleTip results in React error "<div> cannot appear as a descendant of <p>."
### Package

Carbon for IBM Products

### Description

We're implementing a design that requires a `ToggleTip` in the body of the empty state clarifying some vocabulary that might not be clear to the user.  In our product, we see the following error in the console:

```
app-index.tsx:25 Warning: In HTML, <div> cannot be a descendant of <p>.
This will cause a hydration error.
    at div
    at span
    at span
    at PopoverContentRenderFunction (webpack-internal:///(app-pages-
```

I've reproduced our setup in the attached Stackblitz.  Since the code is not being rendered on the server the error is instead:

```
Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.
    at div
    at span
    at span
    at PopoverContentRenderFunction (https://githubkjhwrjfb-wwh0-y58qoqol--5173--4d9fd228.local-credentialless.webcontainer.io/node_modules/.vite/deps/chunk-ZEIYFFZ5.js?v=cb05ba24:10983:5)
```

This appears to be occurring because the empty state's subtitle is rendering in a `<p>` tag, which will cause issues if any Carbon controls are used that have a `<div>`.  Despite console errors to the contrary in my Stackblitz, [it seems](https://github.com/carbon-design-system/ibm-products/blob/ffbdbc4c79b6549028bac4454edc3763b2ebca42/packages/ibm-products/src/components/EmptyStates/EmptyState.tsx#L95C23-L95C32) passing a `ReactNode` is supported for the error state's `subtitle` prop, so there probably shouldn't be restrictions on which elements are used.

Could the error state's implementation be changed to:
* Not use a `<p>`; or
* Only use `<p>` if the `subtitle` is a string; or
* Expose a subtitle-equivalent of `headingAs` (i.e. `subtitleAs`) to let consumers override this default?

Thanks!

### Component(s) impacted

NoDataEmptyState (and likely the other empty state components)

Image

Image

### Browser

Chrome

@carbon/ibm-products@2.11.2

### Suggested Severity

None

### Product/offering

IBM watsonx BI

https://stackblitz.com/edit/github-kjhwrjfb?file=src%2FMain%2FMain.jsx

_No response_

_No response_

### Code of Conduct

jlongshore:
Closes #7490 
A team is implementing a design that adds a ToggleTip in the subtitle. The subtitle was a hardcoded paragraph tag, throwing a warning when elements that are div based are added as children of the <p> tag...

#### What did you change?
Modified the tag to be a div

- Visual A/B testing initially using the dev tools
- Wrote a story including a Carbon tooltip

After cleaning up the isolated preview from the deprecated wrapper, the a11y violations were outside of the component:
image

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/EmptyStates/EmptyStateV2.stories.jsx
+++ b/packages/ibm-products/src/components/EmptyStates/EmptyStateV2.stories.jsx
@@ -7,11 +7,12 @@
 
 import React from 'react';
 import { action } from 'storybook/actions';
-import { Add } from '@carbon/react/icons';
+import { Add, Information } from '@carbon/react/icons';
 import CustomIllustration from './story_assets/empty-state-bright-magnifying-glass.svg';
 import { EmptyStateV2 } from '.';
 import { StoryDocsPage } from '../../global/js/utils/StoryDocsPage';
 import { Annotation } from '../../../.storybook/Annotation';
+import { Tooltip } from '@carbon/react';
 
 export default {
   title: 'Deprecated/Empty state/EmptyStateV2',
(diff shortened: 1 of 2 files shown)


davidmenendez (approved):

matthewgallo (changes requested):
@jlongshore can you also fix this issue in `EmptStateContent.tsx`? It will have the same issue too with the `subtitle`

davidmenendez (approved):

PR #7970, opened 2025-08-01
fix: update ci script to use reporter

### LINKED ISSUE #7605: Investigate why playwright reports are not included in build output
The playwright a11y report is what is used to show a11y statuses from our website, however, these reports are not included in our build output. This requires us to manually generate the report and include that file in our website, a manual step that we haven't been doing causing the a11y statuses on our website to become outdated.

We should identify any missing steps to ensure this report is included in our build output so that the a11y status from our website shows the most up to date information.

davidmenendez:
Closes #7605


matthewgallo (approved):

amal-k-joy (commented):
@davidmenendez ,
One doubt since I am not good at actions. 
We are generating and uploading a merged playwright-report with name` html-report--attempt-1` during CI and  during release process, we are again uploading another report `playwright-avt-report`. I wonder where we use these?

davidmenendez:
@amal-k-joy good questions!

we want to run the AVT tests as part of the CI process for pull requests to make sure there are no regressions. If one of the AVT shards fails then exporting the report we can get a better view of what failed during that particular PR. but those artifacts only exist within the context of the pull request CI. on release we need to ensure that a report is generated and bundled with the library output so that it can easily be used on the website.

if you look at this [currently commented out line of code](https://github.com/carbon-design-system/carbon-for-products-website/blob/b7b6380bb0da53bf5ab305262e515a3588d4f58d/src/components/A11yStatus/A11yStatus.js#L19) 

```
// import * as avtTestData from '@carbon/react/.playwright/INTERNAL_AVT_REPORT_DO_NOT_USE.json';
```

we should just be able to import the report from the build, but because the report isn't being generated on release it has to be manually updated. so, yes, we are generating the report multiple times but we're doing it for different reasons; the CI part is for bug fixing and the release part is for presentation.

amal-k-joy (approved):
Nice work 🎉

PR #7956, opened 2025-07-31
build(performance): `rolldown-vite` integration

matthewgallo:
Closes #7953 

This PR introduces `rolldown-vite` into our storybook build configuration. The initial benchmark comparison between build times looks like a solid improvement:

## Performance comparison
Build times for `@carbon/ibm-products`
| Default vite config | `rolldown-vite` |
|------------------|----------------- |
| ~1 minute 50 seconds | ~23.5 seconds |

~79% decrease in total build time

#### What did you change?
- Renamed all `*.js` files containing jsx syntax to `*.jsx`
- Added `*.jsx` file extension to any import of a file that hasn't migrated to `*.tsx` yet
- Removed unsupported css syntax @property { initial-value }
#### How did you test and verify your work?
Ran storybook locally and successfully built `ibm-products` locally
#### PR Checklist

As the author of this PR, before marking ready for review, confirm you:

- [ ] Reviewed every line of the diff
- [ ] Updated documentation and storybook examples
- [ ] Wrote passing tests that cover this change
- [ ] Addressed any impact on accessibility (a11y)
- [ ] Tested for cross-browser consistency
- [ ] Validated that this code is ready for review and status checks should pass

More details can be found in the [pull request](./CONTRIBUTING.md) section of
our contributing docs.

PR #7990, opened 2025-08-05
fix(styles): use correct syntax for custom CSS props

jhpedemonte:
Use correct syntax for custom CSS property.

#### What did you change?

A couple of custom CSS properties with value "320px" were incorrectly set as `<integer>` instead of `<length>`. The former should only be numbers and not contain any units, therefore should be using the latter.

I noticed this when building a separate project using [lightningcss](https://lightningcss.dev/), which is more strict than other CSS parsers. Lightningcss throws an error for these 2 references.

Compared local storybook to deployed version. Tested `SidePanel` and `Datagrid/Filtering/Panel` to ensure that nothing regressed.

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products-styles/src/__tests__/__snapshots__/styles.test.js.snap
+++ b/packages/ibm-products-styles/src/__tests__/__snapshots__/styles.test.js.snap
@@ -2652,7 +2652,7 @@ p.c4p--about-modal__copyright-text:first-child {
 @property --panel-transform {
   inherits: true;
   initial-value: 320px;
-  syntax: "<integer>";
+  syntax: "<length>";
 }
 .c4p--side-panel--scrolls {
   overflow: auto;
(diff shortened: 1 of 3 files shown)


jhpedemonte:

matthewgallo (changes requested):
Thanks for this contribution @jhpedemonte! I faced a similar issue with this syntax in https://github.com/carbon-design-system/ibm-products/pull/7956 where we're integrating `rolldown-vite` which uses `lightningcss`.

Can you also make these updates in `_filter-summary.scss`?

jhpedemonte:
> I faced a similar issue with this syntax in #7956 where we're integrating `rolldown-vite` which uses `lightningcss`.

Oh, nice!



> Can you also make these updates in `_filter-summary.scss`?

@matthewgallo That one is [already set](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products-styles/src/components/FilterSummary/_filter-summary.scss#L18) to `<length>` from what I can see.

matthewgallo:
Ah yes you're right, sorry about that!

matthewgallo (approved):

PR #8014, opened 2025-08-06
fix: options-tile review updates

### LINKED ISSUE #7999: Options tile followup
We should try to wrap up the epic for this.

Confirmed with @susanacr2 that [#7723](https://github.com/carbon-design-system/ibm-products/issues/7723) is ready for development although Judith is continuing to look at the focus state issues.

## Tasks

* ~~Adjust spacing between title and description~~
  * currently appears to be correct. `$spacing-01` margin in XL version is present. 
* ~~Center align toggle with title & description~~
  * currently appears to be correct
* ~~Body padding~~
  * currently appears to be correct
* Description should be hidden in open state
* Update font to `$body-01` in the storybook example to align with react version
* Update the story to remove the attribute handling in `handleOpen` & `handleClose`
  * the `details` html component can't be controlled like you would expect with an `open` attribute. if you look at the current [story](https://ibm-products-web-components.netlify.app/?path=/story/components-optionstile--default) and use the `open` control it will cause an infinite loop of opening and closing because we assume we can control it with `open`. however this is not the case. `open` should just be treated as a default state and the html element will handle the open and closed state internally. `handleOpen` and `handleClose` should be treated like callbacks and not interfere with the open state.
* Make sure to get design review/approval in PR

davidmenendez:
Closes #7999 

addresses design review items and also includes enhancements to the default `options-tile` story to more closely resemble the react version

### DIFF
--- a/packages/ibm-products-web-components/src/components/options-tile/options-tile.stories.ts
+++ b/packages/ibm-products-web-components/src/components/options-tile/options-tile.stories.ts
@@ -10,13 +10,15 @@
 import { html } from 'lit';
 import './index';
 import '@carbon/web-components/es/components/toggle/index.js';
+import '@carbon/web-components/es/components/dropdown/index.js';
+import styles from './story-styles.scss?lit';
 
 const argTypes = {
   body: {
     control: 'text',
     description: 'Slot body content',
   },
-  open: {
+  defaultOpen: {
     control: 'boolean',
     description: 'If `true` the body of the component is shown',
   },
@@ -39,25 +41,29 @@ const argTypes = {
   },
 };
 
+const blockClass = 'options-tile';
+
 const handleOpen = (evt: Event) => {
-  const tile = document.querySelector('#my-tile');
-  tile?.setAttribute('open', 'true');
+  console.log('open');
 };
 
 const handleClose = (evt: Event) => {
-  const tile = document.querySelector('#my-tile');
-  tile?.removeAttribute('open');
+  console.log('close');
 };
 
 const renderTemplate = (args) => {
-  const { open, size, titleText, titleId } = args;
+  const { defaultOpen, size, titleText, titleId } = args;
   return html`
+    <style>
+      ${styles}
+    </style>
     <c4p-options-tile
+      class="${blockClass}"
+      ?defaultOpen="${defaultOpen}"
       id="my-tile"
-      ?open=${open}
-      size=${size}
-      titleId=${titleId}
-      titleText=${titleText}
+      size="${size}"
+      titleId="${titleId}"
+      titleText="${titleText}"
       @c4p-options-tile-open=${handleOpen}
       @c4p-options-tile-close=${handleClose}
     >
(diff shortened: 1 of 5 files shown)


susanacr2:
Approved by design. It matches what we have on [the site](https://pages.github.ibm.com/carbon/ibm-products/components/options-tile/usage/). 

Thanks @davidmenendez 
Screenshot 2025-08-07 at 8 22 50 AM

packages/ibm-products-web-components/src/components/options-tile/options-tile.stories.ts
```diff
@@ -39,25 +41,29 @@ const argTypes = {
   },
 };
 
+const blockClass = 'options-tile';
+
 const handleOpen = (evt: Event) => {
-  const tile = document.querySelector('#my-tile');
-  tile?.setAttribute('open', 'true');
+  console.log('open');
```
matthewgallo: Remove ^^

packages/ibm-products-web-components/src/components/options-tile/options-tile.stories.ts
```diff
@@ -39,25 +41,29 @@ const argTypes = {
...
 };
 
 const handleClose = (evt: Event) => {
-  const tile = document.querySelector('#my-tile');
-  tile?.removeAttribute('open');
+  console.log('close');
```
matthewgallo: Remove ^^

matthewgallo (changes requested):
Looks good! Just found to console logs. And I also think the `body` slot should be wrapped in the `layer` component. If you compare this implementation vs the react version, the text input and dropdown have different background colors.

davidmenendez:
@matthewgallo thanks for reminding me to figure out how to integrate actions into the web-component stories.

TL;DR https://storybook.js.org/docs/essentials/actions#via-storybooktest-fn-spy-function

davidmenendez:
unrelated, but i noticed these properties show up on all the overview tabs. we should figure out why

Screenshot 2025-08-08 at 10 51 23 AM

matthewgallo (approved):

matthewgallo:
I made an [issue](https://github.com/carbon-design-system/ibm-products/issues/8026) to track the overflow examples showing in the api table so we don't lose track of that

elycheea (approved):
Changes look good, but can you add some clarification in the PR on the reason we’re changing `open` ~prop~ attribute to `defaultOpen`?

davidmenendez:
@elycheea good idea. thank you. i've updated the body to extrapolate on that change.

> also includes an API change: open => defaultOpen this was done to convey the correct intended usage of the component. because of how the underlying html details element works it can't be controlled with open as a convention open is associated with controlled elements. since this is an uncontrolled element i thought changing the prop from open to defaultOpen would help consumers understand that distinction. details can be given an initial open or closed state, but after that it's best to let the element handle it's state internally.

i hope that that makes sense and helps to clarify the change.

elycheea:
Yes, I think it was implied but extra context on the `details` element is helpful too. Also like when our PRs are self-documenting. 😁

PR #8020, opened 2025-08-07
feat(stepflows): introduce new step utility

### LINKED ISSUE #8018: Composable stepping utility
### The problem

Our current stepping experiences are explicitly built-in to `CreateTearsheet` and `CreateFullPage`. 

### The solution

We need a more composable way to orchestrate stepped patterns that will allow for scalability across any component. Another pain point that has been brought up from product teams is that every step is rendered in the DOM, this utility should solve that problem as well. This will also be a differentiating aspect to this utility vs the view stack utility, where all steps are rendered in the DOM.

### Design link

_No response_

### Links to other materials

_No response_

### Owner/main maintainer(s)

Our team

_No response_

### Product/offering

_No response_

### Business priority

None

### Code of Conduct

matthewgallo:
Closes #8018 

Introduces new stepping utility that allows for building composable/headless stepped experiences. This level of flexibility means that steps could be introduced with any component. For demonstration purposes, I've included examples for tearsheet and side panel in the new stories file.

I've not exported anything new since I think the final place for this utility is in `@carbon/utilities-react` and then we surface it via example stories within our storybook (similar to the `overflowHandler`).

#### What did you change?
```
packages/ibm-products/src/components/StepFlow/StepActions.test.js
packages/ibm-products/src/components/StepFlow/StepActions.tsx
packages/ibm-products/src/components/StepFlow/StepContext.ts
packages/ibm-products/src/components/StepFlow/StepFlow.mdx
packages/ibm-products/src/components/StepFlow/StepFlow.stories.jsx
packages/ibm-products/src/components/StepFlow/StepGroup.tsx
packages/ibm-products/src/components/StepFlow/_storybook-styles.scss
packages/ibm-products/src/components/StepFlow/index.ts
packages/ibm-products/src/components/StepFlow/story-assets/*
packages/ibm-products/src/components/StepFlow/types.ts
```
Storybook, tests pass
#### PR Checklist

- [X] Wrote passing tests that cover this change

our contributing docs.


davidmenendez (commented):
i think overall the implementation looks great. one thought: did you give any thought to using a prop for the steps instead of just using them as children?

```
<StepGroup>
  <Step />
  <Step />
  <Step />
</StepGroup>
// vs
const steps = [
  { id: '1', content: <div>...</div> },
  ...
];
<StepGroup steps={steps} />
```

personally i find using `React.Children` apis to be a bit cumbersome at times and if you can just pass in the data as an array you can use native array apis for free.

one other minor thought: can we change `numSteps` to something like `maxSteps`? i was trying to figure out without looking at the code why the number of steps would be relevant until i figured out it was used to determine if you were on the final step. i think just naming it something like `maxSteps` helps to clarify it's purpose.

amal-k-joy:
> i think overall the implementation looks great. one thought: did you give any thought to using a prop for the steps instead of just using them as children?
> 
> ```
> <StepGroup>
>   <Step />
>   <Step />
>   <Step />
> </StepGroup>
> // vs
> const steps = [
>   { id: '1', content: <div>...</div> },
>   ...
> ];
> <StepGroup steps={steps} />
> ```
> 
> personally i find using `React.Children` apis to be a bit cumbersome at times and if you can just pass in the data as an array you can use native array apis for free.

 But i am more inclined to children, since it give more flexibility to user and manage its states.
> 
> one other minor thought: can we change `numSteps` to something like `maxSteps`? i was trying to figure out without looking at the code why the number of steps would be relevant until i figured out it was used to determine if you were on the final step. i think just naming it something like `maxSteps` helps to clarify it's purpose.

davidmenendez:
@amal-k-joy yes it might be slightly easier to interact with state and props when you can just pass them as children, but i'd like to point out that even [react docs](https://react.dev/reference/react/Children) at the very top recommends avoiding `Children` apis when possible. even though right now it's only being used for `count` and `toArray` this is the kind of implementation detail that has a lot of ramifications, which is why i think we should consider it early.

definitely not wrong either way, but i wanted to point out that there are [alternatives to children](https://react.dev/reference/react/Children#alternatives) and that we should definitely consider both sides before sticking to a convention.

amal-k-joy:
> alternatives to children

@davidmenendez ,
Yes, that was very helpful — thanks for sharing.
My only concern is that the content needs seamless access to the parent component’s state. Will passing the content as a render function might solve this ?.

```
const steps = [
  { id: '1', content: () => <div>...</div> },
  // ...
];
```

matthewgallo:
Hey @amal-k-joy and @davidmenendez, I made some changes based on some of the conversations we've been having about the different approaches we could take here. Notably moving the context values into the utility itself so that it isn't something that has to be done on the consumer side.

Now all you have to do is wrap your component with `StepProvider` and that's it, you can get the context same as before via the context hook.

matthewgallo:
> one other minor thought: can we change numSteps to something like maxSteps? i was trying to figure out without looking at the code why the number of steps would be relevant until i figured out it was used to determine if you were on the final step. i think just naming it something like maxSteps helps to clarify it's purpose.

@davidmenendez I also renamed `numSteps` to `totalSteps`.

Also, to your points about `React.Children`, and I know we also discussed this during our huddle, but it feels more appropriate/acceptable from the library POV. There are many instances of this in `@carbon/react`, [primer](https://github.com/primer/react), [material ui](https://github.com/mui/material-ui), [fluent ui](https://github.com/microsoft/fluentui), and in this particular case it's only used to count the children, not make any type of transformations or cloning.

packages/ibm-products/src/components/StepFlow/story-assets/Tearsheet/Example.tsx
```diff
@@ -0,0 +1,324 @@
...
+              <Step1 />
+              <Step2 />
+              <Step3 />
+              {introExample && selectedFlow === 'plus' && <PlusOnly />}
+            </StepGroup>
+          )}
+
+          {/* Step actions */}
+          <div className="c4p--step-actions__button-container my-custom-action-set">
+            <StepActions
```
amal-k-joy, line 253: I was wondering does the utility really need to provide  `StepActions` component, since every component inside `<StepProvider>`  already have access to the context data and they can perform step actions swiftly right?

amal-k-joy (commented):
Utility looks simple and powerful now. 🎉

packages/ibm-products/src/components/StepFlow/story-assets/Tearsheet/Example.tsx
matthewgallo: I agree, this is now something that can be included on the implementation side of things. I moved it to `_story-assets` and removed the export. I did just notice that I forgot to remove it from the docs.

amal-k-joy: @matthewgallo ,
Hey, one last thing, did you miss to remove this from index.js
`export type { StepActionsProps } from './story-assets/StepActions';`

matthewgallo: Yep, I did!

matthewgallo: Removed [here](https://github.com/carbon-design-system/ibm-products/pull/8020/commits/6da55655b48166600517dbeb8868c43016277455)

davidmenendez (approved):
looks great! glad we took the time to discuss this 😄

PR #8070, opened 2025-08-14
fix: add avt report to correct workflow

davidmenendez:
it looks like i put the avt reporting script in the incorrect workflow file. it looks like it needs to be in [release-base](https://github.com/carbon-design-system/ibm-products/actions/runs/16773206981/workflow#L31) and not `release`.

i'm going to open up a follow up issue to clean up our workflow files and remove any that are no longer in use in order to make working with them easier.


matthewgallo (approved):
😅  Totally missed this going into `release.yml` initially!

davidmenendez:
@matthewgallo or should it be in `release-minor` or `release-scheduled` OR `RELEASE-V1`?! who knows! 😂

jlongshore (approved):

PR #8076, opened 2025-08-14
wip

jlongshore:
Closes #

{{short description}}

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/CreateTearsheet/CreateTearsheetStep.tsx
+++ b/packages/ibm-products/src/components/CreateTearsheet/CreateTearsheetStep.tsx
@@ -13,7 +13,6 @@ import React, {
   isValidElement,
   PropsWithChildren,
   useRef,
-  MutableRefObject,
   RefObject,
 } from 'react';
 import PropTypes from 'prop-types';
@@ -197,7 +196,7 @@ export let CreateTearsheetStep = forwardRef(
   ) => {
     const localRef = useRef<HTMLDivElement>(null);
     const stepRef = ref || localRef;
-    const stepRefValue = (stepRef as MutableRefObject<HTMLDivElement>).current;
+    const stepRefValue = (stepRef as RefObject<HTMLDivElement>).current;
     const stepsContext = useContext(StepsContext);
     const stepNumber = useContext(StepNumberContext);
     const [shouldIncludeStep, setShouldIncludeStep] =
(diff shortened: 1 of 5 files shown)


packages/ibm-products/src/global/js/hooks/useCreateComponentFocus.js
```diff
@@ -27,11 +27,52 @@ export const useCreateComponentFocus = ({
...
+      if (!elm) return false;
+      const inertParent = elm.closest('[inert]');
+      return (
+        !inertParent || (!inertParent.hasAttribute('inert') && !elm.disabled)
+      );
+    };
+
+    const getActiveSteps = () => {
+      const allSteps = Array.from(document.querySelectorAll(blockClass));
+      return allSteps.filter((el) => {
```
matthewgallo: What do you think about using `.find` here instead? There's only ever one active step at a time and then when you call it you don't have to grab the first item (`getActiveSteps()` rather than `getActiveSteps()[0]`).

packages/ibm-products/src/components/CreateTearsheet/preview-components/MultiStepTearsheet.js
```diff
@@ -209,6 +212,29 @@ export const MultiStepTearsheet = (
...
+                  <>
+                    <p>first section</p>
+                    {/* <TextInput /> */}
+                  </>
+                ) : (
+                  <>
+                    <p>second section</p>
+                    {/* <TextInput /> */}
+                  </>
+                )}
```
matthewgallo: We probably don't need to keep this in the example story if you were able to verify that it works as expected now.

packages/ibm-products/src/components/CreateTearsheet/CreateTearsheetStep.tsx
```diff
@@ -305,7 +279,10 @@ export let CreateTearsheetStep = forwardRef(
     };
 
     return stepsContext ? (
-      <div ref={stepRef as RefObject<HTMLDivElement>}>
+      <div
+        ref={stepRef as RefObject<HTMLDivElement>}
+        inert={stepNumber !== stepsContext?.currentStep}
```
matthewgallo, line 284: 💡 I like this approach! Similar to what Amal added in interstitial recently too 👏

packages/ibm-products/src/components/CreateTearsheet/CreateTearsheet.test.js
```diff
@@ -270,7 +270,7 @@ describe(CreateTearsheet.displayName, () => {
     ));
 
   it('renders the second step if clicking on the next step button with onNext optional function prop and then clicks cancel button', async () => {
-    jest.useFakeTimers();
+    // jest.useFakeTimers();
```
matthewgallo: If we no longer need this, you can remove it

matthewgallo (commented):
Not sure how all of the babel changes snuck in here. Can you remove those? I really like the approach here of scoping the `blockClass` passed to the hook as well as using `inert`, cleans things up a lot within the step component!

packages/ibm-products/src/global/js/hooks/useCreateComponentFocus.js
jlongshore: Yep, I can do that... probably singular-ize the name too since it should only return 1

config/babel-preset-ibm-cloud-cognitive/index.js
```diff
@@ -25,6 +25,7 @@ module.exports = () => {
     ],
     plugins: [
       'dev-expression',
+      '@babel/plugin-transform-class-static-block',
```
matthewgallo: Do we need this plugin?

package.json
```diff
@@ -64,6 +64,7 @@
   "devDependencies": {
     "@babel/core": "^7.26.10",
     "@babel/eslint-parser": "^7.26.10",
+    "@babel/plugin-transform-class-static-block": "^7.28.3",
```
matthewgallo: Do we need this plugin?

matthewgallo:
There's just one avt test failure, possibly related to the changes here [`@avt-simulate-error-and-focus`](https://github.com/carbon-design-system/ibm-products/blob/main/e2e/components/CreateFlows/CreateTearsheet-test.avt.e2e.js#L158)

packages/ibm-products/src/components/CreateTearsheet/preview-components/MultiStepTearsheet.js
```diff
@@ -78,7 +78,7 @@ export const MultiStepTearsheet = (
   const [isInvalid, setIsInvalid] = useState(false);
   const [shouldIncludeAdditionalStep, setShouldIncludeAdditionalStep] =
     useState(false);
-
+  const [selIndex, setSelIndex] = useState(0);
```
szinta:
```suggestion
```
These values are not used anywhere

package.json
jlongshore: Yeah, I remember there being a good reason I added that... and now that some time has passed I cant remember and removing it doesn't seem to cause any warnings or errors...

davidmenendez:
@jlongshore please resolve merge conflict. thanks!

PR #8081, opened 2025-08-18
feat: implement focus trap in interstitial screen webcomponent

### LINKED ISSUE #8068: InterstitialScreen: implement focus trap in web component
Implement focus trap
Focus first focusable element

amal-k-joy:
Closes #8068 

Created a utility for focus trap in webcomponents.
Native focus trap implemented in cds-modal wouldnt work well for us since the content can rendered in shadow dom or light dom.
Instead of having the parent (cds-modal) attempt to query for children—like you might do in React—we handle this at the component level. The strategy is for child or slotted components to register their own containers with the trapFocus utility, allowing the utility to query focusable elements directly without being blocked by shadow DOM boundaries.

local storybook

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products-web-components/src/utilities/manageFocusTrap/manageFocusTrap.ts
+++ b/packages/ibm-products-web-components/src/utilities/manageFocusTrap/manageFocusTrap.ts
@@ -11,7 +11,17 @@
  * @param {NodeListOf<Element>} containers - The containers to trap focus within.
  * @returns {object} An object with a `cleanup` method to remove event listeners.
  */
-export const trapFocus = (containers) => {
+let _containers: HTMLElement[] = [];
+
+export const registerFocusableContainers = (container) => {
+  if (container) {
+    _containers.push(container);
+  }
+};
+export const clearFocusableContainers = () => {
+  _containers.length = 0;
+};
+export const trapFocus = (containers?: HTMLElement[]) => {
   const prefix = 'cds';
   const selectorTabbable = `
     a[href]:not(#start-sentinel, #end-sentinel), area[href],
@@ -80,8 +90,10 @@ export const trapFocus = (containers) => {
     ${prefix}-side-nav-menu-item,
     ${prefix}-slug
   `;
-
-  if (!containers.length) {
+  if (containers && containers.length > 0) {
+    _containers = containers;
+  }
+  if (!_containers.length) {
     return { cleanup: () => {} };
   }
 
@@ -147,7 +159,7 @@ export const trapFocus = (containers) => {
 
   //Optionally focus first element immediately
   requestAnimationFrame(() => {
-    const elements = containers
+    const elements = _containers
       .flatMap((container) => getFocusableElements(container))
       .filter(Boolean);
     getFocusTarget(elements[0])?.focus({ preventScroll: true });
@@ -159,7 +171,7 @@ export const trapFocus = (containers) => {
     }
 
     // Flatten all focusable elements from all containers
-    const elements = containers
+    const elements = _containers
       .flatMap((container) => getFocusableElements(container))
       .filter(Boolean);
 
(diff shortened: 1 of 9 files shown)


packages/ibm-products-web-components/src/utilities/manageFocusTrap/manageFocusTrap.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    // If active element is a slot
+    if (activeElement?.tagName === 'SLOT') {
+      let assigned: Element[] | undefined;
+
+      if (activeElement instanceof HTMLSlotElement) {
+        assigned = activeElement.assignedElements({ flatten: true });
+      }
+      const candidate = assigned?.find((node) =>
+        node.matches('button, [tabindex]:not([tabindex="-1"])')
```
matthewgallo, line 92: Curious about this choice of selector vs `selectorTabbable`? Could you use the `selectorTabbable` constant above (move it outside `trapFocus` so that it can be used in both places)?

packages/ibm-products-web-components/src/utilities/manageFocusTrap/manageFocusTrap.ts
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    return activeElement;
+  };
+
+  //Optionally focus first element immediately
+  requestAnimationFrame(() => {
+    const elements = _containers
+      .flatMap((container) => getFocusableElements(container))
+      .filter(Boolean);
```
matthewgallo: You could write a utility here to get these focusable elements since this same exact code/filtering is also used a few lines below as well.

matthewgallo (dismissed):
This looks really nice! I left just a few small comments and have a few questions around some edge cases:

- What if a slotted component has slots itself (content within a shadow dom)? Will that component need to register it's focusable elements as well? If that is the case, do you think we need to export this utility?
- What if the focusable elements change dynamically within a component? Does it need to "register" within a `connectedCallback` instead of `firstUpdated`?

packages/ibm-products-web-components/src/utilities/manageFocusTrap/manageFocusTrap.ts
amal-k-joy: Nice catch, moved the logic inside the function

packages/ibm-products-web-components/src/utilities/manageFocusTrap/manageFocusTrap.ts
amal-k-joy: i was worried moving `selectorTabbable` outside will move it to global scope and also it is being used only inside `trapFocus` method.

amal-k-joy:
> This looks really nice! I left just a few small comments and have a few questions around some edge cases:
> 
Thanks for looking into this
> * What if a slotted component has slots itself (content within a shadow dom)? Will that component need to register it's focusable elements as well? If that is the case, do you think we need to export this utility?

This can likely be treated as a rare edge case. Even if certain elements aren’t captured, it would only cause problems if the missing element happens to be the first or last focusable item. 
We can consider exporting this as a utility. Do you think perhaps only after trying it out in our internal components, like the tearsheet, to make sure it works as expected.?


> * What if the focusable elements change dynamically within a component? Does it need to "register" within a `connectedCallback` instead of `firstUpdated`?

Since we are registering the wrapper containers, not the individual elements, it should be able to adapt to dynamic content. Query selection is happening only on keydown event.

matthewgallo (approved):
Ok, sounds like we might be ok for now regarding the edge cases. I also just noticed `prefix` and the same set of selectors (`selectorTabbable`) are available from `@carbon/web-components/es/globals/settings`. I don't think it should block this PR, but might be nice to import directly from there.

amal-k-joy:
> Ok, sounds like we might be ok for now regarding the edge cases. I also just noticed `prefix` and the same set of selectors (`selectorTabbable`) are available from `@carbon/web-components/es/globals/settings`. I don't think it should block this PR, but might be nice to import directly from there.
@matthewgallo 
Ya,... that make sense. I have made changes. Thanks

PR #8090, opened 2025-08-20
fix(makeDraggable): improve accessibility

### LINKED ISSUE #7961: Improve accessibility in makeDraggable
I've a few suggestions related to A11y, you can consider it as out of the scope of this PR.

1. We need a engage and disengage the state between drag. Something like in `Coachmark`, after focusing on the drag-handle user needs to press the `Enter` or `Space` key to engage in dragging, and after drag needs to press the `Enter` or `Space` key to confirm the drag position; otherwise press the `Esc` key to cancel the drag.

https://github.com/user-attachments/assets/ac57296f-9a32-4834-a110-98fd9560d6c1

2. When engaging, show it in the focus outline (like in `Coachmark`).
3. Also, provide appropriate `aria-roledescription` and `aria-label` for the dag-handle button. You can refer to the [dnd-kit example](https://master--5fc05e08a4a65d0021ae0bf2.chromatic.com/?path=/story/core-draggable-hooks-usedraggable--basic-setup).

_Originally posted by @makafsal in https://github.com/carbon-design-system/ibm-products/pull/7908#pullrequestreview-3078220708_

anamikaanu96:
Closes #7961 

Improved accessibility by introducing additional features

1. Implemented engage/disengage drag state toggled via the Enter key.
2. Applied styling to indicate the engaged state.
3. Added accessibility support with aria-label and aria-describedby attributes.
4. Applied the same features in composable coachmark since it uses makeDraggable utility for floating variant.

#### What did you change? 
packages/ibm-products/src/global/js/utils/makeDraggable/makeDraggable.ts
packages/ibm-products/src/global/js/utils/makeDraggable/makeDraggable.stories.tsx
packages/ibm-products/src/global/js/utils/makeDraggable/_storybook-styles.scss
packages/ibm-products/src/components/Coachmark/next/Coachmark/ContentHeader.tsx
packages/ibm-products-styles/src/components/Coachmark/_coachmark.scss

#### PR Checklist

our contributing docs.


packages/ibm-products/src/global/js/utils/makeDraggable/makeDraggable.stories.tsx
```diff
@@ -29,11 +29,29 @@ const DraggableDiv = () => {
...
   useEffect(() => {
     if (dialogRef.current && headerRef.current && dragRef.current) {
-      makeDraggable({
+      const draggable = makeDraggable({
         el: dialogRef.current,
         dragHandle: headerRef.current,
         focusableDragHandle: dragRef.current,
       });
+
+      draggable.subscribe((val) => {
```
amal-k-joy: Hi @anamikaanu96 ,
This looks good.
As we discussed, may be  we can dispatch custom events like `dragstart` and `drag end`, which is a common approach and align more with HTML drag API, instead of notifying to the subscribed methods.

amal-k-joy (dismissed):
LGTM

packages/ibm-products/src/components/Coachmark/next/Coachmark/ContentHeader.tsx
```diff
@@ -72,6 +72,25 @@ const ContentHeader = forwardRef<HTMLDivElement, ContentHeaderProps>(
           dragHandle: handleRef.current,
           focusableDragHandle: dragRef.current,
         });
+
+        const onDragStart = () => {
+          if (contentRef) {
+            contentRef.classList.add(`${contentHeaderBlockClass}--is-dragging`);
+            contentRef.setAttribute('aria-label', 'Coachmark is being dragged');
```
matthewgallo, line 79: I think the `ContentHeader` should accept an `aria-label` or some additional prop so that this text has the ability to be translated.

packages/ibm-products/src/global/js/utils/makeDraggable/makeDraggable.ts
```diff
@@ -49,7 +49,23 @@ export function makeDraggable({
   let offsetX = 0;
   let offsetY = 0;
 
+  function dispatch(type: string, detail: any = {}) {
+    el.dispatchEvent(new CustomEvent(type, { detail, bubbles: true }));
```
matthewgallo, line 53: I like this approach, would love a demo about this in our next huddle!

matthewgallo (dismissed):
We can address the aria label in a follow up, don't think it needs to block these changes.

szinta (approved):

PR #8105, opened 2025-08-22
feat(conditionbuilder): option to disable specific property dynamically

### LINKED ISSUE #7983: [Feature Request]: Ability to disable specific options/properties in the dropdowns in the condition builder component.
### The problem

Ability to disable specific options in the dropdowns in the condition builder component. 

Image

### The solution

Add a `disabled` prop for the individual option/properties. 

### Examples

_No response_

### Package

I'm not sure

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

amal-k-joy:
Closes #7983 

- Added isDisabled callback option to  dynamically disable any specific properties.
- Added isHidden callback option to  dynamically hide any specific properties (as an addon functionality)
- Added story
- Updated test and doc

#### What did you change?

```
properties: [
    {
      id: 'continent',
      label: 'Continent',
      icon: Earth,
      type: 'option',
      description: 'Continent',
      isDisabled: ({ conditionState, group }) => {
       
        return true or false;
      },
       isHidden: ({ conditionState, group }) => {
       
        return true or false;
      },
      config: {
        options: [
          ...
        ],
      },
    },
    ...]
```

local storybook

#### PR Checklist

our contributing docs.


packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderItem/ConditionBuilderItemOption/ItemOption.tsx
```diff
@@ -67,7 +73,10 @@ export const ItemOption = ({
   }, [allOptions]);
 
   const onClickHandler = (evt, option) => {
-    onChange(option.id, evt);
+    if (!evt.currentTarget.hasAttribute('aria-disabled')) {
+      onChange(option.id, evt);
+    }
+    return;
```
jlongshore: Do we need a return here?

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderItem/ConditionBuilderItemOption/ItemOption.tsx
```diff
@@ -113,7 +122,16 @@ export const ItemOption = ({
         {filteredItems?.map((option) => {
           const isSelected = selection === option.id;
           const Icon = (option as option).icon;
-
+          let disabled, hidden;
```
jlongshore: I would type these rather than have them run as `any`:
```ts
let disabled:boolean, hidden:boolean;
```

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderItem/ConditionBuilderItemOption/ItemOption.tsx
```diff
@@ -125,6 +143,8 @@ export const ItemOption = ({
                 return false;
               }}
               onClick={(evt) => onClickHandler(evt, option)}
+              {...(disabled ? { 'aria-disabled': 'true' } : {})}
+              {...(hidden ? { 'aria-hidden': 'true' } : {})}
```
jlongshore, line 146: with those being typed as boolean, you could shorten this to:

```ts
aria-disabled={disabled}
aria-hidden={hidden}
```

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.tsx
```diff
@@ -228,6 +228,8 @@ ConditionBuilder.propTypes = {
           'custom',
         ]).isRequired,
         description: PropTypes.string, //will be displayed on hover of property field
+        isDisabled: PropTypes.func,
+        isHidden: PropTypes.func,
```
davidmenendez: minor comment: can we rename these? using `is` is typically associated with just a boolean value, but here it's being used as a getter function. maybe something like `getDisabledState` or `getIsHidden` would make more sense here.

packages/ibm-products/src/components/ConditionBuilder/assets/sampleInput.js
```diff
@@ -630,9 +630,32 @@ function modifyPropertiesWithoutCustomOperators(inputData) {
...
+        ...property,
+        isHidden: () => true,
+      };
+    }
+
+    return property;
+  });
+
+  return { ...inputData, properties: newProperties };
+};
```
jlongshore: Could this be shortened to:

```ts
const newProperties = inputData.properties.map((property, index) => {
  return {
      ...property,
      isDisabled: () => index === 1 || index === 3,
      isHidden: () => index === 2
    };
});
```

packages/ibm-products/src/components/ConditionBuilder/utils/handleKeyboardEvents.js
```diff
@@ -53,6 +53,12 @@ const excludeKeyPress = (evt) => {
   );
 };
 
+const getVisibleOptions = (parentContainer) => {
+  return Array.from(parentContainer.querySelectorAll(`[role="option"]`))
+    .filter((el) => !el.hasAttribute('aria-disabled'))
+    .filter((el) => !el.hasAttribute('aria-hidden'));
```
jlongshore: Could we handle this in 1 filter pass? Something like:

```js
.filter((el) => !el.hasAttribute('aria-disabled') && !el.hasAttribute('aria-hidden'))
```

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderItem/ConditionBuilderItemOption/ItemOption.tsx
```diff
@@ -113,7 +122,16 @@ export const ItemOption = ({
         {filteredItems?.map((option) => {
           const isSelected = selection === option.id;
           const Icon = (option as option).icon;
-
+          let disabled, hidden;
+          if (
+            'isDisabled' in option &&
```
davidmenendez: this line should be unnecessary. `typeof option.isDisabled === 'function'` should be sufficient here.

you could also just do something like `const disabled = option?.isDisabled();`

packages/ibm-products/src/components/ConditionBuilder/utils/handleKeyboardEvents.js
davidmenendez: could this be combined into a single `filter` check? `!el.hasAttribute(a) && !el.hasAttribute(b)`? that way it doesn't has to run through the array twice.

packages/ibm-products/src/components/ConditionBuilder/utils/handleKeyboardEvents.js
```diff
@@ -109,18 +114,17 @@ const handleKeyPressForPopover = (
...
-              index,
-              allElements,
-              null,
-              null,
-              conditionBuilderRef
-            );
-          });
+        allItems = getVisibleOptions(parentContainer);
+        allItems.forEach((eachElem, index, allElements) => {
+          traverseClockVise(
```
davidmenendez: i _think_ this should be `traverseClosewise`

davidmenendez (changes requested):
looks great! have a few comments & questions 👍

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderItem/ConditionBuilderItemOption/ItemOption.tsx
amal-k-joy: You are right, but I went with this approach because I don’t want the attributes showing up on the element at all when they’re false. That way, it’s cleaner to target in both CSS and JavaScript.

packages/ibm-products/src/components/ConditionBuilder/assets/sampleInput.js
amal-k-joy: Yes, that works

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilder.tsx
amal-k-joy: Thats a good observation, I go with `getIsHidden` and getIsDisabled

packages/ibm-products/src/components/ConditionBuilder/ConditionBuilderItem/ConditionBuilderItemOption/ItemOption.tsx
amal-k-joy: refactored

davidmenendez (dismissed):

jlongshore (approved):

PR #8125, opened 2025-08-26
feat: condition builder read only state

### LINKED ISSUE #7984: [Feature Request]: Ability to make the Condition Builder component read only.
### The problem

The condition builder should be read only. The user should not be able to interactive with the condition builder. 

Image

### The solution

Add a `readOnly` prop for the condition builder. 

### Examples

_No response_

### Package

I'm not sure

### Application/PAL

_No response_

### Business priority

None

### Available extra resources

_No response_

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon/blob/f555616971a03fd454c0f4daea184adf41fff05b/.github/CODE_OF_CONDUCT.md)

amal-k-joy:
Closes #7984 

This wil make condition builder as readonly

#### What did you change?
Added a new prop readOnly
Added test
Updated doc

storybook

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products-styles/src/components/ConditionBuilder/styles/_conditionBuilderCondition.scss
+++ b/packages/ibm-products-styles/src/components/ConditionBuilder/styles/_conditionBuilderCondition.scss
@@ -88,3 +88,23 @@ $block-class: #{c4p-settings.$pkg-prefix}--condition-builder;
   > .#{$block-class}__groupConnector {
   margin-block-end: 0;
 }
+.#{$block-class}__readOnly {
+  .#{$block-class}__conditionInnerBlock,
+  .#{$block-class}__connector,
+  .#{$block-class}__groupConnector {
+    border: 1px solid $border-subtle-00;
+    background: transparent;
+
+    button {
+      background: transparent;
+
+      svg {
+        color: $text-disabled;
+      }
+    }
+  }
+}
+
+.#{$block-class}__conditionInnerBlock {
+  display: flex;
+}
--- a/packages/ibm-products/src/components/ConditionBuilder/ConditionBlock/ConditionBlock.tsx
+++ b/packages/ibm-products/src/components/ConditionBuilder/ConditionBlock/ConditionBlock.tsx
@@ -90,7 +90,7 @@ const ConditionBlock = (props: ConditionBlockProps) => {
     isLastCondition,
   } = props;
   const { property, value, operator } = condition;
-  const { inputConfig, variant, conditionBuilderRef } = useContext(
+  const { inputConfig, variant, conditionBuilderRef, readOnly } = useContext(
     ConditionBuilderContext
   );
 
@@ -276,7 +276,7 @@ const ConditionBlock = (props: ConditionBlockProps) => {
           label={group.statement}
           title={conditionText}
           data-name="connectorField"
-          popOverClassName={`${blockClass}__gap`}
+          popOverClassName={`${blockClass}__gap ${blockClass}__connector`}
           className={`${blockClass}__statement-button`}
           tabIndex={0}
         >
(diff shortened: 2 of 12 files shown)


packages/ibm-products-styles/src/components/ConditionBuilder/styles/_conditionBuilderCondition.scss
```diff
@@ -88,3 +88,23 @@ $block-class: #{c4p-settings.$pkg-prefix}--condition-builder;
   > .#{$block-class}__groupConnector {
   margin-block-end: 0;
 }
+.#{$block-class}__readOnly {
```
szinta: Several class names in this component use camelCase. Consider using kebab-case for class names instead of camelCase to follow BEM conventions.

szinta (commented):
I think, It would be useful to add a new story to showcase the readonly state. Also, could you please review the behavior of Actions in readonly mode? Currently, the close and add icons are still visible in the readonly mode of Actions, whereas they're hidden in the readonly state of Conditions.

packages/ibm-products-styles/src/components/ConditionBuilder/styles/_conditionBuilderCondition.scss
amal-k-joy: Good catch. I have corrected to kebab case for new classes  for now. I can do a complete review and open a separate PR for this.

amal-k-joy:
> I think, It would be useful to add a new story to showcase the readonly state. Also, could you please review the behavior of Actions in readonly mode? Currently, the close and add icons are still visible in the readonly mode of Actions, whereas they're hidden in the readonly state of Conditions.

Good point about the behavior of the actions section when readOnly is enabled, have corrected it.
 I also considered creating a separate story for the read-only state at first, but it started to feel a bit overwhelming — we already have nine stories, and adding a couple more just for read-only might be too much. Since adopters can easily experience the read-only behavior through prop interactions, it seems better to keep it that way.

RichKummer:
Took a look through and looks like the tokens are in! Found a few issues that could be addressed:

1. When I toggle `read-only` on and off, I noticed that the border looks to be shifting the cell elements a bit. Is there a way we can adjust the border so we don't have that `1px` shift? The Figma equivalent would be setting the border to `inside`.

https://github.com/user-attachments/assets/139b0e3e-8cd4-47ab-b0a2-a562238eedf3

2. It looks like the icon is set to `$text-disabled` but it should be `$icon-disabled`.
3. The cursor should remain as arrow when hovering over the cell, but currently it uses the hand (`pointer`). [Ref Carbon docs](https://carbondesignsystem.com/patterns/read-only-states-pattern/#interactions)

(Need to take another look but recording my notes here)

amal-k-joy:
> Took a look through and looks like the tokens are in! Found a few issues that could be addressed:
> 
> 1. When I toggle `read-only` on and off, I noticed that the border looks to be shifting the cell elements a bit. Is there a way we can adjust the border so we don't have that `1px` shift? The Figma equivalent would be setting the border to `inside`.
> 
>  Screen.Recording.2025-08-28.at.11.45.02.AM.mov 
> 2. It looks like the icon is set to `$text-disabled` but it should be `$icon-disabled`.
> 3. The cursor should remain as arrow when hovering over the cell, but currently it uses the hand (`pointer`). [Ref Carbon docs](https://carbondesignsystem.com/patterns/read-only-states-pattern/#interactions)
> 
> (Need to take another look but recording my notes here)

Thanks Rich for looking into this.
Made the changes

szinta (dismissed):
LGTM

RichKummer:
The changes look great @amal-k-joy !

I found one last issue in the _with Actions (Hierarchical)_ – it looks like when switched to `read-only`, the Action button will extend to fill the container:
Screenshot 2025-08-29 at 11 12 40 AM

amal-k-joy:
> The changes look great @amal-k-joy !
> 
> I found one last issue in the _with Actions (Hierarchical)_ – it looks like when switched to `read-only`, the Action button will extend to fill the container: Screenshot 2025-08-29 at 11 12 40 AM

@RichKummer ,
ohh thanks , I really missed it.

PR #8153, opened 2025-08-27
fix: remove requiredIf props helpers

davidmenendez:


matthewgallo (approved):

packages/ibm-products/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.js
```diff
@@ -389,9 +389,7 @@ BreadcrumbWithOverflow.propTypes = {
       /**
        * A string based alternative to the children, required only if children is not of type string
        */
-      title: PropTypes.string.isRequired.if(
-        ({ label }) => typeof label !== 'string'
-      ),
+      title: PropTypes.string,
```
jlongshore, line 392: Do we want to amend the comment here about it being conditionally required?

jlongshore (approved):

PR #8154, opened 2025-08-27
build(web-components): exclude `--cds` values in web component custom prefix build output

matthewgallo:
Fixes the same issue resolved in [this PR](https://github.com/carbon-design-system/carbon/pull/20355) from system squad, but for our web components package.

#### What did you change?
`@carbon/web-components` version bump and updated build to avoid replacing `--cds` values.

While upgrading `@carbon/web-components`, we had to refactor how we use icons in the web components package. This is done now through the `iconLoader` util from `@carbon/web-components/es/globals/internal/icon-loader.js` and importing the icons directly from `@carbon/icons`. I've addressed those changes in this PR as well.
Double checked build output
#### PR Checklist

our contributing docs.


davidmenendez (dismissed):

davidmenendez:
hmmm looks like build failed https://github.com/carbon-design-system/ibm-products/actions/runs/17275779580/job/49031959819?pr=8154#step:6:715

matthewgallo:
We have to make a few more changes since `@carbon/web-components` doesn't include the transformed icons anymore. I'll add the `Do not merge` label until I wrap that up

packages/ibm-products-web-components/tasks/build.js
```diff
@@ -187,13 +187,13 @@ async function postBuild() {
...
-        content = content.replace(/cds/g, 'cds-custom');
-        content = content.replace(
+        const content = await fs.promises.readFile(file, 'utf8');
+        let updatedContent = content.replace(/(?<!--)cds/g, 'cds-custom');
+        updatedContent = content.replace(
           /import\s+['"]@carbon\/web-components\/es\/components\/(.*?)['"]/g,
           "import '@carbon/web-components/es-custom/components/$1'"
         );
-        await fs.promises.writeFile(file, content);
+        await fs.promises.writeFile(file, updatedContent);
```
annawen1: When we do `replace()`, the original string `content` doesn't get mutated so I believe the changes coming from line 191 where we swap out the `cds` to `cds-custom` get overwritten and don't get applied. We should revert back to using the `content` variable

packages/ibm-products-web-components/src/globals/internal/icon-loader-utils.ts
```diff
@@ -0,0 +1,105 @@
+/**
```
annawen1: oh hmm these files are meant to be imported from `@carbon/web-components` even though they are listed under the `internal` folder

`import { iconLoader } from "@carbon/web-components/es/globals/internal/icon-loader.js";`

matthewgallo: Ohhh haha! I didn't know those are included in the build. I'll make some changes to fix that.

matthewgallo: I was wondering how to best handle the examples that are using the `iconLoader`, but importing directly from `@carbon/web-components` will solve that problem too

annawen1 (dismissed):
LGTM! I think we can go ahead and remove the copied over icon loader util files!

jlongshore (approved):

PR #8205, opened 2025-09-05
refactor(PageHeader): change api for tags prop in tab bar

### LINKED ISSUE #8087: [PageHeader]: tab bar `tags` api needs to be more flexible/composable
The current implementation for the `tags` prop within the new `PageHeader` tab bar, only allows for default tags to be rendered. There isn't a way to render an `OperationalTag`. We should change the structure of the `tags` prop to be something more like the following to allow for more flexibility:

```jsx
const tags = [
 {
    text: 'My tag',
    tag: <Tag {...tagProps}>My tag</Tag>
  },
 {
    text: 'My other tag',
    tag: <OperationalTag {...operationalTagProps}>My other tag</OperationalTag>
  },
]
```

The would allow us to introduce added flexibility in the tag rendering, but also in the popover content where it currently maps the hidden tags manually into a `<Tag>`.

_Update_
After reviewing further, this feels like a similar use case to the breadcrumb overflow subcomponent. We could better support this customization and also align more with the other page header subcomponent apis if we follow a similar approach.

Latest idea for highest composability/flexibility:
```jsx
<PageHeader.TabBar
  tags={
    <PageHeader.TagOverflow
      renderOverflowTag={(hiddenItems, handleOverflowClick) => (
        <OperationalTag
          onClick={handleOverflowClick}
          text={`+${hiddenItems.length}`}
        />)
    }
    renderPopoverContent={(hiddenItems) => (
      <>
        {hiddenItems.map(item => (
           // render hidden items in popover
        ))}
      </>
    )}
  }
>
      <Tag>Hello</Tag>
      <OperationalTag>Hello</OperationalTag>
      <Tag>Hello</Tag>
      <OperationalTag>Hello</OperationalTag>
      <Tag>Hello</Tag>
      <OperationalTag>Hello</OperationalTag>
      <Tag>Hello</Tag>
      <OperationalTag>Hello</OperationalTag>
    </PageHeader.TagOverflow>
  }
>
  <TabList>
    <Tab>Tab 1</Tab>
    <Tab>Tab 2</Tab>
  </TabList>
</PageHeader.TabBar>
```

matthewgallo:
Closes #8087 

This PR refactors the current api for the `tags` prop in the tab bar of the new page header. The previous implementation was too opinionated, forced each item in the `tags` array to be a `Tag`, but we should allow for `Operational` tags as well (or any other type of tag that may exist in the future). The changes in this PR allow you to pass any children via a new sub component `PageHeader.TagOverflow` and introduces two render props that give the consumer the rendering responsibility for the overflow tag and the contents within the popover.

#### What did you change?
```
packages/ibm-products-styles/src/components/PageHeader/_page-header.scss
packages/ibm-products/src/components/PageHeader/next/PageHeader.mdx
packages/ibm-products/src/components/PageHeader/next/PageHeader.stories.jsx
packages/ibm-products/src/components/PageHeader/next/PageHeader.tsx
packages/ibm-products/src/components/PageHeader/next/_storybook-styles.scss
packages/ibm-products/src/components/PageHeader/next/index.ts
```
Storybook, updated docs
#### PR Checklist

our contributing docs.


matthewgallo:
Have to re-work some of the existing tests before this is ready

davidmenendez (dismissed):

packages/ibm-products-styles/src/components/PageHeader/_page-header.scss
```diff
@@ -966,6 +968,29 @@ $duration: 1000ms;
...
 
+  .#{$block-class}--tag-overflow-container {
+    align-content: center;
+    text-align: end;
+    @include breakpoint(sm) {
+      transform: translateX(-1 * 24px);
+    }
+    @include breakpoint(md) {
+      transform: translateX(-1 * 8px);
+    }
```
jlongshore: Is it just our convention to do the math on these, or can we just have `-24px` and `-8px`?

packages/ibm-products-styles/src/components/PageHeader/_page-header.scss
```diff
@@ -966,6 +968,29 @@ $duration: 1000ms;
...
+  }
+
+  .#{$block-class}--tag-overflow-popover__hidden {
+    visibility: hidden;
+  }
+
+  .#{$block-class}--tag-overflow-container__has-no-hidden-items {
+    /* stylelint-disable-next-line carbon/layout-use */
+    transform: translateX(32px);
```
jlongshore: can we use the spacing token here? 
```
transform: translateX($spacing-07);
```

packages/ibm-products/src/components/PageHeader/next/PageHeader.tsx
```diff
@@ -817,86 +834,56 @@ const PageHeaderTabBar = React.forwardRef<
...
-    </div>
-  );
+  useEffect(() => {
+    localOverflowHandler({
+      container: tagsContainerRef.current!,
+      onChange: (_, hidden) => {
+        setHiddenTags(hidden);
+      },
+    });
+  }, []);
```
jlongshore, line 849: Does `localOverflowHandler` need to be in the dep array here? Or another local function memo-ized with a useCallback that would call that or something?

packages/ibm-products-styles/src/components/PageHeader/_page-header.scss
matthewgallo: I think I just did it that way because that's how we'd typically do it when including a spacing variable. I included those spacing variables in https://github.com/carbon-design-system/ibm-products/pull/8205/commits/39ba7f9e0f7778c0d14181fe40947402331ff35b

packages/ibm-products/src/components/PageHeader/next/PageHeader.tsx
matthewgallo: No, don't think so because the `overflowHandler` has a resize observer internally that will return the visible and hidden items from the `onChange` callback as they change.

jlongshore (approved):
LGTM

Issue #8253, opened 2025-09-16
Plan for promoting `preview__PageHeader` to stable


matthewgallo:
A question was asked today around a timeline for promoting the page header currently in preview to stable. There are a handful of reviews that will be required before we can do this, as well as planning out the deprecation of the already currently stable page header.

- Final design review
- Final development api review
- Verify test coverage
- How are we going to handle the old page header?
  - Especially when we remove the status from the new page header export? How will we export without the status for the new page header but also (ideally) not break the api for the existing page header?

PR #8260, opened 2025-09-18
test(PageHeader): increases coverage, removes .only in test

matthewgallo:
Addresses part of #8253

Increases preview page header test coverage, there's still some coverage specifically around testing intersection observers that I'll save for later. There were also a handful of `.only`s that I mistakely left in the test file, those have been removed so all tests run.

image

#### What did you change?
```
packages/ibm-products/src/components/PageHeader/next/PageHeader.test.js
packages/ibm-products/src/components/PageHeader/next/PageHeader.tsx
packages/ibm-products/src/components/PageHeader/next/overflowHandler.ts
```
Manually ran tests
#### PR Checklist

- [X] Wrote passing tests that cover this change

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/PageHeader/next/PageHeader.test.js
+++ b/packages/ibm-products/src/components/PageHeader/next/PageHeader.test.js
@@ -50,6 +50,52 @@ const carbonPrefix = 'cds';
 
 describe('PageHeader', () => {
   describe('export configuration', () => {
+    let savedObserverCb;
+
+    beforeEach(() => {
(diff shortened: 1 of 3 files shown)


davidmenendez (approved):

matthewgallo:
@amal-k-joy, I did this because the handler util from carbon utils does not include the margins in the calculation of element sizes. This causes jumping and unexpected behavior specifically with the breadcrumb items since they have 8px of margin each which isn't included in the element size. Ideally, we contribute the enhancements made via the local util back to core. I'll make sure we have an issue open to do that.

PR #8188, opened 2025-09-03
[`Do not merge`] Auto generated Figma tsx

RichKummer:
Closes #

{{short description}}

#### PR Checklist

As the author of this PR, before marking ready for review, confirm you:

- [ ] Reviewed every line of the diff
- [ ] Updated documentation and storybook examples
- [ ] Wrote passing tests that cover this change
- [ ] Addressed any impact on accessibility (a11y)
- [ ] Tested for cross-browser consistency
- [ ] Validated that this code is ready for review and status checks should pass

More details can be found in the [pull request](./CONTRIBUTING.md) section of
our contributing docs.

PR #8264, opened 2025-09-18
feat(UserAvatar): code connect

### LINKED ISSUE #8187: Figma Code Connect - React (UserAvatar)
1. Go through [quick start guide](https://www.figma.com/code-connect-docs/quickstart-guide/#install-the-code-connect-command-line-tool)
2. Create figma.config.json at project root
3. Use the cli tool to scaffold the component's code connect Component.figma.tsx file:
`npx figma connect --token=PERSONAL_ACCESS_TOKEN`
4. Correct any issues in the generated prop naming, there may not be a one to one match between the component's props in code vs in figma
5. Publish to figma library --> `npx figma connect publish --token=PERSONAL_ACCESS_TOKEN`
6. Document any discrepancies within figma kit 

[Code Connect demo file](https://www.figma.com/design/LA6Z4lEowU7B1z3An6bSYz/Code-Connect-demo?node-id=0-1&t=OLyQdsRoaQhkCYZC-1)

matthewgallo:
Closes #8187

Sets up code connect for `UserAvatar`

#### What did you change?
~Add code connect package (only because https://github.com/carbon-design-system/ibm-products/pull/8188 hasn't been merged yet) and~ new code connect file for `UserAvatar`.
Verified missing variants with @RichKummer and got the correct url to the figma component.

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/UserAvatar/UserAvatar.figma.tsx
+++ b/packages/ibm-products/src/components/UserAvatar/UserAvatar.figma.tsx
@@ -18,72 +18,105 @@ const sharedProps = {
     Medium: 'md',
     Small: 'sm',
   }),
+
+  tooltipProps: figma.nestedProps('Tooltip', {
+    text: figma.string('Tooltip text'),
+    tooltipAlignment: figma.enum('🚫 Position', {
+      Top: 'top',
+      Right: 'right',
+      Bottom: 'bottom',
+      Left: 'left',
+    }),
+  }),
+
+  bgProps: figma.nestedProps('Background', {
+    color: figma.enum('Color', {
+      Cyan: figma.enum('Sequence', {
+        Primary: 'order-1-cyan',
+        Secondary: 'order-7-cyan',
+      }),
+      Gray: figma.enum('Sequence', {
+        Primary: 'order-2-gray',
+        Secondary: 'order-8-gray',
+      }),
+      Green: figma.enum('Sequence', {
+        Primary: 'order-3-green',
+        Secondary: 'order-9-green',
+      }),
+      Magenta: figma.enum('Sequence', {
+        Primary: 'order-4-magenta',
+        Secondary: 'order-10-magenta',
+      }),
+      Purple: figma.enum('Sequence', {
+        Primary: 'order-5-purple',
+        Secondary: 'order-11-purple',
+      }),
+      Teal: figma.enum('Sequence', {
+        Primary: 'order-6-teal',
+        Secondary: 'order-12-teal',
+      }),
+    }),
+  }),
 };
 
-figma.connect(
-  UserAvatar,
-  'https://www.figma.com/design/0F9dKH2abAd7gSfvnacfWf/-v11--IBM-Products-%E2%80%93-Carbon-Design-System?node-id=15368-59379&t=lbewdWdJ4JB5izcw-4',
-  {
-    variant: { Type: 'Image' },
-    props: sharedProps,
-    example: (props) => (
-      <UserAvatar
-        image="/path/to/image"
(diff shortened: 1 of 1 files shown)


packages/ibm-products/src/components/UserAvatar/UserAvatar.figma.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { UserAvatar } from './UserAvatar';
+import figma from '@figma/code-connect';
+
+const instances = [
+  {
+    type: 'Image, xl',
```
szinta: it looks like the props definition allows users to select different sizes, so do we need to explicitly include the term `xl` in the type value?  Can we simplify the type options to just Image, Initials, and Icon?

packages/ibm-products/src/components/UserAvatar/UserAvatar.figma.tsx
```diff
@@ -0,0 +1,52 @@
...
+    name: 'Miles Davis',
+  },
+];
+
+instances.forEach(({ url, type, ...rest }) => {
+  figma.connect(UserAvatar, url, {
+    variant: { Type: type },
+    props: {
+      ...rest,
+      name: figma.string('Initials text'),
```
szinta: Do we need to override the name attribute here? It’s already being set via the `...rest` spread operator for each type, so I don’t think we need to assign a default value for all types explicitly.

packages/ibm-products/src/components/UserAvatar/UserAvatar.figma.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const instances = [
+  {
+    type: 'Image, xl',
+    url: 'https://www.figma.com/design/0F9dKH2abAd7gSfvnacfWf/-v11--IBM-Products-%E2%80%93-Carbon-Design-System?node-id=15368-59380',
+    name: 'Miles Davis',
+  },
+  {
+    type: 'Initials, xl',
+    url: 'https://www.figma.com/design/0F9dKH2abAd7gSfvnacfWf/-v11--IBM-Products-%E2%80%93-Carbon-Design-System?node-id=15368-59389',
```
szinta: The only difference in this URL across components is the `node-id` query parameter value. Instead of repeating this long URL in multiple places, could we consider setting it into a config variable for better maintainability?

packages/ibm-products/src/components/UserAvatar/UserAvatar.figma.tsx
matthewgallo: Made some changes in https://github.com/carbon-design-system/ibm-products/pull/8264/commits/441a04ac2aadbbbbb0d52b5f966256df82f9ee50 so that the size is just "connected" from the specified `figma.enum....` declaration

packages/ibm-products/src/components/UserAvatar/UserAvatar.figma.tsx
matthewgallo: You're right, this was incorrect previously. Should be fixed in https://github.com/carbon-design-system/ibm-products/pull/8264/commits/441a04ac2aadbbbbb0d52b5f966256df82f9ee50

packages/ibm-products/src/components/UserAvatar/UserAvatar.figma.tsx
matthewgallo: This is also covered in the refactoring I did, only have two urls now. One that links to an image variant and another for the icon variant.

szinta (approved):

PR #8294, opened 2025-09-23
fix(guidebanner): add prop to control open state

### LINKED ISSUE #8074: [Feature Request]: prop to set collapsed state for Guide Banner
### The problem

The Guide Banner component internalizes the `isCollapsed` state within the component. In our use case, we would like to have a `collapsible` Guide Banner, that is by default in the open state. This is currently impossible, see code [reference](https://github.com/carbon-design-system/ibm-products/blob/8da4133e9680de00a28d6c857caf789f5ab7de8a/packages/ibm-products/src/components/Guidebanner/Guidebanner.tsx#L115)

Image

### The solution

Many solutions are possible. One would expose a prop for the default `isCollapsed` state so that we can set the initial open/closed state for the Guide Banner. 

### Design link

_No response_

### Links to other materials

_No response_

### Owner/main maintainer(s)

Paul Balkin

_No response_

### Product/offering

_No response_

### Business priority

None

### Code of Conduct

anamikaanu96:
Closes #8074

Add prop to allow component's open state to be controlled.

#### PR Checklist

our contributing docs.


packages/ibm-products/src/components/Guidebanner/Guidebanner.tsx
```diff
@@ -106,17 +111,23 @@ export let Guidebanner = React.forwardRef<HTMLDivElement, GuidebannerProps>(
...
     const [showNavigation, setShowNavigation] = useState(false);
-    const [isCollapsed, setIsCollapsed] = useState(collapsible ? true : false);
+    const [isCollapsed, setIsCollapsed] = useState(!open);
 
     const handleClickToggle = () => {
-      setIsCollapsed((prevState) => !prevState);
+      setIsCollapsed((prev) => !prev);
     };
+
+    useEffect(() => {
```
davidmenendez: this effect is unnecessary. if we are changing it to rely on a prop like `open` to control that state then we don't need to keep track of it on our end. we can replace all the uses of `isCollapsed` with `open` and then we can take in another prop to handle the on close method so the user can control the open state on their end. `setIsCollapsed` inside `handleClickToggle` can be removed and replace with something like an `onToggle` prop.

packages/ibm-products/src/components/Guidebanner/Guidebanner.tsx
```diff
@@ -74,6 +75,22 @@ export interface GuidebannerProps {
...
+  /**
+   * A handler for managing the controlled state of open prop. If not passed the open prop will not be honored and an uncontrolled state will be used.
+   */
+  onChange?: (value: boolean) => void;
+
+  /**
+   * Provide a function which will be called each time the user
+   * interacts with the toggle.
+   */
+  onToggle?: (value: boolean) => void;
```
davidmenendez: adding `useControllableState` is actually a better implementation for this use case. well done. since we're sticking with that i think we can remove `onToggle` as it has redundant functionality to `onChange`. if the user needs a callback function for opening and closing the banner then they should be able to handle using `onChange` with a controlled state.

davidmenendez (dismissed):

szinta (approved):

PR #8322, opened 2025-09-25
fix: remove truncation styling from sidepanel title

### LINKED ISSUE #8233: [SidePanel]: Update title to use TruncatedText (React and wc)
Currently `SidePanel` is using `title` to handle text overflow for long titles. It should be removed and the truncation should be handled with `TruncatedText`

Acceptance criteria

- [x] Remove all usage of `title` attribute from `SidePanel`
- [ ] Implement `TruncatedText` component to handle the title
- [x] Make necessary changes to styles and code to verify working implementation

davidmenendez:
Closes #8233

Originally this PR was intended to fix the issue by using the `TruncatedText` component, but I ran into issues there. Due to the the way the animation functionality is implemented, the `TruncatedText` functionality introduces visual bugs. The video below demonstrates the problem:

https://github.com/user-attachments/assets/d99c84a8-3bf3-4136-beb7-1384c96823f6

In this video, the text in the red box represents the initial state of the title text and the blue box text represents the state of the title when you scroll down.

When you hover over the visible title, the popover only fires when your cursor is in an area that's not being covered by the hidden title. The hidden title is only hidden via opacity, so when the cursor is hovering over both the hidden and visible titles simultaneously, the hidden popover is being fired because it has a higher `z-index`. The video below demonstrates how the hidden title gets in the way of the visible title.

https://github.com/user-attachments/assets/2fb89e20-6baa-4b56-a92f-72cf2419cfe4

Because this greatly complicates the existing scrolling animation functionality, I decided the best and easiest solution was to simply remove the truncation styling. Since the "need" for truncation in UI is often the result of a failure to properly handle displaying information, I would argue that this is a case of C4P trying too solve a complicated problem that can simply be resolved by the user not implementing a long title.


davidmenendez:
tagging @matthewgallo because this is a fun one that you might want to weigh in on 😂

amal-k-joy:
@davidmenendez @matthewgallo 
The side panel header animation has an edge case: when the header’s content is scrollable in the expanded state but becomes too small to scroll once collapsed, the scroll events stop firing because there’s no scrollable area left.

I encountered this same issue in the composable tearsheet when using scroll listeners or intersection observers — once the content lost its scroll, the header could no longer respond to scroll-based triggers.

To work around this, I implemented a completely different approach for the collapsible header there: instead of relying on scroll events, I listened to mouse and wheel events directly and detected upward or downward movement to determine the header’s collapsed or expanded state. 

[Deploy preview](https://deploy-preview-8337--carbon-for-ibm-products.netlify.app/?path=/story/preview-tearsheet--default) and [code](https://github.com/amal-k-joy/ibm-products/blob/tearsheet_initial/packages/ibm-products/src/components/Tearsheet/next/TearsheetBody.tsx#L81)

PR #8373, opened 2025-10-03
feat(EmptyState): integration of open in Stackblitz

### LINKED ISSUE #8267: [`Empty states`]: Replace codesandbox / stackblitz example links with `Open in stackblitz`
Some of the empty state components have 404 links for code sandbox and stackblitz. We should replace those links in the storybook docs with the new `Open in stackblitz` integration. This should be done for the following:

- EmptyState
- ErrorEmptyState
- NoDataEmptyState
- NoTagsEmptyState
- NotFoundEmptyState
- NotificationsEmptyState
- UnauthorizedEmptyState

anamikaanu96:
Closes #8267 

Integration of `Open in Stackblitz ` in Overview page of below components:

- EmptyState
- ErrorEmptyState
- NoDataEmptyState
- NoTagsEmptyState
- NotFoundEmptyState
- NotificationsEmptyState
- UnauthorizedEmptyState

#### What did you change?
Added `mdx` file for each component and made the required updates in their story files.

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/EmptyStates/EmptyStates.stories.jsx
+++ b/packages/ibm-products/src/components/EmptyStates/EmptyStates.stories.jsx
@@ -6,11 +6,11 @@
  */
 
 import React from 'react';
-import { action } from 'storybook/actions';
+import { action as storybookAction } from 'storybook/actions';
 import { Add, Information } from '@carbon/react/icons';
 import CustomIllustration from './story_assets/empty-state-bright-magnifying-glass.svg';
 import { EmptyState } from '.';
-import { StoryDocsPage } from '../../global/js/utils/StoryDocsPage';
+import mdx from './EmptyState.mdx';
 import { Tooltip } from '@carbon/react';
 
 export default {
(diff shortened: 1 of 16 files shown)


packages/ibm-products/src/components/EmptyStates/EmptyStates.stories.jsx
```diff
@@ -20,31 +20,72 @@ export default {
...
+      return (
+        <>
+          Click <span>here</span> to upload your data
+          <Tooltip label="Facts and statistics collected together for reference or analysis">
+            <Information size="16" />
+          </Tooltip>
+        </>
+      );
+    }
+  };
```
davidmenendez: this is a nice implementation, but i think we can simplify this a bit. using the `value` parameter as a way to return different output is good, but using `0` and `1` can be confusing out of context. `getSubTitle(1)` doesn't really clearly indicate what you're getting with that input.

instead, i think you could just separate them into two functions with names that clearly identify what they do like `getSubtitle()` and `getSubTitleWithTooltip()`.

packages/ibm-products/src/components/EmptyStates/EmptyStates.stories.jsx
```diff
@@ -20,31 +20,72 @@ export default {
...
+        onClick: () => {
+          sbDocs
+            ? storybookAction('Clicked empty state action button')()
+            : console.log('Clicked empty state action button');
+        },
+        renderIcon: (props) => <Add size={20} {...props} />,
+        iconDescription: 'Add icon',
+      };
+    }
+  };
```
davidmenendez: similar to my comment about `getSubTitle` I think we can simplify this. again, `value` doesn't really help indicate what the parameter does. you could separate these into two separate functions like i suggested above, but with object it can be a little easier to do it in one with just some tweaks to the implementation:

```
getAction(icon: boolean) {
  const obj = {
   text: 'create new'
    onClick: () => { ... }
    ...
  };

  if (icon) return { ...obj, ... { renderIcon, IconDescription} }

  return obj;
}
```

with this implementation you don't need to include any redundant code while making it's intention a little more clear.

packages/ibm-products/src/components/EmptyStates/EmptyStates.stories.jsx
```diff
@@ -55,14 +96,7 @@ Default.args = {
...
   ...emptyStateCommonProps,
-  subtitle: (
-    <>
-      Click <span>here</span> to upload your data
-      <Tooltip label="Facts and statistics collected together for reference or analysis">
-        <Information size="16" />
-      </Tooltip>
-    </>
-  ),
+  subtitle: 1,
```
davidmenendez: i noticed that when you check the controls in storybook that `subtitle` just says `1` so we end up obfuscating these inputs. lets try separating the getter functions from the template and see if we can populate these controls

PR #8406, opened 2025-10-08
fix(coachmark): added missing exports

anamikaanu96:
Closes #

Added missing exports also resolved TypeError: Cannot read properties of undefined (reading 'Header') caused in stackblitz.

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/index.ts
+++ b/packages/ibm-products/src/components/index.ts
@@ -55,6 +55,7 @@ export * from './WebTerminal';
 // Preview
 export * as preview__Coachmark from './Coachmark/next/Coachmark';
 export * as preview__CoachmarkTagline from './Coachmark/next/Coachmark/CoachmarkTagline';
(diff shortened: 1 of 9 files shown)


packages/ibm-products/src/components/index.ts
```diff
@@ -55,6 +55,7 @@ export * from './WebTerminal';
 // Preview
 export * as preview__Coachmark from './Coachmark/next/Coachmark';
 export * as preview__CoachmarkTagline from './Coachmark/next/Coachmark/CoachmarkTagline';
+export * as preview__CoachmarkBeacon from './Coachmark/next/Coachmark/CoachmarkBeacon';
```
matthewgallo, line 58: Can you also update these exports to include the types used in the new coachmark components?

anamikaanu96: Types exports are already added to the index file of each component.

PR #8408, opened 2025-10-08
feat(notificationPanel): code connect

### LINKED ISSUE #8288: Figma Code Connect - React (NotificationPanel)
1. Review [quick start guide](https://www.figma.com/code-connect-docs/quickstart-guide/#install-the-code-connect-command-line-tool)
2. Create `NotificationsPanel.figma.tsx` within packages/ibm-products/src/components/NotificationsPanel` component directory
3. Map props from component to the props within figma, there may not be a one to one match between the component's props in code vs in figma
4. Document any discrepancies within figma kit

szinta:
Closes #8288 

#### What did you change?

packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.figma.tsx

Verified in Figma code connect

#### PR Checklist

our contributing docs.

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -13,7 +13,10 @@ jobs:
     if: github.repository == 'carbon-design-system/ibm-products'
     runs-on: ubuntu-latest
     steps:
-      - uses: github/issue-labeler@c1b0f9f52a63158c4adc09425e858e87b32e9685 # v3.4
+      - name: Checkout
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - name: Label PR
+        uses: github/issue-labeler@c1b0f9f52a63158c4adc09425e858e87b32e9685 # v3.4
         with:
           configuration-path: .github/pr-labeler-config.yml
           not-before: 2025-10-10T00:00:00Z
--- a/.github/workflows/release-base.yml
+++ b/.github/workflows/release-base.yml
@@ -19,13 +19,14 @@ on:
         type: boolean
         default: true
 
+# Required for OIDC publishing https://docs.npmjs.com/trusted-publishers
+permissions:
+  id-token: write
+  contents: write
+
 jobs:
   release:
     runs-on: ubuntu-latest
-    # Recommended by npm for publishing with provenance https://docs.npmjs.com/generating-provenance-statements
-    permissions:
-      id-token: write
-      contents: write
     timeout-minutes: 60
     env:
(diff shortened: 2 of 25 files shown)


matthewgallo (approved):
@RichKummer and I noticed that the `illustrationTheme` is hardcoded in the examples to `light` but I'm now guessing that's because there's no matching figma prop. Rich said he'll look at this first in the empty state itself.

packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.figma.tsx
```diff
@@ -0,0 +1,93 @@
...
+      {
+        id: '3',
+        type: 'warning',
+        title: 'Title',
+        description: 'Message',
+        timestamp: new Date(new Date().getTime() - 120 * 1000),
+        unread: false,
+        onNotificationClick: () => console.log('Clicked on notification'),
+      },
+    ];
```
jlongshore: When the notifications object is inside that example property, you get this kind of output:
image

If we take that array and just dump it in the `data` prop of the NotificationsPanel, it seems to clean up the output in Figma:

image

packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.figma.tsx
```diff
@@ -0,0 +1,93 @@
...
+    return (
```
jlongshore: If we put the array loose in the `data` prop, we can remove the `return` and the encapsulating curly braces

packages/ibm-products/src/components/NotificationsPanel/NotificationsPanel.figma.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <NotificationsPanel
+        open={true}
+        data={notifications}
```
jlongshore: Putting the array directly in the `data` attribute's value eliminates the TS error on the data prop. If we were to leave it, we could remedy this by pulling in the Data and the Link interface from the NotificationsPanel (or exporting them)... I think the simpler solution would be to just put the array in the `data` prop's value.

image

szinta:
> @RichKummer and I noticed that the `illustrationTheme` is hardcoded in the examples to `light` but I'm now guessing that's because there's no matching figma prop. Rich said he'll look at this first in the empty state itself.

Theme is currently controlled using the `Variable modes` feature in Figma and It looks like Code connect doesn't support reading values directly from Variable modes. That's why I thought of using a hard-coded value for that. We can remove hard-coded value if a prop available for Theme switching.

PR #8431, opened 2025-10-09
fix: add code panel back into storybook

### LINKED ISSUE #8310: Re-integrate code tab back to storybook for react and web components
We should re-introduce the code tab back to our storybooks. There is a [PR](https://github.com/carbon-design-system/carbon/pull/20229) from `@carbon/react` that can be used as a reference for this work. You can now specify within storybook's `preview` config to include a `codePanel`, as the `storysource` addon was [removed in v9](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#storysource-addon-removed).

### Acceptance criteria
- [ ] Add `codePanel` config to `packages/ibm-products/.storybook/preview.js`
- [ ] Add `codePanel` config to `packages/ibm-products-web-components/.storybook/preview.ts`
- [ ] Verify output within code panel is expected, accurate, and representative of the given component

davidmenendez:
Closes #8310 

adds the code panel back into storybook


packages/ibm-products-web-components/.storybook/preview.ts
```diff
@@ -159,6 +159,10 @@ export const parameters = {
   },
   docs: {
     theme,
+    source: {
+      excludeDecorators: true,
```
matthewgallo, line 163: It's really nice that they have this as an option, is it defaulted to `true` because in the react storybook I see they're excluded without this option set?

We could also clean some of the react stories up a bit by wrapping more of the stories with decorators, kind of like how the web component page header story does. Not a blocker for this, but did want to mention it.

PR #8469, opened 2025-10-14
feat(PageHeader): add open in stackblitz config

### LINKED ISSUE #8365: Open in stackblitz for page header stories
Rather than creating examples directly for page header, we should integrate the "open in stackblitz" functionality directly within the storybook docs for page header.

matthewgallo:
Closes #8365 

Adds open in stackblitz config to the new/preview page header stories. This required creating a mapping for preview and preview candidate components. @anamikaanu96 I couldn't just add `status` of either `preview` or `previewCandidate` because there could be other components imported from our library that have different statuses. The mapping for status that I ended up with has the following structure:

```tsx
status: {
  preview: ['PageHeader'],
  previewCandidate: [] // optionally include preview candidate components
}
```

This way the status isn't added to every component from our library. I also updated some of the `previewer/configFiles`
- Set `react-is` version to the same major version of react that we're using, otherwise the `Tabs` don't render
- Updated `@carbon/styles/config` to get plex to load
- Updated the [component regex pattern](https://github.com/carbon-design-system/ibm-products/pull/8469/files#diff-5c53a916410a1637b49888d1931b8db0171cfee87ff186c629869245035c19aeR218) to match compound components

> [!NOTE]
>  ~I found that `customFunctionDefs` and `customImports` both only work if there's only one item in the array. That's a small fix should address within `codePreviewer.tsx`~ ◀️ I've fixed this, but another issue with the code previewer is that it doesn't include any icons passed to `renderIcon` (ie rendered via non-jsx usage), that would be a nice enhancement. There is a work-around by manually adding them as specified imports.

#### What did you change?
- Updated code previewer config files
- Updated page header mdx
- Added some missing styles
- Updated code previewer to support a status for specific components
Clicking `Open in stackblitz` for all of the previews on the page header docs page
#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/previewer/codePreviewer.tsx
+++ b/packages/ibm-products/previewer/codePreviewer.tsx
@@ -211,9 +211,7 @@ const appGenerator = async (
   `;
 
   try {
-    const getFormattedCode = async (app: string) =>
-      await formatWithPrettier(app);
-    return await getFormattedCode(app);
+    return await formatWithPrettier(app);
   } catch (error) {
     console.error('Error formatting code:', error);
     throw error;


davidmenendez:
wow this is super impressive! great work.

I would just double check that the links are working. looks like the first one in [overview](https://deploy-preview-8469--carbon-for-ibm-products.netlify.app/?path=/docs/preview-pageheader--overview#overview) breaks due to clashing carbon imports, which i suspect is coming from [customImportsArr](https://github.com/carbon-design-system/ibm-products/pull/8469/files#diff-57368f3cf86badcca10d886d395647d584d0a15399bab95682af2977f318fd33R26)

Screenshot 2025-10-15 at 1 51 09 PM

anamikaanu96:
Looks great! 🎉  I hope this will be suitable for all the scenarios.

Once the above error is resolved, we are good to merge. 👍

matthewgallo:
@davidmenendez @anamikaanu96 ready for another review. I also added some formatting (via `prettier.format(...)`) to the generated stackblitz code since I noticed it had a lot of spacing issues after all of the transforming we do to generate it.

Anamika, I tried to be careful so that any existing usage of the code previewer would _not_ be impacted. I've verified across most of the existing "Open in stackblitz" previews.

davidmenendez (approved):

PR #8471, opened 2025-10-14
fix: remove canary check from truncatedtext component

### LINKED ISSUE #8470: TruncatedText: remove canary check
`checkComponentEnabled` is no longer required and should be removed from `TruncatedText.tsx`

davidmenendez:
Closes #8470


matthewgallo:
Ah, actually @davidmenendez it looks like `TruncatedText` is exported twice. ~Lets keep it's current status `previewCandidate`~ (fine to update since we updated this in the codemod non-stable mapping) and remove the [other export](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products/src/components/index.ts#L46). This way we can focus this PR on just removing the canary check.

davidmenendez:
@matthewgallo fixed the duplicate exports 👍

PR #8513, opened 2025-10-20
feat: init guide banner component

### LINKED ISSUE #8476: Guidebanner: scaffolding and layout
Ticket to represent the scaffolding and layout of the `GuideBanner` and its sub-components independent of the scroll utility. 

Proposed architecture:

`GuideBanner` - main export container

API:

* titleText - for title text
* onClose(evt) - callback function for when the close button is fired
* closeIconDescription - label text for close button
* open - determines if the body is expanded or not
* expand text - text for expand button
* collapse text - text for collapse button
* onChange(evt) - callback function for when the expand / collapse button is fired

Slots:

* title
* body
* footer

`GuideBannerElement` - individual elements that make up the body content

API:

* titleText - title text
* description - description text

Slots:

* title
* description
* button

davidmenendez:
Closes #8476 

Adds the `guide-banner` web component.


davidmenendez:
still a work in progress because tests need to be added and also because I wanted to open this now to address some things that need consideration;

* the original react component is coupled with the carousel for displaying children. is this something we can to continue with moving forward in the web component version or do we let the consumer handle that implementation?
* this point is reliant on the previous point: styling is also coupled with the carousel. eg `.#{$block-class}__carousel .#{$carousel-item} .#{$block-class}__element` which leaves the body content without the carousel unstyled. if we decouple the carousel, we will have to implement custom web component styling separate from `_guidebanner.scss` in the styles package.
* unlike the react component, this implementation uses the native `summary` and `details` HTML components
* unlike `options-tile`, this component doesn't rely on internal state. initially I implemented the open state in `options-tile` to handle issues I was seeing with the native details & summary implementation, which would cause the disclosure to open and close rapidly, but here that doesn't appear to be the case. if this is preferred then we should establish that as an implementation pattern, which I touch more on in the next point.
* Events:

guide banner uses a similar implementation to options tile, but in this implementation i went with a single `onchange` event that's called when the details disclosure is opened and closed. options tile uses two separate [open and close events](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products-web-components/src/components/options-tile/options-tile.ts#L30). carbon core uses a single event, similar to the one in this implementation, but they call it [toggled](https://web-components.carbondesignsystem.com/?path=/docs/components-dropdown--overview#cds-dropdown). checklist, in this repo, uses [toggle](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products-web-components/src/components/checklist/checklist.ts#L32).

I think for consistency sake we should define some basic guidance for event naming and defining. that will probably need to be done separate from this, but I wanted to point it out now as part of the initial review process.

davidmenendez:
ready for review now

packages/ibm-products-web-components/src/components/guide-banner/guide-banner-element.ts
```diff
@@ -0,0 +1,30 @@
...
+class CDSGuideBannerElement extends HostListenerMixin(LitElement) {
+  render() {
+    return html`
+      <div class="${blockClass}">
+        <slot></slot>
+      </div>
+    `;
+  }
+
+  // static styles = styles;
```
jlongshore: commented out code could be removed.

packages/ibm-products-web-components/src/components/guide-banner/guide-banner.scss
```diff
@@ -0,0 +1,26 @@
...
+@use '@carbon/styles/scss/theme' as *;
+@use '@carbon/styles/scss/spacing' as *;
+@use '@carbon/styles/scss/reset';
+@use '@carbon/styles/scss/type';
+@use '@carbon/ibm-products-styles/scss/config';
+@use '@carbon/ibm-products-styles/scss/components/Guidebanner' as *;
+
+$prefix: config.$pkg-prefix;
+$carbon-prefix: config.$carbon-prefix;
+$block-class: #{$prefix}--guidebanner;
```
jlongshore: I think we might be able to remove these two variables, they dont appear to be used in this file or exported.

packages/ibm-products-web-components/src/components/guide-banner/guide-banner.ts
```diff
@@ -0,0 +1,150 @@
...
+            kind="ghost"
+            size="md"
+            @click="${this._handleClose}"
+          >
+            ${iconLoader(Close16, { slot: 'icon' })}
+          </cds-button>
+        </div>
+        <details ?open=${this.open}>
+          <slot name="body"></slot>
+          <summary
```
jlongshore, line 134: A `::marker` element is rendering in the summary tag here...
image
image

I do see in the options-tile scss file, that there is an entry for `::marker` (and seems to be the only place this is referenced in our code). Applying the following to the `guide-banner.scss` seems to fix the issue:

```css
::marker,
::-webkit-details-marker {
  display: none;
  content: '';
}
```

packages/ibm-products-web-components/src/components/guide-banner/guide-banner.scss
```diff
@@ -0,0 +1,26 @@
...
+
+:host(#{$prefix}-guide-banner) {
+  details {
+    display: flex;
+    flex-direction: column-reverse;
+  }
+}
```
jlongshore, line 30: Mentioned in the `guide-banner.ts` file for more detail - suggest adding the following:

```suggestion
:host(#{$prefix}-guide-banner) {
  details {
    display: flex;
    flex-direction: column-reverse;
  }
  ::marker,
  ::-webkit-details-marker {
    display: none;
    content: '';
  }
}
```

jlongshore (changes requested):
Just a couple things... looks good!

jlongshore (approved):

packages/ibm-products-web-components/src/components/guide-banner/guide-banner.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  render() {
+    const classes = classMap({
+      [`${blockClass}`]: true,
+      [`${blockClass}__collapsible-collapsed`]: !this.open,
+    });
+    return html`
+      <div class="${classes}">
+        <div class="${blockClass}__header">
```
amal-k-joy, line 112: Hey @davidmenendez ,
I know `Idea20` is the icon to be rendered based on our guidelines. I was thinking we can wrap this icon inside an `icon` slot, it allows user to override incase they need. Refer [checklist](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products-web-components/src/components/checklist/checklist-item.ts#L114C6-L116C14) for a similar approach

packages/ibm-products-web-components/src/components/guide-banner/guide-banner.ts
```diff
@@ -0,0 +1,150 @@
...
+        </div>
+        <details ?open=${this.open}>
+          <slot name="body"></slot>
+          <summary
+            tabindex="-1"
+            @click=${(evt: Event) => {
+              evt.preventDefault();
+            }}
+          >
+            <div class="${blockClass}__navigation">
```
amal-k-joy, line 140: Here, when the button is clicked, we emit the `eventToggle` event, and the user must listen to it to handle expanding or collapsing the component.

It feels like we're hard-coding this ghost button. I’m thinking about scenarios where users might want a different experience for expand/collapse. IMO, we could move this into the footer slot and provide the ghost button as the default behavior. If the user wants to customize it, they can pass their own footer content.

amal-k-joy (changes requested):
Everything look soo cool.
Added 2 comments.

packages/ibm-products-web-components/src/components/guide-banner/guide-banner.ts
davidmenendez: that's a good point. can't believe i missed something so obvious! ty

davidmenendez:
@amal-k-joy changes made to make the icon and footer default slot behavior. good call on those 👍

davidmenendez:
@amal-k-joy @anamikaanu96 @makafsal @jlongshore please review. i need this merged in ASAP. thanks.

makafsal (approved):
🚀 Good work @davidmenendez

PR #8535, opened 2025-10-23
fix(Feature flag): moved the merge logic from feature-flag js to FeatureFlag

### LINKED ISSUE #8166: [Bug]: Unable to find a feature flag with the name: 'enableSidepanelResizer'
Based on [this comment](https://github.com/carbon-design-system/ibm-products/issues/7969#issuecomment-3231882950)
, the same issue also appears in Tagset. You can reproduce it using the provided StackBlitz examples, and verify by testing with other potential components like SidePanel. Please also check whether this issue exists in other components that rely on feature flags. Additionally, explore whether there’s an alternative fix by looking into the[ root cause](https://github.com/carbon-design-system/ibm-products/pull/5516).

Image

### Refinement points

We’ve added the import (`import '../../../feature-flags';`) inside `src/global/js/hooks/usePortalTarget.js`, expecting this script to run. However, on Next.js app is still throwing the error, which makes us think the script isn’t being executed in the Next.js environment.

Steps to reproduce:
Clone this repo and start the server : https://github.com/abdonrd/carbon-ibm-products-turbopack

amal-k-joy:
Closes #8166 

We previously kept the FeatureFlags.merge logic — which merges the C4P flags with the core flags — in a separate file (feature-flags.js), expecting it to execute automatically on load. While this worked fine in React apps, it failed to run in Next.js.

The fix is to move this logic into the FeatureFlag component, placing it just before the context is created. This should resolve the issue since it will now at the component level and should execute in client side rather than in server.

Tested in Next app:

- Cloned this repo in local : https://github.com/abdonrd/carbon-ibm-products-turbopack
- Made the above changes in C4P repo and run `yarn build` which generated the lib and es folders there.
- Replaced these folders with the ones in Next app node modules under @carbon/ibm-products
- Ran the next app and apears to be working
- Followed the same approach in React app(https://stackblitz.com/edit/github-gyndkvoc) and verified

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/FeatureFlags/index.tsx
+++ b/packages/ibm-products/src/components/FeatureFlags/index.tsx
@@ -182,7 +182,6 @@ function useChangedValue<T>(
  */
 function useFeatureFlag(flag) {
   const scope = useContext(FeatureFlagContext);
-  console.log('scope', scope);
   return scope.enabled(flag);
 }
 


matthewgallo (approved):
Great work here 👏 now our components aren't littered with feature flag imports as well.

PR #8551, opened 2025-10-27
fix: replace yarn with npx for playwright scripts

davidmenendez:
looks like `yarn playwright` is having issues as per https://github.com/microsoft/playwright/issues/17975#issuecomment-1917441952

to alleviate this we will switch all the playwright script commands to use `npx` instead of `yarn`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,7 +142,7 @@ jobs:
         run: yarn
       - name: Install browsers
         working-directory: packages/ibm-products-web-components
-        run: yarn playwright install --with-deps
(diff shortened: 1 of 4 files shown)


matthewgallo:
Hey @davidmenendez, was able to resolve this ([here](https://github.com/carbon-design-system/ibm-products/commit/0b709c0a9b1184bc58166c3db70e4c4264c53bdf)) by adding the playwright install step (including the `--with-deps` flag) from the test script of the web component package. For monorepo usage, the playwright install needs to be  run from the same package as the test script.

davidmenendez:
@matthewgallo thank you for figuring that out!

PR #8537, opened 2025-10-23
chore(Tearsheet): replace `h2` and `h3` tags with `Section` and `Heading`

makafsal:
Closes #8376 

Replace hard-coded tags with `Section` and `Heading` components from `@carbon/react`

#### What did you change?
Replaced `h2` and `h3` tags with `Section` and `Heading`

#### How did you test and verify your work?
Storybook

#### PR Checklist

As the author of this PR, before marking ready for review, confirm you:

- [ ] Reviewed every line of the diff
- [ ] Updated documentation and storybook examples
- [ ] Wrote passing tests that cover this change
- [ ] Addressed any impact on accessibility (a11y)
- [ ] Tested for cross-browser consistency
- [ ] Validated that this code is ready for review and status checks should pass

More details can be found in the [pull request](./CONTRIBUTING.md) section of
our contributing docs.

PR #8563, opened 2025-10-29
feat(Tearsheet): stacking, unit test, design review changes and fixes

### LINKED ISSUE #8140: Tearsheet: add unit tests
Add unit tests for maximum coverage

### LINKED ISSUE #8141: Tearsheet: update doc section and add examples

### LINKED ISSUE #8142: Tearsheet: Implement stepped tearsheet pattern using stepper utility
Create stepped tearsheet pattern and story
Add stackblitz example

### LINKED ISSUE #8562: Tearsheet: design review changes and fixes

amal-k-joy:
Closes #8562 
Closes #8140 
Closes #8141
Closes #8142 

- Stacking
- Updated storybook docs section
- Added open in stackblitz examples
- Stories restructuring
- Added unit tests
- Fixed issues and changes from design review by Rich

#### What did you change?
**Stacking**
This implementation introduces stacking tearsheets using a new composable approach.
Unlike the current tearsheet implementation — where all stacking logic is embedded within the component — the new model externalizes that logic into a dedicated wrapper:
<StackProvider stackStepSize="lg"></StackProvider> //stackStepSize is optional and be defaulted to 1 rem

In this setup, the StackProvider encapsulates all stacking behavior.
Each component simply needs to notify the provider when a tearsheet is opened. The component can then consume the calculated stacking information from the context and apply it as needed.

Styling adjustments will still be handled within the component itself, as these may vary between different components.

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/Tearsheet/next/Tearsheet.test.js
+++ b/packages/ibm-products/src/components/Tearsheet/next/Tearsheet.test.js
@@ -213,10 +213,6 @@ describe('Tearsheet component V2', () => {
     expect(style2.getPropertyValue('--stack-depth')).toBe('1');
     expect(style1.getPropertyValue('--stack-depth')).toBe('2');
 
-    expect(style3.getPropertyValue('--stack-depth')).toBe('0');
-    expect(style2.getPropertyValue('--stack-depth')).toBe('1');
-    expect(style1.getPropertyValue('--stack-depth')).toBe('2');
-
     const scaleFactor1 = style1.getPropertyValue('--scale-factor');
     const scaleFactor2 = style2.getPropertyValue('--scale-factor');
     const scaleFactor3 = style3.getPropertyValue('--scale-factor');


packages/ibm-products/src/components/Tearsheet/next/Tearsheet.stories.jsx
```diff
@@ -25,13 +25,23 @@ import {
   Tabs,
   TextInput,
   Toggletip,
+  ToggletipButton,
```
matthewgallo, line 27: Can you remove the `IconButton` import (on line 16)? It's imported but not used within any stories

packages/ibm-products/src/components/Tearsheet/next/StackContext.tsx
```diff
@@ -0,0 +1,131 @@
...
+ */
+
+import React, {
+  useContext,
+  createContext,
+  ReactNode,
+  useState,
+  useCallback,
+  useRef,
+  RefObject,
```
matthewgallo: Can you remove this line? `RefObject` isn't used anywhere

packages/ibm-products/src/components/Tearsheet/next/StackContext.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const getScaleFactor = (id) => {
+    const depth = getDepth(id);
+    const buffer = bufferMap[stackStepSize];
+    const bufferInPx = remToPx(buffer);
+
+    const scale = depth > -1 ? (width - bufferInPx * 2 * depth) / width : 1;
+    return scale;
+  };
+
+  const remToPx = (rem) => {
```
matthewgallo: Can you specify a type for the `rem` param?

packages/ibm-products/src/components/Tearsheet/next/StackContext.tsx
```diff
@@ -0,0 +1,131 @@
...
+          // remove from stack
+          return prev.filter((i) => i !== id);
+        }
+      });
+    },
+    []
+  );
+
+  // method that calculates spacing factor for the stacked items
+  const getScaleFactor = (id) => {
```
matthewgallo: Can you specify a type for the `id` param?

packages/ibm-products/src/components/Tearsheet/next/StackContext.tsx
```diff
@@ -0,0 +1,131 @@
...
+      const index = stack.indexOf(id);
+      if (index === -1) {
+        return -1;
+      }
+      return stack.length - 1 - index; // topmost → 0
+    },
+    [stack]
+  );
+
+  const getBlockSizeChange = (id) => {
```
matthewgallo: Can you specify a type for the `id` param?

packages/ibm-products/src/components/Tearsheet/next/StackContext.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  return (
+    <StackContext.Provider value={context}>{children}</StackContext.Provider>
+  );
+};
+
+export const useStackContext = (): StackContextType => {
+  const context = useContext<StackContextType | undefined>(StackContext);
+
+  if (context === undefined) {
```
matthewgallo, line 120: This is what I was curious about with my comment from before about the usage of `useStackContext` within the Tearsheet without wrapping it with the stacking provider. I like your approach here of returning some default values when there is no context found.

packages/ibm-products/src/components/Tearsheet/next/Tearsheet.mdx
```diff
@@ -1,6 +1,309 @@
...
+
+## Default
+
+<Canvas
+  of={stories.Default}
+  additionalActions={[
+    {
+      title: 'Open in Stackblitz',
+      onClick: () =>
+        stackblitzPrefillConfig({
```
matthewgallo: None of the "Open in stackblitz" work just yet, only because we haven't published `preview__Tearsheet` yet. Maybe comment out any `stackblitzPrefillConfig` integrations until we've published the preview, then we can add them back in.

I also noticed that some of the stackblitz's (outside of `preview__Tearsheet` not being available) are missing some variables that you will need to include via:

```ts
stackblitzPrefillConfig({
  customFunctionDefs: [`const variant = ['wide', 'narrow', 'wide'];`]
}),
```

In this case, it's the stacking example that needs `variant` defined.

packages/ibm-products/src/components/Tearsheet/next/Tearsheet.test.js
```diff
@@ -6,38 +6,363 @@
...
-import { render } from '@testing-library/react';
+import { render, screen, fireEvent } from '@testing-library/react';
+import { blockClass } from './context';
+import { StackProvider } from './StackContext';
 
-describe('Tearsheet', () => {
-  it('renders Tearsheet v2', async () => {
+const baseFontSize = '16px';
+
+jest.mock('../../../global/js/hooks/useResizeObserver', () => ({
```
matthewgallo, line 19: I noticed some coverage missing for the `createOverflowHandler` within the header actions component. I was able to [mock it](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products/src/components/PageHeader/next/PageHeader.test.js#L42-L46) somewhat similar to this within the page header tests and then [use it](https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products/src/components/PageHeader/next/PageHeader.test.js#L379-L384) to trigger resizing/overflowing.

This is more of a organizational this, but it might help to group some of these tests in different describe blocks based on the different tearsheet sub-component you're testing.

packages/ibm-products/src/components/Tearsheet/next/Tearsheet.tsx
```diff
@@ -13,6 +13,8 @@ import React, {
   ReactNode,
   ForwardedRef,
   FC,
+  useId,
```
matthewgallo: We'll need to take a similar approach as system squad here with their custom `useId` hook. Since `useId` was only published in React 18 but we support 16 and 17 we need to account for that. See [here](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/internal/useId.js).

matthewgallo: We actually have something similar, not quite as robust but it will work for old React versions, but still use `useId` from `React` if it exists.

https://github.com/carbon-design-system/ibm-products/blob/main/packages/ibm-products/src/global/js/utils/useId.js

packages/ibm-products/src/components/Tearsheet/next/TearsheetHeaderContent.tsx
```diff
@@ -40,52 +53,35 @@ const TearsheetHeaderContent = React.forwardRef<
...
-                  Icon && titleIconPosition === 'leading',
-              },
-              {
-                [`${blockClass}__trailing-icon`]:
-                  Icon && titleIconPosition === 'trailing',
-              }
-            )}
-          >
-            {!isSm && Icon && <Icon size={32} />}
+          <h2 className={cx(`${blockClass}__header-title`)}>
```
matthewgallo, line 69: Lets use the `Section` and `Heading` approach to avoid hard coding heading levels.

packages/ibm-products/src/components/Tearsheet/next/StackContext.tsx
amal-k-joy: Hopefully this should prevent intended code to be run as usage end.

packages/ibm-products/src/components/Tearsheet/next/Tearsheet.test.js
amal-k-joy: I tried a bit, but the resizing was not getting triggered on utility side and then I thought this utility itself is already well tested for resize behaviour on its own.
Thanks for sharing this, I can definitely try this one.

packages/ibm-products/src/components/Tearsheet/next/Tearsheet.tsx
amal-k-joy: Oh, you’re right — I actually meant to use the one that David implemented, but it seems the React version was imported instead. Thanks for catching that!

packages/ibm-products/src/components/Tearsheet/next/TearsheetHeaderContent.tsx
amal-k-joy: I was kind of confused  on what is the right level to be used here. Even if we use Section and Heading, its kind of we itself explicitly deciding the level as in this [PR](https://github.com/carbon-design-system/ibm-products/pull/8537/files)

packages/ibm-products/src/components/Tearsheet/next/Tearsheet.mdx
amal-k-joy: Good catch. I can probably comment out Open in stackblitz for now and open another issue for that.

packages/ibm-products/src/components/Tearsheet/next/Tearsheet.test.js
amal-k-joy: It worked

packages/ibm-products/src/components/Tearsheet/next/Tearsheet.test.js
```diff
@@ -6,38 +6,410 @@
...
+    // Grab computed styles
+    const style1 = window.getComputedStyle(t1);
+    const style2 = window.getComputedStyle(t2);
+    const style3 = window.getComputedStyle(t3);
+
+    expect(style3.getPropertyValue('--stack-depth')).toBe('0');
+    expect(style2.getPropertyValue('--stack-depth')).toBe('1');
+    expect(style1.getPropertyValue('--stack-depth')).toBe('2');
+
+    expect(style3.getPropertyValue('--stack-depth')).toBe('0');
```
szinta: This is basically the same as the expect statements above. Looks like accidentally repeated the same tests here.

szinta (approved):
Great work Amal.

packages/ibm-products/src/components/Tearsheet/next/Tearsheet.test.js
amal-k-joy: Good catch

anamikaanu96 (approved):

szinta (approved):

PR #8641, opened 2025-11-17
fix: remove use client directive warnings from build

### LINKED ISSUE #8532: [build]: `use client` directive warnings should be ignored within rollup build
Image

It's possible there's a rollup option we missed when initially implementing this so that we don't receive warning logs about using directives in bundled files.

We should confirm that the warning code is `MODULE_LEVEL_DIRECTIVE`.

davidmenendez:
Closes #8532 

suppresses logs that are showing up during build regarding `use client`

confirmed with `yarn build`

### DIFF
--- a/packages/ibm-products/.storybook/main.js
+++ b/packages/ibm-products/.storybook/main.js
@@ -70,6 +70,18 @@ export default {
     const { mergeConfig } = await import('vite');
 
     return mergeConfig(config, {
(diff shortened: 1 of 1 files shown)


amal-k-joy (approved):
LGTM

davidmenendez:
@anamikaanu96 please review when you have a moment. ty

PR #8720, opened 2025-12-10
feat: snapscroll

### LINKED ISSUE #8475: "Carousel" replacement scroll utility
Investigate the feasibility of creating a utility to replace the old internal Carousel component to be used in the web component of Guidebanner as well as the React version.

davidmenendez:
closes #8475 

still requires tests, but I wanted to get the initial implementation done and agreed upon before writing the tests.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -7,7 +7,7 @@ jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/stale@5f858e3efba33a5ca4407a664cc011ad407f2008 # v10.1.0
+      - uses: actions/stale@997185467fa4f803885201cee163a9f38240193d # v10.1.1
         with:
           any-of-labels:
             'status: needs reproduction,status: waiting for author 💬'
--- a/.github/workflows/update-example-gallery.yml
+++ b/.github/workflows/update-example-gallery.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Create pull request
         if: failure() && steps.up_to_date.outcome == 'failure'
         id: create-pr
-        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
+        uses: peter-evans/create-pull-request@22a9089034f40e5a961c8808d113e2c98fb63676 # v7.0.11
         with:
           token: ${{ secrets.generate_token.outputs.token }}
           commit-message: 'chore: update example gallery'
--- a/.github/workflows/update-v1.yml
+++ b/.github/workflows/update-v1.yml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Create PR
         id: create-pr
-        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
+        uses: peter-evans/create-pull-request@22a9089034f40e5a961c8808d113e2c98fb63676 # v7.0.11
         with:
           token: ${{ secrets.CREATE_PR_ACCESS_TOKEN }}
           commit-message: 'fix: update Carbon 10 compatible versions to latest'
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -29,7 +29,7 @@ jobs:
           private_key: ${{ secrets.APP_PRIVATE_KEY }}
       - name: Create PR
         id: create-pr
-        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
+        uses: peter-evans/create-pull-request@22a9089034f40e5a961c8808d113e2c98fb63676 # v7.0.11
         with:
           token: ${{ steps.generate_token.outputs.token }}
           commit-message:
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,4 @@ blob-report/
 # Actions
 /actions/**/node_modules
 /actions/**/yarn.lock
+figma.config.json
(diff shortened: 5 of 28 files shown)


davidmenendez:
@makafsal @anamikaanu96 please review at your earliest convenience. thank you.

packages/ibm-products-web-components/src/components/guide-banner/guide-banner-element.ts
```diff
@@ -9,20 +9,34 @@
...
 
 export const blockClass = `${prefix}--guidebanner__element`;
 
 @customElement(`${prefix}-guide-banner-element`)
 class CDSGuideBannerElement extends HostListenerMixin(LitElement) {
+  @property()
+  titleText?: string;
+
+  @property()
+  descriptionText?: string;
```
amal-k-joy: Lets keep `descriptionText` as a slot rather than a string property for more flexibility , also lets keep the slot name as `description`

packages/ibm-products-web-components/src/components/guide-banner/guide-banner-element.ts
```diff
@@ -9,20 +9,34 @@
...
 import HostListenerMixin from '@carbon/web-components/es/globals/mixins/host-listener.js';
 import { carbonElement as customElement } from '@carbon/web-components/es/globals/decorators/carbon-element.js';
+import styles from './guide-banner-element.scss?lit';
 
 export const blockClass = `${prefix}--guidebanner__element`;
 
 @customElement(`${prefix}-guide-banner-element`)
 class CDSGuideBannerElement extends HostListenerMixin(LitElement) {
+  @property()
+  titleText?: string;
```
amal-k-joy: Lets keep `titleText` as a slot rather than a string property for more flexibility , also lets keep the slot name as `title`. That way we can give more responsibilities to user like text truncation.

packages/ibm-products-web-components/src/utilities/snapscroll/snapscroll.ts
```diff
@@ -0,0 +1,121 @@
...
+export function snapScroll(body: string, child: string) {
+  const bodyEl = document.querySelector(body);
+
+  if (!bodyEl) {
+    return;
+  }
+
+  const styleElement = document.createElement('style');
+  styleElement.textContent = `
+  .snappy {
```
amal-k-joy: Shall we use the package prefix for all the classes used to avoid any future conflicts?

makafsal:
@davidmenendez Build is failing, please have a look.

davidmenendez:
@amal-k-joy i switched to slots like you recommended. the show more and less button functionality is being controlled via the storybook controls.

and yeah if this looks good i can move forward with docs and testing 👍

amal-k-joy:
> @amal-k-joy i switched to slots like you recommended. the show more and less button functionality is being controlled via the storybook controls.
> 
> and yeah if this looks good i can move forward with docs and testing 👍

@davidmenendez ,
Changes looks very cool.
One more thing noticed, when we reach at scroll end, we need to disable the respective arrow buttons which is not happening now.

davidmenendez:
@amal-k-joy thanks for pointing out the issue with the next and previous not being disabled. i think that will need to be looked at separately because it looks like there's a problem with the underlying `cds-button` when it's using an icon and it's disabled


https://github.com/user-attachments/assets/8b4d3b81-e709-438b-a6d0-11d8b90a9cf3

basically it looks like the native `toggle` event for details gets called even when `disabled` is applied to the button, but interestingly enough it only happens when clicking the icon and not the outer button area. this looks like a bug with `icon-loader` inside `cds-button`.

davidmenendez:
i was able to alleviate this by wrapping the icon in a span with `preventDefault`. feels like a bit more boilerplate code than i'd prefer, but it works.

packages/ibm-products-web-components/src/components/guide-banner/guide-banner.stories.ts
```diff
@@ -25,6 +50,35 @@ const renderTemplate = (args) => {
...
+
+  const nextHandler = () => {
+    scrollNext();
+  };
+
+  const previousHandler = () => {
+    scrollPrevious();
+  };
+
+  const scrollendHandler = () => {
```
davidmenendez, line 66: @amal-k-joy @szinta i included this example code to demonstrate how you could implement the functionality to disable the nav buttons based on if there's a next or previous item relative to the selected item. i'm actually not certain this implementation is the best way to handle it.

using `[remove/set]Attribute()` directly on the `cds-button` i'm seeing this message in the console:

> Element cds-button scheduled an update (generally because a property was set) after an update completed, causing a new update to be scheduled. This is inefficient and should be avoided unless the next update can only be scheduled as a side effect of the previous update. See https://lit.dev/msg/change-in-update for more information.

i believe this is because the storybook stories aren't using `LitElements` and just utilizing the `html` directive, which doesn't provide state management. if i'm missing something or there's a better way to do this please review and let me know!

davidmenendez:
@amal-k-joy i added some initial unit testing for snapscroll. it's not comprehensive, but it's a start. the rest of the testing will need to be AVT tests because of how the functionality is reliant on window events and user interaction. i think we can add that in a follow up issue / pr 👍

amal-k-joy (approved):

amal-k-joy:
> @amal-k-joy i added some initial unit testing for snapscroll. it's not comprehensive, but it's a start. the rest of the testing will need to be AVT tests because of how the functionality is reliant on window events and user interaction. i think we can add that in a follow up issue / pr 👍

Thanks @davidmenendez , sure we can do that.

anamikaanu96 (approved):
Great work @davidmenendez!
I noticed a few issues, but I think it would be better to address them as a separate [issue](https://github.com/carbon-design-system/ibm-products/issues/8823):
1. The right arrow does not get disabled when the scroll reaches the end.
2. The component is breaking on the overview page.
3. Proper documentation is required on the overview page.

PR #8822, opened 2026-01-16
feat(tearsheet): implement `enable-presence` feature

### LINKED ISSUE #8293: Incorporate the enable-presence feature flag into the Tearsheet
A new feature in ComposeModal (introduced in t[his PR](https://github.com/carbon-design-system/carbon/pull/19531)) unmounts the modal from the DOM after its exit animation. This behavior can be enabled via the enable-presence feature flag.

Investigate how to introduce the same capability in Tearsheet V1 as an opt-in setting. Because this is an experimental Carbon feature, users should have the ability to opt in or out.

Ensure exit animation is working as expected with this new feature flag

Note: This feature flag will be available from @carbon/react v1.92.0

## Acceptance criteria

- [ ] update `@carbon/react` to version that supports this feature flag
- [ ] add `enablePresence` prop to toggle the flag
- [ ] update styling, tests, docs, etc as needed
- [ ] verify that other tearsheet variants are working as intended after update
- [ ] double check animation style is correct
- [ ] If for some reason the tearsheet animation does not work with the feature flag, we should use our own `usePresence` hook to achieve the same unmounting behavior

makafsal:
Closes #8293

#### What did you change?
Implemented the `enable-presence` and a feature flag to enable it.

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/Tearsheet/Tearsheet.featureflag.stories.jsx
+++ b/packages/ibm-products/src/components/Tearsheet/Tearsheet.featureflag.stories.jsx
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corp. 2020, 2026
+ * Copyright IBM Corp. 2026
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.


packages/ibm-products/src/components/Tearsheet/Tearsheet.tsx
```diff
@@ -22,6 +22,7 @@ import { allPropTypes } from '../../global/js/utils/props-helper';
 import { getDevtoolsProps } from '../../global/js/utils/devtools';
 import { pkg } from '../../settings';
 import { portalType } from './TearsheetShell';
+import { useTearsheetPresence } from '../../global/js/hooks/useTearsheetPresence';
```
amal-k-joy: Why do we need to import this here as well as in TearsheetNarrow.ts?

amal-k-joy (approved):
Nice work 🎉 🎉

packages/ibm-products/src/components/Tearsheet/Tearsheet.stories.jsx
```diff
@@ -802,6 +857,17 @@ withInfluencer.args = {
   actions: 7,
 };
 
+export const withPresenceMode = TemplateWithPresence.bind({});
+withPresenceMode.storyName = 'FeatureFlags / Tearsheet with presence mode';
```
szinta, line 861: Do we need this story? Looks like this is similar to the `Tearsheet/ Feature Flag/ Enable Presence ` story.

packages/ibm-products/src/components/Tearsheet/Tearsheet.featureflag.stories.jsx
```diff
@@ -0,0 +1,538 @@
+/**
+ * Copyright IBM Corp. 2020, 2026
```
szinta:
```suggestion
 * Copyright IBM Corp. 2026
```

szinta (commented):
LGTM. Just a few comments related to copyright text and story duplication.

packages/ibm-products/src/components/Tearsheet/Tearsheet.stories.jsx
makafsal: Yes, this is how core is organizing `feature-flags`

szinta (approved):

timfuhrmann:
Are you also planning to implement an equivalent to [ComposedModalPresence](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/ComposedModal/ComposedModalPresence.tsx)? While having the tearsheet not mount when it's closed is nice, it alone is not really helping, since the consuming component and all of it's states are still going to remain mounted. 

With `ComposedModalPresence` you are able to wrap you components to move that boundary of what is actually unmounted to another level. In case we do not get something like this for the tearsheet, it kind of defeats the purpose of `enable-presence`, as its goal never was to just not have it in the DOM, but to be able to benefit from the react component lifecycle using the tearsheet. 

It would be extremely helpful if you could achieve a similar dev experience to what carbon is offering.

makafsal:
> Are you also planning to implement an equivalent to [ComposedModalPresence](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/ComposedModal/ComposedModalPresence.tsx)? While having the tearsheet not mount when it's closed is nice, it alone is not really helping, since the consuming component and all of it's states are still going to remain mounted.
> 
> With `ComposedModalPresence` you are able to wrap you components to move that boundary of what is actually unmounted to another level. In case we do not get something like this for the tearsheet, it kind of defeats the purpose of `enable-presence`, as its goal never was to just not have it in the DOM, but to be able to benefit from the react component lifecycle using the tearsheet.
> 
> It would be extremely helpful if you could achieve a similar dev experience to what carbon is offering.

Good point - the limitation you’re calling out is real. `enable-presence` is now only controls the Tearsheet’s own DOM presence and does not reset consumer state the way `ComposedModalPresence` does in Carbon.

We did explore using `ComposedModalPresence` directly with Tearsheet, but it breaks Tearsheet animations because Tearsheet does not participate in Carbon’s presence contract (presence context consumption and exit coordination). As a result, presence has no way to know when the Tearsheet’s exit animation completes and unmounts immediately.

That said, we agree with the underlying goal. Providing a Tearsheet-specific presence boundary that allows consumers to opt into full lifecycle reset—similar in spirit to `ComposedModalPresence`-is a valid improvement area and something we can revisit with a dedicated implementation rather than direct reuse.

@amal-k-joy I'm moving this PR back to draft until the rework is complete.

amal-k-joy:
> > Are you also planning to implement an equivalent to [ComposedModalPresence](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/ComposedModal/ComposedModalPresence.tsx)? While having the tearsheet not mount when it's closed is nice, it alone is not really helping, since the consuming component and all of it's states are still going to remain mounted.
> > With `ComposedModalPresence` you are able to wrap you components to move that boundary of what is actually unmounted to another level. In case we do not get something like this for the tearsheet, it kind of defeats the purpose of `enable-presence`, as its goal never was to just not have it in the DOM, but to be able to benefit from the react component lifecycle using the tearsheet.
> > It would be extremely helpful if you could achieve a similar dev experience to what carbon is offering.
> 
> Good point - the limitation you’re calling out is real. `enable-presence` is now only controls the Tearsheet’s own DOM presence and does not reset consumer state the way `ComposedModalPresence` does in Carbon.
> 
> We did explore using `ComposedModalPresence` directly with Tearsheet, but it breaks Tearsheet animations because Tearsheet does not participate in Carbon’s presence contract (presence context consumption and exit coordination). As a result, presence has no way to know when the Tearsheet’s exit animation completes and unmounts immediately.
> 
> That said, we agree with the underlying goal. Providing a Tearsheet-specific presence boundary that allows consumers to opt into full lifecycle reset—similar in spirit to `ComposedModalPresence`-is a valid improvement area and something we can revisit with a dedicated implementation rather than direct reuse.
> 
> @amal-k-joy I'm moving this PR back to draft until the rework is complete.

@timfuhrmann @makafsal ,
Do you think changes or improvements to the current ComposedModalPresence in Carbon would help enable Tearsheet to consume it as well?

timfuhrmann:
I am not sure if making changes to ComposedModalPresence would help, curious to here what you might suggest though. 

What I was thinking though, that the Tearsheet implements its own `TearsheetPresence`, almost identical to the [ComposedModalPresence](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/ComposedModal/ComposedModalPresence.tsx), but not using it, with its own context etc. 

Both of these internal carbon utils should be sufficient to achieve a similar outcome: 
- https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/internal/usePresenceContext.ts
- https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/internal/usePresence.ts

However, since they are not exposed by carbon yet, we would either have to ask for it or copy/paste. 

Happy to hear your thought or create a first draft to get started, just let me know.

makafsal:
@timfuhrmann Yeah, those hooks are almost identical, but in `Tearsheet`, the entry/exit animations are implemented using the transition property, not animation. In Carbon hooks implementation, it uses `getAnimations`, and the Tearsheet hook `getComputedStyle` to track the entry/exit animation.

PR #8866, opened 2026-01-27
fix: pageheader position on scroll and resize

### LINKED ISSUE #8792: PageHeader position is wrong when scrolling on narrow viewports
### Package

Carbon for IBM Products

### Description

Hi team,

When using the PageHeader component, if the user shrinks the viewport to a narrow size, moves scroll position down from the top and refreshes the page, the component's position is wrong; it appears to stick to the bottom or the middle of the page whereas we expected it to stick to the top - please see the stackblitz for a minimal repro example.

Thank you! :)

### Component(s) impacted

PageHeader (preview)

### Browser

Firefox

2.80.0

### Suggested Severity

### Product/offering

Usage Analytics

https://stackblitz.com/edit/vitejs-vite-u1nssef9?file=src%2FApp.tsx

1. Scroll down a bit
2. Make the viewpoint narrower
3. The PageHeader's position should now be wrong:

Image

_No response_

### Code of Conduct

amal-k-joy:
Closes #8792 

When changing the size of the viewport after scrolling, the top position for the sticky elements is incorrectly set. 

#### What did you change?

- Updated the getHeaderOffset util method to give the correct calculations when the pageHeader is in document level and also inside a scrollable container
- Added a mock page header in the tab bar with tabs and tags story to test 
- Fixed for both react & wc

Storybook (see "Tab bar with tabs and tags" story for an example with a mock application header)

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/package.json
+++ b/packages/ibm-products/package.json
@@ -102,7 +102,7 @@
     "rollup-plugin-strip-banner": "^3.1.0",
     "sass": "^1.93.2",
     "storybook": "^9.0.8",
-    "typescript-config-carbon": "^0.8.0",
+    "typescript-config-carbon": "^0.9.0",
     "vite": "^7.0.0",
     "yargs": "^18.0.0"
   },
--- a/yarn.lock
+++ b/yarn.lock
@@ -2108,7 +2108,7 @@ __metadata:
     rollup-plugin-strip-banner: "npm:^3.1.0"
     sass: "npm:^1.93.2"
     storybook: "npm:^9.0.8"
-    typescript-config-carbon: "npm:^0.8.0"
+    typescript-config-carbon: "npm:^0.9.0"
     vite: "npm:^7.0.0"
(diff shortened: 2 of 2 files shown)


packages/ibm-products/src/components/PageHeader/next/PageHeader.stories.jsx
```diff
@@ -451,128 +451,152 @@ const tabBarTags = [
...
+        backgroundColor: '#fff',
+        position: 'fixed',
+        top: 0,
+        left: 0,
+        inlineSize: '100%',
+        borderBottom: '1px solid #eee',
+        zIndex: 10000,
+      }}
+    >
+      <span>Application header</span>
```
szinta: Do we need this additional header bar in the story? It looks a bit odd to me. Could we use the Header component from UI Shell instead, so the look and feel will be consistent with other stories?

makafsal (approved):

PR #8929, opened 2026-02-05
feat(tearsheetpresence): add tearsheet presence

timfuhrmann:
Closes https://github.com/carbon-design-system/ibm-products/issues/8293

As discussed here https://github.com/carbon-design-system/ibm-products/pull/8822, this is my example implementation to be 100% aligned with carbon @amal-k-joy @makafsal 

Since carbon is not exposing usePresence and usePresenceContext yet, I copy & pasted both hooks. If we can get carbon to expose both, they could be used alomst plug'n'play (prefix for keyframe animations would have to be updated).

```tsx
// Option A: opt-in via feature flag
<FeatureFlags enablePresence>
    <Tearsheet {...args} open={open} />
</FeatureFlags>

// Option B: opt-in via tearsheet presence
<TearsheetPresence open={open}>
    <MyTearsheet {...args} />
</TearsheetPresence>

// Option C: opt-in via higher-order function
const MyTearsheet = withTearsheetPresence(({ onSubmit }) => {
  const [foo, setFoo] = useState();
  return <Tearsheet
  {...args}
  actions={[{
    kind: 'primary',
    label: 'Submit',
    onClick: () => {
      onSubmit(foo);
    }
  }]}
  />;
})

<MyTearsheet open={open} />
```

#### What did you change?
- Implemented TearsheetPresence & withTearsheetPresence
- Updated Tearsheet to use CSS animations for entrance and exit, if feature flag enabled
- Added story for tearsheet presence

- Tested stacking in storybook
- Tested feature flag in storybook
- Tested TearsheetPresence in storybook

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products-styles/src/components/Tearsheet/_tearsheet.scss
+++ b/packages/ibm-products-styles/src/components/Tearsheet/_tearsheet.scss
@@ -83,7 +83,7 @@ $motion-duration: $duration-moderate-02;
         &.#{$block-class}.#{$block-class} .#{$block-class}__container {
           animation: #{$block-class}--presence-tearsheet--container__exit
             $motion-duration
-            motion(exit, expressive)
+            motion(entrance, expressive)
             forwards;
           @media screen and (prefers-reduced-motion: reduce) {
             animation: none;


makafsal (approved):
Excellent work @timfuhrmann 🎉

timfuhrmann:
Thank you! 

> Can you add the presence stories under `Feature flags` as in composeModal and remove `Tearsheet with Presence `story. Also we need a story for stacking tearsheet.

@amal-k-joy I created a new file for the presence stories, including narrow and stacked and made is accessible under. Feel free to update, if that's not what you had in mind: 

Bildschirmfoto 2026-02-10 um 08 34 44

----------------

> Currently you have kept the presence hooks locally, I was wondering if we can consume directly from carbon as below

I am not sure we should be directly importing from their ES module build (`/es/`). My guess is that this could mess up your bundling, specifically your CJS build? However, I am not an expert in this area and it might as well just work?

amal-k-joy:
> Thank you!
> 
> > Can you add the presence stories under `Feature flags` as in composeModal and remove `Tearsheet with Presence `story. Also we need a story for stacking tearsheet.
> 
> @amal-k-joy I created a new file for the presence stories, including narrow and stacked and made is accessible under. Feel free to update, if that's not what you had in mind:
> 
> Bildschirmfoto 2026-02-10 um 08 34 44
> > Currently you have kept the presence hooks locally, I was wondering if we can consume directly from carbon as below
> 
> I am not sure we should be directly importing from their ES module build (`/es/`). My guess is that this could mess up your bundling, specifically your CJS build? However, I am not an expert in this area and it might as well just work?

Storybook examples looks great @timfuhrmann . 
However there are 2 issues I noticed in stacking examples. 
1.  Exit animation in stacking tearsheet is bit off.
2. The width of all stacked tearsheets appears same.

I know the Tearsheet stacking implementation is a bit complex, and this issue may have been introduced when we switched from transitions to animations. I was wondering if the usePresence hook could also listen for and support transition events. If that approach works, we could make the change in Carbon’s usePresence utility and then consume it from there.

You are right, consuming from es/internals can mess up bundling and CJS support. Lets go with local copy for now.

timfuhrmann:
@amal-k-joy Thank you, I didn't notice that. - It was a CSS specificity issue where the stacked transform was overwritten. Width of all stacked Tearsheets look good to me now. Did this also fix your feeling that the stacking animation itself is a bit off? 

> I was wondering if the usePresence hook could also listen for and support transition events. If that approach works, we could make the change in Carbon’s usePresence utility and then consume it from there.

CSS animations are typically used for this, as they are exactly designed for this, since they express enter/exit as an explicit lifecycle, whereas CSS transitions only react to style changes. 

In other words, CSS transitions are state-change driven, while CSS animations are lifecycle driven, which is exactly what we need trying to express the beginning and ending of a Tearsheet's lifecycle. 

In any way, it should not be a problem to use a CSS animation for the entrance and a CSS transition for the stacking - I would event go as far to say, that this actually is the best practice, since entrance is life cycle driven and everything in regards to stacking is state driven. The stacking issue you were experiencing was simply just a specificity problem, that we would have also gotten with CSS transitions and is more due to the setup where we have to support both worlds, with and without presence.

timfuhrmann:
@amal-k-joy I just aligned my use of carbons easing functions with the current impl - this will probably fix "Exit animation in stacking tearsheet is bit off".

However I noticed that you have been using `motion(entrance, expressive)` for both the current enter and exit transition. I updated my impl to do the same thing. - However if `motion(exit, expressive)` would actually be correct for the exit animation, we could easily set that straight for the CSS animations version - just in case, and I understand if y'all want to keep it identical. 

Anyways, it looks identical to the non presence version now - at least to me 😁

amal-k-joy:
> @amal-k-joy Thank you, I didn't notice that. - It was a CSS specificity issue where the stacked transform was overwritten. Width of all stacked Tearsheets look good to me now. Did this also fix your feeling that the stacking animation itself is a bit off?
> 
> > I was wondering if the usePresence hook could also listen for and support transition events. If that approach works, we could make the change in Carbon’s usePresence utility and then consume it from there.
> 
> CSS animations are typically used for this, as they are exactly designed for this, since they express enter/exit as an explicit lifecycle, whereas CSS transitions only react to style changes.
> 
> In other words, CSS transitions are state-change driven, while CSS animations are lifecycle driven, which is exactly what we need trying to express the beginning and ending of a Tearsheet's lifecycle.
> 
> In any way, it should not be a problem to use a CSS animation for the entrance and a CSS transition for the stacking - I would event go as far to say, that this actually is the best practice, since entrance is life cycle driven and everything in regards to stacking is state driven. The stacking issue you were experiencing was simply just a specificity problem, that we would have also gotten with CSS transitions and is more due to the setup where we have to support both worlds, with and without presence.

Thanks @timfuhrmann for sharing all this knowledge.

Issue #8930, opened 2026-02-06
[Tearsheet]: heading sizes change randomly (regression)


wkeese:
### Package

Carbon for IBM Products

### Description

When a Tearsheet is inside one or more `<Section>` elements, its header is too small.

<img width="558" height="632" alt="Image" src="https://github.com/user-attachments/assets/15bd9deb-a714-4279-ada7-33f94687d895" />

This is a separate problem from #8828, but it has the same cause (@makafsal's https://github.com/carbon-design-system/ibm-products/pull/8675).  You should revert the code that sets a Tearsheet's header level dynamically.  

It would also be good to use a CSS class to set the heading text rather than relying on Carbon's CSS for the `H3` tag.

### Component(s) impacted

Tearsheet

### Browser

Chrome

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

2.83.0

### Suggested Severity

None

### Product/offering

IKC

### CodeSandbox or Stackblitz example

https://stackblitz.com/edit/github-drt3dur2?file=src%2FExample%2FExample.jsx

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

PR #8936, opened 2026-02-08
fix(Tearsheet): tearsheet heading level for title and label"

### LINKED ISSUE #8828: [a11y]: Meaningless <section> wrapper around Tearsheet heading (regression)
### Package

@carbon/ibm-products

### Browser

Chrome

### Operating System

MacOS

### Package version

2.81

### React version

18

N/A

### Assistive technology

none

### Description

`Tearsheet` now wraps its heading in two meaningless `<section>` elements:

Image

It used to look like:

Image

This is a regresion from @makafsal's #8675.   While it's generally good to not hardcode heading levels, Tearsheet may be an exception.  And in any case, the idea of `<section>` is to wrap a heading AND all the content associated with that heading.  It doesn't make sense to have a section with nothing but a heading.  Please see the example at https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/section.

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://carbon-for-ibm-products.netlify.app/?path=/story/components-tearsheet--tearsheet

### Steps to reproduce

Just look at https://carbon-for-ibm-products.netlify.app/?path=/story/components-tearsheet--tearsheet in DOM inspector.

### Code of Conduct

wkeese:
Closes #8828, #8930.

PR #8675 caused multiple regressions, both visually and semantically.

In particular, it broke the Tearsheet header size settings by applying the classes to the wrong node.

And, IMO, modals should always be at the H2 level, regardless of where they appear in the VDOM.

It's still questionable for Tearsheet to use *both* H2 and H3, but I didn't change that in this PR.

#### What did you change?

This reverts commit d952a8f256c29d97261c06799e22f63c799c9780.

Storybook

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
+++ b/packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
@@ -35,7 +35,6 @@ import {
   Section,
   usePrefix,
   unstable_FeatureFlags as FeatureFlags,
-  Heading,
 } from '@carbon/react';
 
 import { ActionSet } from '../ActionSet';
@@ -491,7 +490,7 @@ export const TearsheetShell = React.forwardRef(
                   effectiveHasCloseIcon ? closeIconDescription : undefined
                 }
               >
-                <Section
+                <Wrap
                   className={`${bc}__header-content`}
                   element={wide ? Layer : undefined}
                 >
@@ -499,14 +498,15 @@ export const TearsheetShell = React.forwardRef(
                     {/* we create the label and title here instead of passing them
                       as modal header props so we can wrap them in layout divs */}
                     <Wrap className={`${bcModalHeader}__label`}>{label}</Wrap>
-                    <Section
+                    <Wrap
+                      element="h3"
                       className={cx(
                         `${bcModalHeader}__heading`,
                         `${bc}__heading`
                       )}
                     >
-                      <Heading>{title}</Heading>
-                    </Section>
+                      {title}
+                    </Wrap>
                     <Wrap className={`${bc}__header-description`}>
                       {description}
                     </Wrap>
@@ -514,7 +514,7 @@ export const TearsheetShell = React.forwardRef(
                   <Wrap className={`${bc}__header-actions`}>
                     {headerActions}
                   </Wrap>
-                </Section>
+                </Wrap>
                 <Wrap className={`${bc}__header-navigation`}>{navigation}</Wrap>
               </ModalHeader>
             )}


wkeese:
@makafsal / @amal-k-joy - Can you change the PR title?  It's complaining that it's "invalid" and yet it won't allow me to change it.

packages/ibm-products/src/components/Tearsheet/TearsheetShell.tsx
```diff
@@ -491,30 +490,33 @@ export const TearsheetShell = React.forwardRef(
...
+                <Wrap
                   className={`${bc}__header-content`}
                   element={wide ? Layer : undefined}
                 >
                   <Wrap className={`${bc}__header-fields`}>
                     {/* we create the label and title here instead of passing them
                       as modal header props so we can wrap them in layout divs */}
-                    <Wrap className={`${bcModalHeader}__label`}>{label}</Wrap>
-                    <Section
+                    <Wrap element="h2" className={`${bcModalHeader}__label`}>
```
amal-k-joy: Hi @wkeese ,
Thanks for the contribution
It looks like this change renders the header label as h2 and the main heading (title) as h3. Is there a specific reason the label should be an h2 ?

If I remember correctly from the discussion with the accessibility team, we agreed the label could remain a div, and since the tearsheet is a dialog, we could start the heading hierarchy at h1 or h2. 
 Appreciate your thoughts on this

wkeese: I agree it makes more sense for the main heading (`title`) to be H2 and the `label` to be a DIV.  I updated to do that in aa0c475.

amal-k-joy (approved):

amal-k-joy:
@wkeese 
Few tests are failing likely because in main content and influencer, headings are starting at h4. Looks like some changes are required in body levels as well

Also worth considering our new [Tearsheet](https://ibm-products.carbondesignsystem.com/?path=/docs/preview-tearsheet--overview) which is in preview mode and that offers more flexibility.

wkeese:
> @wkeese Few tests are failing likely because in main content and influencer, headings are starting at h4. Looks like some changes are required in body levels as well

OK, I did a quick update so the nested headers to start at H3.  Ideally, all the code would use `<Section>` and `<Heading>`, and levels would be set automatically, but that seems like a waste of time since `Tearsheet` and `EditTearsheet` are deprecated. 

> Also worth considering our new [Tearsheet](https://ibm-products.carbondesignsystem.com/?path=/docs/preview-tearsheet--overview) which is in preview mode and that offers more flexibility.

I see, I think it's good to use for the future.  It's like `ComposedModal` in @carbon/react.  But judging from  judging from https://ibm-products.carbondesignsystem.com/?path=/docs/preview-tearsheet--overview, it will need to be fixed for accessibility too.

amal-k-joy (approved):

wkeese:
I agree it's better for tearsheets to have a single header, but it looks like fixing that caused lots of spurious diffs.  Presumably mainly due to fragile tests/storybook that set styling based on the tag (H3, H4, H5) etc. rather than a class name.  I don't have permission to see the percy output, but can you fix the remaining failures @amal-k-joy?   Alternatively, should we go back to my original commit that just fixes the regression?

As for the code coverage, I don't know what happened there either.  I assume it's a spurious failure?

amal-k-joy:
> #8675

Hi @wkeese ,
That makes sense. I’m a bit worried that switching the title to an h2 might break existing adopters. Let’s stick with your original change: label as a div and the title as an h3.

wkeese:
OK, except my "original change" (1e5e450918de1948e7031b3613aebb393c22a33f) -- which is really just reverting to the original code -- had an H2 and H3.

amal-k-joy:
> OK, except my "original change" ([1e5e450](https://github.com/carbon-design-system/ibm-products/commit/1e5e450918de1948e7031b3613aebb393c22a33f)) -- which is really just reverting to the original code -- had an H2 and H3.

Would keeping the label as a div (as it is today) and using an h3 for the title resolve the issue?

wkeese:
> Would keeping the label as a div (as it is today) and using an h3 for the title resolve the issue?

Yes, I think that still addresses the regressions that I filed as https://github.com/carbon-design-system/ibm-products/issues/8828 and https://github.com/carbon-design-system/ibm-products/issues/8930.

You might get flagged for a11y violations since the tearsheet starts at H3 rather than H2 (essentially skipping over H2) but that was always an issue for tearsheets that don't have a label (like https://carbon-for-ibm-products.netlify.app/?path=/story/components-tearsheet--tearsheet).

wkeese:
> Would keeping the label as a div (as it is today) and using an h3 for the title resolve the issue?

I pushed that change.  For some reason there are a bunch of Percy errors.  What's causing those?   I checked and the label has the `cds--modal-header__label` class that sets the font size, so it shouldn't matter if it's a `<div>` or `<h2>`.

PR #8917, opened 2026-02-03
fix(PageHeader): use TruncatedText in PageHeader subtitle

ruchas31:
Closes #8615 

Use TruncatedText in storybook file of the PageHeader to truncate the subtitle 2 lines.

#### What did you change?

- Use TruncatedText in storybook file of the PageHeader to truncate the subtitle 2 lines
- Add tests for code coverage
- remove the existing overflow logic and reference to DefinitionTooltip component

#### How did you test and verify your work?
Local storybook and ran all PageHeader tests.

#### PR Checklist

As the author of this PR, before marking ready for review, confirm you:

- [ ] Reviewed every line of the diff
- [ ] Updated documentation and storybook examples
- [ ] Wrote passing tests that cover this change
- [ ] Addressed any impact on accessibility (a11y)
- [ ] Tested for cross-browser consistency
- [ ] Validated that this code is ready for review and status checks should pass

More details can be found in the [pull request](./CONTRIBUTING.md) section of
our contributing docs.

PR #8963, opened 2026-02-11
fix: add overflow options for page header

### LINKED ISSUE #2501: PageHeader: Subtitle has no way to view overflow text

- [ ] Legacy/CDAI
- [ ] Legacy/Security

## Detailed description

When one provides a long string to the `subtitle` prop of the `<PageHeader>` component, the text overflows (by design). There's no way for the user to see this text without resizing their browser. I took this screenshot of [the Storybook](https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-components-page-header-pageheader--fully-loaded-and-some) and changed the breakpoints at the top.
image

Obviously, the simple answer is to not put so much text there, but translations make that difficult. Sometimes some languages take up much more space to convey the same idea.

This isn't exactly a bug, but I'd expect to be able to provide the full string so it displays on hover. Unfortunately we can't provide a `<span title="text">` since the `subtitle` only "accepts" string props. My requested behavior is either:
1. allow arbitrary JSX in the `subtitle`
2. add a new prop similar to `actionBarOverflowAriaLabel` and handle the wrapping inside the component
These would look like this to the user:
image

Firefox Developer / Chrome Dev

> package are you using?

latest

> should be aware of?

IBM Security Verify

## Additional information

- This may also apply to `title`, but there's more prop options there I haven't tested.

Ciberramir52:
Closes #2501

Implement a solution where the user can show the subtitle as a ReactNode or as another type, adapting the styles to this behavior and adding a storybook for using the subtitle with an expand option.
This by passing a string as subtitle:
image
This by using Truncated Text with Tooltip:
image
And this by using Truncated Text with expand:
image
image

#### What did you change?
- Updated the PageHeader Option to check the existence and the type of subtitle to adapt the styles.
- Updated the storybooks for showing this new behaviors.

- Tested stacking in storybook
- Tested feature flag in storybook

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/PageHeader/PageHeader.stories.jsx
+++ b/packages/ibm-products/src/components/PageHeader/PageHeader.stories.jsx
@@ -772,7 +772,7 @@ export const withSubtitle = Template.bind({});
 withSubtitle.storyName = 'Page header with title and subtitle';
 withSubtitle.args = {
   title: 2,
-  subtitle: longSubtitle,
+  subtitle: subtitle,
   breadcrumbs: 2,
   navigation: 1,
   ...commonArgs,


amal-k-joy (commented):
Thanks for the contribution @Ciberramir52 

But , it looks like this issue is already fixed part of this [PR](https://github.com/carbon-design-system/ibm-products/pull/8917/changes) and its already merged.

Did you notice any more issues currently?

Ciberramir52:
I read that PR and it does not show anything about the expand option for the page header subtitle. In fact I tested the behavior first and the expansion of the hidden text wasn't working, it was fixed to two lines no matter the entire content of the subtitle

packages/ibm-products/src/components/PageHeader/PageHeader.stories.jsx
```diff
@@ -812,11 +824,28 @@ fullyLoaded.args = {
   ...commonArgs,
 };
 
-export const fullyLoadedAndSome = Template.bind({});
-fullyLoadedAndSome.storyName = 'Page header with long values and many items';
-fullyLoadedAndSome.args = {
+export const fullyLoadedAndSomeWithTooltip = Template.bind({});
```
amal-k-joy, line 827: Hi @Ciberramir52 ,
Since we already have too many stories for PageHeader, it wouldn’t be ideal to add another one. Instead, I would add the expand/collapse subtitle as a Storybook control.- `showExpandableSubtitle`

packages/ibm-products/src/components/PageHeader/PageHeader.tsx
```diff
@@ -1065,9 +1066,15 @@ export const PageHeader = React.forwardRef(
               {subtitle && (
                 <Row className={`${blockClass}__subtitle-row`}>
                   <Column className={`${blockClass}__subtitle`}>
-                    <span className={`${blockClass}__subtitle-text`}>
-                      {subtitle}
-                    </span>
+                    {isValidElement(subtitle) ? (
```
amal-k-joy, line 1069: I was wondering why we would need to add this?
truncated-text component us  already adding the similar styles for clamping.

packages/ibm-products-styles/src/components/PageHeader/_page-header.scss
```diff
@@ -585,6 +585,12 @@ $duration: 1000ms;
     line-clamp: 2;
   }
 
+  .#{$block-class}__subtitle-node {
```
amal-k-joy, line 588: I was wondering why we would need to add this?
truncated-text component us  already adding the similar styles for clamping.

Ciberramir52: I did that in case the user wanted to use a component other than Truncated Text

packages/ibm-products/src/components/PageHeader/PageHeader.tsx
Ciberramir52: I did that in case the user wanted to use a component other than Truncated Text

packages/ibm-products-styles/src/components/PageHeader/_page-header.scss
amal-k-joy: In such cases, adopters can consider applying additional styles on their end if needed.

packages/ibm-products/src/components/PageHeader/PageHeader.tsx
amal-k-joy: In such cases, adopters can consider applying additional styles on their end if needed.

amal-k-joy:
Closing as conflicted and stale. Please feel free to re-open if needed. 
Thanks

Issue #8864, opened 2026-01-27
[TEARSHEET CHILDREN REMOUNTING]: Children in <Tearsheet> remount multiple times


BonbonLemon:
### Package

Carbon for IBM Products

### Description

Children of <Tearsheet> will be mounted>unmounted>mounted.
Note: I've turned off React.strict mode to better demonstrate this.

### Component(s) impacted

<Tearsheet>

### Browser

Chrome

### @carbon/ibm-products (previously @carbon/ibm-cloud-cognitive) version

2.80.0

### Suggested Severity

Severity 4 = The problem is not visible to or noticeable to an average user. Affects minor functionality, no workaround needed.

### Product/offering

StreamSets

### CodeSandbox or Stackblitz example

https://stackblitz.com/edit/github-omwflzs3

### Steps to reproduce the issue (if applicable)

I have 2 buttons that both open up the same Counter component.
The first opens it up in a tearsheet, and another opens it up in a div.
When the tearsheet opens, it mounts the children twice and the counter increments twice.
The div properly only increments the counter once.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/ibm-products/blob/main/.github/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/carbon-design-system/ibm-products/issues) for duplicate issues

amal-k-joy:
Hi @BonbonLemon ,
The double rendering is happening due to how the `portalTarget` prop is implemented. We’ve recently pushed a fix for this—could you try passing any element to   the `portalTarget` prop and check if the double rendering issue is resolved? 
Also this issue is completely addressed in [new Tearsheet](https://ibm-products.carbondesignsystem.com/?path=/docs/preview-tearsheet--overview)
`portalTarget={document.body}`

PR #9024, opened 2026-02-20
fix(usePortalTarget): fix hydration error

timfuhrmann:
@amal-k-joy Reverts https://github.com/carbon-design-system/ibm-products/commit/28f69084db375fd7294c31cff6579f2e6f33b11b, since it's causing hydration errors in SSR environments. However, we still need to make sure that `renderPortalUse` is not returning anything as long as the portal target has not been set, otherwise the component will mount without the portal first, then unmount and mount in the portal again, in other words: it renders twice. 

Previously this went unnoticed most of the times - even though there are issues about it (https://github.com/carbon-design-system/ibm-products/issues/8864) - since this happened during closed state. However now with TearsheetPresence, this is becoming more observable, especially for heavy components, since the Tearsheet mounts just when it opens.  

**Note:** There is an E2E failing due to the NarrowTearsheet's close button not receiving focus. However this seems unrelated, as I can also see it fail for other PRs as well and I guess is because of the close button not being part of the NarrowTearsheet's story anymore. 

#### What did you change?

- Does not render anything until portal target has not been initialised 
- Adds isInitialized state (see [comment](https://github.com/carbon-design-system/ibm-products/pull/9024#discussion_r2836046406) for alternative)
- Moves portal to parent component to prevent effects from being executed before portal itself has mounted

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products/src/components/APIKeyModal/APIKeyModal.tsx
+++ b/packages/ibm-products/src/components/APIKeyModal/APIKeyModal.tsx
@@ -48,7 +48,15 @@ const defaults = {
   customSteps: [],
 };
 
-export const APIKeyModal: React.FC<APIKeyModalProps> = forwardRef(
+export const APIKeyModal: React.FC<APIKeyModalProps> = forwardRef<
+  HTMLDivElement,
+  APIKeyModalProps
+>(({ portalTarget, ...props }, ref) => {
+  const renderPortalUse = usePortalTarget(portalTarget);
+  return renderPortalUse(<APIKeyModalDialog ref={ref} {...props} />);
+});
+
+const APIKeyModalDialog = forwardRef<HTMLDivElement, APIKeyModalProps>(
   (
     {
       // The component props, in alphabetical order (for consistency).
@@ -96,7 +104,6 @@ export const APIKeyModal: React.FC<APIKeyModalProps> = forwardRef(
       onRequestEdit,
       onRequestGenerate,
       open,
-      portalTarget: portalTargetIn,
       previousStepButtonText,
       showAPIKeyLabel,
       helperText,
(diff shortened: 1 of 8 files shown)


packages/ibm-products/src/components/Tearsheet/next/Tearsheet.tsx
```diff
@@ -147,6 +147,13 @@ export type TearsheetComponentType = React.ForwardRefExoticComponent<
 };
 
 export const Tearsheet = forwardRef<HTMLDivElement, TearsheetProps>(
+  (props, ref) => {
+    const renderPortalUse = usePortalTarget(props.portalTarget);
+    return renderPortalUse(<TearsheetDialog ref={ref} {...props} />);
+  }
+) as TearsheetComponentType;
```
timfuhrmann, line 154: This is also were presence could be implemented for tearsheet v2

packages/ibm-products/src/global/js/hooks/usePortalTarget.js
```diff
@@ -12,22 +12,28 @@ import { useFeatureFlag } from '../../../components/FeatureFlags';
...
-  }
+    setIsInitialized(true);
+  }, [portalTargetIn, enablePortalTarget]);
 
   const renderPortalUse = useCallback(
-    (children) =>
-      portalTarget ? createPortal(children, portalTarget) : children,
+    (children) => {
+      if (!isInitialized) return null;
+      return portalTarget ? createPortal(children, portalTarget) : children;
```
timfuhrmann, line 35: Alternatively this could also return children right away if `portalTargetIn` is undefined and portal target is disabled, similar to the current behavior where `portalTarget` would never become defined in this case. 

```suggestion
     if (
        portalTargetIn ||
        (!enablePortalTarget &&
          !pkg.isFeatureEnabled('default-portal-target-body'))
      )
        return children;
      return portalTarget ? createPortal(children, portalTarget) : null;
```

timfuhrmann: I am not sure about the setup of pkg.isFeatureEnabled or enablePortalTarget, but if this is allowed to be called here, we would not need `isInitialized`.

PR #9062, opened 2026-02-27
feat: incorporate usePresence for preview_Tearsheet and doc updates

### LINKED ISSUE #8714: Incorporate usePresenceContext hook into the new Composable Tearsheet
After closing the new Tearsheet, the element currently remains in the DOM instead of being removed.
We need to ensure the element is properly removed once the Tearsheet is closed.

Integrate the new `usePresenceContext` hook into the Tearsheet to handle presence and cleanup correctly.

amal-k-joy:
Closes #8714

This will enable usePresence for preview_Tearsheet. Now component will unmount from dom on closure as a default behaviour. To opt out this behaviour , there is `keepMounted` prop.

Also updated docs with few more details

local

#### PR Checklist

our contributing docs.


packages/ibm-products/src/components/Tearsheet/next/Tearsheet.tsx
```diff
@@ -191,6 +201,13 @@ export const Tearsheet = forwardRef<HTMLDivElement, TearsheetProps>(
 
     const renderPortalUse = usePortalTarget(portalTarget);
 
+    // Use presence hook for enter/exit animations (unless keepMounted is true)
+    const { isPresent, isExiting } = usePresence(
+      presenceRef,
+      keepMounted ? true : open
+    );
```
timfuhrmann: What about moving this to a tiny wrapper component, which just forward all props to this components, but handles this hook and the early return null (similar to tearsheet v1 where `Tearsheet` calls `TearsheetDialog`)? This would make sure that even all this component's effects and states are only initialised when the tearsheet is open/present. This will reduce side effects during closed state and make this components mental model much clearer: it will always be present when it's mounted.

packages/ibm-products/src/components/Tearsheet/next/Tearsheet.tsx
```diff
@@ -255,6 +272,11 @@ export const Tearsheet = forwardRef<HTMLDivElement, TearsheetProps>(
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [stack]);
 
+    // Don't render if not present (after exit animation completes) - unless keepMounted is true
+    if (!keepMounted && !isPresent) {
+      return null;
+    }
+
     return renderPortalUse(
```
timfuhrmann, line 284: We've had this discussion, but I strongly recommend to not do portals internally anymore, now that presence is the default, but let the user handle this instead, similar to @carbon/react and Modal. What do you think?

timfuhrmann (commented):
Awesome!

packages/ibm-products/src/components/Tearsheet/next/Tearsheet.tsx
amal-k-joy: make sense, I have opened a ticket for this change.

packages/ibm-products/src/components/Tearsheet/next/Tearsheet.tsx
amal-k-joy: Good point. I have updated.

PR #9103, opened 2026-03-12
feat(preview_Tearsheet): add support for actions prop for footer

### LINKED ISSUE #9082: preview_Tearsheet:  Tearsheet.Footer to accept actions prop as well
Current implementation of  <Tearsheet.Footer> is what we kept as an open canvas. When this is composable , the complete responsibility of styling action buttons and its order is on user. 

The proposal is to accept actions prop as in old tearsheet and at the same time support children as well.

amal-k-joy:
Closes #9082

We had kept Tearsheet as an open canvas where user can render their own content. This approach is highly scalable at the same time this will create some overhead for existing consumers to migrate to new tearsheet.
This PR will support actions props where they can pass actions as an array as in old Tearsheet, where internally we use ActionSet which will sort and style the buttons as per carbon guidelines. At the same time can pass their custom children as well.

#### PR Checklist

our contributing docs.

### DIFF
--- a/e2e/components/OptionsTile/OptionsTile-test.avt.e2e.js
+++ b/e2e/components/OptionsTile/OptionsTile-test.avt.e2e.js
@@ -9,9 +9,6 @@
 
 import { expect, test } from '@playwright/test';
 import { visitStory } from '../../test-utils/storybook';
-import { pkg } from '../../../packages/ibm-products/src/settings';
-
-const blockClass = `${pkg.prefix}--options-tile`;
 
 test.describe('OptionsTile @avt', () => {
   test('@avt-default-state', async ({ page }) => {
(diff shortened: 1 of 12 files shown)


packages/ibm-products/src/components/Tearsheet/next/Tearsheet.stories.jsx
```diff
@@ -341,36 +341,30 @@ export const Default = ({
             </div>
           </Tearsheet.SummaryContent>
         </Tearsheet.Body>
-        <Tearsheet.Footer>
-          <div className="default__action-buttons">
```
szinta, line 345: Please remove the styles associated with this class from:
`ibm-products/packages/ibm-products/src/components/Tearsheet/next/_storybook-styles.scss`

amal-k-joy: I have added a story to demonstrate Footer with custom buttons , where this style is required

PR #9123, opened 2026-03-16
chore: migrate to tsdown for react and wc

### LINKED ISSUE #8938: Explore and adopt tsdown for builds
This is to continue the initiative of Matt in [this PR.](https://github.com/carbon-design-system/ibm-products/pull/8065) to adopt tsdown for our build.

- Adopt the changes from Matt's PR
- Go through the documentation to adopt ts down for react and wc
- Verify and ensure nothing is broken

amal-k-joy:
Closes #8938

This replaces our rollup build setup with [tsdown](https://tsdown.dev/) for both react and web components and using the TypeScript Compiler API to generate type declarations(.d.ts files) . There was a noticeable improvement in the react build times (~25 seconds down to ~8 seconds) and quality.  The current rollup-plugin-lit-scss uses the deprecated sass.render() API; migrating to the modern sass.compileString() API  would provide significant performance improvements for the web components package's SCSS-to-Lit template transformation.

 ### Build Comparison: Rollup → tsdown (React & Web Components)

**⚛️ React (@carbon/ibm-products)**

What’s working

- ESM (es/) and CJS (lib/) builds generated correctly
- JS, SCSS, and CSS outputs match Rollup
- TypeScript declarations included
- Added .d.ts.map for better IDE navigation

Key differences

- Slightly larger build (due to declaration maps)
- More files for improved TypeScript support
- Cleaner output with a unified runtime

Comparison of a sample component artifat generated using old rollup and tsdown :AboutModal

- Same API, types, and behavior (no breaking changes)
- ~24% smaller bundle, ~50% less code
- Modern syntax (no Babel helpers)

🌐 Web Components (@carbon/ibm-products-web-components)

Improvements

- Cleaner builds (no tests, helpers, or story assets)
- Modern runtime (oxc-project)
- Source maps + declaration maps
- Smaller, more efficient bundles
- Same APIs and functionality

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products-web-components/package.json
+++ b/packages/ibm-products-web-components/package.json
@@ -75,7 +75,7 @@
     "@carbon/motion": "^11.42.0",
     "@figma/code-connect": "^1.4.0",
     "@ibm/telemetry-js-config-generator": "^2.1.0",
-    "@mordech/vite-lit-loader": "^0.37.0",
+    "@mordech/vite-lit-loader": "^0.38.0",
     "@open-wc/testing": "^4.0.0",
     "@stackblitz/sdk": "^1.11.0",
     "@storybook/addon-a11y": "^9.1.15",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2191,7 +2191,7 @@ __metadata:
     "@ibm/telemetry-js-config-generator": "npm:^2.1.0"
     "@lit-labs/signals": "npm:^0.1.2"
     "@lit/context": "npm:^1.1.5"
-    "@mordech/vite-lit-loader": "npm:^0.37.0"
+    "@mordech/vite-lit-loader": "npm:^0.38.0"
     "@open-wc/testing": "npm:^4.0.0"
     "@stackblitz/sdk": "npm:^1.11.0"
     "@storybook/addon-a11y": "npm:^9.1.15"
@@ -3689,6 +3689,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@esbuild/aix-ppc64@npm:0.19.12":
+  version: 0.19.12
+  resolution: "@esbuild/aix-ppc64@npm:0.19.12"
+  conditions: os=aix & cpu=ppc64
+  languageName: node
+  linkType: hard
+
 "@esbuild/aix-ppc64@npm:0.25.0":
   version: 0.25.0
   resolution: "@esbuild/aix-ppc64@npm:0.25.0"
@@ -3700,6 +3707,13 @@ __metadata:
   version: 0.27.2
   resolution: "@esbuild/aix-ppc64@npm:0.27.2"
   conditions: os=aix & cpu=ppc64
+  languageName: node
+  linkType: hard
+
+"@esbuild/android-arm64@npm:0.19.12":
+  version: 0.19.12
+  resolution: "@esbuild/android-arm64@npm:0.19.12"
+  conditions: os=android & cpu=arm64
   languageName: node
   linkType: hard
 
@@ -3717,6 +3731,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@esbuild/android-arm@npm:0.19.12":
+  version: 0.19.12
+  resolution: "@esbuild/android-arm@npm:0.19.12"
+  conditions: os=android & cpu=arm
+  languageName: node
+  linkType: hard
+
 "@esbuild/android-arm@npm:0.25.0":
   version: 0.25.0
   resolution: "@esbuild/android-arm@npm:0.25.0"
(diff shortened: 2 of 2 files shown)


packages/ibm-products-web-components/tasks/build.cjs
kennylam, line 1: Is there a reason this file is CJS and not `.js` (ESM) like the React package? I see that all the output are for Node, tooling, etc., and not the browser, which is fine. But I wonder if this is just debt for an ESM refactor in the future.

packages/ibm-products-web-components/tasks/build.cjs
```diff
@@ -0,0 +1,117 @@
...
+  const parsed = ts.parseJsonConfigFileContent(
+    configFile.config,
+    ts.sys,
+    path.dirname(tsconfigPath),
+    {
+      declaration: true,
+      declarationMap: true,
+      emitDeclarationOnly: true,
+      noEmit: false,
+      noEmitOnError: false,
```
kennylam: You mentioned the decision to use `tsc` because it potentially results in full fidelity maps and typings. You're probably right about this (I'm thinking of changing our build to do this too, now that you point it out). However, the `noEmitOnError` here means type errors only surface as warnings, not errors, which seems to negate any inherent benefits in using `tsc`. Suggest turning this back on (here, and in the other 2 instances in build files), unless there is another step in CI/CD for type-checking, in which case leaving off for builds is fine.

packages/ibm-products-web-components/tsdown.config.ts
```diff
@@ -0,0 +1,214 @@
...
+            return '[name].scss.js';
+          }
+          return '[name].js';
+        },
+        exports: 'named',
+        preserveModules: true,
+        preserveModulesRoot: path.resolve(import.meta.dirname, 'src'),
+        sourcemap: true,
+      };
+    },
```
kennylam: This `outputOptions` block is identical to the one below (CJS build). You could probably extract into `sharedConfig`.

packages/ibm-products-web-components/tools/rollup-plugin-lit-scss.js
kennylam, line 76: This seems to be the last instance of Rollup usage. We handled this query directly with a `resolveId` hook, making `@rollup/pluginutils` redundant.

https://github.com/carbon-design-system/carbon/blob/796bf1ccdf7b392d8e597a4acb281cd40e36b103/packages/web-components/tools/lit-scss-plugin.js#L78

The `transform` hook receives and compiles normally, no alias needed.

https://github.com/carbon-design-system/carbon/blob/796bf1ccdf7b392d8e597a4acb281cd40e36b103/packages/web-components/tools/lit-scss-plugin.js#L119

packages/ibm-products/tsdown.config.ts
```diff
@@ -0,0 +1,67 @@
...
+  platform: 'browser',
+  dts: false,
+  unbundle: true,
+  external: getExternalPatterns(),
+  failOnWarn: false,
+  logLevel: 'warn',
+  loader: {
+    '.js': 'jsx',
+  },
+  target: 'es2020',
```
kennylam: The web components config uses `2022`. Whichever is correct, they should be consistent I think?

packages/ibm-products-web-components/tsdown.config.ts
kennylam, line 1: In `@carbon/web-components`, this config was added directly to the [build file](https://github.com/carbon-design-system/carbon/blob/main/packages/web-components/tasks/build.js). We found this approach a bit more manageable since the web components build is a bit more complex than React.

If this approach works for you, you could remove this file entirely (and ignore my comment below about the duplicated outputOptions).

packages/ibm-products-web-components/tasks/build.js
kennylam, line 1: Is this file being used? It seems like a duplicate of `build.cjs`. See my comment in `packages/ibm-products-web-components/tsdown.config.ts` about consolidating the config and build files like the core monorepo.

packages/ibm-products-web-components/tasks/build.cjs
amal-k-joy: You are right, I made it to .cjs when I hit wit a ESM error due to the usage of require. But now I updated `require` with `import` and able to resolve it.

packages/ibm-products-web-components/tasks/build.cjs
amal-k-joy: Thanks for pointing that out.

packages/ibm-products-web-components/tasks/build.js
amal-k-joy: This indeed was a duplicate.

packages/ibm-products-web-components/tsdown.config.ts
amal-k-joy: Yes. makes sense. I have deleted the tsdown.config files and used build.js for both react and wc . I believe the config is much more aligned with core now.

packages/ibm-products-web-components/tools/lit-scss-plugin.js
```diff
@@ -33,7 +68,7 @@ function transformToTemplate(css) {
  * @returns {object} The rollup plugin to transform an `.scss` file to a `lit-html` template.
```
kennylam:
```suggestion
 * @returns {object} The plugin to transform an `.scss` file to a `lit-html` template.
```

anamikaanu96 (approved):

PR #9136, opened 2026-03-18
fix(side-panel): Parity work

### LINKED ISSUE #8734: [Side panel] WC visual parity bug
#8454 — These are the visual bugs we discovered in the visual parity work stream.

## Tasks

- [ ] Review the documented examples in the [Figma file ](https://www.figma.com/design/UuGd8Ua7ovPZc3U2GlubPv/WC-Design-Review?node-id=139-1085&t=0MXLAPKpOoO2fkND-11).
- [x] **Sizes**: The XL size is missing from the options
- [ ] **Scrolling title animation**: visual overlap stuck in between states. When the panel has minimal scroll, it prevents the completion of animation. 
- [ ] **Scrolling title animation**: The close button and AI icon are lost (scrolled out of view) when the content is scrolled. 
- [x] **Header**: The size of the header, when scrolled (condensed), does not resize to hug the title text as it has a fix size. This seems to be part of the title or subtitle, as it retain the spacing when actions are enabled. 
- [x] **Action toolbar**: Missing spacing between primary and ghost actions. Add `8px` gap after primary button. 
- [x] **AI background**: The AI background layer is not blending with the panel layer and the header. The blue aura fades into white instead of matching the panel layer. This issue can be seen in the following themes: White, G90, G100. 
- [x] **Panel border**: The panel right border `border-inline-end` is using incorrect token. Edit to be `border-subdle-01`. 
- [x] **Panel border**: The panel right border, should only be visible on left panel placement. Currently is visible also on the default view. 
- [ ] **Action items**: Across the large and xtra large size, the Ghost button—when paired with other options such as "three buttons with ghost"—does not resize properly to fill the available area. This should take the full width available and have the text 16px from the left. 
- [x] **Action items**: Tertiary button is not an option, use ghost instead.
- [x] **Action items**: The primary button should always sit on top when multiple actions are in place.
- [x] **Action items**: When more than 1 action items are chosen, there should be a separator between buttons.
- [x] **Accessibility**: Focus trap skip the header area and goes directly within the content. 
- [x] **Accessibility**: When focusing+enter to open the AI label, the keyboard navigation gets stuck and focused on the AI icon (after closing the popover). This restrict navigation and there is no way to move to the next item. 
- [ ] **Storybook**: Missing feature flag option
- [x] **Storybook**: Disparity of options across `action items` and `sizes`.

ruchas31:
Closes #8734 

Side Panel web component parity work

#### What did you change?
All points in the issue description are addressed except for:

- Animation points
- Action items - buttons to resize properly to fill the available area. The react component uses ActionSet utility and there is no such utility for WC so this will need to be developed separately. New issue created #9139.
- Storybook: Missing feature flag option - the react component uses Resizer utility from carbon labs. This is not available in web component version. So this will need to be developed separately. New issue created #9138.

Storybook

#### PR Checklist

our contributing docs.

### DIFF
--- a/packages/ibm-products-web-components/src/components/side-panel/side-panel.test.ts
+++ b/packages/ibm-products-web-components/src/components/side-panel/side-panel.test.ts
@@ -342,7 +342,7 @@ describe('c4p-side-panel', () => {
       `.${prefix}--action-set__action-button`
     );
 
-    expect(actionItems).toHaveLength(3);
+    expect(actionItems).toHaveLength(2);
   });
 
   it('should display a close button by default', async () => {


packages/ibm-products-web-components/src/components/side-panel/side-panel.scss
```diff
@@ -86,7 +86,7 @@ $block-class-action-set: #{$prefix}--action-set;
       @extend .#{$block-class}--left-placement;
 
       /* remove if https://github.com/carbon-design-system/ibm-products/pull/3983 merged */
```
amal-k-joy: Can we remove this comment , if this is not relevant now?
`remove if https://github.com/carbon-design-system/ibm-products/pull/3983 merged */`

PR #9243, opened 2026-04-10
fix(EditInPlace): resolve focus missing and few other issue

### LINKED ISSUE #6106: EditInPlace: focus lost when clicking on disabled save button
### Package

Carbon for IBM Products

### Description

when `EditInPlace` is in focus, if you click `cancel` the focus remains on the component, but if you click `cancel` focus is lost entirely.

i don't believe this is the intended behavior. i'm also not sure if clicking `cancel` should take away focus from the actual input. we should verify what the intended behavior is before fixing.

### Component(s) impacted

`EditInPlace`

### Browser

Chrome

2.49.1

### Suggested Severity

### Product/offering

ibm products

https://carbon-for-ibm-products.netlify.app/?path=/story/ibm-products-patterns-edit-and-update-editinplace--default&globals=viewport:basic

1. click `EditInPlace` in order to focus the input. the `cancel` and `save` buttons should appear from the right of the input. `save` should be disabled.
2. without altering the input, click the disabled `save` button
3. observe how focus from the component is lost entirely

_No response_

### Code of Conduct

makafsal:
Closes #6106 

Fixed the focus issue, also resolved a few issues mentioned in this [comment](https://github.com/carbon-design-system/ibm-products/issues/6106#issuecomment-4222935472).

#### What did you change?

image

image

image

image

Invalid state:

image

Storybook

#### PR Checklist

our contributing docs.


packages/ibm-products/src/components/EditInPlace/EditInPlace.tsx
```diff
@@ -342,7 +391,43 @@ export const EditInPlace = forwardRef<HTMLDivElement, EditInplaceProps>(
         ) : (
           inputElement
         )}
-        <div className={`${blockClass}__toolbar`}>
+        {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions -- This div is not interactive; onMouseDown is used only to detect clicks on disabled buttons within to prevent blur */}
+        <div
+          className={`${blockClass}__toolbar`}
+          onMouseDown={(e) => {
```
amal-k-joy, line 397: Can we move this function to a handler?

amal-k-joy (changes requested):
Everything looks good 🎉 
Added 1 minor comment

amal-k-joy (approved):

szinta (changes requested):
As per the [requirement](https://github.com/carbon-design-system/ibm-products/issues/6106#issuecomment-4025298413), after saving we should remove focus from the component, which is not implemented.

> Case 3: After saving, focus should move to the next item — Move out/tab to next

makafsal:
> As per the [requirement](https://github.com/carbon-design-system/ibm-products/issues/6106#issuecomment-4025298413), after saving we should remove focus from the component, which is not implemented.
> 
> > Case 3: After saving, focus should move to the next item — Move out/tab to next

Deciding which element is next will make this tiny component complex. So, better keep the focus on the component, and the user needs to press the `Tab` key to move focus to the next element.

szinta:
> Deciding which element is next will make this tiny component complex. So, better keep the focus on the component, and the user needs to press the `Tab` key to move focus to the next element.


I don’t think deciding which element should receive focus next falls within the scope of this component. As you mentioned, the next item will be focused when the user presses the Tab key. However, once the user clicks the Save button, we should remove focus from the component and render it in its default state, similar to how the component behaves [now](https://ibm-products.carbondesignsystem.com/?path=/story/components-editinplace--default). Can we recheck this with the design team?

szinta (approved):

PR #9381, opened 2026-05-10
fix(ibm-products): enable tree-shaking of unused components

it-rec:
Closes #

Module-scope side effects were preventing bundlers from dropping unused components from `@carbon/ibm-products`, even though the package's `sideEffects` array already declared the component files as side-effect free. Two patterns were responsible.

#### What did you change?

**`PageHeader.tsx`** — removed two stale lines at module scope:
- `pkg._silenceWarnings(true)` — that helper is documented as test/storybook-only ("should only be used... in internal test suites and storybook builds"), and shipping it in production code silenced every package warning whenever PageHeader was imported.
- `pkg.component.ActionBar = true` — dead code. No call site reads `pkg.component.ActionBar`, and `ActionBar` has no `checkComponentEnabled` gate anywhere in the codebase. The assignment existed only because the `_silenceWarnings(true)` above suppressed the Proxy's "enabled via feature flags" warning it would otherwise emit.

**29 component files** — annotated `Foo = pkg.checkComponentEnabled(Foo, componentName)` reassignments with `/* @__PURE__ */` so bundlers (Rollup, esbuild, Webpack/Terser) treat the call as pure and can eliminate the module when the export is unused. Behavior is unchanged when the component IS used — the wrap still happens and the live binding still flows through to consumers.

Affected components: `BreadcrumbWithOverflow`, `ButtonMenu`, `ButtonMenuItem`, `Decorator`, `DecoratorBase/DecoratorIcon`, `DecoratorDualButton`, `DecoratorLink`, `DecoratorSingleButton`, `DescriptionList`, `DescriptionListBody`, `DescriptionListCell`, `DescriptionListRow`, `ExampleComponent`, `ExampleDeprecatedComponent`, `FilterPanel`, `FilterPanelAccordion`, `FilterPanelAccordionItem`, `FilterPanelCheckbox`, `FilterPanelCheckboxWithOverflow`, `FilterPanelGroup`, `FilterPanelLabel`, `FilterPanelSearch`, `Nav`, `NavItem`, `NavList`, `ScrollGradient`, `StatusIndicator`, `StatusIndicatorStep`, `StringFormatter`.

- `yarn build:js` succeeds and the `/* @__PURE__ */` annotation is preserved in the emitted ESM output (verified `es/components/StringFormatter/StringFormatter.js`).
- Verified `es/components/PageHeader/` no longer contains any reference to `_silenceWarnings` or `component.ActionBar`.
- `tsc --noEmit` produces no new errors in the changed files (only pre-existing errors in `src/patterns/*/example/` apps, unrelated to this PR).
- All edits are non-behavioral: removed lines were either dead (`ActionBar`) or test-only (`_silenceWarnings`), and `/* @__PURE__ */` is a bundler hint that doesn't change runtime semantics.

#### PR Checklist

- [ ] Updated documentation and storybook examples — ~not applicable; no public API change~
- [ ] Wrote passing tests that cover this change — ~tree-shaking effectiveness isn't covered by the existing jest suite; verified manually via the built output~
- [ ] Addressed any impact on accessibility (a11y) — ~not applicable~
- [ ] Tested for cross-browser consistency — ~not applicable; bundler-time change~

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/ibm-products/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.jsx
+++ b/packages/ibm-products/src/components/BreadcrumbWithOverflow/BreadcrumbWithOverflow.jsx
@@ -355,7 +355,7 @@ export let BreadcrumbWithOverflow = ({
 };
 
 // Return a placeholder if not released and not enabled by feature flag
-BreadcrumbWithOverflow = pkg.checkComponentEnabled(
+BreadcrumbWithOverflow = /* @__PURE__ */ pkg.checkComponentEnabled(
   BreadcrumbWithOverflow,
   componentName
 );
--- a/packages/ibm-products/src/components/ButtonMenu/ButtonMenu.jsx
+++ b/packages/ibm-products/src/components/ButtonMenu/ButtonMenu.jsx
@@ -99,7 +99,10 @@ ButtonMenu.deprecated = {
 };
 
 // Return a placeholder if not released and not enabled by feature flag
-ButtonMenu = pkg.checkComponentEnabled(ButtonMenu, componentName);
+ButtonMenu = /* @__PURE__ */ pkg.checkComponentEnabled(
+  ButtonMenu,
+  componentName
+);
 
 // The display name of the component, used by React. Note that displayName
 // is used in preference to relying on function.name.
--- a/packages/ibm-products/src/components/ButtonMenu/ButtonMenuItem.jsx
+++ b/packages/ibm-products/src/components/ButtonMenu/ButtonMenuItem.jsx
@@ -27,6 +27,9 @@ export let ButtonMenuItem = React.forwardRef((props, ref) => (
 ));
 
 // Return a placeholder if not released and not enabled by feature flag
-ButtonMenuItem = pkg.checkComponentEnabled(ButtonMenuItem, componentName);
+ButtonMenuItem = /* @__PURE__ */ pkg.checkComponentEnabled(
+  ButtonMenuItem,
+  componentName
+);
 
 ButtonMenuItem.propTypes = OverflowMenuItem.propTypes;
--- a/packages/ibm-products/src/components/Decorator/Decorator.jsx
+++ b/packages/ibm-products/src/components/Decorator/Decorator.jsx
@@ -42,7 +42,7 @@ export let Decorator = React.forwardRef((props, ref) => {
 });
 
 // Return a placeholder if not released and not enabled by feature flag
-Decorator = pkg.checkComponentEnabled(Decorator, componentName);
+Decorator = /* @__PURE__ */ pkg.checkComponentEnabled(Decorator, componentName);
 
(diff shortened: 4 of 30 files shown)


it-rec:

it-rec:
Not a GitHub user? LOL 😂

amal-k-joy (changes requested):
Hi @it-rec ,
Thank you for this contribution! While we appreciate the effort to enable tree-shaking, the /* @__PURE__ */ annotations won't provide the intended benefit in our current architecture. Our barrel export pattern (export * from './components') means all components are evaluated at import time regardless of which ones are used, and we already have sideEffects configuration in package.json that handles tree-shaking for side-effect-free modules. I have opened a broader ticket to evaluate and remove this helper code as such. 

However the removal of these two lines from PageHeader.tsx is a legitimate improvement we can keep.

amal-k-joy:
@it-rec 
Looks like the DCO is still failing. I think **claude** is being counted as a committer/co-author identity and has not signed, which is why the check still fails even though you already signed.

it-rec:
What do you suggest going forward?

amal-k-joy:
> What do you suggest going forward?

I have opened this [issue](https://github.com/carbon-design-system/carbon/issues/22264) to deprecate and remove this set up

Closing this PR as the required code, i have included in esm upgrade PR

it-rec:
Got it. Thank you!

it-rec:
> Hi @it-rec , Thank you for this contribution! While we appreciate the effort to enable tree-shaking, the /* @**PURE** */ annotations won't provide the intended benefit in our current architecture. Our barrel export pattern (export * from './components') means all components are evaluated at import time regardless of which ones are used, and we already have sideEffects configuration in package.json that handles tree-shaking for side-effect-free modules. I have opened a broader ticket to evaluate and remove this helper code as such.
> 
> However the removal of these two lines from PageHeader.tsx is a legitimate improvement we can keep.

@amal-k-joy 

Two clarifications on the bundler reasoning, since this issue will be the long-term record:

1. **`/* @__PURE__ */` is not redundant with `sideEffects: false`.** The package field tells the bundler *"if no exports of this module are used, drop the module."* The annotation tells it *"this specific top-level call has no side effects — drop the call if its result is unused."* A statement like `Foo = pkg.checkComponentEnabled(Foo, name)` at module scope is opaque to bundlers; Rollup, esbuild, and Terser keep it conservatively, which then keeps the whole module alive. This is the documented use case for the annotation.

2. **Barrel re-exports do tree-shake** when the underlying modules are marked side-effect-free — that's the entire purpose of the `sideEffects` field. lodash-es, RxJS, and most modern libraries depend on this. The barrel pattern itself isn't what's preventing elimination; the top-level call in each component file is.

That said — removing `checkComponentEnabled` outright is strictly better than annotating around it, so I'm fully behind this direction. Happy to help review the deprecation PR.

PR #9514, opened 2026-06-07
fix(CreateFullPage): show step error state in pure pattern progress i…

### LINKED ISSUE #9389: CreateFullPage: Missing error state in progress indicator for CreateFullPage React pattern
In the [Create Full Page With Step In Error State ](https://ibm-products.carbondesignsystem.com/?path=/story/patterns-prebuilt-patterns-create-flows-createfullpage--create-full-page-with-step-in-error-state)story under CreateFullPage prebuilt pattern, the progress indicator displays error state when a step fails. However, this behaviour is missing in the equivalent pure pattern implementation, leading to an inconsistency between the two patterns.

Prebuilt pattern
Image

Pure pattern
Image

ayobamiseun:
Closes #9389

Adds the missing step error state to the pure (composable) CreateFullPage
pattern's progress indicator, matching the prebuilt pattern.

#### What did you change?

The progress indicator's error state is rendered by the shared
`CreateInfluencer`, which reads an `invalid` flag from each step's data.

- The **prebuilt** pattern feeds `invalid` into step data via
  `useRetrieveStepData`, so the error icon shows correctly.
- The **pure** pattern builds its step data manually in
  `CreateFullPage.jsx` and was dropping the step's `invalid` prop, so the
  indicator never reflected the error state — an inconsistency between the
  two patterns.

Changes (both in `patterns/CreateFullPage/example/components/`):
- `CreateFullPage.jsx`: pass `invalid: child.props.invalid` through when
  building `stepData`.
- `CreateInfluencer.tsx`: declare `invalid?: boolean` on the `Step`
  interface and drop the now-unneeded `(step as any)` cast.

Ran the pattern locally with `yarn storybook` and opened
*Patterns → CreateFullPage → Create Full Page With Step In Error State*.
The "Partition" step now shows the red error icon in the progress
indicator while invalid, and clears once a valid value is entered —
matching the prebuilt pattern's behavior.

#### PR Checklist

- [ ] ~Wrote passing tests that cover this change~ (pattern example components have no unit tests)


ayobamiseun:

amal-k-joy (approved):
Thanks @ayobamiseun for the contribution. 🎉

szinta (approved):

PR #9618, opened 2026-07-02
fix: story lag in set of patterns web components

devadula-nandan:
Closes N/A

Removes the significant lag i observed in `set-of-*` overflow stories.

Ever since the code panel was enabled in Preview Head, styles were also being rendered inside the code panel, which blocked browser resources and caused performance issues. this may have impacted other pages too. in case the stories render style tag in code panel. This PR moves those styles to story decorators for `set-of-*` overflow stories. other components need to be re-checked and remediated.

#### What did you change?

Moved styles to story decorators, which are ignored by the code panel. This prevents the styles from being rendered there and fixes the lag.

This PR also includes some necessary changes to account for gaps in overflow calculations in `packages/ibm-products-web-components/examples/set-of-breadcrumbs/src/overflowHandler.ts`
although breadcrumbs do not currently use gaps, these changes will later be contributed upstream to the `@carbon/utilities` package.

Ideally, we should remove all instances of relative paths and reference the utilities package directly in the near future.

We also have duplicated `set-of-*` example patterns in `page-header-*-set`, for example:

`packages/ibm-products-web-components/src/components/page-header/page-header-actions-set/page-header-actions-set.ts`

There is a lot more opportunity for cleanup and to better follow the principles described in [Atomic Design](https://atomicdesign.bradfrost.com/table-of-contents/?utm_source=chatgpt.com). Doing so would improve reusability and simplify maintenance. I'll try to discuss this with the team later.

storybook all 4 examples works as expected, no lag or calculation issues.

#### PR Checklist

our contributing docs.


amal-k-joy (dismissed):
Cool 🎉

anamikaanu96 (approved):

devadula-nandan:
not needed since we have recently added a story for the overflow handler utility in core.
[react](https://react.carbondesignsystem.com/?path=/story/utilities-overflowhandler--default)
[wc](https://web-components.carbondesignsystem.com/?path=/story/utilities-overflowhandler--default)

PR #9621, opened 2026-07-03
fix: design rerview changes react

### LINKED ISSUE #9542: Tearsheet design feedback
Changes requested based on design review of the preview Tearsheet, for both wide and narrow.

# Wide Tearsheet

## Collapsed header

Image
- The collapsed header should be `$spacing-08` / `40px` height, but currently appears to be `55px` height when collapsed.
- The close button should size down to `40x40px` to match the new height.

Image
- The collapsed header should also adjust the spacing between the leading/trailing icon with the title. By default, the spacing between the icon and the title should be `$spacing-05` / `16px`. When collapsed, the spacing should shrink to `$spacing-03` / `8px`. Currently it looks like the spacing is the same whether collapsed or default.

## Smaller screen sizes

Image

- This might be more to do with the Storybook demo setup, but the buttons overlap with the main content area titles once the Tearsheet's width causes it to lose the grid influencers. Instead, can we have the main content area text content fall below the buttons in that situation? If the title begins `$spacing-03` / `8px` below the large buttons that should give it enough space.
- Let's also size down these buttons that trigger the grid influencer in the demo to Medium instead of Large (these should appear as `40x40px` to match the Tab buttons that may appear, such as the expand button.

## Storybook specific

Image
- This is probably more a Storybook issue than anything, but the "Tab panel 1" label should align with the start of "Tab 1" This way we hint at the alignment of interior elements in the main content area.
- For the main content area, can we change the type token for the "Main content area title" to use the `$heading-03`? The larger size than the title might throw some folks off.

# Narrow

https://github.com/user-attachments/assets/809898e1-bc7b-41f5-8f9b-46edc6fd2803

- Referencing a [message from Laura ](https://github.com/user-attachments/assets/809898e1-bc7b-41f5-8f9b-46edc6fd2803) – the width of the Narrow Tearsheet should be set to `480px` and only respond with its width when it gets smaller (`320px`). It looks like currently the width shifts as the viewport shifts.

- The header area and main content area should use different background color tokens to mirror the design of the wide Tearsheet. Header area should use `$layer-01` while the main content area should use `$background`.

amal-k-joy:
Closes #9542 

Design review changes 

#### What did you change?
* Reduced the header height to 40px in the collapsed state, along with smaller close button and AI label sizes.
* Updated the menu button to use the `xs` size when collapsed.
* Enhanced the `onHeaderCollapse` callback so adopters can conditionally adjust action button sizes based on the collapse state.
* Set stories to be open by default.
* Added additional Storybook controls across all stories for better configurability.

#### PR Checklist

our contributing docs.


packages/ibm-products-styles/src/components/Tearsheet/_tearsheet_next.scss
```diff
@@ -368,7 +386,7 @@ $transform-start: translate3d(0, calc(min(95vh, 500px)), 0);
 
     .#{$block-class__next}__header-content-wrapper {
       min-block-size: $spacing-08;
-      padding-block-end: $spacing-04;
+      // padding-block-end: $spacing-04;
```
makafsal, line 389: We can remove the unwanted style instead of commenting it out.

makafsal (approved):

anamikaanu96 (approved):

RichKummer:
Took a look and saw much of this was resolved. Seeing a handful of design updates to make:


1. The influencer trigger buttons that appear on smaller screen sizes are rendering as `40x48px`. These should use the button size corresponding to `40x40px` so they don't look misshapen. This looks true for the left grid influencer trigger button and the one used for the right grid influencer.

Screenshot 2026-07-17 at 12 45 19 PM

2. When the trigger button appears on smaller screen sizes, it looks like there's no spacing between the button and the top of the main content area heading. Can we add `16px` spacing just so these elements are not flush?

Screenshot 2026-07-17 at 12 45 07 PM

3. The form elements that display in the main content area also don't have any spacing between the form fields. `32px` spacing between the form elements should give us enough space but I'll double check when the fixes are in.

Screenshot 2026-07-17 at 12 47 45 PM

PR #9667, opened 2026-07-15
fix(coachmark): remove unwanted code and added launcherbuttonref

### LINKED ISSUE #9649: Coachmark: dev review issues
- [x] Resolve duplicate keydown handlers — In CoachmarkContent.tsx, there are two nearly identical keydown event listeners handling the Escape key (lines 86–105 and 108–121). The first also restores focus to the trigger, while the second is redundant. Remove the duplicate and consolidate the logic into a single handler.

- [x] Simplify focus management in Coachmark — Remove the firstFocusable logic from Coachmark.tsx. Instead, introduce a new optional prop launcherButtonRef that allows adopters to pass a reference to the trigger element. When provided, focus should return to this element on close.

- [x] Fix setContentRef typing in context — Update the type of setContentRef in context.ts from (value: any) => void to (value: HTMLElement | null) => void, and ensure the default implementation aligns with this type.\

- [x] Make aria-label optional — Update the CoachmarkContent props to make the 'aria-label' attribute optional instead of required.

anamikaanu96:
Closes #9649 

- [ ] Removed duplicate keydown handlers from CoachmarkContent
- [ ] Remove the firstFocusable logic from Coachmark.tsx and introduced new optional prop launcherButtonRef 
- [ ] Updated the type of setContentRef in context.ts
- [ ] Updated the CoachmarkContent prop 'aria-label' as optional

#### PR Checklist

our contributing docs.


packages/ibm-products/src/components/Coachmark/next/Coachmark/Coachmark.tsx
```diff
@@ -155,32 +160,6 @@ export const Coachmark = forwardRef<HTMLDivElement, CoachmarkPropsNext>(
...
-
-      if (firstFocusable) {
-        triggerRef.current = firstFocusable;
-      }
-    }, [children, triggerRef]);
-
-    useEffect(() => {
-      const el = triggerRef.current;
-      if (el) {
-        el.setAttribute('aria-expanded', String(!!open));
```
amal-k-joy, line 180: we can set  aria-expanded here if launcherButtonRef is passed
` el.setAttribute('aria-expanded', String(!!open)); , `

amal-k-joy (changes requested):

amal-k-joy (approved):

szinta (approved):

PR #9817, opened 2026-08-18
feat(ConditionBuilder): composability review — value/onChange API and bug fixes

### LINKED ISSUE #7912: ConditionBuilder: composability review
Things to review within the current `ConditionBuilder` component to evaluate composability and address any existing pitfalls:

- What are the ways, if any, this component can be customized?
- Audit the number of props this component accepts?
  - Are there props that accept an array of objects or a single object (these may be potential red flags for composability)?
    - This could highlight areas where the component could introduce sub-components --> handing the responsibility of rendering to the adopter
- Are there sub-components usable by consumer?
  - If there aren't, what areas of the component would you consider breaking out into consumable sub-components to move towards a more composable api
- Based on these findings, draft a rough outline of what you would consider a strong developer experience in delivering a composable version of this component (keeping in mind inversion of control, handing rendering responsibility to the adopter as much as possible).

amal-k-joy:
Closes #7912

This PR addresses the composability review for `ConditionBuilder` ahead of stabilisation. Changes span the public API, state management, prop type fixes, and documentation.

---

#### New `value` / `onChange` API (replaces `initialState` / `getConditionState`)

- Added `value` prop — accepts a `ConditionBuilderState` object as either an **uncontrolled seed** (no `onChange`) or as a **controlled state** (paired with `onChange`).
- Added `onChange` prop — called with the full new state on every user change. Use alongside `value` for fully controlled mode, or standalone as a replacement for `getConditionState`.
- `getConditionState` and `initialState` remain functional during the deprecation window but emit a console warning when used. `value`/`onChange` take precedence when both are supplied.
- `getEmptyState` is now exported from the package barrel (`@carbon/ibm-products`) so controlled consumers can reset state externally.
- Stories migrated to use `value` instead of `initialState`. A new **"Controlled mode — value + onChange (Hierarchical)"** story demonstrates external state driving with Reset/Update buttons and a live state preview panel.

#### State management fixes

- `setRootState` now uses a functional updater (`setInternalState(prev => ...)`) so `onChange` always receives the committed value — eliminates stale-closure bugs when two mutations fire in the same render cycle.
- `onStartConditionBuilder` correctly handles all combinations of `value`, `onChange`, and `startActive`:
  - **Controlled** (`value` + `onChange`): just reveals the builder, never calls `setRootState`.
  - **Uncontrolled seed** (`value` only, groups still present, e.g. `startActive=false`): just reveals the builder.
  - **Uncontrolled seed, cleared** (`value` only, user removed all conditions): calls `getEmptyState()` to reinitialise the empty row.
  - **No seed**: existing `initialState` / `getEmptyState()` path unchanged.
- `initialState.enabledDefault` is skipped when `value` is also provided, preventing it from overwriting the controlled state.

#### New `startActive` prop

- Added `startActive` (default `true`) — when `false`, the builder starts behind the "Add condition" button even if `value` contains groups. The user must click the button to activate.
- Replaces the role previously played by `initialState.enabledDefault: false`.

#### TypeScript / PropTypes fixes

- `getConditionState` and `initialState` moved to a `deprecatedProps` export so they appear in a separate "deprecated" category in Storybook controls.
- `popOverSearchThreshold` is now optional (default `4`); was previously `isRequired`.
- `statementConfigCustom` is now optional; was previously required.
- `onAddItem` and `onRemoveItem` return type widened to `{ preventAdd/preventRemove: boolean } | void` to match real-world usage.
- `ConditionBuilderTextKeys` type exported; two missing translation keys (`searchPropertiesText`, `conditionRemovedText`, `conditionBuilderGroupText`) added.

---

### How did you test and verify your work?

- Tested all stories in Storybook: fresh builder, seed with `startActive=true` (default), seed with `startActive=false`, fully controlled mode (Reset/Update buttons), custom operators, dynamic options, hierarchical variant.
- Verified controlled mode: external state changes (reset, swap) update the builder; user edits propagate back via `onChange`.
- Verified uncontrolled seed + `startActive=false`: clicking "Add condition" renders the seeded state.
- Verified uncontrolled seed after clearing all conditions: clicking "Add condition" renders a fresh empty row with the property popover open.
- All existing tests pass; 331 lines of new tests covering `value`, `onChange`, `startActive`, and `getConditionState` callback.

---

#### PR Checklist


anamikaanu96 (dismissed):

packages/ibm-products/src/components/index.ts
```diff
@@ -112,6 +110,7 @@ export type {
   AddSelectProps as preview__AddSelectProps,
   AddSelectComponentType,
   AddSelectBodyProps,
+  AddSelectContentProps,
```
szinta:
```suggestion
```

szinta: This is not required

packages/ibm-products/src/components/index.ts
```diff
@@ -47,7 +47,6 @@ export * from './WebTerminal';
 export { AddSelectData } from '@carbon/ibm-products-utilities';
 export type {
   AddSelectItem,
```
szinta, line 49:
```suggestion
  AddSelectItem,
  ItemDetailEntry,
```

szinta (changes requested):
Noticed some changes related to AddSelect in index file which is not required/incorrect.  Please fix those. Other than that everything else looks good.

szinta (approved):

anamikaanu96 (approved):

Issue #9743, opened 2026-07-29
[preview__PageHeade]  Regression Testing & Code Review (React + Web Components)


amal-k-joy:
The new composable preview__PageHeader (React) and c4p-page-header (Web Components) are ready for promotion to Stable. Before the promotion , both implementations must undergo a full regression pass at every Carbon breakpoint and a complete code review. All issues discovered must be documented here and open separate issues  and resolved.

 _This ticket covers only the new (next/) PageHeader. The legacy PageHeader (deprecated) is out of scope._

### Scope
React: All PageHeader (next/) sub-components
Web Components: c4p-page-header and all related elements
Breakpoints: sm, md, lg, xlg, max (+ boundary checks)

**What to Validate**

1. Layout & Responsiveness

- [ ] No overflow or layout issues
- [ ] Correct hero image ratios (sm/md: 3×2 → lg+: 2×1)
- [ ] Grid alignment and responsive behaviour
- [ ] Proper truncation/overflow for actions, tags, breadcrumbs

2. Behaviour

- [ ] React callbacks fire correctly (collapse, clipping, overflow)
- [ ] WC context, offsets, and events work as expected
- [ ] Sticky/scroll behaviour consistent

3. React ↔ WC Parity

- [ ] Visual + functional parity across all features

4. Accessibility

- [ ] Keyboard navigation works
- [ ] No axe violations
- [ ] Correct ARIA roles and focus handling

5. Code Review

- [ ] API, types, and props validated
- [ ] No hardcoded styles (use Carbon tokens)
- [ ] No console errors
- [ ] Test coverage ≥ 90%

makafsal:
@szinta @amal-k-joy 

<img width="2040" height="82" alt="Image" src="https://github.com/user-attachments/assets/50e12153-c628-4496-bb9c-fc4d67413390" />

The `preview__PageHeader` doesn't seem to work as described by its equivalent approach. Also, an adopter has already raised a concern in the [support channel](https://ibm-studios.slack.com/archives/CQGR0HC05/p1785776566563169). Either we've to work onto achieve the behaviour as per the prop description or provide an additional prop to achieve this.

szinta:
## Regression review complete — child issues created

A full code review of both `preview__PageHeader` (React `next/`) and `c4p-page-header` (Web Components) was carried out against every checkpoint in this ticket. 6 child issues have been opened to track the findings:

---

### React (`preview__PageHeader`)

[...]

PR #9833, opened 2026-08-21
fix(PageHeader): code quality fixes and fullWidthGrid/narrowGrid support

### LINKED ISSUE #9795: [preview__PageHeader]: code quality and behaviour bugs
## Summary

Several code-quality and runtime behaviour issues found during pre-promotion review of `preview__PageHeader` (`next/`). Tracked from #9743.

---

## Bug 1 — Functional `pageActions` in `PageHeaderContent` always receives zeroed observer state

**File:** `packages/ibm-products/src/components/PageHeader/next/PageHeaderContent.tsx:142–148`

```tsx
// Current — always passes false/false/false regardless of scroll position
{typeof pageActions === 'function'
  ? pageActions({
      fullyCollapsed: false,
      titleClipped: false,
      contentActionsClipped: false,
    })
  : pageActions}
```

The render-prop variant of `pageActions` in `PageHeaderContent` is called with a hardcoded zeroed state object. `PageHeaderBreadcrumbBar` correctly reads live state from context via `observerState`. Consumers using `pageActions` as a function inside `PageHeaderContent` will never see state-driven updates.

**Fix:** Read `observerState` from `usePageHeader()` and pass it through.

---

## Bug 2 — Stale `subtitle` prop in `PageHeaderContent` PropTypes

**File:** `packages/ibm-products/src/components/PageHeader/next/PageHeaderContent.tsx:183`

`propTypes` declares `subtitle: PropTypes.string` but the TypeScript interface has no `subtitle` prop and nothing in the component body uses it. `subtitle` belongs on `PageHeaderContentText`, not `PageHeaderContent`. This produces misleading auto-generated API docs in Storybook.

**Fix:** Remove `subtitle` from `PageHeaderContent.propTypes`.

---

## Bug 3 — Global CSS variable `--page-header-title-grid-width` is unprefixed and unscoped

**File:** `packages/ibm-products/src/components/PageHeader/next/PageHeaderContentPageActions.tsx:88–90`

```ts
document.documentElement.style.setProperty(
  '--page-header-title-grid-width',
  `${width}px`
);
```

The variable is set on `document.documentElement` without the `pkg.prefix` (`c4p`) namespace. If multiple PageHeaders are rendered on the same page they will overwrite each other's value, causing layout corruption.

**Fix:** Use `--${pkg.prefix}-page-header-title-grid-width` and scope it to the component's own DOM node, not `document.documentElement`.

---

## Bug 4 — Hardcoded inline styles in `PageHeaderBreadcrumbPageActions`

**File:** `packages/ibm-products/src/components/PageHeader/next/PageHeaderBreadcrumbPageActions.tsx:110–136`

Two inline styles are applied directly in JSX:
- `inlineSize: '50%'` on the `<ul>` container
- `position: 'relative'` on the overflow `<li>`

Per component guidelines, layout values must use Carbon tokens via SCSS, not hardcoded JS inline styles.

**Fix:** Move both to the component's SCSS file using appropriate Carbon spacing/layout tokens.

---

## Acceptance Criteria

- [ ] `pageActions` render prop in `PageHeaderContent` receives live `observerState` from context
- [ ] `subtitle` removed from `PageHeaderContent.propTypes`
- [ ] CSS variable is prefixed with `pkg.prefix` and scoped to the component node, not `document.documentElement`
- [ ] `inlineSize: '50%'` and `position: 'relative'` moved to SCSS using Carbon tokens
- [ ] All existing tests pass; new test added for live `observerState` in `pageActions`

### LINKED ISSUE #9802: [preview__PageHeader]: missing fullWidthGrid / narrowGrid grid control
## Summary

The deprecated `PageHeader` supported `fullWidthGrid` and `narrowGrid` props that controlled the Carbon grid mode for the entire header. The new `preview__PageHeader` composable API does not expose an equivalent mechanism. An adopter has already raised this in the support channel.

Reference: #9743 (comment https://github.com/carbon-design-system/ibm-products/issues/9743#issuecomment-5188469697)

---

## Background — what the old props did

In the deprecated `PageHeader`, `fullWidthGrid` and `narrowGrid` were passed directly to Carbon's `<FlexGrid>` which wrapped the entire header:

```tsx
// Old PageHeader.tsx
<FlexGrid
  fullWidth={fullWidthGrid === true || fullWidthGrid === 'xl'}
  narrow={narrowGrid}
  className={cx({ [`${blockClass}--width--xl`]: fullWidthGrid === 'xl' })}
>
  {/* all header sub-sections: breadcrumb bar, content, tab bar */}
</FlexGrid>
```

- `fullWidthGrid={true}` — grid spans full viewport width (no max-width cap)
- `fullWidthGrid='xl'` — full width + `xl` override class for largest breakpoint
- `narrowGrid={true}` — Carbon narrow grid mode (content aligns to gutter edge)

---

## Problem in the new implementation

Each sub-component (`PageHeaderBreadcrumbBar`, `PageHeaderContent`, `PageHeaderTabBar`) renders its **own internal hardcoded `<Grid>`** with no way to configure grid mode from outside:

```tsx
// PageHeaderBreadcrumbBar.tsx — hardcoded, no fullWidth/narrow support
<Grid>
  <Column lg={16} md={8} sm={4}>
    ...
  </Column>
</Grid>

// PageHeaderContent.tsx — same
<Grid>
  <Column lg={16} md={8} sm={4}>
    ...
  </Column>
</Grid>

// PageHeaderTabBar.tsx — same
<Grid condensed>
  <Column lg={16} md={8} sm={4}>
    ...
  </Column>
</Grid>
```

The current migration guide says:
> `fullWidthGrid` / `narrowGrid` — Wrap with Carbon `<Grid>` / `<Column>` as needed

This is **not a true equivalent**. Wrapping `<PageHeader.Root>` in an outer `<Grid fullWidth>` has no effect on the internal grids rendered inside each sub-component.

---

## Proposed solution

Add `fullWidthGrid` and `narrowGrid` (or equivalent) props to `PageHeader.Root` and propagate them via context to all sub-components, which pass them through to their internal `<Grid>` instances:

```tsx
// PageHeader.Root
<PageHeader.Root fullWidthGrid narrowGrid={false}>
  <PageHeader.BreadcrumbBar>...</PageHeader.BreadcrumbBar>
  <PageHeader.Content title="My Page" />
  <PageHeader.TabBar>...</PageHeader.TabBar>
</PageHeader.Root>
```

The `fullWidthGrid` / `narrowGrid` values should be stored in `PageHeaderContext` and consumed by each sub-component's internal `<Grid>`.

The `fullWidthGrid='xl'` variant should also be supported for parity with the deprecated API.

---

## Acceptance Criteria

- [ ] `PageHeader.Root` accepts `fullWidthGrid?: boolean | 'xl'` and `narrowGrid?: boolean` props
- [ ] Both values are distributed via `PageHeaderContext` to all sub-components
- [ ] `PageHeaderBreadcrumbBar`, `PageHeaderContent`, and `PageHeaderTabBar` pass `fullWidth` / `narrow` through to their internal `<Grid>` components
- [ ] `fullWidthGrid='xl'` adds the `--width--xl` class (matching deprecated behaviour)
- [ ] Migration guide in `PageHeader.mdx` updated to document the new props
- [ ] Storybook stories demonstrate full-width and narrow variants
- [ ] Existing tests pass; new tests cover both props

szinta:
## Summary

Fixes and feature work for `preview__PageHeader` (`next/`), tracked from #9743.

---

### Bug 1 — Live `observerState` in functional `pageActions`
`PageHeaderContent` was calling the `pageActions` render prop with a hardcoded zeroed state object (`fullyCollapsed: false` etc.) instead of reading live state from context. Fixed by destructuring `observerState` from `usePageHeader()` and passing it through.

### Bug 2 — Stale `subtitle` in `PageHeaderContent.propTypes`
Removed the dead `subtitle: PropTypes.string` entry. The prop belongs on `PageHeaderContentText`, not `PageHeaderContent`, and was generating misleading Storybook API docs.

### Bug 3 — Global CSS variable scoping and naming
`PageHeaderContentPageActions` was setting `--page-header-title-grid-width` on `document.documentElement`, causing layout corruption when multiple `PageHeader` instances coexist. Fixed by:
- Renaming the variable to `--c4p-page-header-title-grid-width` (using `pkg.prefix`)
- Scoping the `setProperty` call to the nearest `PageHeader` root node via `.closest()`
- Updating the `var()` reference in `_page-header.scss` accordingly

### Bug 4 — Hardcoded inline styles moved to SCSS
Removed `style={{ inlineSize: '50%' }}` from the `<ul>` and `style={{ position: 'relative' }}` from the overflow `<li>` in `PageHeaderBreadcrumbPageActions`. Both are now in `_page-header.scss` under the `__next` block.

---

## Commit 2 — `fullWidthGrid` / `narrowGrid` props (closes #9802)

The deprecated `PageHeader` supported `fullWidthGrid` and `narrowGrid` props to control Carbon's grid mode across the entire header. The new composable API had no equivalent, and the migration guide incorrectly suggested wrapping with an outer `<Grid>`, which has no effect on the internal grids.

### Changes
- **`PageHeader.Root`** accepts `fullWidthGrid?: boolean | 'xl'` and `narrowGrid?: boolean`
- Both values are stored in `PageHeaderContext` and consumed by `PageHeaderBreadcrumbBar`, `PageHeaderContent`, and `PageHeaderTabBar`, which pass them to their internal `<Grid>` instances
- `fullWidthGrid='xl'` applies the `--width--xl` modifier class (`padding-inline: $spacing-07`) inside the `__next` SCSS block, matching deprecated parity
- MDX migration table updated — the old "Wrap with Carbon `<Grid>`" guidance is replaced with the correct `<PageHeader.Root fullWidthGrid narrowGrid>` equivalent
- `fullWidthGrid` (select: `false | true | 'xl'`) and `narrowGrid` (boolean) added as controls on the existing **Default** story

---

## Commit 3 — Accessibility fixes: high-priority (WCAG 2.2 AA)

Static code review identified accessibility issues across the composable `next/` PageHeader family. High-priority fixes:

### A11y Fix 1 — Empty `role="toolbar"` landmark (`PageHeaderBreadcrumbBar`)
The actions wrapper always rendered `role="toolbar"` even when both `contentActions` and `pageActions` were `undefined`. A `toolbar` landmark must contain at least one focusable control per ARIA spec (APG); an empty one confuses screen readers.

**Fix:** Conditionally render the `role="toolbar"` div only when at least one of `contentActions` or `pageActions` is present.

> WCAG 4.1.2 · ARIA 4.3.2

### A11y Fix 2 — Icon wrapper missing `aria-hidden` (`PageHeaderContent`)
The `renderIcon` wrapper `<div>` had no `aria-hidden="true"`, unlike the identical pattern already used in `PageHeaderBreadcrumbBar`. Screen readers would announce the SVG (or an empty element) alongside the visible title.

**Fix:** Added `aria-hidden="true"` to the icon container `<div>`.

> WCAG 1.1.1

---

### A11y Fix 3 — Duplicate DOM `id` on `TruncatedText` (`PageHeaderContent`)
A static hardcoded `id` string was passed to `<TruncatedText>` on every render. If two `PageHeaderContent` instances exist in the same document, duplicate IDs break `aria-labelledby`/`aria-controls` associations and fail HTML validity.

**Fix:** Replaced the static string with a `useId()`-generated unique id.

> WCAG 4.1.1 · HTML spec

### A11y Fix 4 — Popover content lacks list semantics (`PageHeaderTagOverflow`)
The hidden-tags popover rendered overflow tags inside a plain `<div>`, giving screen readers no structural context for the group.

**Fix:** Changed the container to `<ul aria-label="Hidden tags">` so the items are announced as a list.

> WCAG 1.3.1

### A11y Fix 5 — Focus lost on programmatic popover close (`PageHeaderTagOverflow`)
When the window resized, the popover was forcibly closed but focus was silently dropped rather than returned to the overflow trigger, leaving keyboard users stranded.

**Fix:** Before calling `setOpenPopover(false)` in the resize handler, the trigger element is resolved via its well-known `id` and focused.

> WCAG 2.4.3

---

### A11y Fix 6 — Dev warning for empty `collapseText`/`expandText` (`PageHeaderScrollButton`)
The default English strings `"Collapse"` / `"Expand"` are the sole accessible label for the `IconButton`. If a consumer passes an empty string, the button becomes unlabelled with no visible warning.

**Fix:** Added `pconsole.warn()` calls (the project's standard dev-only utility — no-ops in production) when either prop is an empty string.

> WCAG 2.4.6 · i18n

### A11y Fix 7 — Missing `key` props on `OverflowMenuItem` in stories (`PageHeader.stories.jsx`)
All four `OverflowMenuItem` `.map()` callbacks were missing `key` props, producing React reconciliation warnings and potentially incorrect DOM ordering / broken focus restoration in overflow menus.

**Fix:** Added `key={el.innerText}` to all four instances.

> WCAG 3.2.2

---

## Tests
- New test: `pageActions` render prop receives live `observerState` (Bug 1)
- New tests: `fullWidthGrid=true` propagates `--css-grid--full-width` to all 3 sub-component grids, `fullWidthGrid='xl'` adds `--width--xl` class, `narrowGrid=true` propagates `--css-grid--narrow` (58 tests total, all passing)
- SCSS snapshot updated


amal-k-joy (dismissed):

amal-k-joy (dismissed):

szinta:
During further review, I found several additional visual inconsistencies in the PageHeader and resolved them. The details are provided below.

#### 1. `fix(PageHeader): remove trailing slash from breadcrumbs in stories`
Corrected story-level usage of `PageHeaderBreadcrumbOverflow` — `noTrailingSlash` was being ignored because the prop wasn't wired to the underlying Carbon `Breadcrumb`.

#### 2. `fix(PageHeader): use BreadcrumbOverflow in all stories`
Updated remaining stories to use the composable `PageHeader.BreadcrumbOverflow` pattern consistently, so breadcrumb overflow behaviour is demonstrated correctly across all story variants.

#### 3. `fix(PageHeader): use MenuItem instead of OverflowMenuItem for v12`
`OverflowMenuItem` is deprecated in Carbon v12; replaced with `MenuItem` in the stories' overflow menus to suppress console warnings and align with v12 usage patterns.

#### 4.  expand icon / tag overflow overlap 
In the `TabBarWithTabsAndTags` story, the expand/collapse `PageHeaderScrollButton` (chevron icon) was overlapping the tag overflow button (`+N`) at smaller viewports.

**Root cause:** The `.scroller-button-container` was rendered as a sibling of the Carbon `<Grid>` inside `.tab-bar`, and used `position: absolute; inset-inline-end: 0` (anchored to the full-width sticky `.tab-bar`). The tag overflow container only had `padding-inline-end: $spacing-06` (24px) to create clearance, but the `IconButton size="md"` scroller is 40px wide — leaving 16px of uncovered overlap at `sm`–`md` breakpoints.

**Fixes applied**
Replaced the fragile absolute-positioning approach with a proper flex row layout — all CSS-only, no structural JSX changes:

| Before | After |
|--------|-------|
| `.tab-bar` was `display: block` at `sm+` | `.tab-bar` is `display: flex` at **all** breakpoints |
| Scroller used `position: absolute; inset-inline-end: 0` | Scroller is always **in-flow**, `flex-shrink: 0` |
| Grid had no flex sizing | Grid has `flex: 1; min-inline-size: 0` so it fills remaining space |
| Tags needed `padding-inline-end` / `transform` hacks to clear scroller | No clearance hacks needed — each element occupies its natural slot |

The result is a true three-column flex row: **[tabs — flex:1] [tag overflow] [scroller — fixed 40px]**. The scroller can never overlap the tag overflow area at any viewport width.

**Confirmed no impact on the deprecated `PageHeader`** — it uses entirely different class names (`__navigation-row`, `__navigation-tabs`, `__navigation-tags`) with zero overlap against the changed selectors.

amal-k-joy (approved):

makafsal (approved):

PR #9869, opened 2026-09-09
fix(Tearsheet next): correct Carbon layer hierarchy — body elevated instead of header

### LINKED ISSUE #9868: bug(Tearsheet next): body is elevated instead of the header, causing contrast failures in dark mode
## Problem

In the preview (`next`) Tearsheet, the **body is elevated to the wrong layer level** — the header should be the elevated surface, but instead the body is. This inversion has a concrete accessibility consequence in dark mode:

- In **dark mode**, layers get progressively *lighter* as they go deeper. Because the body already starts at the wrong (elevated) layer, any content placed inside a `<Layer>` within the body hits the ceiling of Carbon's 3-level layer system immediately. This leaves insufficient contrast between the body background and nested content — links and text can fail WCAG contrast requirements.
- In **light mode**, layer 3 cycles back to a lighter colour, so the problem is visually hidden — designs appear to work but are structurally wrong.

## Root cause

`Tearsheet.tsx` wraps the body in Carbon's `<ModalBody>`, which internally renders as a `<Layer>`. Combined with `ComposedModal` already rendering as `<Layer level={0}>`, the body content starts at layer context **2** before the consumer adds anything. Carbon's layer system caps at level **2** (three levels: 0 → 1 → 2), so the body is immediately at the ceiling with no room for consumers to layer content on top.

The legacy `TearsheetShell` avoids this by using a plain `<div className="cds--modal-content">` — never consuming a layer level — leaving the full 1 → 2 → 3 progression available to consumers, with the header correctly as the elevated surface.

amal-k-joy:
Closes #9868

In the preview (`next`) Tearsheet, the body was elevated to the wrong Carbon layer level — the header should be the elevated surface, but the body was instead. This caused nested `<Layer>` components inside the body to immediately hit the layer ceiling, leaving no contrast between the body background and nested content in dark mode.

#### What did you change?

**`packages/ibm-products/src/components/Tearsheet/next/Tearsheet.tsx`**

Replaced Carbon's `<ModalBody>` with a plain `<div className="cds--modal-content ...">`. `ModalBody` internally renders as a `<Layer>`, which consumed one layer level before the consumer added anything. Since `ComposedModal` itself renders as `<Layer level={0}>`, the body content was starting at layer context **2** — one step from the ceiling. A plain `div` matches what the legacy `TearsheetShell` does and keeps the full layer progression (1 → 2 → 3) available to consumers.

**`packages/ibm-products/src/components/Tearsheet/next/TearsheetBody.tsx`**

- `MainContent`: removed `<Layer withBackground>` wrapper, replaced with a plain `<div>`. The background is now set via CSS (`background-color: $background`) 

**`packages/ibm-products-styles/src/components/Tearsheet/_tearsheet_next.scss`**

- Added `background-color: $background` to `.__main-content` — body starts at base level.

- Verified in the `WithInfluencer` story using nested `<Layer>` components inside `MainContent`:
  ```jsx
  <TestComponent />           {/* base level — $background */}
  <Layer><TestComponent /></Layer>           {/* level 1 — $layer */}
  <Layer><Layer><TestComponent /></Layer></Layer>  {/* level 2 — $layer-02 */}
  ```
  All three render with visually distinct backgrounds in both White and G100 themes.
- Confirmed header background is elevated (`$layer`) and body background is at base (`$background`) in both light and dark themes.
- Verified influencer layout is unaffected by the `display: contents` Layer wrapper.

#### PR Checklist

### DIFF
--- a/packages/ibm-products-styles/src/components/Tearsheet/_tearsheet_next.scss
+++ b/packages/ibm-products-styles/src/components/Tearsheet/_tearsheet_next.scss
@@ -176,6 +176,7 @@ $transform-start: translate3d(0, calc(min(95vh, 500px)), 0);
     display: flex;
     flex-wrap: wrap;
     align-content: space-between;
+    background-color: $layer;
     border-block-end: 1px solid $border-subtle-01;
     margin-block: 0;
     max-block-size: 50vh;
@@ -672,9 +673,15 @@ $transform-start: translate3d(0, calc(min(95vh, 500px)), 0);
   }
 }
 
+.#{$block-class__next}__layer {
+  display: contents;
+}
+
 .#{$block-class__next}__main-content {
   position: relative;
+  background-color: $background;
 }
+
 .#{$block-class__next}__side-panel {
   position: absolute;
   z-index: 9999;
--- a/packages/ibm-products-styles/src/components/Tearsheet/_tearsheet_next.scss
+++ b/packages/ibm-products-styles/src/components/Tearsheet/_tearsheet_next.scss
@@ -176,7 +176,6 @@ $transform-start: translate3d(0, calc(min(95vh, 500px)), 0);
     display: flex;
     flex-wrap: wrap;
     align-content: space-between;
-    background-color: $layer;
     border-block-end: 1px solid $border-subtle-01;
     margin-block: 0;
     max-block-size: 50vh;
@@ -671,10 +670,6 @@ $transform-start: translate3d(0, calc(min(95vh, 500px)), 0);
   &:focus {
     @include focus-outline.focus-outline('outline');
   }
-}
-
-.#{$block-class__next}__layer {
-  display: contents;
 }
 
 .#{$block-class__next}__main-content {
--- a/packages/ibm-products/src/components/Tearsheet/next/Tearsheet.tsx
+++ b/packages/ibm-products/src/components/Tearsheet/next/Tearsheet.tsx
@@ -24,6 +24,7 @@ import {
   ComposedModal,
   type ComposedModalProps,
   unstable_FeatureFlags as FeatureFlags,
+  ModalBody,
   usePrefix,
 } from '@carbon/react';
 import { blockClass, TearsheetContext } from './context';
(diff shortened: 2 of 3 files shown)


gonzoblasco:
Hi @amal-k-joy, I've just reviewed this diff against the legacy `TearsheetShell` and Carbon's modal primitives.

**The approach restores legacy layer semantics.** Verified in Carbon source: `ComposedModal` renders `Layer level={0}` and `ModalBody` wraps its content in another `Layer` (packages/react/src/components/ComposedModal/ComposedModal.tsx), which is what pushed body content to the level-2 ceiling. The replacement body (`cds--modal-content` plain div) matches the legacy pattern, and the influencer `Layer` + `display: contents` wrapper mirrors the legacy `__layer` handling, preserving the influencer's existing layer context. The explicit backgrounds (`$layer` header, `$background` main-content) also match legacy values for the wide variant.

Two questions:

1. **Scrolling-content behavior.** `ModalBody` detects scrollable content and adds `role="region"` + `tabIndex` + aria-label (`hasScrollingContentProps`). The plain `div` drops this. If the next Tearsheet body can scroll, that announcement is lost. Is dropping it intentional?

2. **Narrow variant.** Legacy sets `__main` to `$layer` in the narrow variant and `$background` for wide (_tearsheet.scss). This PR sets `__main-content` to `$background` unconditionally, so narrow dark mode will differ from legacy. Was that considered?

Also: light-mode rendering is identical before/after by token design (layer-03 cycles back to the background value in light), so light-only visual tests won't exercise this change - a dark-mode visual pass over the published Storybook would be the real check.

amal-k-joy:
> Hi @amal-k-joy, I've just reviewed this diff against the legacy `TearsheetShell` and Carbon's modal primitives.
> 
> **The approach restores legacy layer semantics.** Verified in Carbon source: `ComposedModal` renders `Layer level={0}` and `ModalBody` wraps its content in another `Layer` (packages/react/src/components/ComposedModal/ComposedModal.tsx), which is what pushed body content to the level-2 ceiling. The replacement body (`cds--modal-content` plain div) matches the legacy pattern, and the influencer `Layer` + `display: contents` wrapper mirrors the legacy `__layer` handling, preserving the influencer's existing layer context. The explicit backgrounds (`$layer` header, `$background` main-content) also match legacy values for the wide variant.
> 
> Two questions:
> 
> 1. **Scrolling-content behavior.** `ModalBody` detects scrollable content and adds `role="region"` + `tabIndex` + aria-label (`hasScrollingContentProps`). The plain `div` drops this. If the next Tearsheet body can scroll, that announcement is lost. Is dropping it intentional?
> 2. **Narrow variant.** Legacy sets `__main` to `$layer` in the narrow variant and `$background` for wide (_tearsheet.scss). This PR sets `__main-content` to `$background` unconditionally, so narrow dark mode will differ from legacy. Was that considered?
> 
> Also: light-mode rendering is identical before/after by token design (layer-03 cycles back to the background value in light), so light-only visual tests won't exercise this change - a dark-mode visual pass over the published Storybook would be the real check.

Thanks @gonzoblasco for the review. 
1 . I have restored `ModalBody` component with level passed thereby maintaining the legacy behaviour
2. This behaviour is intentional. The wide and narrow need to follow the same layer structure unlike the legacy one.

anamikaanu96 (approved):

szinta (approved):

PR #9878, opened 2026-09-09
fix(CreateFullPage): focus first element of visible step on step change

### LINKED ISSUE #9864: [a11y]: CreateFullPage: keyboard focus moves to "Skip to main content" instead of new step content after clicking Next (regression since v2.44)
### Package

@carbon/ibm-products

### Browser

Chrome

### Operating System

MacOS

### Package version

2.80

### React version

1.99.0

Keyboard navigation

### Assistive technology

_No response_

### Description

When navigating between steps in CreateFullPage using the keyboard (pressing Next), focus jumps to the browser's "Skip to main content" link instead of the first interactive element of the newly activated step. This is a regression introduced between v2.43 and v2.44.

Actual: Focus moves to the "Skip to main content" anchor at the top of the page, or is lost entirely.

Root cause
The regression was introduced in useCreateComponentFocus.js when a new getActiveStep() helper was added to auto-detect the currently visible step.
getActiveStep walks every step element and returns the first one whose ancestor chain contains no [inert] attribute:
const getActiveStep = () => {
  const allSteps = Array.from(document.querySelectorAll(blockClass));
  return allSteps.find(el => {
    let currentStep = el;
    while (currentStep) {
      if (currentStep.hasAttribute('inert')) return false;
      currentStep = currentStep.parentElement;
    }
    return true;
  });
};

The assumption is that hidden steps are marked with the inert attribute. However, CreateFullPageStep hides non-active steps exclusively via a CSS class (display: none), never via inert:
// CreateFullPageStep.js
{
  [`${blockClass}__step--hidden-step`]: stepNumber !== stepsContext?.currentStep,
  [`${blockClass}__step--visible-step`]: stepNumber === stepsContext?.currentStep
}

.c4p--create-full-page__step__step--hidden-step { display: none; }

Because no step ever has
[inert], getActiveStep() always returns the first step in DOM order — regardless of which step is currently active.
getFocusableElement then queries for interactive elements inside that first (hidden) display: none step. While querySelectorAll does find elements inside display: none subtrees, calling .focus() on such an element is a browser no-op. No focus movement occurs, the focus stays on the Next button or drifts to <body>, and the next tab stop (often "Skip to main content") receives focus.

Expected behaviour (v2.43)
In v2.43 the hook had a simple guard — if firstFocusElement was not supplied, it did nothing and left focus management to the browser. No incorrect focus movement occurred.

Proposed fix
Replace the [inert]-based search with a direct selector targeting the visible-step CSS modifier, which is already reliably applied by CreateFullPageStep:
// useCreateComponentFocus.js
const getActiveStep = () => {
  // Select the step that is currently visible using its CSS modifier class,
  // not by absence of [inert] (which is never set on hidden steps).
  return document.querySelector(`${blockClass}__step--visible-step`);
};

This is O(1), unambiguous, and consistent with how the component already signals step visibility.

Affected versions
VersionBehaviour2.43.x:white_check_mark: No incorrect focus movement2.44.0 and later (tested up to 2.80.0):x: Focus jumps to "Skip to main content"
Environment

@carbon/ibm-products:2.80.0(regression present),2.43.1(working)
Component:CreateFullPage/CreateFullPageStep
Trigger: keyboard activation ofNextbutton withoutfirstFocusElementprop
Browsers affected: Chromium (Windows), Firefox (macOS) — any browser where.focus()is a no-op ondisplay: noneelements

### WCAG 2.1 Violation

_No response_

### Reproduction/example

https://ibm-products-web-components.carbondesignsystem.com/?path=/story/patterns-create-flows-createfullpage--create-full-page

### Steps to reproduce

Steps to reproduce

Render aCreateFullPagewith two or more steps, each containing interactive elements (e.g. radio buttons, inputs).
Donotsupply thefirstFocusElementprop.
Navigate to the first step and pressEnterorSpaceon theNextbutton using keyboard only.
Observe where focus lands.
Expected: Focus moves to the first interactive element in the newly visible step (e.g. the first radio button or input field).

### Code of Conduct

gonzoblasco:
Closes #9864

## Root cause

`getActiveStep()` in `useCreateComponentFocus.js` walked the DOM looking for a step whose ancestor chain contained no `[inert]` attribute. `CreateFullPageStep` never sets `inert` - it hides non-active steps exclusively with the `__step--hidden-step` CSS class (`display: none`). The walk therefore always returned the first step in DOM order, regardless of which step was active.

`getFocusableElement` then found interactive elements inside that hidden (`display: none`) step. Calling `.focus()` on an element inside a `display: none` subtree is a browser no-op, so focus never moved to the newly activated step and drifted to the next tab stop - often the "Skip to main content" link. Regression introduced between v2.43 and v2.44.

## Fix

Replace the `[inert]`-based walk with a direct query for the visible-step modifier class that `CreateFullPageStep` already applies:

```js
return document.querySelector(`${blockClass}__step--visible-step`);
```

The `blockClass` prop passed from `CreateFullPage` ends with `__step` (`.prefix--create-full-page .prefix--create-full-page__step`), so the composed selector targets the real DOM class `.prefix--create-full-page__step__step--visible-step`.

## Testing

- Added a regression test: renders a two-step `CreateFullPage` with text inputs, activates `Next` with the keyboard, and asserts focus lands on the first input of the newly visible step (not on the "Skip to main content" link or `document.body`).
- The test fails on `main` (focus never lands, waitFor timeout) and passes with this fix.
- Full `CreateFullPage.test.jsx` suite: 28/28 passing.

### DIFF
--- a/packages/ibm-products/src/components/CreateFullPage/CreateFullPage.test.jsx
+++ b/packages/ibm-products/src/components/CreateFullPage/CreateFullPage.test.jsx
@@ -84,10 +84,10 @@ const renderComponent = ({ ...rest } = {}) =>
       {...rest}
     >
       <CreateFullPageStep title="Title 1" subtitle="Subtitle 1">
-        <p>1</p>
+        {stepFormField}
       </CreateFullPageStep>
       <CreateFullPageStep title="Title 2" description="2">
-        <p>2</p>
+        {stepFormField}
       </CreateFullPageStep>
     </CreateFullPage>
   );
@@ -689,5 +689,40 @@ describe(componentName, () => {
         `${carbon.prefix}--progress-step--current`
       )
     ).toBe(true);
+  });
+
+  it('moves focus to the first input of the visible step after clicking Next (issue #9864)', async () => {
+    const { click } = userEvent;
+    const { container } = renderCreateFullPage(defaultFullPageProps);
+
+    // Step 1 is visible on mount; its input should not be focused before Next.
+    const firstStepInput = container.querySelector(
+      `.${blockClass}__step__step--visible-step input`
+    );
+    expect(firstStepInput).toBeTruthy();
+
+    // Advance to step 2.
+    await act(() => click(screen.getByText(nextButtonText)));
+    await waitFor(() => expect(onNextStepFn).toHaveBeenCalled());
+
+    // The visible step is now step 2.
+    const visibleStep = container.querySelector(
+      `.${blockClass}__step__step--visible-step`
+    );
+    expect(visibleStep).toBeTruthy();
+
+    // The hook awaits a 10ms timeout before focusing; wait for focus to land.
+    await waitFor(() => {
+      const focusedInput = container.querySelector(
+        `.${blockClass}__step__step--visible-step input`
+      );
+      expect(focusedInput).toBeTruthy();
+      expect(document.activeElement).toBe(focusedInput);
+    });
+
+    // Focus must NOT be on the "Skip to main content" button or the body.
+    const skipLink = container.querySelector('a');
+    expect(document.activeElement).not.toBe(skipLink);
+    expect(document.activeElement).not.toBe(document.body);
   });
 });
--- a/packages/ibm-products/src/global/js/hooks/useCreateComponentFocus.js
+++ b/packages/ibm-products/src/global/js/hooks/useCreateComponentFocus.js
@@ -36,16 +36,16 @@ export const useCreateComponentFocus = ({
       );
     };
 
+    // The currently visible step is the one carrying the visible modifier
+    // class. NOTE: the `blockClass` prop ends with `__step` (`.prefix--create-
+    // full-page .prefix--create-full-page__step`), so appending
+    // `__step--visible-step` targets
+    // `.prefix--create-full-page__step__step--visible-step`.
+    // (Fix for carbon-design-system/ibm-products#9864: the previous inert-walk
+    // logic never matched because CreateFullPageStep hides steps with
+    // display:none instead of the `inert` attribute.)
     const getActiveStep = () => {
-      const allSteps = Array.from(document.querySelectorAll(blockClass));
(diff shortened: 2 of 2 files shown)


amal-k-joy (approved):
@gonzoblasco 
Please note that CreateFullPage is a prebuilt pattern component that has been identified for [deprecation](https://ibm-products.carbondesignsystem.com/?path=/docs/deprecated-prebuilt-patterns-overview--overview&globals=theme:g100). We are in the path of deprecating all our prebuilt patterns in favour of true example patterns.
We’ve provided an equivalent, true pattern example for CreateFullPage [here](https://ibm-products.carbondesignsystem.com/?path=/docs/patterns-create-flows-create-full-page--overview&globals=theme:g100) ([stackblitz app ](https://stackblitz.com/github/carbon-design-system/ibm-products/tree/main/packages/ibm-products/src/patterns/CreateFullPage/example)). You can open the example app and copy the pattern code into your source code.
This approach should be more effective and scalable, as it allows you to adopt and maintain the fix directly in your application.

gonzoblasco:
Thanks for the context on the deprecation path.

I'd like to keep this fix in the prebuilt component rather than closing it. The bug is live for anyone on the published package today, and the change is one line in the shared hook plus a regression test - no new API surface, no maintenance burden beyond what the component already carries. Users hitting this on 2.44 through current get correct focus behavior without having to migrate first.

On the pattern example: happy to follow up separately if it's useful. My read is that the example doesn't ship the same focus-on-step-change logic, so there may not be an equivalent bug to fix there - but if I'm wrong about that, tell me and I'll open a PR against the pattern instead.

One thing worth flagging on scope: `useCreateComponentFocus` is also used by `CreateTearsheet`, whose steps set `inert` on the outer element in addition to the visible-step class. Both paths still resolve there, so no regression - but that's the blast radius of the change, and it's not spelled out in the PR description.

amal-k-joy (approved):

makafsal (approved):

